Issue #76, opened 2020-06-30
Add signature computation support for arbitrary logical plans


pirz:
### Describe the issue
Currently, Hyperspace only supports generating plan signature for logical plans consisting of only logical relation nodes (Check [FileBasedSignatureProvider](https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/index/FileBasedSignatureProvider.scala)).
Moving forward, Hyperspace needs to support generating signature for arbitrary logical plans. 

### To Reproduce
NA

### Expected behavior
Signature computation code supports arbitrary logical plans and not just plans consisting of logical relation nodes.

### Environment 
All

pirz:
This issue is addressed by PR #77

Issue #31, opened 2020-06-24
Remove Test duplication, e.g. creation of test object IndexLogEntry is highly duplicated


apoorvedave1:
### Describe the issue
Remove Test duplication, e.g. creation of test object IndexLogEntry is highly duplicated. This is required only for code cleanup. No functionality change.

### To Reproduce
NA

### Expected behavior
NA

### Environment 
NA

clee704:
Is it still open?

sezruby:
Closed by #77 & #89

PR #77, opened 2020-06-30
Add signature computation support for arbitrary logical plans

pirz:
Currently, Hyperspace has one signature provider: `FileBasedSignatureProvider`, which only covers LogicalRelation nodes in a plan to compute the plan signature.
This limits the DataFrames that could potentially be indexed to only those whose plans consist of only logical relation nodes.
Moving forward, Hyperspace needs to support generating signature for DataFrames with arbitrary logical plans. 

Below changes address issue #76 
1- Adds new signature provider which computes a signature for arbitrary logical plans.
2- Changes existing Hyperspace rules to comply with changes in the signature provider.
3- Changes the API for signature() in Signature provider to return an Option (instead of a default value) to avoid potential issues due to using a default signature.
4- It adds test cases for the new signature provider and refactors existing test code.

Below change addresses Issue #31 
5- It refactors the test code in JoinIndex and FilterIndex rules tests.

Supporting signature computation for arbitrary logical plans.

Refactoring and cleaning up test code.

No

Test cases are added under new files: IndexSignatureProviderTest and RulesHelperTest.
Backward combability tested manually. A sample application was used to generate indexes with current master (0.1.0 code without this change) and new application code using "current master + this change" used to read and leverage those indexes. The output was verified manually to confirm the indexes created with an older version of bits are still usable with new bits including this change.


src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Generate the signature of logical plan.
+   *
+   * @param logicalPlan logical plan.
+   * @return signature if there is at least one operator in the plan; Otherwise None.
+   */
+  def signature(logicalPlan: LogicalPlan): Option[String] = {
+    var signature = ""
+    logicalPlan.foreachUp(o => signature += HashingUtils.md5Hex(o.nodeName))
```
apoorvedave1, line 38: could you please add a final HashingUtils.md5Hex(signature) on the overall signature?

src/main/scala/com/microsoft/hyperspace/index/rules/RulesHelper.scala
```diff
@@ -0,0 +1,41 @@
...
+  def getLogicalRelation(logicalPlan: LogicalPlan): Option[LogicalRelation] =
+    logicalPlan match {
+      case r: LogicalRelation => Some(r)
+      case p: LogicalPlan =>
+        if (p.children.size == 1) {
+          getLogicalRelation(p.children.head)
+        } else {
+          None // plan is non-linear or it does not have any LogicalRelation
+        }
+    }
```
apoorvedave1, line 40: can simplify to 
```
val lrs = logicalPlan.collect {
    case r: LogicalRelation => r
}
if (lrs.count == 1) Some(lrs.head) else None
```
or something like 
```
val leaves = logicalPlan.collectLeaves
leaves match {
   case r : LogicalRelation :: Nil => r
   case _ => None
}
```

src/test/scala/com/microsoft/hyperspace/index/IndexSignatureProviderTest.scala
```diff
@@ -0,0 +1,141 @@
...
+    "Verify signature inequality for plans with same logical relation and different structure.") {
+    val r1 = createSimplePlan(fileLength1, fileModificationTime1, filePath1)
+    val r2 = createLinearPlan(fileLength2, fileModificationTime2, filePath2)
+    assert(!createIndexSignature(r1).equals(createIndexSignature(r2)))
+  }
+
+  private def createSimplePlan(
+                                length: Long,
+                                modificationTime: Long,
+                                path: Path): LogicalRelation = {
```
apoorvedave1, line 67: formatting

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -149,42 +149,49 @@ object FilterIndexRule extends Rule[LogicalPlan] with Logging {
...
-        val signature = LogicalPlanSignatureProvider
-          .create(sourcePlanSignature.provider)
-          .signature(project)
-        signatureMap.put(sourcePlanSignature.provider, signature)
-      }
-
-      signatureMap(sourcePlanSignature.provider).equals(sourcePlanSignature.value)
-    }
+    RulesHelper.getLogicalRelation(project) match {
+      case Some(r) =>
```
apoorvedave1, line 153: Could you please explain the reason for this change? To me, this looks like hard-coding signature calculation logic to a signature provider which works only on LogicalRelation. I think here we should keep it relaxed to any "plan" node and not just logical relation.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -329,27 +329,36 @@ object JoinIndexRule extends Rule[LogicalPlan] with Logging {
...
-      val sourcePlanSignature = sourcePlanSignatures.head
-
-      if (!signatureMap.contains(sourcePlanSignature.provider)) {
-        val signature: String = LogicalPlanSignatureProvider
-          .create(sourcePlanSignature.provider)
-          .signature(plan)
-        signatureMap.put(sourcePlanSignature.provider, signature)
-      }
+    RulesHelper.getLogicalRelation(plan) match {
+      case Some(r) =>
```
apoorvedave1, line 333: same comment as in FilterIndexRule, copy pasting here.

Could you please explain the reason for this change? To me, this looks like hard-coding signature calculation logic to a signature provider which works only on LogicalRelation. I think here we should keep it relaxed to any "plan" node and not just logical relation. I also don't think it saves on calculation so some explanation could help. Thanks.

imback82:
Under `How was this patch tested?` in the description, can you add a comment whether you checked the backward compatibility?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -60,30 +61,31 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+            CoveringIndex.Properties(
+              CoveringIndex.Properties
+                .Columns(indexConfig.indexedColumns, indexConfig.includedColumns),
+              IndexLogEntry.schemaString(schema),
+              numBuckets)),
+          Content(path.toString, Seq()),
+          Source(SparkPlan(sourcePlanProperties), Seq(Hdfs(sourceDataProperties))),
+          Map())
+
+      case None => throw HyperspaceException("Invalid plan for index dataFrame.")
```
imback82, line 87: "Invalid plan for creating an index"?

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
imback82: also `p` instead of `o` as in "plan"?

src/main/scala/com/microsoft/hyperspace/index/FileBasedSignatureProvider.scala
```diff
@@ -26,26 +26,27 @@ import com.microsoft.hyperspace.util.HashingUtils
...
    * Generate the signature of logical plan.
    *
    * @param logicalPlan logical plan of data frame.
-   * @return signature.
+   * @return signature, if the logical plan has some LogicalRelation operator(s); Otherwise None.
    */
-  def signature(logicalPlan: LogicalPlan): String = {
-    HashingUtils.md5Hex(fingerprintVisitor(logicalPlan))
+  def signature(logicalPlan: LogicalPlan): Option[String] = {
+    fingerprintVisitor(logicalPlan).map(HashingUtils.md5Hex(_))
```
imback82, line 40: I think `_` is not required: `fingerprintVisitor(logicalPlan).map(HashingUtils.md5Hex)`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -33,27 +34,54 @@ object RuleUtils {
...
+        val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
+        allIndexes.filter(index => index.created && signatureValid(index))
+
+      case None => Seq[IndexLogEntry]()
+    }
   }
+
+  /**
+   * Extract LogicalRelation node from a given logical plan.
+   * The assumption is the input plan is linear and has one scan node.
```
imback82, line 72: Looks like this is not an assumption, but you are checking this in this function?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -33,27 +34,54 @@ object RuleUtils {
...
-          .create(sourcePlanSignature.provider)
-          .signature(plan)
-        signatureMap.put(sourcePlanSignature.provider, signature)
-      }
-      signatureMap(sourcePlanSignature.provider).equals(sourcePlanSignature.value)
-    }
+    getLogicalRelation(plan) match {
+      case Some(r) =>
+        // Map of a signature provider to a signature generated for the given plan.
+        val signatureMap: mutable.Map[String, String] = mutable.Map()
```
imback82, line 40: why has this changed from `val signatureMap = mutable.Map[String, String]()`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -33,27 +34,54 @@ object RuleUtils {
...
+          assert(sourcePlanSignatures.length == 1)
+          val sourcePlanSignature = sourcePlanSignatures.head
+
+          if (!signatureMap.contains(sourcePlanSignature.provider)) {
+            val signature = LogicalPlanSignatureProvider
+              .create(sourcePlanSignature.provider)
+              .signature(r)
+              .get // It is safe to call Option.get as signature() is
+            // called on a LogicalRelation node which is
+            // valid for signature computation
```
imback82, line 54: Hmm, should we just call `signature` with the given logical plan and not put it into the map if it returns None (instead of relying on `getLogicalRelation`)? Then this function will just return empty Seq, which translates to no candidate indexes, which makes sense.

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
pirz: @imback82 `o` is actually added for "Operator" because for each operator we add hash of its node name to the plan.

imback82: I didn't think "operator" when I saw `o` since technically it's `logicalPlan.foreachUp(p: LogicalPlan => ...)`. But it's nit anyway.

pirz: Got it, I will change it in the next push.

src/main/scala/com/microsoft/hyperspace/index/rules/RulesHelper.scala
pirz: Thnx, It is changed and this code is now in `RuleUtils.scala`.

src/test/scala/com/microsoft/hyperspace/index/IndexSignatureProviderTest.scala
pirz: Fixed, thnx.

src/main/scala/com/microsoft/hyperspace/index/IndexSignatureProvider.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Generate the signature of logical plan.
+   *
+   * @param logicalPlan logical plan.
+   * @return signature, if both [[FileBasedSignatureProvider]] and [[PlanSignatureProvider]]
+   *        can generate signature for the logical plan; Otherwise None.
+   */
+  def signature(logicalPlan: LogicalPlan): Option[String] =
+    fileBasedSignatureProvider.signature(logicalPlan) match {
```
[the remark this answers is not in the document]
pirz: Fixed, thnx.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/index/FileBasedSignatureProvider.scala
pirz: Fixed.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
pirz: Reworded the comment. I really meant that if a plan is non-linear, this function wont return the logical relations.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
pirz: Fixed it. It had happened due to resolving merge conflicts and an older code had replaced the newer one.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
pirz: This is not done for saving calculations, but it is needed for correctness.
Assume we have a query plan as:
	QueryPlan (P):  Scan(R: a, b, c) -> Filter(a>5) -> Project(b)
and we have two indexes defined as:	
	Covering Index 1 (IX1): R(a, b)
	Index 2 (IX2): Scan(a, b, c) -> Filter(a>100) -> Project(b, c)	
As the LogicalRelation node is the same for all 3 plans, `FileBasedSignatureProvider` generates the same signature for all three of them.
`IndexSignatureProvider` (introduced by this change) includes node names in signature calculation and as a result it generate the same signature for P and IX2 and a different signature for IX1.

In FilterIndex and JoinIndex rules, a candidate index has to meet below two conditions:
- Index has to be defined on the same LogicalRelation as the one query uses,
- Index has to cover all records (i.e. index plan should not have a Filter, Join, Agg or any other node that removes or modifies original relation's records).
    
As a result, when we look for indexes in FilterIndex and JoinIndex rules we need to: "Extract a single logical relation node from original query plan and compare the signature of that subplan with the signature of covering indexes".
This way the signature comparison will satisfy both conditions mentioned above for these rules and only returns covering indexes that include all records of original relation.
In the above example, this procedure of extracting logical relation from query plan and comparing its signature with index signatures (which are calculated and stored using IndexSignatureProvider) results in only marking IX1 as a candidate index.

Please let me know if above explanation makes sense; we can discuss it further if you prefer.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
pirz: I added an explanation to the previous comment in FilterIndex rule. Please check that and let me know if further discussion is required. Thnx

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
pirz: Using `getLogicalRelation` here is required for correctness as we need to extract the relevant subplan for the query plan to calculate the signature and compare it with covering indexes' signatures.
Here is an explanation on why we need to extract the logical relation. Plz let me know if it makes sense or we need to further discuss it
:
Assume we have a query plan as:
	QueryPlan (P):  Scan(R: a, b, c) -> Filter(a>5) -> Project(b)
and we have two indexes defined as:	
	Covering Index 1 (IX1): R(a, b)
	Index 2 (IX2): Scan(a, b, c) -> Filter(a>100) -> Project(b, c)	
As the LogicalRelation node is the same for all 3 plans, `FileBasedSignatureProvider` generates the same signature for all three of them.
`IndexSignatureProvider` (introduced by this change) includes node names in signature calculation and as a result it generate the same signature for P and IX2 and a different signature for IX1.

In FilterIndex and JoinIndex rules, a candidate index has to meet below two conditions:
- Index has to be defined on the same LogicalRelation as the one query uses,
- Index has to cover all records (i.e. index plan should not have a Filter, Join, Agg or any other node that removes or modifies original relation's records).
    
As a result, when we look for indexes in FilterIndex and JoinIndex rules we need to: "Extract a single logical relation node from original query plan and compare the signature of that subplan with the signature of covering indexes".
This way the signature comparison will satisfy both conditions mentioned above for these rules and only returns covering indexes that include all records of original relation.
In the above example, this procedure of extracting logical relation from query plan and comparing its signature with index signatures (which are calculated and stored using IndexSignatureProvider) results in only marking IX1 as a candidate index.

imback82: Sounds like this would only apply to existing filter / join rule? In that case, those rules should call `getLogicalRelation` and call `getCandidateIndexes` with the result of `getLogicalRelation`. `getCandidateIndexes` doesn't seem like the right place to have a tight dependency on `getLogicalRelation`.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -33,27 +34,51 @@ object RuleUtils {
...
+        //  the code to support indexes in transitioning states as well.
+        val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
+        allIndexes.filter(index => index.created && signatureValid(index))
+
+      case None => Seq[IndexLogEntry]()
+    }
+  }
+
+  /**
+   * Extract LogicalRelation node from a given linear logical plan.
```
imback82, line 71: We are not "given` a linear plan, right? You meant "Extract a LogicalRelation node if the given logical plan is linear"?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -33,27 +34,51 @@ object RuleUtils {
...
+   *
+   * @param logicalPlan given logical plan to extract LogicalRelation from.
+   * @return if the plan is linear, the LogicalRelation node; Otherwise None.
+   */
+  def getLogicalRelation(logicalPlan: LogicalPlan): Option[LogicalRelation] = {
+    val lr = logicalPlan.collect { case r: LogicalRelation => r }
```
imback82, line 77: nit: `lrs`

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -93,4 +94,24 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+  test("Verify get logical relation for non-linear plan.") {
+    val joinNode = Join(t1ProjectNode, t2ProjectNode, JoinType("inner"), None)
+    val r = RuleUtils.getLogicalRelation(Project(Seq(t1c3, t2c3), joinNode))
+    assert(r.isEmpty)
+  }
+
+  private def validateLogicalRelation(plan: LogicalPlan): Unit = {
+    val r = RuleUtils.getLogicalRelation(plan)
+    assert(r.isDefined)
+    assert(r.get.equals(t1ScanNode))
```
imback82, line 115: It should be `private def validateLogicalRelation(plan: LogicalPlan, expected: LogicalRelation)` and not tightly depend on `t1ScanNode`

src/test/scala/com/microsoft/hyperspace/index/SignatureProviderTestHelper.scala
```diff
@@ -0,0 +1,64 @@
...
+import scala.collection.mutable
+
+import org.apache.hadoop.fs.{FileStatus, Path}
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
+import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex, PartitionSpec}
+import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
+import org.apache.spark.sql.types.StructType
+
+object SignatureProviderTestHelper {
```
imback82, line 28: nit: can we rename this to `SignatureProviderTestUtils`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
pirz: That is correct and your suggestion makes sense. I changed the code accordingly.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
pirz: true, I fixed it.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
pirz: changed it.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
pirz: Fixed it.

src/test/scala/com/microsoft/hyperspace/index/SignatureProviderTestHelper.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -42,10 +43,13 @@ object RuleUtils {
...
       if (!signatureMap.contains(sourcePlanSignature.provider)) {
-        val signature = LogicalPlanSignatureProvider
+        LogicalPlanSignatureProvider
           .create(sourcePlanSignature.provider)
-          .signature(plan)
-        signatureMap.put(sourcePlanSignature.provider, signature)
+          .signature(plan) match {
+          case Some(s) =>
+            signatureMap.put(sourcePlanSignature.provider, s)
+          case None => return false
```
imback82, line 51: I think we should still put the value if the signature generation is already tried. You probably need to change the map to `val signatureMap = mutable.Map[String, Option[String]]()`.

pirz: Makes sense, changed it. Thnx

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -151,18 +150,25 @@ object FilterIndexRule
...
+        candidateIndexes.filter { index =>
+          indexCoversPlan(
+            outputColumns,
+            filterColumns,
+            index.indexedColumns,
+            index.includedColumns,
+            fsRelation.fileFormat)
+        }
+
+      case None =>
```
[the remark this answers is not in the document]
pirz: Done.

src/main/scala/com/microsoft/hyperspace/index/IndexSignatureProvider.scala
```diff
@@ -0,0 +1,50 @@
...
+   * @param logicalPlan logical plan.
+   * @return signature, if both [[FileBasedSignatureProvider]] and [[PlanSignatureProvider]]
+   *         can generate signature for the logical plan; Otherwise None.
+   */
+  def signature(logicalPlan: LogicalPlan): Option[String] =
+    fileBasedSignatureProvider.signature(logicalPlan).flatMap { f =>
+      planSignatureProvider.signature(logicalPlan).map { p =>
+        HashingUtils.md5Hex(f + p)
+      }
+    }
```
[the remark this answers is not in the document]
pirz: Done.

pirz:
@imback82 I verified backward compatibility and updated PR description to reflect that with details.

imback82:
> @imback82 I verified backward compatibility and updated PR description to reflect that with details.

Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -47,7 +48,10 @@ object RuleUtils {
           .signature(plan)
         signatureMap.put(sourcePlanSignature.provider, signature)
```
imback82, line 49: can we use `getOrElseUpdate` to remove double look up?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -228,41 +228,46 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
+        val sourcePlanProperties = SparkPlan.Properties(
+          serializedPlan,
+          LogicalPlanFingerprint(
+            LogicalPlanFingerprint.Properties(
+              Seq(Signature(
+                LogicalPlanSignatureProvider.create().name,
+                s)))))
+        val sourceFiles = df.queryExecution.optimizedPlan.collect {
+          case LogicalRelation(
           HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
```
imback82, line 244: indentation looks off? is this from auto-format?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -228,41 +228,46 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
+              IndexConstants.INDEX_NUM_BUCKETS_DEFAULT)),
+          Content(
+            s"$indexStorageLocation/${indexConfig.indexName}" +
+              s"/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=0",
+            Seq()),
+          Source(SparkPlan(sourcePlanProperties), Seq(Hdfs(sourceDataProperties))),
+          Map())
+        entry.state = state
+        entry
+      case None => throw HyperspaceException("Invalid plan for index dataFrame.")
```
imback82, line 269: should we just do assert?

src/test/scala/com/microsoft/hyperspace/index/IndexSignatureProviderTest.scala
```diff
@@ -0,0 +1,141 @@
...
+    val aggExpression =
+      Seq(t1c2, t1c3, Alias(Count(Seq(t2c1)), "count(t2c1)")())
+
+    Aggregate(grpExpression, aggExpression, projectNode)
+  }
+
+  private def createIndexSignature(logicalPlan: LogicalPlan): String =
+    new IndexSignatureProvider().signature(logicalPlan) match {
+      case Some(s) => s
+      case None => throw HyperspaceException("Invalid plan for signature generation.")
```
imback82, line 133: just `fail` here? unless you are testing the scenario where exception is thrown

src/test/scala/com/microsoft/hyperspace/index/IndexSignatureProviderTest.scala
```diff
@@ -130,7 +130,7 @@ class IndexSignatureProviderTest extends SparkFunSuite with SparkInvolvedSuite {
   private def createIndexSignature(logicalPlan: LogicalPlan): String =
     new IndexSignatureProvider().signature(logicalPlan) match {
       case Some(s) => s
-      case None => throw HyperspaceException("Invalid plan for signature generation.")
+      case None => fail // Invalid plan for signature generation.
```
imback82, line 133: I think you can do `fail("Invalid plan for signature generation")`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
pirz: nice, added it.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
pirz: fixed.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
pirz: Done.

src/test/scala/com/microsoft/hyperspace/index/IndexSignatureProviderTest.scala
pirz: Done.

src/test/scala/com/microsoft/hyperspace/index/IndexSignatureProviderTest.scala
pirz: Done.

imback82:
@apoorvedave1 Any other comments?

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Generate the signature of logical plan.
+   *
+   * @param logicalPlan logical plan.
+   * @return signature if there is at least one operator in the plan; Otherwise None.
+   */
+  def signature(logicalPlan: LogicalPlan): Option[String] = {
+    var signature = ""
+    logicalPlan.foreachUp(p => signature += HashingUtils.md5Hex(p.nodeName))
```
apoorvedave1, line 38: nit: we can avoid this calculation by just doing md5Hex in the last step in match-case statement. It would be faster and still correct. Right?

apoorvedave1: cc @imback82

pirz: This is an approach that we take in [FileBasedSignatureProvider](https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/index/FileBasedSignatureProvider.scala#L59) as well and it makes the final value non-sensitive to the order of strings (or files in case of FileBasedSignatureProvider). This is important as for example for a plan with a  join node the signature should be independent of order of visiting join children. Otherwise the signature computation would not be reproducible if the order of visiting left and right children swap.

pirz: @imback82 @apoorvedave1 Can you please check above comment and merge this PR if you are fine with it?

imback82: > Visiting plan as Join,LeftChild,RightChild should produce the same signature as visiting the plan as Join,RightChild,LeftChild. Calculating Hash value per node name and summing them up makes this achievable but concatenating node names and calling hash() on final String breaks this.

@pirz are you sure about this? `hash(a) + hash(b)` is the same as `hash(b) + hash(a)` when `hash` returns a string?

@apoorvedave1 I am not too sure about the performance improvement here (<100 nodes) and would depend on how we append the strings. I would say it would be negligible. Do you have some numbers to back up?

pirz: @imback82 My understanding of @apoorvedave1 's comment was: When visiting `a,b,c` instead of doing `hash(a) + hash(b) + hash(c)` do it as first concat `abc` and then call hash once as `hash(abc)` to save time. My point was that if we do it that way we hurt correctness as `hash(abc)` is different from `hash(acb)`. That is why we should call hash per node and add the hash values.

imback82: Well, `hash(a) + hash(b) + hash(c)` is different from `hash(a) + hash(c) + hash(b)` since `hash` returns a string, so if there is a correctness issue in `hash(abc)`, the current approach also has an issue.

pirz: @imback82 Actually you are right - that order does not matter if we treat the hash value as a number and add up numbers and hash the final number to a String. We do not follow that approach here and also not in `FileBasedSignatureProvider`.

apoorvedave1: @imback82 no I don't have numbers for this, I just said it based on the understanding that hash(hash(a)+hash(b)+hash(c)) will take more time than hash(a+b+c). Other than that, my concern was the second approach looks slightly simpler and we don't lose anything. But either is fine actually.

imback82: well, `a + b + c` involves string copies, so depending on the size of those strings, the above statement may not be true.

imback82: OK, can we do this instead?
```scala
var signature = ""
logicalPlan.foreachUp(p => signature = HashingUtils.md5Hex(signature + p.nodeName))
signature match {
  case "" => None
  case _ => signature
}
```
We will always bound the string size and also consistent with file based signature.

imback82: Since this PR is getting long, I will merge this now. @pirz Can you do a follow up PR to address @apoorvedave1's concern? (if we are going with one md5hex, let's use `StringBuffer`).

pirz: Makes sense, I updated it. @imback82 and @apoorvedave1 Can you plz check/approve again?

apoorvedave1: approved. Thanks!

pirz: @imback82 Sure, Thanks. I guess we first need some numbers to see if calling Hash multiple times on the plans (which are normally <100 nodes as you said) is a perf bottleneck or not. If that turned out to be an issue, we will switch.

Issue #25, opened 2020-06-23
Add Alias column support in join condition for JoinIndex rule


apoorvedave1:
**Describe the issue**
Add support for `org.apache.spark.sql.catalyst.expressions.Alias` as column type in `JoinIndexRule.scala`. Currently, this rule only supports Join Conditions on `org.apache.spark.sql.catalyst.expressions.Attribute`s . Consider a case where a column `c` from the base table has an alias `aliasC`. A Join query with `c` will be able to utilize indexes but a query with `aliasC` won't be.

**To Reproduce**
Create a query plan of the form 

```sql
select T1.A, T2Temp.B 
from T1 
INNER JOIN (
   select c as aliasC 
   from T2
) as T2Temp 
where T1.C = T2Temp.aliasC
```

**Expected behavior**
Query utilizes index

**Screenshots**
NA
**Desktop (please complete the following information):**
NA
**Additional context**
NA

rapoth:
CC: @sezruby

rapoth:
CC: @imback82

imback82:
I think this should have been fixed by https://github.com/apache/spark/pull/26943 in Spark 3.0.

For the following:
```scala
spark.conf.set("spark.sql.autoBroadcastJoinThreshold", "-1")
sql("drop table t1")
sql("drop table t2")
val df1 = (0 until 100).map(i => (i % 5, i % 11)).toDF("i1", "j1")
val df2 = (0 until 100).map(i => (i % 7, i % 11)).toDF("i2", "j2")
df1.write.format("parquet").bucketBy(8, "i1").saveAsTable("t1")
df2.write.format("parquet").bucketBy(8, "i2").saveAsTable("t2")
sql("SELECT t1.i1, t2Temp.aliasC FROM t1 INNER JOIN (SELECT i2 as aliasC from t2) as t2Temp WHERE t1.i1 = t2Temp.aliasC").explain
```

Spark 2.4.5 outputs:
```
== Physical Plan ==
*(5) SortMergeJoin [i1#31], [aliasC#37], Inner
:- *(2) Sort [i1#31 ASC NULLS FIRST], false, 0
:  +- Exchange hashpartitioning(i1#31, 200)
:     +- *(1) Project [i1#31]
:        +- *(1) Filter isnotnull(i1#31)
[...]

PR #78, opened 2020-07-01
Case Sensitivity: Support Case Insensitive Index Column Names

### LINKED ISSUE #27: Handle Case sensitivity in indexed and included column names appropriately
### Describe the issue
Handle case-sensitivity in indexed and included column names appropriately based on the value of `"spark.sql.caseSensitive"`.

Here's the desired behavior
Q. If case-sensitive = true, index config is wrong case, will createIndex work? No. Fail, exception.
Q. If case-sensitive = false, index config is wrong case, will createIndex work? Yes.
Q. If case-sensitive = true, an index exists with wrong case, will this index be used? No. Ignore this index
Q. If case-sensitive = false, an index exists with wrong case, will this index be used? Yes.

### To Reproduce
NA

### Expected behavior
Here's the desired behavior
Q. If case-sensitive = true, index config is wrong case, will createIndex work? No. Fail, exception.
Q. If case-sensitive = false, index config is wrong case, will createIndex work? Yes.
Q. If case-sensitive = true, an index exists with wrong case, will this index be used? No. Ignore this index
Q. If case-sensitive = false, an index exists with wrong case, will this index be used? Yes.

### Environment 
NA

apoorvedave1:
This fixes #27 

Currently index creation and usage is case sensitive. This PR addresses the issue by using "spark.sql.caseSensitivity" to determine whether to support case sensitivity or not. It adds support to use indexes if the original query is case insensitive.

Currently index creation and usage is case sensitive. 
If base table has columns "c1", "c2" and the query is "SELECT C1, C2 FROM T WHERE...." the indexes cannot be used beause the rules do a case-sensitive matching irrespective of the value of "spark.sql.caseSensitivity".

No

Added unit tests to cover case-insensitive scenarios


src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -49,8 +51,19 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
     val signatureProvider = LogicalPlanSignatureProvider.create()
 
+    val (resolvedIndexedColumns, resolvedIncludedColumns) = {
+      val dfColumnNames = df.schema.fieldNames
+      try {
+        // Try creating log entry with resolved column names.
+        (indexConfig.indexedColumns.map(resolve(spark, _, dfColumnNames)),
+          indexConfig.includedColumns.map(resolve(spark, _, dfColumnNames)))
+      } catch {
+        // Try creating index log entry with whatever the user passed.
```
apoorvedave1, line 61: "Try creating" -> "Create"

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -102,20 +115,33 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
     repartitionedIndexDataFrame.write
       .saveWithBuckets(
         repartitionedIndexDataFrame,
         indexDataPath.toString,
         numBuckets,
-        indexColNames)
+        resolvedIndexedColumns)
+  }
+
+  private def resolve(spark: SparkSession, dst: String, srcs: Array[String]): String = {
```
apoorvedave1, line 139: remove this resolve and use IndexNameUtils.resolve instead

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -49,8 +51,19 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
 
     val signatureProvider = LogicalPlanSignatureProvider.create()
 
+    val (resolvedIndexedColumns, resolvedIncludedColumns) = {
```
apoorvedave1, line 54: add comments what "resolved" means

src/main/scala/com/microsoft/hyperspace/util/IndexNameUtils.scala
```diff
@@ -16,6 +16,9 @@
 
 package com.microsoft.hyperspace.util
 
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.analysis.Resolver
+
 /**
  * [[IndexNameUtils]] provides utility functions to format the index name as need.
```
apoorvedave1, line 23: rename class to IndexConfigUtils

src/main/scala/com/microsoft/hyperspace/util/IndexNameUtils.scala
```diff
@@ -31,4 +34,44 @@ object IndexNameUtils {
   def normalizeIndexName(indexName: String): String = {
     indexName.trim.replaceAll("\\s+", "_")
   }
+
+  def resolve(spark: SparkSession, requiredString: String, availableStrings: String): Boolean = {
```
apoorvedave1, line 38: rename "resolve" to "isResolved"

src/main/scala/com/microsoft/hyperspace/util/IndexNameUtils.scala
```diff
@@ -31,4 +34,44 @@ object IndexNameUtils {
...
+  }
+
+  def resolve(
+      spark: SparkSession,
+      requiredStrings: Iterable[String],
+      availableStrings: Iterable[String]): Boolean = {
+    requiredStrings.forall(resolve(spark, _, availableStrings))
+  }
+
+  def resolvedString(
```
apoorvedave1, line 57: rename "resolvedString" to "resolve"

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -586,10 +593,13 @@ object JoinIndexRule extends Rule[LogicalPlan] with Logging {
       lIndex: IndexLogEntry,
       rIndex: IndexLogEntry,
       columnMapping: Map[String, String]): Boolean = {
-    require(columnMapping.keys.toSet.equals(lIndex.indexedColumns.toSet))
-    require(columnMapping.values.toSet.equals(rIndex.indexedColumns.toSet))
+    val spark = SparkSession.getActiveSession.get
```
apoorvedave1, line 596: add a def spark or def resolver to deduplicate "get spark, resolve" parts in the code

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,137 @@
...
+   * @param spark Spark session.
+   * @param requiredStrings List of strings to resolve.
+   * @param availableStrings List of available strings to resolve from.
+   * @return Sequence of optional values of resolved strings or None's.
+   */
+  def resolve(
+      spark: SparkSession,
+      requiredStrings: Iterable[String],
+      availableStrings: Iterable[String]): Seq[Option[String]] = {
+    requiredStrings.map(resolve(spark, _, availableStrings)).toSeq
```
imback82, line 135: This API is a bit confusing since the behavior is different from
```
  def isResolved(
      spark: SparkSession,
      requiredStrings: Iterable[String],
      availableStrings: Iterable[String]): Boolean = {
    requiredStrings.forall(isResolved(spark, _, availableStrings))
  }
```
Can we just return `Option[Seq[String]]`? Do we have a case where we want to match partially?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,137 @@
...
+   *
+   * @param spark Spark session.
+   * @param requiredStrings List of strings to resolve.
+   * @param availableStrings List of available strings to resolve from.
+   * @return Sequence of optional values of resolved strings or None's.
+   */
+  def resolve(
+      spark: SparkSession,
+      requiredStrings: Iterable[String],
+      availableStrings: Iterable[String]): Seq[Option[String]] = {
```
imback82, line 134: Can we just use `Seq` instead of `Iterable` to match the return type as well? Was there a case where `Iterable` was required?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,137 @@
...
+object ResolverUtils {
+
+  /**
+   * Normalize the index name by trimming space(s) at both ends, and replacing internal space(s)
+   * with an underscore.
+   *
+   * @param indexName the name of index passed in by user.
+   * @return a normalized index name.
+   */
+  def normalizeIndexName(indexName: String): String = {
```
imback82, line 36: This function seems to be out of place?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -518,8 +514,12 @@ object JoinIndexRule extends Rule[LogicalPlan] with Logging {
       allRequiredCols: Seq[String]): Seq[IndexLogEntry] = {
     indexes.filter { idx =>
       val allCols = idx.indexedColumns ++ idx.includedColumns
-      requiredIndexCols.toSet.equals(idx.indexedColumns.toSet) &&
-      allRequiredCols.forall(allCols.contains)
+
+      // All required index columns should match one-to-one with all indexed columns and
+      // vice-versa. All required columns must be present in the available index columns.
+      isResolved(spark, requiredIndexCols, idx.indexedColumns) &&
+      requiredIndexCols.distinct.size == idx.indexedColumns.distinct.size &&
```
imback82, line 521: distinct needs to take care of case sensitivity or no?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * Resolve two strings based on spark's resolver. Returns true if both are resolved.
+   *
+   * @param spark Spark session.
+   * @param requiredString First string.
+   * @param availableString Second string.
+   * @return True if both are equivalent based on the current spark resolver.
+   */
+  def isResolved(
+      spark: SparkSession,
```
imback82, line 49: should this take in `Resolver`?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,137 @@
...
+   * true if all the requiredStrings can be resolved with some string from availableStrings.
+   * Returns false otherwise.
+   *
+   * @param spark Spark session.
+   * @param requiredStrings List of strings to resolve.
+   * @param availableStrings List of availble strings to resolve from.
+   * @return True if every one of requiredStrings can be resolved from availableStrings. False
+   *         otherwise.
+   */
+  def isResolved(
```
imback82, line 83: how about `canResolve`? or just unify the API so that people can just use `resolve(...).isDefined`? (after addressing the different behavior below)

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: it should. Thanks for the catch. Let me fix that

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: thanks @imback82 I will remove this. Seems like it is not used anywhere

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
imback82: meaning we don't have a test coverage for this?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: yeah In `JoinIndexRule` there's a key-value Map and I need to `resolve` on map keys. If I use iterable, I can avoid `map.keys.toSeq`. Or I can call resolve(spark, map.keys.toSeq, ...)

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: done, thanks @imback82

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: do you mean for all functions? I have changed the innermost function to take Resolver. Let me know if the intention was for all functions

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: done, thanks @imback82

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: changed to Seq

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: yeah, I will add a test where join condition is
```sql
T1.a = T2.a and T1.A = T2.B
```
where `T1.a` is the same as `T1.A`
and it should not pick index. WIth the current (bug) logic, this case will pick any T1 index with 2 columns, one being `a`

imback82: cool thanks!

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: @imback82 do you mean for all functions? I have changed the innermost function to take Resolver. Let me know if the intention was for all functions

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -49,8 +51,18 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
 
     val signatureProvider = LogicalPlanSignatureProvider.create()
 
+    // Resolve the passed column names with existing column names from the dataframe if possible.
+    val (resolvedIndexedColumns, resolvedIncludedColumns) = {
+      try {
+        // Try creating log entry with resolved column names.
```
imback82, line 57: I think we should always use the casing that the user provided.

src/test/scala/com/microsoft/hyperspace/index/HyperspaceSuite.scala
```diff
@@ -72,4 +72,14 @@ trait HyperspaceSuite extends SparkFunSuite with SparkInvolvedSuite {
       }
     }
   }
+
+  protected def withSparkConf(confName: String, confValue: Any)(f: => Unit): Unit = {
```
AFFogarty, line 76: Might be worth adding a comment that this isn't thread-safe?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
apoorvedave1: changed this behavior to not handle unresolved case. if `resolve` fails, it will directly throw exception. No "CREATING" file gets created at this point.

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -68,8 +68,8 @@ class CreateAction(
     val validColumnNames = schema.fieldNames
     val indexedColumns = indexConfig.indexedColumns
     val includedColumns = indexConfig.includedColumns
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @AFFogarty , done

src/test/scala/com/microsoft/hyperspace/index/HyperspaceSuite.scala
apoorvedave1: We can add it, but looks like for testing, Spark doesn't explicitly mention this. ([e.g. here](https://github.com/apache/spark/blob/7955b3962ac46b89564e0613db7bea98a1478bf2/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/plans/PlanTest.scala#L182)))

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -102,20 +106,42 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    val resolvedIncludedColumns =
+      ResolverUtils.resolve(spark, indexConfig.includedColumns, dfColumnNames)
+
+    val allColumns = resolvedIndexedColumns ++ resolvedIncludedColumns
+
+    (resolvedIndexedColumns, resolvedIncludedColumns) match {
+      case (Some(indexed), Some(included)) => (indexed, included)
+      case _ =>
+        throw HyperspaceException(
+          s"Unexpected Exception: Some of the columns from $allColumns could not be resolved " +
```
imback82, line 143: Can we pinpoint the exact column that failed? Since this is during create, it will be helpful for the user.

apoorvedave1: thanks @imback82, added

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,75 @@
...
+ * [[ResolverUtils]] provides utility functions to resolve strings based on spark's resolver.
+ */
+object ResolverUtils {
+
+  /**
+   * Return available string if required string can be resolved with it, based on spark resolver.
+   *
+   * @param resolver Resolver.
+   * @param requiredString The string that requires resolution.
+   * @param availableString Available list of strings to resolve from.
```
pirz, line 32: is this a `list` or a single String?

apoorvedave1: it's single. Thanks for the catch @pirz

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * Finds all resolved strings for requiredStrings, from the list of availableStrings. Returns a
+   * sequence of Optional values for matches, None for unmatched strings.
+   *
+   * @param spark Spark session.
+   * @param requiredStrings List of strings to resolve.
+   * @param availableStrings All available strings to resolve from.
+   * @return Sequence of optional values of resolved strings or None's.
```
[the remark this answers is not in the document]
apoorvedave1: Thanks for the catch @sezruby . Fixed

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,75 @@
...
+   * @param resolver Resolver.
+   * @param requiredString The string that requires resolution.
+   * @param availableString Available string to resolve from.
+   * @return Optional available string if resolution is successful, else None
+   */
+  def resolve(
+      resolver: Resolver,
+      requiredString: String,
+      availableString: String): Option[String] = {
+    if (resolver(requiredString, availableString)) Some(availableString) else None
```
pirz, line 39: I am just curious if this one-liner is consistent with our formatting rules. I thought we `if` always needs `{}`.

apoorvedave1: thanks @pirz , I had my doubts with this one too. In [databricks scala guide](https://github.com/databricks/scala-style-guide#curly-braces), I saw that as long as everything fits in a line, it's acceptable. Our scala-style also did not complain for this so I kept it.

src/test/scala/com/microsoft/hyperspace/index/rules/JoinIndexRuleTest.scala
```diff
@@ -320,6 +332,23 @@ class JoinIndexRuleTest extends HyperspaceRuleTestSuite with SQLHelper {
...
+    val joinCondition = And(EqualTo(t1c1, t2c1), EqualTo(t1c1Caps, t2c1Caps))
+    val originalPlan = Join(t1ProjectNode, t2ProjectNode, JoinType("inner"), Some(joinCondition))
+    val updatedPlan = JoinIndexRule(originalPlan)
+    assert(!updatedPlan.equals(originalPlan))
+
+    val indexPaths = Seq(getIndexDataFilesPath("t1i1"), getIndexDataFilesPath("t2i1"))
+    verifyUpdatedIndex(originalPlan, updatedPlan, indexPaths)
+  }
+
+
```
imback82, line 351: remove empty line

src/test/scala/com/microsoft/hyperspace/index/rules/JoinIndexRuleTest.scala
```diff
@@ -320,6 +332,23 @@ class JoinIndexRuleTest extends HyperspaceRuleTestSuite with SQLHelper {
...
+    val t1ProjectNode = Project(Seq(t1c1, t1c3), t1FilterNode)
+    val t2ProjectNode = Project(Seq(t2c1, t2c3), t2FilterNode)
+
+    val t1c1Caps = t1c1.withName("T1C1")
+    val t2c1Caps = t2c1.withName("T2C1")
+
+    val joinCondition = And(EqualTo(t1c1, t2c1), EqualTo(t1c1Caps, t2c1Caps))
+    val originalPlan = Join(t1ProjectNode, t2ProjectNode, JoinType("inner"), Some(joinCondition))
+    val updatedPlan = JoinIndexRule(originalPlan)
+    assert(!updatedPlan.equals(originalPlan))
```
imback82, line 345: this contradicts with the test description?

src/test/scala/com/microsoft/hyperspace/index/CreateIndexTests.scala
```diff
@@ -87,6 +85,27 @@ class CreateIndexTests extends SparkFunSuite with SparkInvolvedSuite {
...
+    assert(
+      indexes.head.getAs[WrappedArray[String]]("indexedColumns").head.equals("Query"),
+      "Indexed columns with wrong case are stored in metadata")
+    assert(
+      indexes.head.getAs[WrappedArray[String]]("includedColumns").head.equals("imprs"),
+      "Included columns with wrong case are stored in metadata")
+  }
+
+  test("Index creation fails with columns of different case if case-sensitivity is true.") {
+    withSparkConf("spark.sql.caseSensitive", true) {
```
imback82, line 101: Can we use withSQLConf? https://github.com/microsoft/hyperspace/blob/48b3332bf5aaca0c2daf890c9156113b62a65843/src/test/scala/com/microsoft/hyperspace/index/rules/JoinIndexRuleTest.scala#L115

src/test/scala/com/microsoft/hyperspace/index/CreateIndexTests.scala
```diff
@@ -87,6 +85,27 @@ class CreateIndexTests extends SparkFunSuite with SparkInvolvedSuite {
     assert(exception.getMessage.contains("Index config is not applicable to dataframe schema"))
   }
 
+  test("Index creation passes with columns of different case if case-sensitivity is false.") {
+    hyperspace.createIndex(df, IndexConfig("index1", Seq("qUeRy"), Seq("ImpRS")))
+    val indexes = hyperspace.indexes.where(s"""name = "${indexConfig1.indexName}" """)
```
imback82, line 90: nit: does the following work?
```
val indexes = hyperspace.indexes.where(s"name = '${indexConfig1.indexName}'")
```

src/test/scala/com/microsoft/hyperspace/index/CreateIndexTests.scala
```diff
@@ -87,6 +85,27 @@ class CreateIndexTests extends SparkFunSuite with SparkInvolvedSuite {
...
+    assert(indexes.count == 1)
+    assert(
+      indexes.head.getAs[WrappedArray[String]]("indexedColumns").head.equals("Query"),
```
imback82, line 93: nit: `== "Query"` is easier to read, esp. when you have a long chain.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -91,6 +91,18 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite {
...
 
+  test("E2E test for case insensitive filter query utilizing indexes.") {
+    val df = spark.read.parquet(sampleParquetDataLocation)
+    val indexConfig = IndexConfig("filterIndex", Seq("C3"), Seq("C1"))
+
+    hyperspace.createIndex(df, indexConfig)
+
+    def query(): DataFrame = df.filter("C3 == 'facebook'").select("C3", "c1")
+
+    // verify if case-insensitive index works with case-insensitive query
```
imback82, line 102: nit: `// Verify if case-insensitive index works with case-insensitive query.`

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,75 @@
...
+  def resolve(
+      spark: SparkSession,
+      requiredString: String,
+      availableStrings: Seq[String]): Option[String] = {
+    availableStrings.find(resolve(spark.sessionState.conf.resolver, requiredString, _).isDefined)
+  }
+
+  /**
+   * Finds all resolved strings for requiredStrings, from the list of availableStrings. Returns a
+   * sequence of Optional values for matches, None for unmatched strings.
```
imback82, line 61: This sounds like you are returning `Seq[Option[String]]`

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -0,0 +1,75 @@
...
+   * @param requiredStrings List of strings to resolve.
+   * @param availableStrings All available strings to resolve from.
+   * @return Optional Seq of resolved strings if all required strings are resolved. Else, None.
+   */
+  def resolve(
+      spark: SparkSession,
+      requiredStrings: Seq[String],
+      availableStrings: Seq[String]): Option[Seq[String]] = {
+    val resolved = requiredStrings.map(resolve(spark, _, availableStrings))
+    if (resolved.forall(_.nonEmpty)) Some(resolved.map(_.get)) else None
```
imback82, line 73: nit: Will the following work? (looks like we are iterating seq three times):
```scala
Some(requiredStrings.map(resolve(spark, _, availableStrings).getOrElse { return None }))
```

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -104,20 +107,43 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    val resolvedIncludedColumns = ResolverUtils.resolve(spark, includedColumns, dfColumnNames)
+
+    (resolvedIndexedColumns, resolvedIncludedColumns) match {
+      case (Some(indexed), Some(included)) => (indexed, included)
+      case _ =>
+        val unresolvedColumns = (indexedColumns ++ includedColumns)
+          .map(c => (c, ResolverUtils.resolve(spark, c, dfColumnNames)))
+          .collect { case c if c._2.isEmpty => c._1 }
+        throw HyperspaceException(
+          s"Unexpected Exception: Columns $unresolvedColumns could not be resolved " +
```
imback82, line 145: nit: we can just remove "Unexpected Exception:"

src/test/scala/com/microsoft/hyperspace/index/rules/JoinIndexRuleTest.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/CreateIndexTests.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/CreateIndexTests.scala
apoorvedave1: yes it works. Fixed here and other places.thanks @imback82 ,

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , fixed

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: thanks @imback82 , fixed

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
apoorvedave1: thanks @imback82 , fixed

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
apoorvedave1: Thanks, this is much better. Fixed, @imback82

src/test/scala/com/microsoft/hyperspace/index/CreateIndexTests.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/rules/JoinIndexRuleTest.scala
apoorvedave1: thanks @imback82 , fixed

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -17,13 +17,14 @@
...
+import scala.util.Try
 
 import org.apache.hadoop.fs.Path
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.analysis.CleanupAliases
 import org.apache.spark.sql.catalyst.catalog.BucketSpec
-import org.apache.spark.sql.catalyst.expressions.{And, Attribute, AttributeReference, EqualTo, Expression}
-import org.apache.spark.sql.catalyst.plans.logical.{Join, LogicalPlan}
+import org.apache.spark.sql.catalyst.expressions.{Alias, And, AttributeReference, AttributeSet, EqualTo, Expression}
+import org.apache.spark.sql.catalyst.plans.logical.{Join, LogicalPlan, Project}
```
imback82, line 27: is `Project` used? same for `AttributeSet` above

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -234,9 +236,6 @@ object JoinIndexRule
    *    WHERE newT1.aliasCol = T2.b
    * Here, aliasCol is not directly from the base relation T1
    *
-   * TODO: add alias resolver for supporting aliases in join condition. Until then,
```
imback82, line 237: which change is handling the aliases?

imback82: add a test if this is fixed?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -287,35 +286,45 @@ object JoinIndexRule
...
       r: LogicalPlan,
       condition: Expression): Boolean = {
-
+    val conditions = extractConditions(condition)
     // Output attributes from base relations. Join condition attributes must belong to these
-    // attributes
-    val lBaseAttrs = l.collectLeaves().filter(_.isInstanceOf[LogicalRelation]).flatMap(_.output)
-    val rBaseAttrs = r.collectLeaves().filter(_.isInstanceOf[LogicalRelation]).flatMap(_.output)
+    // attributes. We work on canonicalized forms to make sure we support case-sensitivity.
+    val lBaseAttrs = l
```
imback82, line 292: nit: create a nested `def baseAttrs(p: LogicalPlan):`?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -287,35 +286,45 @@ object JoinIndexRule
       l: LogicalPlan,
       r: LogicalPlan,
       condition: Expression): Boolean = {
-
+    val conditions = extractConditions(condition)
```
imback82, line 289: can we define this close to where it's being used?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
imback82: hmm. looks like something similar is being used below. Maybe we can create
```scala
def relationOutputs(l: logicalPlan): ??? = {
  l.collectLeaves().filter(_.isInstanceOf[LogicalRelation]).flatMap(_.output)
}
```
so that this can be reused?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -340,17 +349,24 @@ object JoinIndexRule
...
+    // Map of left resolved columns with their corresponding right resolved
+    // columns from condition.
+    val lRMap = getLRColumnMapping(lBaseAttrs, rBaseAttrs, joinCondition)
+    val lRequiredIndexedCols = lRMap.keys.toSeq
+    val rRequiredIndexedCols = lRMap.values.toSeq
 
-    val lRequiredAllCols = allRequiredCols(left)
-    val rRequiredAllCols = allRequiredCols(right)
+    // All required columns resolved with base relation.
+    val lRequiredAllCols: Seq[String] = resolve(spark, allRequiredCols(left), lBaseAttrs).get
```
imback82, line 364: nit: remove type

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -558,9 +561,6 @@ object JoinIndexRule
       lIndex: IndexLogEntry,
       rIndex: IndexLogEntry,
       columnMapping: Map[String, String]): Boolean = {
-    require(columnMapping.keys.toSet.equals(lIndex.indexedColumns.toSet))
```
imback82, line 561: any reason for removing this?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -359,27 +375,6 @@ object JoinIndexRule
...
-   * Returns list of column names which must be in the indexed columns config of a selected
-   * index.
-   *
-   * @param plan logical plan
-   * @param condition join condition used to pull out required indexed columns in the plan
-   * @return list of column names from this plan which must be part of indexed columns in a chosen
-   *         index
-   */
-  private def requiredIndexedCols(plan: LogicalPlan, condition: Expression): Seq[String] = {
-    val cleanedPlan = CleanupAliases(plan)
```
imback82, line 372: no need to clean up aliases now?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -287,35 +286,45 @@ object JoinIndexRule
...
+    // This ensures uniqueness of attributes in case of case-insensitive system. E.g. We want to
+    // store just one copy of column 'A' when join condition contains column 'A' as well as 'a'.
     val attrMap = new mutable.HashMap[Expression, Expression]()
 
-    extractConditions(condition).forall {
+    conditions.forall {
       case EqualTo(c1, c2) =>
-        // c1 and c2 should belong to l and r respectively, or r and l respectively.
-        if (!fromDifferentBaseRelations(c1, c2)) {
+        val (c1Canonicalized, c2Canonicalized) = (c1.canonicalized, c2.canonicalized)
```
imback82, line 316: How about something like:
```
  case EqualTo(e1, e2) =>
    val (c1, c2) = (e1.canonicalized, e2.canonicalized)
```
to minimize the diff and to align comments as well.

imback82:
@apoorvedave1 Can you resolve conflicts please?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: removed Project. AttributeSet is used in scaladoc as [[AttributeSet]]

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: I thought this was the case https://github.com/microsoft/hyperspace/issues/25 and we decided not to fix it until 3.0?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: Removed it for https://github.com/microsoft/hyperspace/issues/25. Let me revert alias related changes and take it in a separate PR

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
imback82: Got it.

In this case, can we add something like `The following scenario will be supported in Spark 3.0 (SPARK-30298)` for the above block (starting with `This will not be true`)?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
imback82: Sounds good. Let's minimize the scope for this PR.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: Let me revert alias related changes and take it in a separate PR

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: thanks @imback82, changed as suggested

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: realized removing this was a mistake. reverting.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: based on other discussions and closed PR #102 , aliases aren't supported in Join rule. So keeping this change as discussed

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -104,20 +107,43 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+
+    (resolvedIndexedColumns, resolvedIncludedColumns) match {
+      case (Some(indexed), Some(included)) => (indexed, included)
+      case _ =>
+        val unresolvedColumns = (indexedColumns ++ includedColumns)
+          .map(c => (c, ResolverUtils.resolve(spark, c, dfColumnNames)))
+          .collect { case c if c._2.isEmpty => c._1 }
+        throw HyperspaceException(
+          s"Columns $unresolvedColumns could not be resolved " +
+            s"from available source columns $dfColumnNames")
```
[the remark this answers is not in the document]
apoorvedave1: there is no e2e test for this case because wrong column names will fail in `validate` phase most of the time. So this is more like an unexpected exception or illegal state exception. Should I try to add a test case?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -287,32 +289,33 @@ object JoinIndexRule
...
+        // Check 1: c1 and c2 should belong to l and r respectively, or r and l respectively.
         if (!fromDifferentBaseRelations(c1, c2)) {
           return false
         }
-        // The following validates that c1 is compared only against c2 and vice versa
+        // Check 2: c1 is compared only against c2 and vice versa.
         if (attrMap.contains(c1) && attrMap.contains(c2)) {
-          attrMap(c1).equals(c2) && attrMap(c2).equals(c1)
+          attrMap(c1).equals(c2) &&
+          attrMap(c2).equals(c1)
```
imback82, line 318: nit: this change is from auto-format?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -91,6 +91,18 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite {
     verifyIndexUsage(query, Seq(getIndexFilesPath(indexConfig.indexName)))
   }
 
+  test("E2E test for case insensitive filter query utilizing indexes.") {
```
imback82, line 94: Can you add a test where changing the sensitivity config changes the behavior?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: yeah looks like. Will revert

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , added

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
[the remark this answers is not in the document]
apoorvedave1: thanks @imback82 for the suggestion, added test case and fixed the output line as well.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -127,6 +158,27 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite {
...
+    def query(): DataFrame = {
+      leftDf.join(rightDf, leftDf("c3") === rightDf("C3")).select(leftDf("C1"), rightDf("c4"))
+    }
+
+    verifyIndexUsage(
+      query,
+      Seq(
+        getIndexFilesPath(leftDfIndexConfig.indexName),
+        getIndexFilesPath(rightDfIndexConfig.indexName)))
+
```
imback82, line 179: nit: empty line

apoorvedave1: thanks @imback82 , removed

PR #82, opened 2020-07-07
Creating index fails for csv data where data is materialized before index creation

apoorvedave1:
Fixes #81 

### What changes were proposed in this pull request?

Fixes bug with creating index on csv data where data is materialized beforehand

### Why are the changes needed?

bug occurs when data is materialized, it materializes some internal objects which can't be serialized. This fails during df serialization phase of index creation.
the fix uses wrappers for un-serializable objects
### Does this PR introduce _any_ user-facing change?
no

### How was this patch tested?

manually tested and unit tests added

PR #83, opened 2020-07-08
Creating index fails for csv data where data is materialized before index creation

### LINKED ISSUE #81: Hyperspace Create Index Fails for csv if data is materialized before
### Describe the issue
Probable root cause of https://github.com/dotnet/spark/issues/580
Hyperspace index with dataset (CSV) and fails during the creation if data is materialized just before create index

### To Reproduce

I could repro the issue only if I call df.show() before hs.createIndex(). If I don't call df.show(), it works fine and creates index without issue.
```
object Test extends App {
  val spark: SparkSession = ...
  
  val df = spark.read.csv("csv/*/*")
  df.show()                            // commenting this line eliminates the issue
  hs.createIndex(df, new IndexConfig("ic2", Seq("_c0"), Seq()))
}
```
### Expected behavior

New index created and app will show in the log.
### Additional Context

```
com.esotericsoftware.kryo.KryoException: java.util.ConcurrentModificationException
Serialization trace:
classes (sun.misc.Launcher$AppClassLoader)
classloader (java.security.ProtectionDomain)
context (java.security.AccessControlContext)
acc (org.apache.spark.util.MutableURLClassLoader)
classLoader (org.apache.hadoop.conf.Configuration)
conf (org.apache.hadoop.io.compress.BZip2Codec)
codecs (org.apache.hadoop.io.compress.CompressionCodecFactory)
codecFactory (org.apache.spark.sql.execution.datasources.csv.CSVFileFormat)
fileFormat (com.microsoft.hyperspace.index.serde.package$HadoopFsRelationWrapper)
relation (com.microsoft.hyperspace.index.serde.package$LogicalRelationWrapper)
at com.esotericsoftware.kryo.serializers.ObjectField.write(ObjectField.java:101)
at com.esotericsoftware.kryo.serializers.FieldSerializer.write(FieldSerializer.java:508)
at com.esotericsoftware.kryo.Kryo.writeObject(Kryo.java:575)
at com.esotericsoftware.kryo.serializers.ObjectField.write(ObjectField.java:79)
...
Caused by: java.util.ConcurrentModificationException
at java.util.Vector$Itr.checkForComodification(Unknown Source)
at java.util.Vector$Itr.next(Unknown Source)
at com.esotericsoftware.kryo.serializers.CollectionSerializer.write(CollectionSerializer.java:99)
at com.esotericsoftware.kryo.serializers.CollectionSerializer.write(CollectionSerializer.java:40)
at com.esotericsoftware.kryo.Kryo.writeObject(Kryo.java:575)
at com.esotericsoftware.kryo.serializers.ObjectField.write(ObjectField.java:79)
```

apoorvedave1:
Same as PR https://github.com/microsoft/hyperspace/pull/82 with cleaner commit history.
Fixes #81

Fixes bug with creating index on csv data where data is materialized beforehand

bug occurs when data is materialized, it materializes some internal objects which can't be serialized. This fails during df serialization phase of index creation.
the fix uses wrappers for un-serializable objects

no

manually tested and unit tests added

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
@@ -26,6 +26,7 @@ import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.execution.datasources._
 import org.apache.spark.sql.execution.datasources.csv.CSVFileFormat
 import org.apache.spark.sql.execution.datasources.json.JsonFileFormat
+import org.apache.spark.sql.execution.datasources.orc.OrcFileFormat
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
 
 import com.microsoft.hyperspace.HyperspaceException
@@ -225,7 +226,7 @@ object LogicalPlanSerDeUtils {
     def unapply(fileFormat: FileFormat): Option[FileFormat] = fileFormat match {
       case CSVFileFormatWrapper => Some(new CSVFileFormat)
       case JsonFileFormatWrapper => Some(new JsonFileFormat)
-      case f: ParquetFileFormat => Some(f)
+      case _: ParquetFileFormat | _: OrcFileFormat => Some(fileFormat)
       case _ => throw HyperspaceException("Unsupported File Format found.")
     }
   }
@@ -234,7 +235,7 @@ object LogicalPlanSerDeUtils {
     def unapply(fileFormat: FileFormat): Option[FileFormat] = fileFormat match {
       case _: CSVFileFormat => Some(CSVFileFormatWrapper)
       case _: JsonFileFormat => Some(JsonFileFormatWrapper)
-      case f: ParquetFileFormat => Some(f)
+      case _: ParquetFileFormat | _: OrcFileFormat => Some(fileFormat)
       case _ => throw HyperspaceException("Unsupported File Format found.")
     }
   }
--- a/src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
@@ -24,6 +24,8 @@ import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.expressions.{Exists, InSubquery, ListQuery, ScalarSubquery, ScalaUDF}
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.execution.datasources._
+import org.apache.spark.sql.execution.datasources.csv.CSVFileFormat
+import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
 
 /**
  * Utility functions for logical plan serialization/deserialization.
@@ -119,7 +121,7 @@ object LogicalPlanSerDeUtils {
             partitionSchema,
             dataSchema,
             bucketSpec,
-            fileFormat,
+            fileFormat: ParquetFileFormat,
             options),
           output,
           catalogTable,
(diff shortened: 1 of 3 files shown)


src/test/scala/com/microsoft/hyperspace/index/LogicalPlanSerDeTests.scala
```diff
@@ -73,6 +78,50 @@ class LogicalPlanSerDeTests extends SparkFunSuite with SparkInvolvedSuite {
     verifyPlanSerde(scanNode, "hadoopFsRelation.plan")
   }
 
+  private def verifyFileFormat(
+      fileFormat: FileFormat,
+      expectSerializationError: Boolean): Unit = {
+    val relation: HadoopFsRelation =
```
imback82, line 84: can we remove the type to be consistent?

src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
```diff
@@ -215,4 +221,22 @@ object LogicalPlanSerDeUtils {
             e.query.childOutputs))
     }
   }
+
+  object ExtractFileFormat {
+    def unapply(fileFormat: FileFormat): Option[FileFormat] = fileFormat match {
+      case CSVFileFormatWrapper => Some(new CSVFileFormat)
+      case JsonFileFormatWrapper => Some(new JsonFileFormat)
+      case _: ParquetFileFormat | _: OrcFileFormat => Some(fileFormat)
+      case f => throw HyperspaceException(s"Unsupported File Format found: ${f.toString}.")
```
imback82, line 230: usually `other` is used for the variable name (in Spark, at least)

imback82: also, either `FileFormat` or `file format`

src/test/scala/com/microsoft/hyperspace/index/LogicalPlanSerDeTests.scala
apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
apoorvedave1: thanks @imback82 , done

imback82:
@apoorvedave1, this is not backward compatible right since we have new wrappers?

apoorvedave1:
yes that is correct

imback82:
Can you update the PR description (Does this PR introduce any user-facing change?)

apoorvedave1:
done

imback82:
What I meant was if we can call `refresh` on the index created from v0.1 (for csv) since we introduced new "wrapper" class.

apoorvedave1:
@imback82 thanks for explaining.
Yes this is breaking change. If index was created with v0.1 (for csv), and if we call `refresh` v0.2, refresh will fail with "Unsupported file format found: CSV" error when trying to deserialize CSVFileFormat.
On the other hand, we can make it backward compatible with a minor fix: If serialized "CSVFileFormat found, return CSVFileformat" when deserializing. i.e.
```
object ExtractFileFormat {
    def unapply(fileFormat: FileFormat): Option[FileFormat] = fileFormat match {
      case CSVFileFormatWrapper | _: CSVFileFormat => Some(new CSVFileFormat)
      case JsonFileFormatWrapper => Some(new JsonFileFormat)
      case _: ParquetFileFormat | _: OrcFileFormat => Some(fileFormat)
      case other =>
        throw HyperspaceException(s"Unsupported file format found: ${other.toString}.")
      case f => Some(f)
    }
  }
```
I tested this change and it can support backward compatibility. Let me know if this is OK.

imback82:
Yea, this approach sounds reasonable.

PR #85, opened 2020-07-09
[WIP] Support for Spark 3.0

imback82:

This PR proposes to support Spark 3.0. I will share more about branching strategies, etc. as a separate PR.

To support Spark 3.0.

Yes. The user needs to consume this library along with Spark 3.0. This is not compatible with Scala 2.11 any longer. 

Updated existing tests.


src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -60,7 +61,31 @@ object LogicalPlanFingerprint {
 // IndexLogEntry-specific SparkPlan that represents the source plan.
 case class SparkPlan(properties: SparkPlan.Properties) {
   val kind = "Spark"
+
+  override def equals(o: Any): Boolean = o match {
+    case that: SparkPlan =>
+      // When a logical plan is serialized, some of the fields can be 'lazy' (e.g.,
+      // Metadata._hashCode), meaning that the same logical plan can result in different serialized
+      // stings. Thus, the serialized strings are deserialized to logical plans and compared.
+      val isPlanEqual = if (!properties.rawPlan.isEmpty) {
```
imback82, line 70: I will update the tests to put in a valid serialized string instead of doing this check.

sezruby:
Do you have any plan for code management of both spark 2.4.6 and 3.0.0?
I think we should manage different codebases for spark2.4.6 and 3.0.0 and provide build.sbt option accordingly.

imback82:
The current plan is to have 3.0 support in `master` and create a `branch-2.4` for 2.4 support. The development process is to create a PR against master and merge it to both `master`/`branch-2.4`.

sezruby:
That should be fine for developers :)

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: this is no longer relevant

build.sbt
```diff
@@ -16,15 +16,9 @@
...
+sparkVersion := "3.0.0"
 
-lazy val scala212 = "2.12.8"
-lazy val scala211 = "2.11.12"
-lazy val supportedScalaVersions = List(scala212, scala211)
-
-scalaVersion := scala212
-
-crossScalaVersions := supportedScalaVersions
+scalaVersion := "2.12.10"
```
pirz, line 21: Does it mean that switching to Spark 3.0, Hyperspace only runs on Scala 2.12?

imback82: Yes. Spark 3.0 is compiled only with Scala 2.12.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -46,8 +46,8 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
   override def allFiles: Seq[FileStatus] = plan.relation match {
     case HadoopFsRelation(location: TahoeLogFileIndex, _, _, _, _, _) =>
       location
-        .getSnapshot(stalenessAcceptable = false)
-        .filesForScan(projection = Nil, location.partitionFilters, keepStats = false)
+        .getSnapshot
+        .filesForScan(projection = Nil, location.partitionFilters)
```
imback82, line 50: @sezruby in Delta 8.0, there were some changes in the API. Could you confirm if these changes are fine?

imback82:
@andrei-ionescu In Spark 3.0, `DataSourceV2Relation` signature has changed and takes in `Table` now. Could you help fixing the it?
```

[error] /home/vsts/work/1/s/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala:58: value source is not a member of org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
[error]   override def allFiles: Seq[FileStatus] = plan.source match {
[error]                                                 ^
[error] /home/vsts/work/1/s/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala:66: value source is not a member of org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
[error]   override def partitionBasePath: Option[String] = plan.source match {
[error]                                                         ^
[error] /home/vsts/work/1/s/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala:85: value source is not a member of org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
[error]     plan.source match {
[error]          ^
[error] /home/vsts/work/1/s/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala:87: not found: type DataSourceOptions
[error]         val dsOpts = new DataSourceOptions(plan.options.asJava)
[error]                          ^

```

imback82:
closed by #421

PR #86, opened 2020-07-09
Backward compatibility support for json and csv index creation

apoorvedave1:
This is an extension to PR https://github.com/microsoft/hyperspace/pull/83
The previous PR broke backward compatibility such that if an index was created with hyperspace v 0.1, it could not be `refresh`ed when above mentioned PR was merged.
This PR improves on previous PR and fixes CSV bugs as well as supports backward compatibility of `refresh`ing indexes created with v0.1

Backward compatibility between v0.2-SNAPSHOT and v0.1 of hyperspace fails with PR https://github.com/microsoft/hyperspace/pull/83. This PR is a fix for that.

No

manually tested the following
- created index with hs v0.1
- tried refreshing index with hs v0.2-snapshot >> Fails with "Unsupported file format found: CSV"
- added the fix and tried refreshing again >> refresh works as expected

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
@@ -224,10 +224,11 @@ object LogicalPlanSerDeUtils {
 
   object ExtractFileFormat {
     def unapply(fileFormat: FileFormat): Option[FileFormat] = fileFormat match {
+      case CSVFileFormatWrapper => Some(new CSVFileFormat)
+      case JsonFileFormatWrapper => Some(new JsonFileFormat)
       // Add CSVFileFormat and JsonFileFormat in below checks for backward compatibility.
-      case CSVFileFormatWrapper | _: CSVFileFormat => Some(new CSVFileFormat)
-      case JsonFileFormatWrapper | _: JsonFileFormat => Some(new JsonFileFormat)
-      case _: ParquetFileFormat | _: OrcFileFormat => Some(fileFormat)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/microsoft/hyperspace/index/serde/LogicalPlanSerDeUtils.scala
```diff
@@ -224,8 +224,9 @@ object LogicalPlanSerDeUtils {
 
   object ExtractFileFormat {
     def unapply(fileFormat: FileFormat): Option[FileFormat] = fileFormat match {
-      case CSVFileFormatWrapper => Some(new CSVFileFormat)
-      case JsonFileFormatWrapper => Some(new JsonFileFormat)
+      // Add CSVFileFormat and JsonFileFormat in below checks for backward compatibility.
+      case CSVFileFormatWrapper | _: CSVFileFormat => Some(new CSVFileFormat)
```
imback82, line 228: remove `_`?

imback82: or
```scala
case CSVFileFormatWrapper => Some(new CSVFileFormat)
case JsonFileFormatWrapper => Some(new JsonFileFormat)
case ParquetFileFormat | OrcFileFormat | CSVFileFormat | JsonFileFormat => Some(fileFormat)
```

apoorvedave1: `_` can't be removed for classes. Only for case objects like CSVFileFormatWrapper. Let me know if I should go with 
```
case _: ParquetFileFormat | _: OrcFileFormat | _: CSVFileFormat | _: JsonFileFormat => Some(fileFormat)
```
But to some

PR #87, opened 2020-07-10
Support for hyperspace event logging

apoorvedave1:

Add support for custom hyperspace events and event handling.

To support custom event handlers for hyperspace events

no

na


src/main/scala/com/microsoft/hyperspace/logging/HyperspaceEvent.scala
```diff
@@ -0,0 +1,216 @@
...
+import com.microsoft.hyperspace.index.IndexLogEntry
+
+/**
+ * Base class for all Hyperspace events.
+ */
+class HyperspaceEvent(
+    sparkUser: String = StringUtils.EMPTY,
+    appId: String = StringUtils.EMPTY,
+    appName: String = StringUtils.EMPTY) {
+  def dimensions: Seq[String] = Seq[String](sparkUser, appId, appName)
```
imback82, line 31: I don't think we need to expose this? The logger implemenation can do
```
def logEvent(event: HyperspaceEvent) = {
  event match {
    case HyperspaceIndexCRUDEvent => _
    ...
  }
}
```
?

src/main/scala/com/microsoft/hyperspace/logging/GenericHyperspaceLogging.scala
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+}
+
+/**
+ * Logger that logs Hyperspace events to spark logs.
+ */
+class GenericHyperspaceLogging extends HyperspaceLogging {
+  override def logEvent(event: HyperspaceEvent): Unit = {
+    logInfo(event.dimensions.mkString("; "))
```
imback82, line 58: I think no op should be fine.

apoorvedave1: Thanks @imback82 , 
I added logInfo so that if nothing is provided, we will see the events at least in spark application logs. With no-op, the events won't be seen. Is that the expected behavior?

If we go with no-op, this will be the behavior:
- for generic logging (info, error, warn, debug etc), use `HyperspaceLogging.logInfo(..), logError(...)` which logs to spark application logs
- for event logging, use `HyperspaceLogging.logEvent` which would be a no-op if no external HyperspaceLogging implementation is provided

If this is ok, I can make this change. Thanks

imback82: yes. if the user wants to capture the event, they can provide the event logger. btw, do we want `HyperspaceLogging` to extend `Logging`? or should we decouple it  as a `HyperspaceEventLogging` so that you can do `class A extends Logging with HyperspaceEventLoggin`?

imback82: cc: @AFFogarty with event logging backgound

src/main/scala/com/microsoft/hyperspace/logging/HyperspaceEvent.scala
```diff
@@ -0,0 +1,216 @@
...
+    sparkUser: String,
+    appId: String,
+    appName: String,
+    indexes: Seq[IndexLogEntry],
+    message: String = "")
+    extends HyperspaceEvent(sparkUser, appId, appName) {
+
+  override def dimensions: Seq[String] = {
+    commonDimensions ++ Seq("", "") ++ Seq(message)
+  }
```
AFFogarty, line 53: Confused about what's going on in this line.

src/main/scala/com/microsoft/hyperspace/logging/GenericHyperspaceLogging.scala
```diff
@@ -0,0 +1,60 @@
...
+import scala.util.{Success, Try}
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.util.hyperspace.Utils
+
+import com.microsoft.hyperspace.logging.Constants._
+
+trait HyperspaceLogging extends Logging {
+  def logEvent(event: HyperspaceEvent): Unit = HyperspaceLogging.logger.logEvent(event)
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @AFFogarty for the infinite loop catch!

src/main/scala/com/microsoft/hyperspace/logging/GenericHyperspaceLogging.scala
apoorvedave1: Thanks @imback82 , @AFFogarty 
Here's the summary from both your comments:
1. Let's keep logging (info/error etc) and emitting events separate
2. Logging will be handled by spark.internal.Logging class. We don't need to implement this logging or add a redirection to this logging through HyperspaceLogging
3. Hyperspace only handles event emission.
 
Let me know if this needs change.

src/main/scala/com/microsoft/hyperspace/logging/Constants.scala
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.logging
+
+object Constants {
+  val HYPERSPACE_LOGGER_CLASS = "spark.hyperspace.logger"
```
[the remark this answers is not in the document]
apoorvedave1: I can rename this to `_KEY`

src/main/scala/com/microsoft/hyperspace/logging/HyperspaceEvent.scala
imback82: Can we just remove `dimensions` as suggested here: https://github.com/microsoft/hyperspace/pull/87#discussion_r453082594 ?

src/main/scala/com/microsoft/hyperspace/logging/GenericHyperspaceLogging.scala
AFFogarty: That approach sounds good to me.

src/main/scala/com/microsoft/hyperspace/logging/HyperspaceEvent.scala
apoorvedave1: done, thanks

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
```diff
@@ -0,0 +1,71 @@
...
+object EventLogger extends Logging {
+  // Singleton logger instance for event logging.
+  private var logger: EventLogger = _
+
+  def getLogger: EventLogger = {
+    if (logger != null) {
+      return logger
+    }
+
+    val className: String = SparkSession.getActiveSession
```
imback82, line 50: How about the following flow? I think if the user specified the config for the logger and if it fails to load, we should fail early instead of falling back:
```scala
    SparkSession.getActiveSession
      .map(_.conf.get(HYPERSPACE_EVENT_LOGGER_CLASS_KEY))
      .map { className =>
        Try(Utils.classForName(className).newInstance) match {
          case Success(logger: EventLogger) =>
            logInfo("")
            logger
          case Failure(exception) =>
            logError("")
            throw exception
        }
      }
      .getOrElse{
        logInfo("")
        new NoOpEventLogger
      }
  }
```

src/main/scala/com/microsoft/hyperspace/telemetry/Constants.scala
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.telemetry
+
+object Constants {
+  val HYPERSPACE_EVENT_LOGGER_CLASS_KEY = "spark.hyperspace.eventloggerclass"
+  val DEFAULT_HYPERSPACE_EVENT_LOGGER_CLASS = NoOpEventLogger.getClass.getName
```
imback82, line 21: not used.

src/main/scala/com/microsoft/hyperspace/telemetry/Constants.scala
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.telemetry
+
+object Constants {
+  val HYPERSPACE_EVENT_LOGGER_CLASS_KEY = "spark.hyperspace.eventloggerclass"
```
imback82, line 20: nit: `eventLoggerClass`

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
```diff
@@ -0,0 +1,65 @@
...
+          logInfo(s"Setting hyperspace event logger to $className")
+          logger
+        case _ =>
+          logError(s"Unable to instantiate hyperspace logger from provided class $className")
+          throw HyperspaceException(
+            "Unable to instantiate hyperspace logger from provided class $className")
+      }
+    }
+    .getOrElse {
+      logInfo(s"Setting hyperspace event logger to $NoOpEventLogger")
```
imback82, line 58: `logInfo("Defaulting to the NoOpEventLogger event logger")`

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
```diff
@@ -0,0 +1,65 @@
...
+    .flatMap(_.conf.getOption(HYPERSPACE_EVENT_LOGGER_CLASS_KEY))
+    .map { className =>
+      Try(Utils.classForName(className).newInstance) match {
+        case Success(logger: EventLogger) =>
+          logInfo(s"Setting hyperspace event logger to $className")
+          logger
+        case _ =>
+          logError(s"Unable to instantiate hyperspace logger from provided class $className")
+          throw HyperspaceException(
+            "Unable to instantiate hyperspace logger from provided class $className")
```
imback82, line 54: you need `s` here

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
```diff
@@ -0,0 +1,65 @@
...
+}
+
+object EventLogger extends Logging {
+  // Singleton logger instance for event logging.
+  private[telemetry] lazy val logger: EventLogger = SparkSession.getActiveSession
+    .flatMap(_.conf.getOption(HYPERSPACE_EVENT_LOGGER_CLASS_KEY))
+    .map { className =>
+      Try(Utils.classForName(className).newInstance) match {
+        case Success(logger: EventLogger) =>
+          logInfo(s"Setting hyperspace event logger to $className")
```
imback82, line 49: I guess we can remove hyperspace here (and below).

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
imback82: and `event logger`

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
```diff
@@ -0,0 +1,65 @@
...
+ * Event Logger interface. Concrete implementations of this class implement event handling logic
+ * for hyperspace events.
+ */
+trait EventLogger {
+  def logEvent(event: HyperspaceEvent): Unit
+}
+
+object EventLogger extends Logging {
+  // Singleton logger instance for event logging.
+  private[telemetry] lazy val logger: EventLogger = SparkSession.getActiveSession
```
imback82, line 44: nit: shall we wrap this with `{ }` to be a bit more readable?:

```
private[telemetry] lazy val logger: EventLogger = {
  SparkSession.getActiveSession
    .flatMap(_.conf.getOption(HYPERSPACE_EVENT_LOGGER_CLASS_KEY))
    ...
}
```

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -0,0 +1,102 @@
...
+ */
+class HyperspaceIndexCRUDEvent(
+    sparkUser: String,
+    appId: String,
+    appName: String,
+    index: IndexLogEntry,
+    message: String = StringUtils.EMPTY)
+    extends HyperspaceEvent
+
+case class CreateActionEvent(
```
imback82, line 45: do you want to add a comment for each event?

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+package com.microsoft.hyperspace.telemetry
+
+import org.apache.commons.lang.StringUtils
+
+import com.microsoft.hyperspace.index.IndexLogEntry
+
+/**
+ * Base class for all Hyperspace events.
```
imback82, line 24: trait?

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
imback82, line 44: Did you want to scope this at the `object` level instead? `private[telemetry] EventLogger`. Not sure of your intention here.

apoorvedave1: Oh I made a slight change. Instead of using a getLogger public method, I directly used the logger object in the `HyperspaceLogging class. So I thought of making it private[telemetry]

apoorvedave1: I can revert it to
```
private lazy val logger: EventLogger = ....

def getLogger: EventLogger = logger
```

imback82: Oh this is fine. I was just curious if we should apply the scoping at the object level.

apoorvedave1: sorry I misread the comment. Will scope it to object level

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/telemetry/Constants.scala
apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/telemetry/Constants.scala
apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/actions/DeleteAction.scala
```diff
@@ -40,4 +43,17 @@ class DeleteAction(final override protected val logManager: IndexLogManager) ext
   }
 
   final override def op(): Unit = { /* Do nothing */ }
+
+  final override protected def event(message: String): HyperspaceEvent = {
+    val sc = SparkSession.getActiveSession.getOrElse {
```
imback82, line 48: I think we need a `trait` like `ActiveSparkSession` to remove duplication. And it sounds reasonable:
```scala
class DeleteAction(...) extends Action with ActiveSparkSession
```

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -91,6 +147,18 @@ case class CancelActionEvent(
...
+ * Index usage event. This event is emitted when an index is picked instead of original data
+ * source by one of the hyperspace rules.
+ *
+ * @param sparkUser
+ * @param appId
+ * @param appName
+ * @param indexes
+ * @param planBeforeRule
+ * @param planAfterRule
+ * @param message
```
imback82, line 160: missing?

src/main/scala/com/microsoft/hyperspace/actions/VacuumAction.scala
```diff
@@ -49,4 +52,17 @@ class VacuumAction(
       }
     }
   }
+
+  final override protected def event(message: String): HyperspaceEvent = {
```
imback82, line 56: why `protected`?

src/main/scala/com/microsoft/hyperspace/actions/VacuumAction.scala
```diff
@@ -49,4 +52,17 @@ class VacuumAction(
...
+    val sc = SparkSession.getActiveSession.getOrElse {
+      throw HyperspaceException("No spark session found")
+    }.sparkContext
+
+    VacuumActionEvent(
+      sc.sparkUser,
+      sc.applicationId,
+      sc.appName,
```
imback82, line 64: looks like these are repeating. can you try to refactor?

src/main/scala/com/microsoft/hyperspace/actions/VacuumAction.scala
apoorvedave1: This is implementation of the base class method `Action.event()`.  Are you saying to make this method public? Sorry did not understand he comment.

src/main/scala/com/microsoft/hyperspace/actions/DeleteAction.scala
apoorvedave1: thanks @imback82 , added.
Since all Actions have this, is it ok to just change 
`trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession`

src/main/scala/com/microsoft/hyperspace/actions/VacuumAction.scala
imback82: I meant whether this can be `private`.

src/main/scala/com/microsoft/hyperspace/actions/DeleteAction.scala
imback82: Yes, that sounds good.

src/main/scala/com/microsoft/hyperspace/actions/VacuumAction.scala
apoorvedave1: changed event() method to take in spark user, appid, app name. Could you please have a look ?

src/main/scala/com/microsoft/hyperspace/actions/VacuumAction.scala
apoorvedave1: Oh private is not possible. It becomes inaccessible to base `Action` class. I will then keep it protected. Thanks.

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -0,0 +1,169 @@
...
+ * @param appName Spark application name.
+ * @param index Related indexes.
+ * @param message Message about event.
+ */
+class HyperspaceIndexCRUDEvent(
+    sparkUser: String,
+    appId: String,
+    appName: String,
+    index: IndexLogEntry,
+    message: String = StringUtils.EMPTY)
```
imback82, line 42: why default?

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -0,0 +1,169 @@
...
+ * General index CRUD event.
+ *
+ * @param sparkUser User of spark application.
+ * @param appId Spark application id.
+ * @param appName Spark application name.
+ * @param index Related indexes.
+ * @param message Message about event.
+ */
+class HyperspaceIndexCRUDEvent(
+    sparkUser: String,
```
imback82, line 38: looks like we can capture `sparkUser`, `appId` and `appName` as `AppInfo`?

imback82: and it can be used in `protected def event` as well

apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
apoorvedave1: removed default. thanks @imback82 , done

imback82:
@apoorvedave1 Is this still a draft?

apoorvedave1:
> @apoorvedave1 Is this still a draft?

changed to ready. Thanks

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -120,10 +125,21 @@ object FilterIndexRule extends Rule[LogicalPlan] with Logging {
         val newOutput =
           logicalRelation.output.filter(attr => index.schema.fieldNames.contains(attr.name))
 
-        project.copy(
+        val updatedPlan = project.copy(
           child =
             filter.copy(child = logicalRelation.copy(relation = newRelation, output = newOutput)))
 
+        // TODO: implement scrubber to remove PII from plans before adding plans to events.
```
imback82, line 132: why? it should be the logger implementation which should worry about this.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -120,10 +125,21 @@ object FilterIndexRule extends Rule[LogicalPlan] with Logging {
...
+        logEvent(
+          HyperspaceIndexUsageEvent(
+            AppInfo(sparkContext.sparkUser, sparkContext.applicationId, sparkContext.appName),
+            Seq(index),
+            "", // Blank until scrubber implementation
```
imback82, line 137: ditto

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -59,6 +64,19 @@ object JoinIndexRule extends Rule[LogicalPlan] with Logging {
             case (lIndex, rIndex) =>
               val updatedPlan = join
                 .copy(left = getReplacementPlan(lIndex, l), right = getReplacementPlan(rIndex, r))
+
+              // TODO: implement scrubber to remove PII from plans before adding plans to events.
```
imback82, line 68: ditto

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -50,7 +51,11 @@ import com.microsoft.hyperspace.index.rankers.JoinIndexRanker
  * These indexes are indexed by the join columns and can improve the query performance by
  * avoiding full shuffling of T1 and T2.
  */
-object JoinIndexRule extends Rule[LogicalPlan] with Logging {
+object JoinIndexRule
+    extends Rule[LogicalPlan]
+    with Logging
+    with HyperspaceEventLogging
+    with ActiveSparkSession {
```
imback82, line 58: if you are using this, you must be removing some existing usage?

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -0,0 +1,119 @@
...
+ */
+trait HyperspaceEvent
+
+/**
+ * General index CRUD event.
+ *
+ * @param appInfo AppInfo for spark application.* @param index Related indexes.
+ * @param message Message about event.
+ */
+class HyperspaceIndexCRUDEvent(appInfo: AppInfo, index: IndexLogEntry, message: String)
```
imback82, line 41: looks like you are using this only for grouping anyway? then just make this as trait and remove fields? I see you have fields in case class and here so you are duplicating? (not quite sure about your intention here)

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -0,0 +1,119 @@
...
+    extends HyperspaceEvent
+
+/**
+ * Index creation event. Emitted on index creation.
+ *
+ * @param appInfo AppInfo for spark application.* @param index Related indexes.
```
imback82, line 47: can you fix param here and other places?

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
```diff
@@ -0,0 +1,67 @@
...
+            logInfo(s"Setting event logger to $className")
+            logger
+          case _ =>
+            logError(s"Unable to instantiate event logger from provided class $className")
+            throw HyperspaceException(
+              s"Unable to instantiate event logger from provided class $className")
+        }
+      }
+      .getOrElse {
+        logInfo(s"Defaulting to the $NoOpEventLogger event logger")
```
imback82, line 59: curious. what does it print here `$NoOpEventLogger`? I guess it prints better than the following?
```
scala> object NoOpEventLogger {
     |   def logEvent(event: Int): Unit = {}
     | }
defined object NoOpEventLogger

scala> println(s"hello: $NoOpEventLogger")
hello: $line14.$read$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$NoOpEventLogger$@806a687
```

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
apoorvedave1: Oh yes that's correct. Let me fix this one. Thanks

src/main/scala/com/microsoft/hyperspace/ActiveSparkSession.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package com.microsoft.hyperspace
+
+import org.apache.spark.SparkContext
+import org.apache.spark.sql.SparkSession
+
+trait ActiveSparkSession {
+  def spark: SparkSession = {
+    SparkSession.getActiveSession.getOrElse {
+      throw HyperspaceException("No spark session found")
```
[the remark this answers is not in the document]
apoorvedave1: Thanks, fixed

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
```diff
@@ -81,16 +82,24 @@ trait Action extends Logging {
...
+      logEvent(event(message = "Operation Succeeded"))
     } catch {
       case e: Exception =>
+        logEvent(event(message = s"Operation Failed: ${e.getMessage}."))
         throw e
     }
   }
+
+  protected def event(
+      appInfo: AppInfo =
```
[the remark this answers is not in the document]
apoorvedave1: Thanks, fixed

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -72,4 +75,40 @@ class CreateAction(
...
   //   This needs to be refactored to mark this as protected.
   final override def op(): Unit = write(spark, df, indexConfig)
+
+  final override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
+    // logEntry may or may not exist the first time an index is being created.
+    // This can happen, for example, if index config we receive is incompatible with the
+    // dataframe. So we use Try here. We still need an index object with empty values for event
+    // logging.
+    val index: IndexLogEntry = Try {
+      // if a logEntry exists, we can create a valid index object directly
```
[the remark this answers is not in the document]
apoorvedave1: Thanks, fixed

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
apoorvedave1: Thanks, fixed

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: Thanks, fixed

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: Yes, thanks, my changes got reverted during a merge conflict fix. Fixed.

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
apoorvedave1: Thanks, fixed

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
apoorvedave1: Thanks, fixed

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
apoorvedave1: Thanks for the catch. It prints 
`com.microsoft.hyperspace.telemetry.NoOpEventLogger$@3cb5cdba`

I replaced it to print NoOpEventLogger.getClass.getName

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
```diff
@@ -0,0 +1,68 @@
...
+            logger
+          case _ =>
+            logError(s"Unable to instantiate event logger from provided class $className")
+            throw HyperspaceException(
+              s"Unable to instantiate event logger from provided class $className")
+        }
+      }
+      .getOrElse {
+        logInfo(
+          s"Defaulting to the ${NoOpEventLogger.getClass.getName.stripSuffix("$")} event logger")
```
imback82, line 60: Why not just `"Defaulting to the NoOpEventLogger event logger"`, which was my first suggestion?

apoorvedave1: I can change it to that, I feel using string for class name directly may not be right. Assuming if class name changes, we might find it difficult to correct this in future. That is my only concern when using direct strings.

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -72,4 +75,40 @@ class CreateAction(
   // TODO: The following should be protected, but RefreshAction is calling CreateAction.op().
   //   This needs to be refactored to mark this as protected.
   final override def op(): Unit = write(spark, df, indexConfig)
+
+  final override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
+    // logEntry may or may not exist the first time an index is being created.
```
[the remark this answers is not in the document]
apoorvedave1: Sorry I didn't phrase it right. I am just referring to the `logEntry` variable in this class.

So vacuum scenario won't occur because `val logEntry` is freshly calculated from new data: `df` and `indexConfig`. It doesn't read `logEntry` from older operation logs.

So there are only 2 cases:
1. logEntry creation fails because of incorrect config
2. logEntry creation passes because of correct config

And since logEntry is `lazy val`, it won't need any more creation in future `def event()` calls (which is where I meant it exists)

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
apoorvedave1: Please let me know if that's still ok and I can change it to `"Defaulting to the NoOpEventLogger event logger"`

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * General index CRUD event.
+ */
+trait HyperspaceIndexCRUDEvent extends HyperspaceEvent
+
+/**
+ * Index creation event. Emitted on index creation.
+ *
+ * @param appInfo AppInfo for spark application.
+ * @param index Related indexes.
```
imback82, line 44: why index"es"?

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEventLogging.scala
imback82: OK, sounds good.

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
```diff
@@ -80,17 +81,24 @@ trait Action extends Logging {
...
     try {
+      logEvent(event(appInfo, "Operation Started."))
       validate()
 
       begin()
 
       op()
 
       end()
+      logEvent(event(appInfo, message = "Operation Succeeded"))
```
imback82, line 95: nit: "." to be consistent?

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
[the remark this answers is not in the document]
apoorvedave1: @imback82

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
```diff
@@ -80,17 +81,24 @@ trait Action extends Logging {
   }
 
   def run(): Unit = {
+    val appInfo =
+      AppInfo(sparkContext.sparkUser, sparkContext.applicationId, sparkContext.appName)
```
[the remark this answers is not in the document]
apoorvedave1: I have seen it have issues with unit tests, e.g. in JoinRUle and FilterRule unit tests, where once this val is fixed, it somehow refers to incorrect spark session leading to test failures. (This was seen when individually tests would pass, but the test suite when run as a whole would fail)

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -72,4 +75,40 @@ class CreateAction(
   // TODO: The following should be protected, but RefreshAction is calling CreateAction.op().
   //   This needs to be refactored to mark this as protected.
   final override def op(): Unit = write(spark, df, indexConfig)
+
+  final override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
+    // LogEntry may or may not exist the first time an index is being created.
+    // This can happen, for example, if index config we receive is incompatible with the
+    // dataframe. So we use Try here. We still need an index object with empty values for event
```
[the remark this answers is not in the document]
apoorvedave1: Oh it's not completely empty. It still contains index config, index name. My intention was to at least emit an event with info " trying to create an index with this config "
Could help in debugging.

If we go with option, do you mean the logged event will contain null?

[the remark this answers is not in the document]
imback82: In that case, I think `CreateActionEvent` should capture `indexConfig`

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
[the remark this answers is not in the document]
imback82: I think we need to understand why (something doesn't seem to be set up correct). Can you create an issue after this PR is merged?

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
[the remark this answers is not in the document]
apoorvedave1: got it. added. Thanks @imback82

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
apoorvedave1: Thanks @imback82

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
apoorvedave1: Thanks @imback82

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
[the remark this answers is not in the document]
apoorvedave1: got it. I will add an issue

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -72,4 +75,16 @@ class CreateAction(
...
```
imback82, line 82: can we update this comment?

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -72,4 +75,16 @@ class CreateAction(
...
+    // logging.
+    val index: Option[IndexLogEntry] = Try {
+      // If a logEntry exists, we can create a valid index object directly.
```
imback82, line 85: this comment doesn't seem to be useful anymore?

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -72,4 +75,16 @@ class CreateAction(
   // TODO: The following should be protected, but RefreshAction is calling CreateAction.op().
   //   This needs to be refactored to mark this as protected.
   final override def op(): Unit = write(spark, df, indexConfig)
+
+  final override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
+    // LogEntry may or may not exist the first time an index is being created.
+    // This can happen, for example, if index config we receive is incompatible with the
+    // dataframe. So we use Try here. We still need an index object with empty values for event
+    // logging.
+    val index: Option[IndexLogEntry] = Try {
```
imback82, line 84: remove type?

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * General index CRUD event.
+ */
+trait HyperspaceIndexCRUDEvent extends HyperspaceEvent
+
+/**
+ * Index creation event. Emitted on index creation.
+ *
+ * @param appInfo AppInfo for spark application.
+ * @param index Optional index. It can be None if create index fails with invalid config.
```
imback82, line 44: missing indexConfig

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
apoorvedave1: removed. Thanks @imback82

imback82:
Can you rebase with master?

Issue #79, opened 2020-07-01
Create a sample Hyperspace "HelloWorld" application


pirz:
### Describe the issue
Write a sample Hyperspace `HelloWorld` application code with some simple and reasonable scenario.
Minimally the app code should create some example indexes on some sample data and demonstrate how those indexes can be maintained and used.

Main goal for the app is to show how one can import and use Hyperspace within a Spark application. 

### To Reproduce
NA

### Expected behavior
An example (and ideally configurable) Spark application code which demonstrate core features and capabilities of Hyperspace.

### Environment 
All environments.

pherbel:
Hello, I'm looking forward to it.

Could you share a c# .net spar version too?

rapoth:
CC: @AFFogarty for the C# version

AFFogarty:
Once #90 is completed, I'll open a PR with the C# version.

AFFogarty:
@pirz Should we close this issue now?

imback82:
Closed by #94

PR #90, opened 2020-07-14
Hyperspace "HelloWorld" application

pirz:
This change adds a sample Scala application to show how Hyperspace can be used as a library by other applications.
It addresses issue #79  . 

This change demonstrates how an application can use Hyperspace by adding a dependency to a project and import Hyperspace in application's code. This example helps users understand the details by showing required changes to import Hyperspace and use it to create and leverage indexes on sample data.  

No.

This change consists of an standalone example application and is verified by running it manually and validating the output and its behavior.


examples/build.sbt
```diff
@@ -0,0 +1,22 @@
...
+
+libraryDependencies += "com.microsoft.hyperspace" %% "hyperspace-core" % "0.1.0"
+
+scalacOptions ++= Seq(
+  "-target:jvm-1.8"
+)
+
+javaOptions += "-Xmx1024m"
+
+publishMavenStyle := true
```
AFFogarty, line 22: Let's end with a newline.

examples/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.spark.sql.{DataFrame, SparkSession}
+
+import com.microsoft.hyperspace._
+import com.microsoft.hyperspace.Hyperspace
+import com.microsoft.hyperspace.index.{IndexConfig, IndexConstants}
+
+object App {
+  def main(args: Array[String]): Unit = {
+    // Create Spark session
+    val sparkConf = new SparkConf().setMaster("local[*]")
```
AFFogarty, line 27: Does it need to be local?  What if I try to run this app on a cluster?

examples/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+    val spark = SparkSession.builder.config(sparkConf).getOrCreate()
+    spark.sparkContext.setLogLevel("ERROR")
```
AFFogarty, line 29: Why error loglevel?

examples/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    // Sample department records
+    val departments = Seq(
+      (10, "Accounting", "New York"),
+      (20, "Research", "Dallas"),
+      (30, "Sales", "Chicago"),
+      (40, "Operations", "Boston"))
+
+    // Sample employee records
+    val employees = Seq(
```
AFFogarty, line 39: Can we move the data to JSON or CSV files so that we can re-use them for the C# sample app?

examples/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    // Save example data records as Parquet
+    import spark.implicits._
+    val deptLocation = "departments"
+    val empLocation = "employees"
+    departments
+      .toDF("deptId", "deptName", "location")
+      .write
+      .mode("overwrite")
+      .parquet(deptLocation)
```
AFFogarty, line 63: If we store the sample data in files, then we don't need this section.

examples/src/main/scala/App.scala
rapoth: I believe he's making it easy to submit the app to a cluster (as opposed to asking folks to copy csv files into storage accounts). What do you think? 

If we want to do this on the C# side, we can use `CreateDataFrame` [API](https://github.com/dotnet/spark/pull/338). Any thoughts?

examples/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+    val deptIndexConfig: IndexConfig = IndexConfig("deptIndex", Seq("deptId"), Seq("deptName"))
+    hyperspace.createIndex(empDF, empIndexConfig)
+    hyperspace.createIndex(deptDF, deptIndexConfig)
+
+    // List all indexes
+    hyperspace.indexes.show
+
+    // Enable Hyperspace to leverage indexes
+    spark.enableHyperspace()
+    val deptDFrame: DataFrame = spark.read.parquet("departments")
```
rapoth, line 87:
```suggestion
    val deptDFrame: DataFrame = spark.read.parquet(deptLocation)
```

examples/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+    val empDFrame: DataFrame = spark.read.parquet(empLocation)
```
rapoth, line 88: I'm curious. Why did you define `empDFrame` and `deptDFrame` as opposed to just using the `empDF` and `deptDF` above?

examples/src/main/scala/App.scala
AFFogarty: Sure, that's fine.

AFFogarty: Deleted my other comment regarding files.

examples/build.sbt
```diff
@@ -0,0 +1,22 @@
+name := "hyperspaceApp"
```
AFFogarty, line 1: Can you move the whole app to `examples/scala/...` so that we can put the C# example in `examples/csharp/...`?

examples/build.sbt
pirz: Hmm, I have added one before pushing; however I dont see it here. I am not sure if it gets removed automatically or sth else is happening.

examples/src/main/scala/App.scala
pirz: I had added Local to make the code runnable right after checkout. However, I looked at how Spark does it for its examples and followed the same pattern. As an example, check [here](https://github.com/apache/spark/blob/master/examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala#L39).

examples/src/main/scala/App.scala
pirz: I had added it to make the output less noisy with Info logs. Removed it.

examples/src/main/scala/App.scala
pirz: It is just to show that index usage has nothing to do with which DataFrame the query uses (i.e. For a query, user does not need to use the same DF he used for indexing). We follow the same pattern in our notebook example. Plz let me know if you prefer to drop these two DFs and directly write queries against original DFs used to create indexes.

examples/build.sbt
pirz: Done.

examples/src/main/scala/App.scala
rapoth: I would probably simplify it - since there are no comments anyways to clarify what you mentioned.

pirz: @rapoth I removed new DFs and used same DFs for both creating indexes and queries.

examples/scala/build.sbt
```diff
@@ -0,0 +1,22 @@
...
+
+libraryDependencies += "com.microsoft.hyperspace" %% "hyperspace-core" % "0.1.0"
+
+scalacOptions ++= Seq(
+  "-target:jvm-1.8"
+)
+
+javaOptions += "-Xmx1024m"
+
+publishMavenStyle := true
```
imback82, line 22: Do we need these?

examples/scala/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+import com.microsoft.hyperspace.Hyperspace
+import com.microsoft.hyperspace.index.IndexConfig
+
+object App {
+  def main(args: Array[String]): Unit = {
+    // Create Spark session
+    val spark = SparkSession
+      .builder()
+      .appName("Hyperspace example")
+      .config("spark.some.config.option", "some-value")
```
imback82, line 29: do we need this?

examples/scala/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    val deptDF: DataFrame = spark.read.parquet(deptLocation)
+    val empDF: DataFrame = spark.read.parquet(empLocation)
+
+    val deptIndexConfig: IndexConfig = IndexConfig("deptIndex", Seq("deptId"), Seq("deptName"))
+    val empIndexConfig: IndexConfig = IndexConfig("empIndex", Seq("deptId"), Seq("empName"))
+    hyperspace.createIndex(deptDF, deptIndexConfig)
+    hyperspace.createIndex(empDF, empIndexConfig)
+
+
```
imback82, line 83: nit: extra empty line.

examples/scala/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+    employees
+      .toDF("empId", "empName", "deptId")
+      .write
+      .mode("overwrite")
+      .parquet(empLocation)
+
+    // Create Hyperspace indexes
+    val hyperspace = new Hyperspace(spark)
+
+    val deptDF: DataFrame = spark.read.parquet(deptLocation)
```
imback82, line 75: nit: do we want types for local variables?

examples/scala/src/main/scala/App.scala
```diff
@@ -0,0 +1,103 @@
...
+    eqFilter.show()
+    hyperspace.explain(eqFilter)
+
+    // Example of index usage for join
+    val eqJoin: DataFrame =
+      empDF
+        .join(deptDF, empDF("deptId") === deptDF("deptId"))
+        .select(empDF("empName"), deptDF("deptName"))
+    eqJoin.show()
+    hyperspace.explain(eqJoin)
```
imback82, line 101: add `spark.stop` to be explicit?

examples/scala/src/main/scala/App.scala
pirz: Which line do you exactly mean? 
We need to create a SparkSession and instead of using local mode I switched to this way as Spark examples use this style. As an example, please look at [here](https://github.com/apache/spark/blob/master/examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala#L39).

examples/scala/build.sbt
pirz: removed them.

examples/scala/src/main/scala/App.scala
pirz: fixed it.

examples/scala/src/main/scala/App.scala
pirz: dropped type decelerations for local ones.

examples/scala/src/main/scala/App.scala
pirz: added it. Thnx!

examples/scala/src/main/scala/App.scala
imback82: Usually, the line right above the comment: `config` in this case.

examples/scala/src/main/scala/App.scala
```diff
@@ -0,0 +1,104 @@
...
+      (7876, "ADAMS", 20),
+      (7900, "JAMES", 30),
+      (7934, "MILLER", 10),
+      (7902, "FORD", 20),
+      (7654, "MARTIN", 30))
+
+    // Save example data records as Parquet
+    import spark.implicits._
+    val deptLocation = "departments"
+    val empLocation = "employees"
```
AFFogarty, line 59: Move this down to right above where it is first used.

examples/scala/src/main/scala/App.scala
```diff
@@ -0,0 +1,104 @@
...
+
+    // Enable Hyperspace to leverage indexes
+    spark.enableHyperspace()
+
+    // Example of index usage for filtered selection
+    val eqFilter = deptDF.filter("deptId = 20").select("deptName")
+    eqFilter.show()
+    hyperspace.explain(eqFilter)
+
+    // Example of index usage for join
```
AFFogarty, line 94: Let's end single-sentence comments with a `.`.

examples/scala/src/main/scala/App.scala
```diff
@@ -0,0 +1,104 @@
...
+      .mode("overwrite")
+      .parquet(empLocation)
+
+    // Create Hyperspace indexes
+    val hyperspace = new Hyperspace(spark)
+
+    val deptDF = spark.read.parquet(deptLocation)
+    val empDF = spark.read.parquet(empLocation)
+
+    val deptIndexConfig = IndexConfig("deptIndex", Seq("deptId"), Seq("deptName"))
```
AFFogarty, line 78: Wondering if these configs could just be inline.

examples/scala/src/main/scala/App.scala
pirz: Got it; yes I guess it is a good idea to keep this line to highlight the fact that user can change this to set master and create the Spark session in his desired mode, or add any other config override. But if you think it should be removed, I will drop it.

examples/scala/src/main/scala/App.scala
pirz: Done.

examples/scala/src/main/scala/App.scala
pirz: Done.

examples/scala/src/main/scala/App.scala
pirz: I think it is a good idea to keep it separate to improve readability as the main purpose of this App is helping newbie users of Hyperspace. We have followed similar separation of configs and creation step in our tutorial Notebook.

examples/scala/src/main/scala/App.scala
```diff
@@ -0,0 +1,105 @@
...
+    spark.enableHyperspace()
+
+    // Example of index usage for filtered selection.
+    val eqFilter = deptDF.filter("deptId = 20").select("deptName")
+    eqFilter.show()
+    hyperspace.explain(eqFilter)
+
+    // Example of index usage for join.
+    val eqJoin = empDF
+        .join(deptDF, empDF("deptId") === deptDF("deptId"))
```
imback82, line 97: doesn't auto-format give you two space indentations?

pirz: Thanks, it is fixed.

Issue #56, opened 2020-06-24
Add stable state check in IndexLogManager.createLatestStableLog()


apoorvedave1:
### Describe the issue
In   IndexLogManager.createLatestStableLog() make sure the log being copied has a   stable state. The api currently makes a copy of whatever log id it's passed, without 'assert'ing that the state of this log is stable (i.e. ACTIVE, DELETED, DOESNOTEXIST)

### To Reproduce
NA

### Expected behavior
NA
### Environment 
NA

saucam:
Hi,

Let me give this a shot!

saucam:
Here is the PR, please review https://github.com/microsoft/hyperspace/pull/92

sezruby:
@apoorvedave1 I think we can close this issue now. Thanks @saucam!

imback82:
Fixed by #92

PR #92, opened 2020-07-17
Add stable check in IndexLogManager.createLatestStableLog

saucam:

This PR addresses: #56 , added stability state check in ```IndexLogManager.createLatestStableLog```

No

Added a new unit test and updated the existing test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -19,6 +19,7 @@ package com.microsoft.hyperspace.index
 import java.util.UUID
 
 import scala.util.{Failure, Success, Try}
+import scala.util.control.NonFatal
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileStatus, FileSystem, FileUtil, Path}
@@ -71,7 +72,13 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
       return None
     }
     val contents = FileUtils.readContents(fs, path)
-    Try(LogEntry.fromJson(contents)).toOption
+    Try(LogEntry.fromJson(contents)) match {
+      case Success(s) => Some(s)
+      case Failure(NonFatal(e)) =>
+        logError(s"Error reading LogEntry json from $path, $e")
+        None
+      case Failure(e) => throw e
+    }
   }
 
   override def getLog(id: Int): Option[LogEntry] = {
@@ -110,19 +117,23 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
     }
   }
 
-  override def createLatestStableLog(id: Int): Boolean = {
+  override def createLatestStableLog(id: Int): Boolean =
     getLog(id) match {
       case Some(logEntry) if Constants.STABLE_STATES.contains(logEntry.state) =>
         Try(FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration))
-          match {
-            case Success(v) => v
-            case Failure(e) =>
-              logError(s"Failed to create the latest stable log with id = '$id'", e)
-               false
-          }
-      case _ => false
+        match {
+          case Success(v) => v
+          case Failure(e) =>
+            logError(s"Failed to create the latest stable log with id = '$id'", e)
+            false
+        }
+      case Some(logEntry) =>
+        logError(s"Found LogEntry with non stable state = ${logEntry.state}")
+        false
+      case None =>
+        logError(s"Unable to get LogEntry for id = $id")
+        false
     }
-  }
 
   override def deleteLatestStableLog(): Boolean = {
     try {
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -72,13 +72,7 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
       return None
     }
     val contents = FileUtils.readContents(fs, path)
-    Try(LogEntry.fromJson(contents)) match {
-      case Success(s) => Some(s)
-      case Failure(NonFatal(e)) =>
-        logError(s"Error reading LogEntry json from $path, $e")
-        None
-      case Failure(e) => throw e
-    }
+    Some(LogEntry.fromJson(contents))
   }
 
   override def getLog(id: Int): Option[LogEntry] = {
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -18,7 +18,7 @@ package com.microsoft.hyperspace.index
 
 import java.util.UUID
 
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileStatus, FileSystem, FileUtil, Path}
@@ -71,7 +71,7 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
       return None
     }
     val contents = FileUtils.readContents(fs, path)
-    Some(LogEntry.fromJson(contents))
+    Try(LogEntry.fromJson(contents)).toOption
   }
 
   override def getLog(id: Int): Option[LogEntry] = {
@@ -111,13 +111,16 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
   }
 
   override def createLatestStableLog(id: Int): Boolean = {
-    // TODO: make sure log with the id has a stable state.
-    try {
-      FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration)
-    } catch {
-      case ex: Exception =>
-        logError(s"Failed to create the latest stable log with id = '$id'", ex)
-        false
+    getLog(id) match {
+      case Some(logEntry) if Constants.STABLE_STATES.contains(logEntry.state) =>
+        Try(FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration))
+          match {
+            case Success(v) => v
+            case Failure(e) =>
+              logError(s"Failed to create the latest stable log with id = '$id'", e)
+               false
+          }
+      case _ => false
     }
   }
 
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -19,7 +19,6 @@ package com.microsoft.hyperspace.index
 import java.util.UUID
 
 import scala.util.{Failure, Success, Try}
-import scala.util.control.NonFatal
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileStatus, FileSystem, FileUtil, Path}
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -110,7 +110,7 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
     }
   }
 
-  override def createLatestStableLog(id: Int): Boolean =
+  override def createLatestStableLog(id: Int): Boolean = {
     getLog(id) match {
       case Some(logEntry) if Constants.STABLE_STATES.contains(logEntry.state) =>
         Try(FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration))
@@ -121,12 +121,13 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
             false
         }
       case Some(logEntry) =>
-        logError(s"Found LogEntry with non stable state = ${logEntry.state}")
+        logError(s"Found LogEntry with non stable state = ${logEntry.state} for id = '$id'")
         false
       case None =>
-        logError(s"Unable to get LogEntry for id = $id")
+        logError(s"Unable to get LogEntry for id = '$id'")
         false
     }
+  }
 
   override def deleteLatestStableLog(): Boolean = {
     try {
--- a/src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
@@ -55,6 +55,10 @@ class IndexLogManagerImplTest extends SparkFunSuite with SparkInvolvedSuite with
             Content.Directory("dir2", Seq("1.json", "2.json"), NoOpFingerprint()))))))),
     Map())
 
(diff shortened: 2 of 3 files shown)


src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -71,7 +71,7 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging
       return None
     }
     val contents = FileUtils.readContents(fs, path)
-    Some(LogEntry.fromJson(contents))
+    Try(LogEntry.fromJson(contents)).toOption
```
imback82, line 74: If `fromJson` throws a non-fatal exception, this will return `None` instead of throwing an exception, and thus cannot differentiate it with the `!fs.exists` case above. Can we revert this change for now? We can come back to this if required.

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -111,13 +111,16 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging
...
-        logError(s"Failed to create the latest stable log with id = '$id'", ex)
-        false
+    getLog(id) match {
+      case Some(logEntry) if Constants.STABLE_STATES.contains(logEntry.state) =>
+        Try(FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration))
+          match {
+            case Success(v) => v
+            case Failure(e) =>
+              logError(s"Failed to create the latest stable log with id = '$id'", e)
+               false
```
imback82, line 121: nit: indentation.

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -111,13 +111,16 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging
...
+          }
+      case _ => false
```
imback82, line 123: I think we need to catch two cases: 1) non stable state log entry and 2) `getLog` returned `None`. Can you also add a logging for each case?

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -144,11 +144,27 @@ class IndexLogManagerImplTest extends SparkFunSuite with SparkInvolvedSuite with
   test("testUpdateLatestStableLog passes if latestStable.json can be created") {
     val path = new Path(testRoot, UUID.randomUUID().toString)
     val fs = path.getFileSystem(new Configuration)
-    FileUtils.createFile(fs, new Path(path, s"$HYPERSPACE_LOG/0"), "file contents")
+    FileUtils.createFile(
+      fs,
+      new Path(path, s"$HYPERSPACE_LOG/0"),
+      JsonUtils.toJson(getEntry("ACTIVE"))
+    )
     new IndexLogManagerImpl(path).createLatestStableLog(0)
```
imback82, line 152: Can you check the return value?

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -144,11 +144,27 @@ class IndexLogManagerImplTest extends SparkFunSuite with SparkInvolvedSuite with
...
 
+  test("testUpdateLatestStableLog fails if log state is not stable") {
+    val path = new Path(testRoot, UUID.randomUUID().toString)
+    val fs = path.getFileSystem(new Configuration)
+    FileUtils.createFile(
+      fs,
+      new Path(path, s"$HYPERSPACE_LOG/0"),
+      JsonUtils.toJson(getEntry("CANCELLING"))
+    )
+    new IndexLogManagerImpl(path).createLatestStableLog(0)
```
imback82, line 164: ditto

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -144,11 +144,27 @@ class IndexLogManagerImplTest extends SparkFunSuite with SparkInvolvedSuite with
   test("testUpdateLatestStableLog passes if latestStable.json can be created") {
     val path = new Path(testRoot, UUID.randomUUID().toString)
     val fs = path.getFileSystem(new Configuration)
-    FileUtils.createFile(fs, new Path(path, s"$HYPERSPACE_LOG/0"), "file contents")
+    FileUtils.createFile(
+      fs,
+      new Path(path, s"$HYPERSPACE_LOG/0"),
+      JsonUtils.toJson(getEntry("ACTIVE"))
+    )
```
imback82, line 151: Can we put this in the previous line? (that's how auto-format is done)

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -144,11 +144,27 @@ class IndexLogManagerImplTest extends SparkFunSuite with SparkInvolvedSuite with
...
   }
 
+  test("testUpdateLatestStableLog fails if log state is not stable") {
+    val path = new Path(testRoot, UUID.randomUUID().toString)
+    val fs = path.getFileSystem(new Configuration)
+    FileUtils.createFile(
+      fs,
+      new Path(path, s"$HYPERSPACE_LOG/0"),
+      JsonUtils.toJson(getEntry("CANCELLING"))
+    )
```
imback82, line 163: ditto

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
saucam: So we want to catch only Non-fatal exceptions here ?

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
saucam: ok

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
saucam: ok

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
saucam: ok

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
saucam: ok

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
saucam: ok

saucam:
@imback82 I have addressed most of your comments! kindly please review again,  Thanks !

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
saucam: done

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
saucam: done

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -55,6 +55,10 @@ class IndexLogManagerImplTest extends SparkFunSuite with SparkInvolvedSuite with
             Content.Directory("dir2", Seq("1.json", "2.json"), NoOpFingerprint()))))))),
     Map())
 
+  def getEntry(state: String): LogEntry = {
```
apoorvedave1, line 58: could you please make it private

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -71,7 +72,13 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging
       return None
     }
     val contents = FileUtils.readContents(fs, path)
-    Some(LogEntry.fromJson(contents))
+    Try(LogEntry.fromJson(contents)) match {
+      case Success(s) => Some(s)
+      case Failure(NonFatal(e)) =>
+        logError(s"Error reading LogEntry json from $path, $e")
+        None
+      case Failure(e) => throw e
```
apoorvedave1, line 80: From this it looks like if `contents` is corrupted, and json conversion fails, it will not throw the error as it is Non-Fatal.
Could you change it to throw in case of all failures? We would not want file to be present but unreadable.

cc @imback82

imback82: right, this change should be just reverted as mentioned here: https://github.com/microsoft/hyperspace/pull/92#discussion_r456736210

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
saucam: ok done

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
saucam: ok reverted, but it does not look like valid code :
We are returning ```Some``` of something that can throw, but we can essentially never get a ```None``` here

saucam:
Thanks for the comments @apoorvedave1 , fixed your comments, kindly help review again!

cc @imback82

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -110,16 +110,23 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging
...
+      case Some(logEntry) if Constants.STABLE_STATES.contains(logEntry.state) =>
+        Try(FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration))
+        match {
+          case Success(v) => v
+          case Failure(e) =>
+            logError(s"Failed to create the latest stable log with id = '$id'", e)
+            false
+        }
+      case Some(logEntry) =>
+        logError(s"Found LogEntry with non stable state = ${logEntry.state}")
```
imback82, line 124: "for id = '$id'"

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -110,16 +110,23 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging
...
+        false
+      case None =>
+        logError(s"Unable to get LogEntry for id = $id")
```
imback82, line 127: quote '$id' to be consistent?

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -110,16 +110,23 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging
...
   }
 
-  override def createLatestStableLog(id: Int): Boolean = {
-    // TODO: make sure log with the id has a stable state.
-    try {
-      FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration)
-    } catch {
-      case ex: Exception =>
-        logError(s"Failed to create the latest stable log with id = '$id'", ex)
+  override def createLatestStableLog(id: Int): Boolean =
```
imback82, line 113: nit: can we put `{` `}` around this `def`?

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
imback82: I meant the parenthesis:
```
FileUtils.createFile(
  fs, new Path(path, s"$HYPERSPACE_LOG/0"), JsonUtils.toJson(getEntry("ACTIVE")))
```

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
imback82: same here

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -144,11 +144,36 @@ class IndexLogManagerImplTest extends SparkFunSuite with SparkInvolvedSuite with
...
+    assert(result === true)
     assert(fs.exists(new Path(path, s"$HYPERSPACE_LOG/latestStable")))
   }
 
+  test("testUpdateLatestStableLog fails if log state is not stable") {
+    val path = new Path(testRoot, UUID.randomUUID().toString)
+    val fs = path.getFileSystem(new Configuration)
+    FileUtils.createFile(
+      fs, new Path(path, s"$HYPERSPACE_LOG/0"), JsonUtils.toJson(getEntry("CANCELLING"))
+    )
```
imback82, line 160: ditto

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -144,11 +144,36 @@ class IndexLogManagerImplTest extends SparkFunSuite with SparkInvolvedSuite with
...
+
+  test("testUpdateLatestStableLog fails with exception if unable to find a valid log entry") {
+    val path = new Path(testRoot, UUID.randomUUID().toString)
+    val fs = path.getFileSystem(new Configuration)
+    FileUtils.createFile(
+      fs, new Path(path, s"$HYPERSPACE_LOG/0"), "Invalid Log Entry"
+    )
+    assertThrows[com.fasterxml.jackson.core.JsonParseException](
+      new IndexLogManagerImpl(path).createLatestStableLog(0)
+    )
```
imback82, line 174: ditto

saucam: done

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
saucam: done

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
saucam: done

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
saucam: ok done

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
saucam: added

saucam:
@imback82 sorry for so many style issues!! All fixed

apoorvedave1:
LGTM, thanks @saucam

saucam:
Thanks!!

PR #94, opened 2020-07-22
CSharp "Hello World" application

AFFogarty:
This change is a recreation of #90 in CSharp. It addresses issue #79 .

To provide a sample application in .NET.

No.

This change consists of an standalone example application and is verified by running it manually and validating the output and its behavior.


examples/csharp/HyperspaceApp/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            // Create Spark session.
+            SparkSession spark = SparkSession
+                .Builder()
+                .AppName("Hyperspace example")
+                .Config("spark.some.config.option", "some-value")
```
apoorvedave1, line 33: remove this?

examples/csharp/HyperspaceApp/.gitignore
```diff
@@ -0,0 +1,369 @@
+## Ignore Visual Studio temporary files, build results, and
```
apoorvedave1, line 1: can we merge this .gitignore file to <root>/.gitignore ? Unless it's required here?

examples/csharp/HyperspaceApp/Program.cs
AFFogarty: This is copied from the Scala example:  https://github.com/microsoft/hyperspace/blob/master/examples/scala/src/main/scala/App.scala#L29

examples/csharp/HyperspaceApp/.gitignore
AFFogarty: Hmm... the rules in this file are all specific to Visual Studio / .NET so I'm wondering if we should keep them separate?

imback82: I am fine leaving these separate (maybe under csharp)?

examples/csharp/HyperspaceApp/Program.cs
pirz: @apoorvedave1 This follows what Spark examples do. Check [here](https://github.com/apache/spark/blob/master/examples/src/main/scala/org/apache/spark/examples/sql/SparkSQLExample.scala#L39) as an example.

pirz:
@AFFogarty LGTM.

examples/csharp/HyperspaceApp/.gitignore
AFFogarty: Moved to `csharp` folder.

imback82:
Should we update the azure pipeline to make sure this compiles? :)

examples/csharp/HyperspaceApp/Program.cs
apoorvedave1: oh interesting. Thanks!

apoorvedave1:
LGTM, Thanks @AFFogarty

imback82:
Looks like the Scala example #90 also missed setting up the build. I will create an issue and we can do a follow up PR.

Issue #29, opened 2020-06-24
Incremental Indexing: Support for incremental changes to index when new data is added


apoorvedave1:
### Describe the issue
Currently when new data is added, the indexes become obsolete. The only way to make them usable is through the `refresh()` api which recreates complete index from scratch. Add functionality to support creation of indexes only on additional new data instead of recreating from scratch.

### Expected behavior
New additional data to old data doesn't require rebuilding of indexes on complete data. Only additional data gets indexed

Intermediate issues and PRs
 * [x]  Capture source file info to detect the change of the dataset #120 
 * [x]  The root paths captured in index log entry should always be absolute paths #126 
 * [x]  Store index file names and file sizes in operation log #107
 * [x]  Add full list of index files to index log entry #103
 * [x]  Index metadata should always store absolute paths (alternate solution) #128, #127 
 * [x]  Store file name, size and modified time of index files in metadata #122
 * [x]  New Tree based Structure for index metadata #139 
 * [x]  Index merge functionality PR: #162 
 * [x]  Incremental indexing support for Append-only data #110 , PR: #163

imback82:
Can you update the description so that the current `refresh` does a full rebuild?

apoorvedave1:
> Can you update the description so that the current `refresh` does a full rebuild?

added description mentioning current `refresh()` does a full rebuild.

apoorvedave1:
closed by #163

PR #99, opened 2020-07-29
Replace LogicalPlanSerDeUtils

### LINKED ISSUE #98: KryoSerializer is not compatible with the different versions of Spark
### Describe the issue
KryoSerializer is not fully compatible with the different versions of Spark. It's used to serialize/deserialize the logical plan of Index to support `refresh` operation. This problem will be addressed more generally with #95.
For now, we should replace the ser/de implementation based on the current limitation - Covering Index on LogicalRelation/HadoopFsRelation.

### To Reproduce

1. create an index with Spark 2.4.6
2. `hs.refreshIndex()` or `hs.indexes.show` with Spark 3.0.0

```
scala> hs.refreshIndex("testindex")
com.esotericsoftware.kryo.KryoException: java.lang.ArrayIndexOutOfBoundsException
Serialization trace:
catalogTable (com.microsoft.hyperspace.index.serde.package$LogicalRelationWrapper2)
  at com.esotericsoftware.kryo.serializers.ObjectField.read(ObjectField.java:144)
  at com.esotericsoftware.kryo.serializers.FieldSerializer.read(FieldSerializer.java:543)
  at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:813)
  at com.microsoft.hyperspace.index.serde.KryoSerDeUtils$.deserialize(KryoSerDeUtils.scala:60)
  at com.microsoft.hyperspace.index.serde.LogicalPlanSerDeUtils$.deserialize(LogicalPlanSerDeUtils.scala:71)
  at com.microsoft.hyperspace.actions.RefreshAction.df$lzycompute(RefreshAction.scala:48)
  at com.microsoft.hyperspace.actions.RefreshAction.df(RefreshAction.scala:46)
  at com.microsoft.hyperspace.actions.RefreshAction.logEntry$lzycompute(RefreshAction.scala:59)
  at com.microsoft.hyperspace.actions.RefreshAction.logEntry(RefreshAction.scala:58)
  at com.microsoft.hyperspace.actions.RefreshAction.event(RefreshAction.scala:81)
  at com.microsoft.hyperspace.actions.Action.run(Action.scala:98)
  at com.microsoft.hyperspace.actions.Action.run$(Action.scala:83)
  at com.microsoft.hyperspace.actions.RefreshAction.run(RefreshAction.scala:31)
  at com.microsoft.hyperspace.index.IndexCollectionManager.$anonfun$refresh$1(IndexCollectionManager.scala:75)
  at com.microsoft.hyperspace.index.IndexCollectionManager.$anonfun$refresh$1$adapted(IndexCollectionManager.scala:72)
  at com.microsoft.hyperspace.index.IndexCollectionManager.withLogManager(IndexCollectionManager.scala:131)
  at com.microsoft.hyperspace.index.IndexCollectionManager.refresh(IndexCollectionManager.scala:72)
  at com.microsoft.hyperspace.index.CachingIndexCollectionManager.refresh(CachingIndexCollectionManager.scala:97)
  at com.microsoft.hyperspace.Hyperspace.refreshIndex(Hyperspace.scala:77)
  ... 51 elided
Caused by: java.lang.ArrayIndexOutOfBoundsException
```

sezruby:

Instead of using KryoSerializer, reconstruct a DataFrame with {`rootPaths`, `dataSchema`}. In order to do this, we need to keep the necessary info in IndexLogEntry

Note that this approach can only be applied to Covering Index & HadoopFsRelation; limitations of the current version of Hyperspace.

KryoSerializer is not compatible with the different versions of Spark; so refreshing an index built with another spark version would fail.

Fixes #98 

Yes, indexes created before this change should be removed & recreated. 
IndexLogEntry (metadata) will be changed; new `"relations"` field:
```
...
"source" : {
    "plan" : {
      "properties" : {
        "relations" : [ {
          "rootPaths" : [ "file:/<>/table2" ],
          "data" : {
            "properties" : {
              "content" : {
                "root" : "",
                "directories" : [ {
                  "path" : "",
                  "files" : [ "file:/<>/table2/part-00000-2547e7e7-2577-4e07-b796-0a36e3e5444e-c000.snappy.parquet", "file:/<>/table2/part-00001-2547e7e7-2577-4e07-b796-0a36e3e5444d-c000.snappy.parquet" ],
                  "fingerprint" : {
                    "kind" : "NoOp",
                    "properties" : { }
                  }
                } ]
              }
            },
            "kind" : "HDFS"
          },
          "dataSchemaJson" : "{\"type\":\"struct\",\"fields\":[{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}},{\"name\":\"name\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}},{\"name\":\"age\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}]}",
          "fileFormat" : "parquet"
        } ],
        "rawPlan" : null,
        "sql" : null,
        "fingerprint" : {
          "properties" : {
            "signatures" : [ {
              "provider" : "com.microsoft.hyperspace.index.IndexSignatureProvider",
              "value" : "921391350ed46e5ac8761bc4fc5a466a"
            } ]
          },
          "kind" : "LogicalPlan"
        }
      },
      "kind" : "Spark"
    }
```

refreshIndex tested both spark 2.4.6 + scala 2.11 <=> spark 3.0.0 + scala 2.12 locally

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -57,7 +57,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
     signatureProvider.signature(df.queryExecution.optimizedPlan) match {
       case Some(s) =>
         val relations = sourceRelations(df)
-        // Currently we only support to create an index on a LogicalRelation
+        // Currently we only support to create an index on a LogicalRelation.
         assert(relations.size == 1)
 
         val sourcePlanProperties = SparkPlan.Properties(
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -16,6 +16,8 @@
 
 package com.microsoft.hyperspace.actions
 
+import java.util.Locale
+
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
@@ -91,7 +93,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
           location.rootPaths.map(_.toString),
           location.allFiles.map(_.getPath.toString),
           dataSchema.json,
-          fileFormat.toString)
+          fileFormat.toString.toLowerCase(Locale.ROOT))
     }
 
   protected def write(spark: SparkSession, df: DataFrame, indexConfig: IndexConfig): Unit = {
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -93,20 +93,20 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
   protected def sourceFiles(df: DataFrame): Seq[String] =
     df.queryExecution.optimizedPlan.collect {
       case LogicalRelation(
-      HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
-      _,
-      _,
-      _) =>
-      location.allFiles.map(_.getPath.toString)
+          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+          _,
+          _,
+          _) =>
+        location.allFiles.map(_.getPath.toString)
     }.flatten
 
   protected def sourceRelations(df: DataFrame): Seq[RelationMetadataEntry] =
     df.queryExecution.optimizedPlan.collect {
       case LogicalRelation(
-      HadoopFsRelation(location: PartitioningAwareFileIndex, _, dataSchema, _, fileFormat, _),
-      _,
-      _,
-      _) =>
+          HadoopFsRelation(location: PartitioningAwareFileIndex, _, dataSchema, _, fileFormat, _),
+          _,
+          _,
+          _) =>
         RelationMetadataEntry(
           location.allFiles.map(_.getPath.toString),
           dataSchema.json,
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -61,10 +61,13 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
           LogicalPlanFingerprint(
             LogicalPlanFingerprint.Properties(Seq(Signature(signatureProvider.name, s)))))
 
+        val relations = sourceRelations(df)
+        val allFiles = relations.map(_.files).flatten
+
         // Note: Source files are fingerprinted as part of the serialized logical plan currently.
         val sourceDataProperties =
           Hdfs.Properties(
-            Content("", Seq(Content.Directory("", sourceRelations(df), NoOpFingerprint()))))
+            Content("", Seq(Content.Directory("", allFiles, NoOpFingerprint()))))
 
         IndexLogEntry(
           indexConfig.indexName,
@@ -75,7 +78,8 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
               IndexLogEntry.schemaString(schema),
               numBuckets)),
           Content(path.toString, Seq()),
-          Source(SparkPlan(sourcePlanProperties), Seq(Hdfs(sourceDataProperties))),
+          Source(
+            SparkPlan(sourcePlanProperties), Seq(Hdfs(sourceDataProperties)), relations),
           Map())
 
       case None => throw HyperspaceException("Invalid plan for creating an index.")
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -40,8 +40,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       spark: SparkSession,
       df: DataFrame,
       indexConfig: IndexConfig,
-      path: Path,
-      sourceFiles: Seq[String]): IndexLogEntry = {
+      path: Path): IndexLogEntry = {
     val numBuckets = spark.sessionState.conf
       .getConfString(
         IndexConstants.INDEX_NUM_BUCKETS,
@@ -58,8 +57,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
     // Here we use the unanalyzed logical plan returned by the parser, before analysis and
     // optimization. Analyzed and optimized plans have additional ser/de issues that are not
     // fully covered by the current ser/de framework.
-    val serializedPlan =
-      LogicalPlanSerDeUtils.serialize(df.queryExecution.logical, spark)
+    val serializedPlan = LogicalPlanSerDeUtils.serialize(df.queryExecution.logical, spark)
 
     signatureProvider.signature(df.queryExecution.optimizedPlan) match {
       case Some(s) =>
@@ -70,7 +68,11 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
 
         // Note: Source files are fingerprinted as part of the serialized logical plan currently.
         val sourceDataProperties =
-          Hdfs.Properties(Content("", Seq(Content.Directory("", sourceFiles, NoOpFingerprint()))))
+          Hdfs.Properties(
+            Content(
+              "",
+              Seq(
+                Content.Directory("", sourceFiles(df), sourceRelations(df), NoOpFingerprint()))))
 
         IndexLogEntry(
           indexConfig.indexName,
@@ -91,12 +93,25 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
   protected def sourceFiles(df: DataFrame): Seq[String] =
     df.queryExecution.optimizedPlan.collect {
       case LogicalRelation(
-          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
-          _,
-          _,
-          _) =>
-        location.allFiles.map(_.getPath.toString)
+      HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+      _,
+      _,
+      _) =>
+      location.allFiles.map(_.getPath.toString)
     }.flatten
+
+  protected def sourceRelations(df: DataFrame): Seq[RelationMetadataEntry] =
+    df.queryExecution.optimizedPlan.collect {
+      case LogicalRelation(
+      HadoopFsRelation(location: PartitioningAwareFileIndex, _, dataSchema, _, fileFormat, _),
+      _,
+      _,
+      _) =>
+        RelationMetadataEntry(
+          location.allFiles.map(_.getPath.toString),
+          dataSchema.json,
+          fileFormat.toString())
+    }
 
   protected def write(spark: SparkSession, df: DataFrame, indexConfig: IndexConfig): Unit = {
     val numBuckets = spark.sessionState.conf
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -16,11 +16,10 @@
 
 package com.microsoft.hyperspace.actions
 
-import java.util.Locale
-
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
+import org.apache.spark.sql.sources.DataSourceRegister
 
 import com.microsoft.hyperspace.HyperspaceException
 import com.microsoft.hyperspace.index._
@@ -57,17 +56,16 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
 
     signatureProvider.signature(df.queryExecution.optimizedPlan) match {
       case Some(s) =>
+        val relations = sourceRelations(df)
+        // Currently we only support to create an index on a LogicalRelation
+        assert(relations.size == 1)
+
         val sourcePlanProperties = SparkPlan.Properties(
+          relations,
+          null,
+          null,
           LogicalPlanFingerprint(
             LogicalPlanFingerprint.Properties(Seq(Signature(signatureProvider.name, s)))))
-
-        val relations = sourceRelations(df)
-        val allFiles = relations.map(_.files).flatten
-
-        // Note: Source files are fingerprinted as part of the serialized logical plan currently.
-        val sourceDataProperties =
-          Hdfs.Properties(
-            Content("", Seq(Content.Directory("", allFiles, NoOpFingerprint()))))
 
         IndexLogEntry(
           indexConfig.indexName,
@@ -79,25 +77,33 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
               numBuckets)),
           Content(path.toString, Seq()),
           Source(
-            SparkPlan(sourcePlanProperties), Seq(Hdfs(sourceDataProperties)), relations),
+            SparkPlan(sourcePlanProperties)),
           Map())
 
       case None => throw HyperspaceException("Invalid plan for creating an index.")
     }
   }
 
-  protected def sourceRelations(df: DataFrame): Seq[RelationMetadataEntry] =
+  protected def sourceRelations(df: DataFrame): Seq[Relation] =
     df.queryExecution.optimizedPlan.collect {
       case LogicalRelation(
-          HadoopFsRelation(location: PartitioningAwareFileIndex, _, dataSchema, _, fileFormat, _),
-          _,
-          _,
-          _) =>
-        RelationMetadataEntry(
+      HadoopFsRelation(location: PartitioningAwareFileIndex, _, dataSchema, _, fileFormat, _),
+      _,
+      _,
+      _) =>
+        val files = location.allFiles.map(_.getPath.toString)
+        // Note: Source files are fingerprinted as part of the serialized logical plan currently.
+        val sourceDataProperties =
+          Hdfs.Properties(Content("", Seq(Content.Directory("", files, NoOpFingerprint()))))
+        val fileFormatName = fileFormat match {
+          case d: DataSourceRegister => d.shortName
+          case f: Any => throw HyperspaceException(s"Unsupported file format $f")
+        }
+        Relation(
           location.rootPaths.map(_.toString),
-          location.allFiles.map(_.getPath.toString),
+          Hdfs(sourceDataProperties),
           dataSchema.json,
-          fileFormat.toString.toLowerCase(Locale.ROOT))
+          fileFormatName)
     }
 
   protected def write(spark: SparkSession, df: DataFrame, indexConfig: IndexConfig): Unit = {
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -86,7 +86,13 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
   protected def sourceRelations(df: DataFrame): Seq[Relation] =
     df.queryExecution.optimizedPlan.collect {
       case LogicalRelation(
-          HadoopFsRelation(location: PartitioningAwareFileIndex, _, dataSchema, _, fileFormat, _),
+          HadoopFsRelation(
+            location: PartitioningAwareFileIndex,
+            _,
+            dataSchema,
+            _,
+            fileFormat,
+            options),
           _,
           _,
           _) =>
@@ -103,7 +109,8 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
           location.rootPaths.map(_.toString),
           Hdfs(sourceDataProperties),
           dataSchema.json,
-          fileFormatName)
+          fileFormatName,
+          options)
     }
 
   protected def write(spark: SparkSession, df: DataFrame, indexConfig: IndexConfig): Unit = {
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -23,7 +23,6 @@ import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRela
 import com.microsoft.hyperspace.HyperspaceException
 import com.microsoft.hyperspace.index._
 import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
-import com.microsoft.hyperspace.index.serde.LogicalPlanSerDeUtils
 
 /**
  * CreateActionBase provides functionality to write dataframe as covering index.
@@ -54,25 +53,16 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       df.select(allColumns.head, allColumns.tail: _*).schema
     }
 
-    // Here we use the unanalyzed logical plan returned by the parser, before analysis and
-    // optimization. Analyzed and optimized plans have additional ser/de issues that are not
-    // fully covered by the current ser/de framework.
-    val serializedPlan = LogicalPlanSerDeUtils.serialize(df.queryExecution.logical, spark)
-
     signatureProvider.signature(df.queryExecution.optimizedPlan) match {
       case Some(s) =>
         val sourcePlanProperties = SparkPlan.Properties(
-          serializedPlan,
           LogicalPlanFingerprint(
             LogicalPlanFingerprint.Properties(Seq(Signature(signatureProvider.name, s)))))
 
         // Note: Source files are fingerprinted as part of the serialized logical plan currently.
         val sourceDataProperties =
           Hdfs.Properties(
-            Content(
-              "",
-              Seq(
-                Content.Directory("", sourceFiles(df), sourceRelations(df), NoOpFingerprint()))))
+            Content("", Seq(Content.Directory("", sourceRelations(df), NoOpFingerprint()))))
 
         IndexLogEntry(
           indexConfig.indexName,
@@ -90,16 +80,6 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
     }
   }
 
-  protected def sourceFiles(df: DataFrame): Seq[String] =
-    df.queryExecution.optimizedPlan.collect {
-      case LogicalRelation(
-          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
-          _,
-          _,
-          _) =>
-        location.allFiles.map(_.getPath.toString)
-    }.flatten
-
   protected def sourceRelations(df: DataFrame): Seq[RelationMetadataEntry] =
     df.queryExecution.optimizedPlan.collect {
       case LogicalRelation(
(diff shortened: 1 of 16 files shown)


sezruby:
TODO: test fix & add

@imback82 Could you do a code review simply? and I'm not sure we should keep the previous version for backward compatibility or not..

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -70,7 +68,11 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
 
         // Note: Source files are fingerprinted as part of the serialized logical plan currently.
         val sourceDataProperties =
-          Hdfs.Properties(Content("", Seq(Content.Directory("", sourceFiles, NoOpFingerprint()))))
+          Hdfs.Properties(
+            Content(
+              "",
+              Seq(
+                Content.Directory("", sourceFiles(df), sourceRelations(df), NoOpFingerprint()))))
```
sezruby, line 75: seems need to retrieve & keep the root path and use the path for spark.read.load at RefreshAction
let me update this + clean up the previous ver.

imback82: Can you clarify this?

If `df` was created from `spark.read.parquet("/tmp/1/1.parquet", "/tmp/2/2.parquet")`, what will be the root path used for refresh? (Note that there could be `/tmp/1/3.parquet`).

cc: @apoorvedave1

sezruby: I'm going to collect the root paths by using `location.rootPaths.map(_.toString),` and add "rootPaths: Seq[String]" to RelationMetadataEntry.

imback82: Ah OK. You meant `rootPaths` from `FileIndex`.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -168,7 +168,7 @@ private[hyperspace] object IndexSummary {
       entry.numBuckets,
       entry.derivedDataset.properties.schemaString,
       entry.content.root,
-      entry.plan(spark).toString,
+      entry.derivedDataset.properties.schemaString, // TODO
```
sezruby, line 171: @imback82 which property would be here? or just remove "queryPlan"?
``` * TODO: Finalize about adding these: data location, signatures, file lists etc.```

Comment (imback82) [reply src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala:171]: I think we can just remove it for now (and add additional info based on feedback)

Comment (sezruby) [reply src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala:171]: ok I'll remove this along with the next commit :)

And I think it's better not to remove com/microsoft/hyperspace/index/serde and its test as the test seems useful later ..?

Comment (imback82) [reply src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala:171]: yea, we can remove it as a separate PR if needed.

File: src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala [hunk 3]
```diff
@@ -42,12 +42,11 @@ class RefreshAction(
...
-  // Deserialize the plan and create a df.
+  // Reconstruct a df from schema
   private lazy val df = {
-    val serializedPlan = previousIndexLogEntry.source.plan.properties.rawPlan
-    val plan = LogicalPlanSerDeUtils.deserialize(serializedPlan, spark)
-    val qe = spark.sessionState.executePlan(plan)
-    new Dataset[Row](spark, plan, RowEncoder(qe.analyzed.schema))
+    val rel = previousIndexLogEntry.relation
+    val dataSchema = DataType.fromJson(rel.dataSchemaJson).asInstanceOf[StructType]
+    spark.read.schema(dataSchema).load(rel.rootPaths: _*)
```
apoorvedave1, line 49: seems like default data source is parquet ([link](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L898)).
Could you add file format as well to this statement (to support csv etc.)

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -16,23 +16,31 @@
...
 
 // IndexLogEntry-specific fingerprint to be temporarily used where fingerprint is not defined.
 case class NoOpFingerprint() {
   val kind: String = "NoOp"
   val properties: Map[String, String] = Map()
 }
 
+case class RelationMetadataEntry(
+    rootPaths: Seq[String],
+    files: Seq[String],
```
apoorvedave1, line 33: I couldn't find this `files` variable used. Is it used somewhere or it can be removed?

src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
sezruby: Thanks for pointing out this! I updated the pr and added a test for it.

sezruby:
@ reviewers, Should I modify the below CreateActionEvent log accordingly?

```scala
   final override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
    // LogEntry instantiation may fail if index config is invalid. Hence the 'Try'.
    val index = Try(logEntry.asInstanceOf[IndexLogEntry]).toOption
    CreateActionEvent(appInfo, indexConfig, index, df.queryExecution.logical.toString, message)
  }```

File: src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala [hunk 5]
```diff
@@ -16,23 +16,31 @@
...
+    dataSchemaJson: String,
+    fileFormat: String)
```
imback82, line 35: The source info from which index is created such as `dataSchemaJson`, `fileFormat`, etc. should be stored somewhere under `Source`. `Content` is a generic representation of directories (used by both `IndexLogEntry` and `Hdfs`, for example).

src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
```diff
@@ -42,12 +42,11 @@ class RefreshAction(
...
-  // Deserialize the plan and create a df.
+  // Reconstruct a df from schema
   private lazy val df = {
-    val serializedPlan = previousIndexLogEntry.source.plan.properties.rawPlan
-    val plan = LogicalPlanSerDeUtils.deserialize(serializedPlan, spark)
-    val qe = spark.sessionState.executePlan(plan)
-    new Dataset[Row](spark, plan, RowEncoder(qe.analyzed.schema))
+    val rel = previousIndexLogEntry.relation
+    val dataSchema = DataType.fromJson(rel.dataSchemaJson).asInstanceOf[StructType]
+    spark.read.schema(dataSchema).format(rel.fileFormat).load(rel.rootPaths: _*)
```
imback82, line 49: What will be the rootPaths in [this](https://github.com/microsoft/hyperspace/pull/99#discussion_r462728758) example?

sezruby: It seems it depends on how the df created:

```
scala> val df2 = spark.read.parquet("/Users/eunjin/repo/spark/bin/testdir/2/b.snappy.parquet","/Users/eunjin/repo/spark/bin/testdir/2/c.snappy.parquet", "/Users/eunjin/repo/spark/bin/testdir/1/a.snappy.parquet")
df2: org.apache.spark.sql.DataFrame = [id: int, name: string ... 1 more field]
==
"rootPaths" : [ "file:/Users/eunjin/repo/spark/bin/testdir/2/b.snappy.parquet", "file:/Users/eunjin/repo/spark/bin/testdir/2/c.snappy.parquet", "file:/Users/eunjin/repo/spark/bin/testdir/1/a.snappy.parquet" ],
```

```
scala> val df2 = spark.read.parquet("/Users/eunjin/repo/spark/bin/testdir/2", "/Users/eunjin/repo/spark/bin/testdir/1/a.snappy.parquet")
df2: org.apache.spark.sql.DataFrame = [id: int, name: string ... 1 more field]
==
 "rootPaths" : [ "file:/Users/eunjin/repo/spark/bin/testdir/2", "file:/Users/eunjin/repo/spark/bin/testdir/1/a.snappy.parquet" ],
"files" : [ "file:/Users/eunjin/repo/spark/bin/testdir/2/b.snappy.parquet", "file:/Users/eunjin/repo/spark/bin/testdir/2/c.snappy.parquet", "file:/Users/eunjin/repo/spark/bin/testdir/1/a.snappy.parquet" ],
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: Then how about "source" - `locations` - "content" - ...

other options for `locations`: baseRelation, baseDirectories, ..?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -61,10 +61,13 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
           LogicalPlanFingerprint(
             LogicalPlanFingerprint.Properties(Seq(Signature(signatureProvider.name, s)))))
 
+        val relations = sourceRelations(df)
+        val allFiles = relations.map(_.files).flatten
```
apoorvedave1, line 65: `.flatmap(...)`?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -81,22 +78,27 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+          Source(
+            SparkPlan(sourcePlanProperties), Seq(Hdfs(sourceDataProperties)), relations),
           Map())
 
       case None => throw HyperspaceException("Invalid plan for creating an index.")
     }
   }
 
-  protected def sourceFiles(df: DataFrame): Seq[String] =
+  protected def sourceRelations(df: DataFrame): Seq[RelationMetadataEntry] =
```
imback82, line 89: We support covering index from a single source relation. Should we just return `RelationMetadataEntry`. Multiple relations are needed for indexed views, but they cannot work if we remove the serialized df anyway.
Thoughts? @apoorvedave1 / @pirz / @rapoth

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -81,22 +78,27 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
           _,
           _,
           _) =>
-        location.allFiles.map(_.getPath.toString)
-    }.flatten
+        RelationMetadataEntry(
+          location.rootPaths.map(_.toString),
+          location.allFiles.map(_.getPath.toString),
+          dataSchema.json,
+          fileFormat.toString.toLowerCase(Locale.ROOT))
```
imback82, line 100: `fileFormat.shortName` is better since it's an abstract method? (also used in `DataSource.lookupDataSource`)

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -16,19 +16,23 @@
...
 import com.microsoft.hyperspace.actions.Constants
-import com.microsoft.hyperspace.index.serde.LogicalPlanSerDeUtils
 
 // IndexLogEntry-specific fingerprint to be temporarily used where fingerprint is not defined.
 case class NoOpFingerprint() {
   val kind: String = "NoOp"
   val properties: Map[String, String] = Map()
 }
 
+case class RelationMetadataEntry(
```
imback82, line 30: Just `RelationMetadata`? Also, I think this should capture only the info to recreate the DF, meaning `files` don't need to be here.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -74,7 +78,7 @@ object Hdfs {
 }
 
 // IndexLogEntry-specific Source that uses SparkPlan as a plan and Hdfs as data.
-case class Source(plan: SparkPlan, data: Seq[Hdfs])
+case class Source(plan: SparkPlan, data: Seq[Hdfs], relations: Seq[RelationMetadataEntry])
```
imback82, line 81: how about putting this under `SparkPlan.Properties` (same place as `rawPlan`)?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
apoorvedave1: I had the same thought but incremental refresh (https://github.com/microsoft/hyperspace/issues/29) will require `files` as well. Do we want to tackle it later?

imback82: `files` will still be stored in `Source.data` right? I think the current approach is duplicating the `files`.

apoorvedave1: got it. you are right. Thanks @imback82

imback82: Actually, I had a discussion with @rapoth as well and came up with the following:
```
case class Relation(
    rootPaths: Seq[String],
    data: Hdfs,
    dataSchemaJson: String,
    fileFormat: String)

object SparkPlan {
  case class Properties(
    relations: Seq[Relation], // To support indexed views which can contain multiple relations.
    rawPlan: String, // null for now (will be used when we have IR to serialize a plan)
    sql: String, // null for now (will be used when we capture SQL)
    fingerprint: LogicalPlanFingerprint)
}

case class Source(plan: SparkPlan)
```

WDYT @sezruby?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: This is not relevant after https://github.com/microsoft/hyperspace/pull/99/files#r463881967

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
imback82: This is not relevant after https://github.com/microsoft/hyperspace/pull/99/files#r463881967

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -81,22 +78,27 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    }
```
imback82, line 101: Can we assert the size is 1 here until we introduce indexed views?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: Looks good and it's a good idea to define new fields in advance 😀

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
sezruby: Yes but seems it's not accessible with FileFormat..?

imback82: Can you try:
```scala
fileFormat match {
  case d: DataSourceRegister => d.shortName
  case _ => throw HyperspaceException("blah")
}
```
?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -96,8 +100,12 @@ case class IndexLogEntry(
 
   def numBuckets: Int = derivedDataset.properties.numBuckets
 
-  def plan(spark: SparkSession): LogicalPlan = {
-    LogicalPlanSerDeUtils.deserialize(source.plan.properties.rawPlan, spark)
+  def relation: RelationMetadataEntry = source.relations match {
```
imback82, line 103: This can just return `relations` and the caller side can do the check. (The `def`s here are mostly aliases to provide easy accesses to fields)

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
sezruby: It woks well thanks :)

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -81,22 +76,35 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
               IndexLogEntry.schemaString(schema),
               numBuckets)),
           Content(path.toString, Seq()),
-          Source(SparkPlan(sourcePlanProperties), Seq(Hdfs(sourceDataProperties))),
+          Source(
+            SparkPlan(sourcePlanProperties)),
```
imback82, line 80: can you run "reformat code"? (Did you configure your IDE to use /dev/.scalafmt.conf?)

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -81,22 +76,35 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
-          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
-          _,
-          _,
-          _) =>
-        location.allFiles.map(_.getPath.toString)
-    }.flatten
+      HadoopFsRelation(location: PartitioningAwareFileIndex, _, dataSchema, _, fileFormat, _),
+      _,
+      _,
+      _) =>
```
imback82, line 93: I think reformat should help here as well

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -81,22 +76,35 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+        val files = location.allFiles.map(_.getPath.toString)
+        // Note: Source files are fingerprinted as part of the serialized logical plan currently.
```
imback82, line 95: Can we update this now that we don't have a serialized plan?
`// Note that source files are currently fingerprinted when the optimized plan is fingerprinted by LogicalPlanFingerprint.`

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -81,22 +76,35 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+        val sourceDataProperties =
+          Hdfs.Properties(Content("", Seq(Content.Directory("", files, NoOpFingerprint()))))
+        val fileFormatName = fileFormat match {
+          case d: DataSourceRegister => d.shortName
+          case f: Any => throw HyperspaceException(s"Unsupported file format $f")
```
imback82, line 100: nit: `case other => throw HyperspaceException(s"Unsupported file format: $other")`

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -110,7 +123,8 @@ class IndexLogEntryTest extends SparkFunSuite {
           schema.json,
           200)),
       Content("rootContentPath", Seq()),
-      Source(SparkPlan(expectedSourcePlanProperties), Seq(Hdfs(expectedSourceDataProperties))),
+      Source(
+        SparkPlan(expectedSourcePlanProperties)),
```
imback82, line 127: can you also run "reformat code" in this file?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -223,6 +224,44 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
+    val refreshTestLocation = sampleParquetDataLocation + "refresh2"
+    FileUtils.delete(new Path(refreshTestLocation))
+    import spark.implicits._
+    SampleData.testData
+      .toDF("Date", "RGUID", "Query", "imprs", "clicks")
+      .limit(10)
+      .write
+      .option("header", "true")
+      .csv(refreshTestLocation)
+    val df = spark.read.option("header", "true").csv(refreshTestLocation)
```
imback82, line 237: looks like this is the only difference compared to the parquet test? If so can we refactor common code or do
```scala
Seq(("parquet", None), ("csv", Some(Map("header" -> "true"))).foreach {
  case (format, option) =>
    ...
}
```
Either one is fine as long as duplicate code is removed.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -231,22 +270,33 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
+                location: PartitioningAwareFileIndex, _, dataSchema, _, fileFormat, _),
+              _,
+              _,
+              _) =>
+                val files = location.allFiles.map(_.getPath.toString)
+                val sourceDataProperties =
+                  Hdfs.Properties(Content("", Seq(Content.Directory("", files, NoOpFingerprint()))))
+                val fileFormatName = fileFormat match {
+                  case d: DataSourceRegister => d.shortName
+                  case f: Any => throw HyperspaceException(s"Unsupported file format $f")
```
imback82, line 285: ditto

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
sezruby: While trying to fix this test, I realized that we also need to capture `options` field.
Let me know if there is any other field of HadoopFsRelation that might be useful later..?

```
case class HadoopFsRelation(
    location: FileIndex,
    partitionSchema: StructType,
    dataSchema: StructType,
    bucketSpec: Option[BucketSpec],
    fileFormat: FileFormat,
    options: Map[String, String])(val sparkSession: SparkSession)
```

imback82: Good catch on capturing `options`! I think that should be all. `partitionSchema` would be populated based on the `rootPaths` during read, so I think we should be good. @apoorvedave1 can double-confirm.

src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
```diff
@@ -42,12 +42,19 @@ class RefreshAction(
...
   private lazy val df = {
-    val serializedPlan = previousIndexLogEntry.source.plan.properties.rawPlan
-    val plan = LogicalPlanSerDeUtils.deserialize(serializedPlan, spark)
-    val qe = spark.sessionState.executePlan(plan)
-    new Dataset[Row](spark, plan, RowEncoder(qe.analyzed.schema))
+    val rels = previousIndexLogEntry.relations
+    // Currently we only support to create an index on a LogicalRelation
+    assert(rels.size == 1)
+    val dataSchema = DataType.fromJson(rels.head.dataSchemaJson).asInstanceOf[StructType]
+    // "path" key in options incurs to read data twice unexpectedly
```
imback82, line 51: nit: can we end a sentence with "."? Btw, did you meaning listing files (not reading data)`// Remove "path" key in options to prevent listing files twice.`?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -187,40 +188,49 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
-    indexCount = spark.read
-      .parquet(s"$indexStorageLocation/index1" +
-        s"/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=1")
-      .count()
-
-    // Check if index got updated
-    assert(indexCount == 3)
-
-    FileUtils.delete(new Path(refreshTestLocation))
+
```
imback82, line 191: nit: remove empty line

src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
imback82: Oh I see. The data is actually read twice.

imback82: Ah, I see what's happening and I think this could be a bug in Spark. I have a potential fix and will follow up on this.

```
// Both are same and will read the data twice.
spark.read.option("path", "/tmp/parquet_data/").parquet("/tmp/parquet_data/").show
spark.read.parquet("/tmp/parquet_data/", "/tmp/parquet_data/").show
// But the following is fine:
spark.read.option("path", "/tmp/parquet_data/").format("parquet").load("/tmp/parquet_data/").show
```

Putting the above behavior aside, if the user created a `df` with:
```
// Intention is to read three times.
val df = spark.read.option("path", "/tmp/parquet_data").format("parquet").load("/tmp/parquet_data/", "/tmp/parquet_data/")
```
If the user intention was reading the data three times, we can't remove "path" in option?

sezruby: I think this could be a bug. In our refresh-full test, only the index of csv format has the "path" key in its options even I read it without "path" option.

sezruby: And it seems `rootPaths` are stored correctly if the user put multiple paths at load:
```
val df = spark.read.option("path", "/tmp/parquet_data").format("parquet").load("/tmp/parquet_data/", "/tmp/parquet_data/")
```
In this case `rootPaths` might be ["/tmp/parquet_data", "/tmp/parquet_data/", "tmp/parquet_data"]

So I think excluding "path" key here would be okay.

sezruby: There is an inconsistent behavior (Spark 2.4.6):

```
scala> val df = spark.read.option("path","table2").load("table2","table2")
df: org.apache.spark.sql.DataFrame = [id: int, name: string ... 1 more field]

scala> df.show
+---+-----+---+
| id| name|age|
+---+-----+---+
|  1|name1|  4|
|  2|name2|  3|
|  1|name1|  4|
|  2|name2|  3|
|  1|name1|  4|
|  2|name2|  3|
+---+-----+---+


scala> val df = spark.read.option("path","table2").load("table2")
df: org.apache.spark.sql.DataFrame = [id: int, name: string ... 1 more field]

scala> df.show
+---+-----+---+
| id| name|age|
+---+-----+---+
|  1|name1|  4|
|  2|name2|  3|
+---+-----+---+
```

imback82: > There is an inconsistent behavior (Spark 2.4.6/3.0.0):

Yup, that's what I meant [here](https://github.com/microsoft/hyperspace/pull/99#discussion_r464039124), which I am creating a PR against OSS Spark.

Here are the follow up action items:
1. Should we move removing `path` to the creation side so that it's not in the log entry? The `path` "can" be added by [Spark](https://github.com/apache/spark/blob/71aea02e9ffb0c6f7c72c91054c2a4653e22e801/sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala#L232) (not by user), so it is not useful for debugability anyway. Let me know if there is a use case for storing `path` in the log entry.
2. Can we add a test that tests against all the scenarios (useful for reference as well)?
```
spark.read.format("parquet").load("f1")
spark.read.format("parquet").load("f1", "f2")
spark.read.format("parquet").option("path", "blah").load("f1")
spark.read.format("parquet").option("path", "blah").load("f1", "f2")
spark.read.parquet("f1")
spark.read.parquet("f1", "f2")
spark.read.option("path", "blah").parquet("f1")
spark.read.option("path", "blah").parquet("f1", "f2")
```

imback82:
@sezruby I merged this PR, but can you address [this](https://github.com/microsoft/hyperspace/pull/99#discussion_r464098937) as a follow up PR? Thanks!

PR #101, opened 2020-08-03
Remove "path" key in DF read options

sezruby:

The follow up PR to address https://github.com/microsoft/hyperspace/pull/99#discussion_r464098937
- Remove "path" key in IndexLogEntry
- Tests for "path" key in options for spark.read

- "path" key in options can be added by "Spark", not only by the user and it might cause confusion. Therefore, it's better not to write it at creation.
- Test enhancement

- "path" key is not included in "options" field in IndexLogEntry

IndexManagerTest , CreateActionTest

### DIFF
--- a/src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
@@ -16,7 +16,7 @@
 
 package com.microsoft.hyperspace.actions
 
-import org.apache.hadoop.fs.Path
+import org.apache.hadoop.fs.{FileContext, Path}
 import org.apache.spark.SparkFunSuite
 import org.apache.spark.sql.DataFrame
 import org.mockito.Mockito._
@@ -35,6 +35,10 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite {
 
   private val mockLogManager: IndexLogManager = mock(classOf[IndexLogManager])
   private val mockDataManager: IndexDataManager = mock(classOf[IndexDataManager])
+
+  object CreateActionBaseMock extends CreateActionBase(mockDataManager) {
+    def getSourceRelations(df: DataFrame): Seq[Relation] = sourceRelations(df)
+  }
 
   override def beforeAll(): Unit = {
     super.beforeAll()
@@ -103,5 +107,58 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite {
     val ex = intercept[HyperspaceException](action.validate())
     assert(
       ex.getMessage.contains(s"Another Index with name ${indexConfig.indexName} already exists"))
+  }
+
+  test("Verify rootPaths for given LogicalRelations") {
+    val path1 = sampleParquetDataLocation + "table1"
+    val path2 = sampleParquetDataLocation + "table2"
+    FileUtils.delete(new Path(path1))
+    FileUtils.delete(new Path(path2))
+
+    import spark.implicits._
+    SampleData.testData
+      .toDF("Date", "RGUID", "Query", "imprs", "clicks")
+      .limit(2)
+      .write
+      .parquet(path1)
+
+    SampleData.testData
+      .toDF("Date", "RGUID", "Query", "imprs", "clicks")
+      .limit(3)
+      .write
+      .parquet(path2)
+
+    val workDir = FileContext.getFileContext.getWorkingDirectory.toString
+
+    Seq(
+      (spark.read.format("parquet").load(path1), Seq(path1), 2),
+      (spark.read.format("parquet").load(path1, path2), Seq(path1, path2), 5),
+      (spark.read.parquet(path1), Seq(path1), 2),
+      (spark.read.parquet(path1, path2), Seq(path1, path2), 5),
+      (spark.read.parquet(path1, path1, path1), Seq(path1, path1, path1), 6),
+      (spark.read.format("parquet").option("path", path1).load(path1), Seq(path1), 2),
+      (spark.read.format("parquet").option("path", path1).load(path2), Seq(path2), 3),
+      (spark.read.option("path", path1).parquet(path1), Seq(path1, path1), 4),
+      (spark.read.option("path", path1).parquet(path2), Seq(path1, path2), 5),
+      (
+        spark.read.format("parquet").option("path", path1).load(path1, path2),
+        Seq(path1, path1, path2),
+        7),
+      (spark.read.option("path", path1).parquet(path1, path2), Seq(path1, path1, path2), 7))
+      .foreach {
+        case (df, expectedRootPaths, expectedCount) =>
+          val relation = CreateActionBaseMock.getSourceRelations(df)(0)
+          val rootPaths = relation.rootPaths.map {
+            case path if path.startsWith(workDir) => path.drop(workDir.size + 1) // slash
+            case path => path
+          }
+          assert(rootPaths == expectedRootPaths)
+          assert(df.count == expectedCount)
+          assert(relation.options.isDefinedAt("path") == false)
+        case _ => fail("invalid test")
+      }
+
+    FileUtils.delete(new Path(path1))
+    FileUtils.delete(new Path(path2))
   }
 }
--- a/src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
@@ -16,9 +16,10 @@
 
 package com.microsoft.hyperspace.actions
 
-import org.apache.hadoop.fs.{FileContext, Path}
+import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkFunSuite
 import org.apache.spark.sql.DataFrame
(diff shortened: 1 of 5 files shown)


src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -51,7 +51,7 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
       .write
       .parquet(sampleParquetDataLocation)
 
-    df = spark.read.parquet(sampleParquetDataLocation)
+    df = spark.read.option("path", sampleParquetDataLocation).parquet(sampleParquetDataLocation)
```
imback82, line 54: What's the reason for adding this `option` here? I think the new tests you are adding should cover this scenario?

sezruby: Fixed

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -229,6 +229,7 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
         assert(indexCount == 3)
 
         FileUtils.delete(new Path(refreshTestLocation))
+      case _ => assert(false, "invalid test")
```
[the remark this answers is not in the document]
sezruby: Thanks :)

src/test/scala/com/microsoft/hyperspace/actions/RefreshActionTest.scala
```diff
@@ -95,4 +96,43 @@ class RefreshActionTest extends SparkFunSuite with SparkInvolvedSuite {
...
+    // path option ignored
+    assert(spark.read.format("parquet").option("path", path1).load(path1).count == 2)
+    assert(spark.read.format("parquet").option("path", path1).load(path2).count == 3)
+
+    // path option is not ignored
+    assert(spark.read.option("path", path1).parquet(path1).count == 4)
+    assert(spark.read.option("path", path1).parquet(path2).count == 5)
+
+    assert(spark.read.format("parquet").option("path", path1).load(path1, path2).count == 7)
+    assert(spark.read.option("path", path1).parquet(path1, path2).count == 7)
```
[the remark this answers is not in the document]
sezruby: Thanks for the review! I updated PR.

btw don't you think this is a bug in Spark.. ?
```
    // path option ignored
     assert(spark.read.format("parquet").option("path", path1).load(path1).count == 2)
     assert(spark.read.format("parquet").option("path", path1).load(path2).count == 3)

     // path option is not ignored
     assert(spark.read.option("path", path1).parquet(path1).count == 4)
     assert(spark.read.option("path", path1).parquet(path2).count == 5)
```

src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
```diff
@@ -104,4 +108,57 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite {
...
+      (
+        spark.read.format("parquet").option("path", path1).load(path1, path2),
+        Seq(path1, path1, path2),
+        7),
+      (spark.read.option("path", path1).parquet(path1, path2), Seq(path1, path1, path2), 7))
+      .foreach {
+        case (df, expectedRootPaths, expectedCount) =>
+          val relation = CreateActionBaseWrapper.getSourceRelations(df)(0)
+          val rootPaths = relation.rootPaths.map {
+            case path if path.startsWith(workDir) => path.drop(workDir.size + 1) // slash
```
imback82, line 152: Do you need this if you change expectedRootPaths to absolute paths?

src/test/scala/com/microsoft/hyperspace/actions/RefreshActionTest.scala
```diff
@@ -26,6 +26,7 @@ import org.mockito.Mockito.{mock, when}
 import com.microsoft.hyperspace.{HyperspaceException, SampleData, SparkInvolvedSuite}
 import com.microsoft.hyperspace.actions.Constants.States.{ACTIVE, CREATING}
 import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.util.FileUtils
```
imback82, line 29: this can be reverted?

src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
```diff
@@ -104,4 +108,57 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite {
...
+            case path => path
+          }
+          assert(rootPaths == expectedRootPaths)
+          assert(df.count == expectedCount)
+          assert(!relation.options.isDefinedAt("path"))
+        case _ => fail("invalid test")
```
imback82, line 158: I think we can just remove this.

src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
```diff
@@ -104,4 +108,57 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite {
...
+      (spark.read.option("path", path1).parquet(path1), Seq(path1, path1), 4),
+      (spark.read.option("path", path1).parquet(path2), Seq(path1, path2), 5),
+      (
+        spark.read.format("parquet").option("path", path1).load(path1, path2),
+        Seq(path1, path1, path2),
+        7),
+      (spark.read.option("path", path1).parquet(path1, path2), Seq(path1, path1, path2), 7))
+      .foreach {
+        case (df, expectedRootPaths, expectedCount) =>
+          val relation = CreateActionBaseWrapper.getSourceRelations(df)(0)
```
imback82, line 150: nit: `.head`

src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
```diff
@@ -104,4 +108,57 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite {
     assert(
       ex.getMessage.contains(s"Another Index with name ${indexConfig.indexName} already exists"))
   }
+
+  test("Verify rootPaths for given LogicalRelations") {
+    val path1 = sampleParquetDataLocation + "table1"
+    val path2 = sampleParquetDataLocation + "table2"
+    FileUtils.delete(new Path(path1))
+    FileUtils.delete(new Path(path2))
```
imback82, line 116: Can we do
```scala
    withTempPath { p =>
      val path1 = p + "table1"
      val path2 = p + "table2"
```
? You need update this class with `with SQLHelper`.

src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
imback82: This may not be needed after https://github.com/microsoft/hyperspace/pull/101#discussion_r464622936.

src/test/scala/com/microsoft/hyperspace/actions/RefreshActionTest.scala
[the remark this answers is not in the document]
imback82: Yes, I started the discussion around this. The solution suggested is to always ignore `path` option if paths (one or more) are passed to `load` (or throw an exception if both are defined). I am not sure yet but I think it will most likely land in 3.1 (most likely not backported to 2.4).

src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
sezruby: With this fix, `rootPaths` has `file:/` prefix:

```
List("file:/private/var/folders/sd/ynbvcl0120g_16_wm0g56zc40000gn/T/spark-2fef90f7-9b0e-4a67-a3f4-bb158d704fcdtable1")
did not equal
List("/private/var/folders/sd/ynbvcl0120g_16_wm0g56zc40000gn/T/spark-2fef90f7-9b0e-4a67-a3f4-bb158d704fcdtable1") (CreateActionTest.scala:163)
[info]
```

src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
```diff
@@ -110,55 +111,53 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite {
...
-        case _ => fail("invalid test")
-      }
-
-    FileUtils.delete(new Path(path1))
-    FileUtils.delete(new Path(path2))
+    withTempPath { p =>
+      val path1 = p + "table1"
+      val path2 = p + "table2"
+
+      FileUtils.delete(new Path(path1))
```
imback82, line 118: we can remove these `delete`s here and at the end; that was the reason for using `withTempPath`.

src/test/scala/com/microsoft/hyperspace/actions/RefreshActionTest.scala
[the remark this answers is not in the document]
sezruby: Great, thanks!

PR #117, opened 2020-08-06
Fix CreateAction test where path comparison fails on Windows

imback82:

This PR fixes `CreateAction` test where paths comparison causes a failure on Windows.

This PR addresses the issue where temp path is not cleaned up correctly.

To fix a test that fails on Windows.

No.

Updated existing test.

### DIFF
--- a/src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
@@ -123,8 +123,8 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite with SQLHel
 
   test("Verify rootPaths for given LogicalRelations") {
     withTempPath { p =>
-      val path1 = p + "table1"
-      val path2 = p + "table2"
+      val path1 = new Path(p.getCanonicalPath, "t1").toString
+      val path2 = new Path(p.getCanonicalPath, "t2").toString
 
       import spark.implicits._
       SampleData.testData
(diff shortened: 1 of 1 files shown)


src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala
```diff
@@ -123,8 +123,8 @@ class CreateActionTest extends SparkFunSuite with SparkInvolvedSuite with SQLHel
 
   test("Verify rootPaths for given LogicalRelations") {
     withTempPath { p =>
-      val path1 = p + "table1"
-      val path2 = p + "table2"
+      val path1 = new Path(p.getCanonicalPath, "t1").toString
+      val path2 = new Path(p.getCanonicalPath, "t2").toString
```
imback82, line 127: The existing code doesn't clean up the temp path `p`.

sezruby: Thanks for handling this issue. I think we can use withTempDir instead of withTempPath.

imback82: Isn't `withTempDir` inside `SQLTestUtils` which is `private[sql]` in Spark 2.4? There is `SparkFunSuite.withTempDir` in Spark 3.0, but the implementation is same as `withTempPath`.

Also a path represents either a file or directory, so I think it's fine here. For reference, the comment from `withTempPath`:
https://github.com/apache/spark/blob/e93b8f02cd706bedc47c9b55a73f632fe9e61ec3/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/plans/SQLHelper.scala#L62

sezruby: Yes I checked that `(p + "/table1")` or your change cleans up the path well. Thanks 👍

imback82: Yea, I initially did `/`, but switched to the current one since there is no confusion on the delimiter.

PR #121, opened 2020-08-11
Add lineage column(s) to index records

### LINKED ISSUE #104: Add lineage to covering index records
### Describe the issue
For each index record, add _lineage_ to capture the data record it is coming from. Assume data records are organized in Hive partitioned manner, where all the records which have the same value for a given partitioning key are stored in files under the same directory whose name has `PartitionKey=VALUE`. By adding lineage to index records, we want to add extra columns to each record and store the value of partition key(s) and file name according to the data file a record is coming from.

### To Reproduce
N/A

### Expected behavior
Once an index is created successfully, each index record should have columns to capture its source data record's partition key(s"' values and data file name.
 
### Environment 
All environments.

pirz:
This PR proposes to fix #104 by adding lineage columns to index records along with test cases for validating the change.

For non-partitioned data, lineage is captured thru adding a single column to each index record which saves the source file name. For partitioned data, beside file name, we add columns for partition key(s) if they are not already among user-specified columns in index config.

Adding lineage to index is *optional* (by default it is disabled for now); It can be enabled/disabled via below spark config:

- `INDEX_LINEAGE_ENABLED = "spark.hyperspace.index.lineage.enabled"`

**Example.** Assume input data has a schema with four columns: `Date, RGUID, Query, imprs, clicks` and it is partitioned on two columns: `Date` and `Query`. For an index created with `IndexConfig("index1", Seq("RGUID"), Seq("Date"))`:
* Sample index records without lineage (current behavior in Hyperspace) look like as:

| RGUID &nbsp;                            	| Date       	|
|----------------------------------	|------------	|
| 50d690516ca641438166049a6303650c 	| 2018-09-03 	|
| 576ed96b0d5340aa98a47de15c9f87ce 	| 2018-09-03 	|

* Sample index records with lineage look like as below. `_date_file_name` and `Query` (a partitioning key which was not already among index config columns) are added to capture lineage along with the `Date` columns.

| RGUID &nbsp;&nbsp;                         	| Date       	| Query  	| _data_file_name                                             	|
|----------------------------------	|------------	|--------	|-------------------------------------------------------------	|
| 50d690516ca641438166049a6303650c 	| 2018-09-03 	| donde  	| part-00001-20a62dfd-f0e0-4f17-8a70-93b90523ac38-c000.snappy 	|
| 576ed96b0d5340aa98a47de15c9f87ce 	| 2018-09-03 	| ibraco 	| part-00001-77a49590-b385-454c-9877-51ab3b626e89-c000.snappy 	|
 
Lineage is required for supporting DELETE during index refresh when one or more source data files are removed.

No.

Unit test cases are added in `CreateIndexTests.scala`.


imback82:
@pirz Can you fix the build please?

pirz:
@imback82 It is green now

imback82:
Could you add the output when you do `df.show` where `df` is created with index files, compared with the previous version of index? (both for partitioned/non-partitioned source data)

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +171,17 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+      case LogicalRelation(HadoopFsRelation(_, pSchema, _, _, _, _), _, _, _) => pSchema
+    }
+
+    val spark = df.sparkSession
+    partitionSchema.head
+      .map(_.name)
+      .filter(ResolverUtils.resolve(spark, _, indexConfigColumns).isEmpty)
+  }
+
+  private val fileName = udf((fullFilePath: String) => FilenameUtils.getBaseName(fullFilePath))
```
imback82, line 186: can you add a type since this is `private val`? Also, should we change the variable name since this is a function?

imback82, line 186: Question. I guess we wouldn't need the path the file ever right?

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -47,4 +47,6 @@ object IndexConstants {
     val PLAIN_TEXT = "plaintext"
     val HTML = "html"
   }
+
+  val DATA_FILE_NAME_COLUMN = "_data_file_name"
```
imback82, line 51: `private[index] val DATA_FILE_NAME_COLUMN`?

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -109,7 +109,7 @@ object FilterIndexRule
         val newRelation = HadoopFsRelation(
           newLocation,
           new StructType(),
-          index.schema,
+          StructType(index.schema.filter(fsRelation.schema.contains(_))),
```
imback82, line 112: nit: `(_)` is needed?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -70,7 +70,11 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
       200,
-      StructType(Seq(StructField("RGUID", StringType), StructField("Date", StringType))).json,
+      StructType(
+        Seq(
+          StructField("RGUID", StringType),
+          StructField("Date", StringType),
+          StructField(IndexConstants.DATA_FILE_NAME_COLUMN, StringType))).json,
       s"$indexStorageLocation/index1/v__=0",
       Constants.States.ACTIVE)
     assert(actual.equals(expected))
```
imback82, line 80: Can you update e2e tests for filter/join rules with partitioned source data if we already don't have them?

src/test/scala/com/microsoft/hyperspace/index/CreateIndexTests.scala
```diff
@@ -128,14 +164,66 @@ class CreateIndexTests extends HyperspaceSuite with SQLHelper {
...
+  test("Check lineage in index records for partitioned data when partition key is in load path.") {
+    val dataDF =
+      spark.read.parquet(s"$samplePartitionedParquetDataLocation/$partitionKey1=2017-09-03")
+    hyperspace.createIndex(dataDF, indexConfig3)
+    val indexRecordsDF = spark.read.parquet(
+      s"$systemPath/${indexConfig3.indexName}/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=0")
+
+    // As data load path includes first partition key, index schema should only contain
+    // file name column and second partition key column as lineage columns.
+    indexRecordsDF.schema.fields.corresponds(
```
imback82, line 225: you don't need to assert?

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
pirz: We can not make this constant private and package specific as its definition here is in `IndexConstants.scala` which is in index package while `CreateActionBase.scala` is the place the constant is used and it is in actions package. I

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -47,4 +47,8 @@ object IndexConstants {
...
+  val INDEX_LINEAGE_ENABLED = "spark.hyperspace.index.lineage.enabled"
+  val INDEX_LINEAGE_ENABLED_DEFAULT = "false"
```
pirz, line 53: For now I have disabled lineage for indexes - I already made sure that all test cases pass irrespective of this default value (i.e. all tests pass with this default value be true or false).

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -136,26 +136,26 @@ object JoinIndexRule
...
+        val bucketSpec = BucketSpec(
+          numBuckets = index.numBuckets,
+          bucketColumnNames = index.indexedColumns,
+          sortColumnNames = index.indexedColumns)
+
+        val location = new InMemoryFileIndex(spark, Seq(new Path(index.content.root)), Map(), None)
+        val relation = HadoopFsRelation(
+          location,
+          new StructType(),
+          StructType(index.schema.filter(baseRelation.schema.contains(_))),
```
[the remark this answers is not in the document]
pirz: Currently, we only have one extra column and your suggested code works fine. However, I think it is better if we rely on the source data schema to pick schema for an index usage. This way if we add more extra column(s) later; this code works as expected with no change required. 

There is another advantage in relying on the source data schema, at the query time, to prune columns from the index schema: Imagine the source data was partitioned on two columns: A and B with a path like:

- source/A=a1/B=b1

Therefore, these two columns are added to index schema as part of lineage. Now at the query time, if user passes the data path as `source/A=a1`, Column A will no longer be part of data schema. If a rule decides to use the index then by relying on source data schema we make sure column A is excluded from the index usage too, but if simply use `field.name != IndexConstants.DATA_FILE_NAME_COLUMN` then column A will be added to the logical relation node that is replaced by the index and the transformed query plan fails with analyze error as A was not originally present in the data schema, at the query time, and hence it was not resolved.

Therefore, I think it is better if we always rely on source data schema at the query time. 
@sezruby  Please let me know if above is clear and makes sense to you.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -126,8 +131,11 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
       .toInt
 
     val (resolvedIndexedColumns, resolvedIncludedColumns) = resolveConfig(df, indexConfig)
-    val selectedColumns = resolvedIndexedColumns ++ resolvedIncludedColumns
-    val indexDataFrame = df.select(selectedColumns.head, selectedColumns.tail: _*)
+    val indexConfigColumns = resolvedIndexedColumns ++ resolvedIncludedColumns
+    val selectedColumns = indexConfigColumns ++ internalColumns(df, indexConfigColumns)
+    val indexDataFrame = df
+      .select(selectedColumns.head, selectedColumns.tail: _*)
+      .withColumn(IndexConstants.DATA_FILE_NAME_COLUMN, fileName(input_file_name()))
```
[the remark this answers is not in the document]
pirz: I refactored it in the `getIndexDataFrame` function.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +171,17 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
             s"from available source columns '${dfColumnNames.mkString(",")}'")
     }
   }
+
+  private def internalColumns(df: DataFrame, indexConfigColumns: Seq[String]): Seq[String] = {
```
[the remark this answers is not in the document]
pirz: I changed the code and add comments on line 178.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

pirz: Which path do you mean? This function is an auxiliary that extract file name from a full path. It makes no assumption about its input args origin or format.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
pirz: removed it.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
pirz: I added extra test cases there.

src/test/scala/com/microsoft/hyperspace/index/CreateIndexTests.scala
pirz: Oops, thanks for catching it.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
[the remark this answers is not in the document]
sezruby: Thanks for the detailed explanation, but it's a bit unclear to me. The current version also just uses `index.schema`, so it means the current version cannot handle the case you described? 
I could be wrong, but this schema of `HadoopFsRelation` seems for reading index data and `updatedOutput` will filter unused columns out..?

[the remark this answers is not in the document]
pirz: As current version uses index.schema it has a similar issue, however I updated my comment as I realized this behavior is actually desired. The reason for that is if the source data at the query time uses a subset of the original source data then we should not use index as it could have extra rows that were part of original data and now excluded from the source data at query time. Imagine the index was created on source data with two partitions: `.../data/A=a1` and `.../data/A=a2`. An index created on `.../data` path has records from both partitions `a1` and `a2`. However, if data path at query time is set to `.../data/A=a1` then it means records from the `a2` partition should be excluded from the final results and if we use the index during query we may get those records back depending on the query and index config (i.e. potentially wrong query results). Moreover, our current signature computation/comparison does not mark such an index as a candidate for the query - and this is a correct behavior.
Regarding your point about HadoopFsRelation, if I understood what you mentioned correctly: Yes that schema is used for reading data; however any column in the schema is expected to be already resolved (i.e. has to have an `ExprId`) by the time we switch to an index. Once we switch to an index, we still use the same `ExprId` that Spark had assigned to the columns when analyzing the plan with source data. Hence, if the index schema has some column which was not present during former analyze phase, the query fails after transformation as such a column has no `ExprId`.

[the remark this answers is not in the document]
sezruby: Ok, I got the point. I might also need to investigate `ExprId` related things for hybrid scan. Thanks :)

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
imback82: then `[hyperspace]`. The intention is to make it internal to the project.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -61,19 +61,12 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
-    val expected = new IndexSummary(
-      indexConfig1.indexName,
-      indexConfig1.indexedColumns,
-      indexConfig1.includedColumns,
-      200,
-      StructType(Seq(StructField("RGUID", StringType), StructField("Date", StringType))).json,
-      s"$indexStorageLocation/index1/v__=0",
-      Constants.States.ACTIVE)
-    assert(actual.equals(expected))
+  test("Verify that indexes() returns the correct dataframe without lineage.") {
```
imback82, line 64: How about:
```scala
test("Verify that indexes() returns the correct dataframe with/without lineage.") {
  Seq("true", "false").foreach { enableLineage =>
    withSQLConf((IndexConstants.INDEX_LINEAGE_ENABLED -> enableLineage) {
      import spark.implicits._
      hyperspace.createIndex(df, indexConfig1)
      ..
    }
  }
}
```

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +161,44 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+      df.select(dataFrameColumns.head, dataFrameColumns.tail: _*)
+        .withColumn(IndexConstants.DATA_FILE_NAME_COLUMN, getFileName(input_file_name()))
+    } else {
+      df.select(indexConfigColumns.head, indexConfigColumns.tail: _*)
+    }
+
+    (indexDF, resolvedIndexedColumns, resolvedIncludedColumns)
+  }
+
+  private def getPartitioningColumns(df: DataFrame): Seq[String] = {
```
imback82, line 193: nit: `getPartitionColumns`

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +161,44 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    // Extract partitioning keys, if original data is partitioned.
+    val partitionSchema = df.queryExecution.optimizedPlan.collect {
+      case LogicalRelation(HadoopFsRelation(_, pSchema, _, _, _, _), _, _, _) => pSchema
+    }
+
```
imback82, line 198: Can we `assert(partitionSchemas == 1)`? (Note that it should be `val partitionSchemas` since `collect` returns `Seq`.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +161,44 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
             s"from available source columns '${dfColumnNames.mkString(",")}'")
     }
   }
+
+  private def getIndexDataFrame(
```
imback82, line 165: nit: how about `prepareIndexDataFrame`?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +161,44 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+      spark: SparkSession,
+      df: DataFrame,
+      indexConfig: IndexConfig): (DataFrame, Seq[String], Seq[String]) = {
+    val (resolvedIndexedColumns, resolvedIncludedColumns) = resolveConfig(df, indexConfig)
+    val indexConfigColumns = resolvedIndexedColumns ++ resolvedIncludedColumns
```
imback82, line 170: nit: `columnsFromIndexConfig`?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +161,44 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+      .getConfString(
+        IndexConstants.INDEX_LINEAGE_ENABLED,
+        IndexConstants.INDEX_LINEAGE_ENABLED_DEFAULT)
+      .toBoolean
+
+    val indexDF = if (addLineage) {
+      // Lineage is captured using two sets of columns:
+      // 1. DATA_FILE_NAME_COLUMN column contains source data filename for each index record.
+      // 2. If source data is partitioned, all partitioning key(s) are added to index schema.
+      val additionalPartitioningColumns = getPartitioningColumns(df).filter(
```
imback82, line 181: nit: `missingPartitionColumns`?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +161,44 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    val addLineage = spark.sessionState.conf
+      .getConfString(
+        IndexConstants.INDEX_LINEAGE_ENABLED,
+        IndexConstants.INDEX_LINEAGE_ENABLED_DEFAULT)
+      .toBoolean
+
+    val indexDF = if (addLineage) {
+      // Lineage is captured using two sets of columns:
+      // 1. DATA_FILE_NAME_COLUMN column contains source data filename for each index record.
+      // 2. If source data is partitioned, all partitioning key(s) are added to index schema.
```
imback82, line 180: `// 2. If source data is partitioned, all partitioning key(s) are added to index schema as columns if they are not already part of the schema.`

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +161,44 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+      val additionalPartitioningColumns = getPartitioningColumns(df).filter(
+        ResolverUtils.resolve(spark, _, indexConfigColumns).isEmpty)
+      val dataFrameColumns = indexConfigColumns ++ additionalPartitioningColumns
```
imback82, line 183: nit: `allColumns` or `allIndexColumns`.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -109,7 +109,7 @@ object FilterIndexRule
         val newRelation = HadoopFsRelation(
           newLocation,
           new StructType(),
-          index.schema,
+          StructType(index.schema.filter(fsRelation.schema.contains)),
```
imback82, line 112: Question. If we just use `fsRelation.schema` here instead of `index.schem`, it would still work since we already verified that all the columns being referenced are covered?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -27,25 +27,34 @@ import com.microsoft.hyperspace.{Hyperspace, Implicits, SampleData}
...
     spark.conf.set("spark.sql.autoBroadcastJoinThreshold", -1)
-
-    import spark.implicits._
     hyperspace = new Hyperspace(spark)
-
-    fileSystem.delete(new Path(sampleParquetDataLocation), true)
-
-    val dfFromSample = sampleData.toDF("c1", "c2", "c3", "c4", "c5")
-    dfFromSample.write.parquet(sampleParquetDataLocation)
+    fileSystem.delete(new Path(sampleNonPartitionedParquetDataLocation), true)
```
imback82, line 43: Is this needed? I see `fileSystem.delete(new Path(testDir), true)` below.

src/test/scala/com/microsoft/hyperspace/SampleData.scala
```diff
@@ -25,10 +27,43 @@ object SampleData {
...
+        .distinct()
+        .collect()
+        .foreach { q =>
+          val query = q.get(0)
+          val partitionPath =
+            s"$path/$partitionKey1=$date/$partitionKey2=$query"
+          df.filter($"Date" === date && $"Query" === query)
+            .select("RGUID", "imprs", "clicks")
+            .write
+            .parquet(partitionPath)
```
imback82, line 65: huh? Are you trying to do `df.write.partitionBy("Date", "Query").parquet(path)` in this block or something else?

src/test/scala/com/microsoft/hyperspace/SampleData.scala
```diff
@@ -25,10 +27,43 @@ object SampleData {
...
+    ("2019-10-03", "187696fe0a6a40cc9516bc6e47c70bc1", "facebook", 4, 3000))
+
+  // column names and partition keys for partitioned case
+  val colnames = Seq("Date", "RGUID", "Query", "imprs", "clicks")
+  val partitionKey1 = "Date"
+  val partitionKey2 = "Query"
+
+  def saveTestDataNonPartitioned(spark: SparkSession, path: String, colNames: String*): Unit = {
+    import spark.implicits._
+    assert(colNames.length == 5)
```
imback82, line 44: why do you limit this?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -137,13 +162,35 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
 
-  test("E2E test for join query.") {
-    val leftDf = spark.read.parquet(sampleParquetDataLocation)
+  test(
+    "E2E test for filter query when all columns are selected on partitioned data with lineage.") {
+    spark.conf.set(IndexConstants.INDEX_LINEAGE_ENABLED, "true")
+
+    val df = spark.read.parquet(samplePartitionedParquetDataLocation)
+    val indexConfig =
+      IndexConfig("filterIndex", Seq("imprs", "Query"), Seq("Date", "RGUID", "clicks"))
```
imback82, line 171: Is it possible to use `c1`, `c2`, `c3`, etc. similar to other tests in this suite?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -163,4 +161,44 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    // Extract partitioning keys, if original data is partitioned.
+    val partitionSchema = df.queryExecution.optimizedPlan.collect {
+      case LogicalRelation(HadoopFsRelation(_, pSchema, _, _, _, _), _, _, _) => pSchema
+    }
+
+    partitionSchema.head.map(_.name)
+  }
+
+  private val getFileName: UserDefinedFunction = udf(
+    (fullFilePath: String) => FilenameUtils.getBaseName(fullFilePath))
```
imback82, line 203: this will strip off extension as well. is that what we want?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
pirz: Makes sense - I added it.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
pirz: Modified all lineage tests to use SqlHelper.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: Done.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
pirz: I guess we can not rely on a previous run being terminated gracefully by calling afterAll(). If a previous run crashes or is forced to terminate when debugging, sometimes left-over data remains. I think it is safe to explicitly delete those potential left-overs, given that these delete calls are not that expensive and are called only once when running the suite.

src/test/scala/com/microsoft/hyperspace/SampleData.scala
pirz: Yes, I changed it.

src/test/scala/com/microsoft/hyperspace/SampleData.scala
pirz: I wanted to make sure passed columns count matches with the DF's column count. I checked the `toDF()` call and realized it already has `require(schema.size == colNames.size`. So I removed my assert.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
imback82: Well, my point is that you can just delete the root, and not an individual sub-dirs.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
pirz: Fixed it.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: You are correct - I can not think of a case where dropping the extension could cause an issue. But adding it does not hurt either as most likely it wont cause an issue given that it will be repeated per file name and most likely Parquet's compression will take care of storing it efficiently. I can make sure we wont miss it if you have preference to capture it.

imback82: Since these are "source" data, I could have "/tmp/1.csv", and "/tmp/1.parquet". If I tried to create an index on dataframe created by `spark.read.csv("/tmp/1.csv")`, would this be OK?

src/test/scala/com/microsoft/hyperspace/SampleData.scala
```diff
@@ -25,10 +27,26 @@ object SampleData {
...
-    ("2017-09-03", "ff60e4838b92421eafc3e6ee59a9e9f1", "mi perro", 2, 2000),
-    ("2017-09-03", "187696fe0a6a40cc9516bc6e47c70bc1", "multicines", 4, 3000))
+    ("2018-09-03", "e90a6028e15b4f4593eef557daf5166d", "ibraco", 2, 3000),
+    ("2018-09-03", "576ed96b0d5340aa98a47de15c9f87ce", "facebook", 2, 3000),
+    ("2018-09-03", "50d690516ca641438166049a6303650c", "ibraco", 2, 1000),
+    ("2019-10-03", "380786e6495d4cd8a5dd4cc8d3d12917", "facebook", 2, 3000),
+    ("2019-10-03", "ff60e4838b92421eafc3e6ee59a9e9f1", "mi perro", 2, 2000),
+    ("2019-10-03", "187696fe0a6a40cc9516bc6e47c70bc1", "facebook", 4, 3000))
+
+  def saveTestDataNonPartitioned(spark: SparkSession, path: String, colNames: String*): Unit = {
```
imback82, line 37: How about combining these two methods into
```scala
def save(spark: SparkSession, path: String, columns: Seq[String], partitionColumns: Option[Seq[String]] = None): Unit
```

pirz: sure, merged them.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: I switched to `getName` to keep the extension.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
pirz: Oh ok - made the change.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -63,19 +64,31 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
-      StructType(Seq(StructField("RGUID", StringType), StructField("Date", StringType))).json,
-      s"$indexStorageLocation/index1/v__=0",
-      Constants.States.ACTIVE)
-    assert(actual.equals(expected))
+  test("Verify that indexes() returns the correct dataframe with and without lineage.") {
+    Seq(true, false).foreach { enableLineage =>
+      withSQLConf(IndexConstants.INDEX_LINEAGE_ENABLED -> enableLineage.toString) {
+        import spark.implicits._
+        hyperspace.createIndex(df, indexConfig1)
+        val actual = hyperspace.indexes.as[IndexSummary].collect()(0)
```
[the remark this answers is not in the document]
pirz: Changed it.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -63,19 +64,31 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
...
+          indexConfig1.indexedColumns,
+          indexConfig1.includedColumns,
+          200,
+          expectedSchema.json,
+          s"$indexStorageLocation/index1/v__=0",
+          Constants.States.ACTIVE)
+        assert(actual.equals(expected))
+      }
+      hyperspace.deleteIndex(indexConfig1.indexName)
+      FileUtils.delete(new Path(indexStorageLocation))
```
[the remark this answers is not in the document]
pirz: Changed the class to use Hypespace suite

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -123,48 +135,71 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+        withSQLConf(IndexConstants.INDEX_LINEAGE_ENABLED -> enableLineage.toString) {
+          val indexNames = Seq("leftIndex", "rightIndex")
+          withIndex(indexNames: _*) {
+            val leftDf = spark.read.parquet(loc)
+            val leftDfIndexConfig = IndexConfig(indexNames.head, Seq("c3"), Seq("c1"))
+
+            hyperspace.createIndex(leftDf, leftDfIndexConfig)
+
+            val rightDf = spark.read.parquet(loc)
+            val rightDfIndexConfig = IndexConfig(indexNames(1), Seq("c3"), Seq("c4"))
```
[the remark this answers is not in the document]
pirz: Done.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -123,48 +135,71 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+  test(
+    "E2E test for filter query when all columns are selected on partitioned and " +
+      "non-partitioned data with and without lineage.") {
+    Seq(nonPartitionedDataPath, partitionedDataPath).foreach { loc =>
+      Seq(true, false).foreach { enableLineage =>
+        withSQLConf(IndexConstants.INDEX_LINEAGE_ENABLED -> enableLineage.toString) {
+          val indexName = "filterIndex"
+          withIndex(indexName) {
+            val df = spark.read.parquet(loc)
+            val indexConfig = IndexConfig("filterIndex", Seq("c4", "c3"), Seq("c1", "c2", "c5"))
```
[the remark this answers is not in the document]
pirz: Revert it back.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -284,7 +291,7 @@ class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
               IndexLogEntry.schemaString(schema),
               IndexConstants.INDEX_NUM_BUCKETS_DEFAULT)),
           Content(
-            s"$indexStorageLocation/${indexConfig.indexName}" +
+            s"${systemPath.toUri.toString}/${indexConfig.indexName}" +
```
imback82: This can be just `$systemPath`. I pushed a commit to change these.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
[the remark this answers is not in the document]
imback82: Thanks!

imback82:
@pirz can you update the description to include extension for the filename as well?

Issue #107, opened 2020-08-05
Store index file names and file sizes in operation log


apoorvedave1:
### Describe the issue

Upcoming features require all index files to be present in the index metadata. Also, file sizes of individual files is required. A sample of this could be:
```
{
 ...
 "files": ["root/indexfile1:100MB", "root/indexFile2:1MB"...]
...
}
```
### To Reproduce

Enhancement
### Expected behavior
Index creation / modification adds list of created index files to the operation log along with their file size

### Environment 

na

imback82:
is this similar to #103? cc @pirz

apoorvedave1:
yeah that's right. Only this one also mentions sizes which will be used in optimizations

imback82:
Then, can we merge into one issue so it's easier to track?

pirz:
Yes they are the same, we can go with this one that @apoorvedave1 has created and close #103

Issue #110, opened 2020-08-05
Add RefreshIncrementalAction class to support index creation on newly appended data 


apoorvedave1:
### Describe the issue

Implement incremental indexing support for append-only data. Implementation outline:
- identify newly added data files
- create new index version on these files
- update metadata to reflect the latest snapshot of index to include old files and new files

New Api: `refresh(mode = quick)`
### To Reproduce
na

### Expected behavior
na

### Environment 
na

sezruby:
I think `file size` and `modification time` for each file are also required to ensure that the files are not modified. Currently we identify this with the signature value, but supporting append data (+ hybrid scan) needs to check if the existing index can cover the subset of dataset or not.

PR #122, opened 2020-08-12
Store file name, size and modified time of index files in metadata

### LINKED ISSUE #120: Capture source file info to detect the change of the dataset
**Describe the problem**

As we are extending the features like dataset with append or delete, it's required to detect the change of the source files to figure out which files are appended or deleted. This could be done only with the list of file paths, but we need additional information to check the integrity of the dataset. We can calculate a new signature only with the files in the captured file paths, but it's hard to detect accidently modified files without size or timestamp. 
Although Hyperspace assumes that user won't tamper the dataset, we should avoid using the invalid index data.

- For append-only dataset without delete, the modified file (same file name, different size or timestamp)
  - query time
    - will invalidate the whole indexes as the current (immutable dataset) does
  - refresh time
    - should force full rebuild
  - these can be done with the current signature value (only with paths)
- For append-only dataset with delete, the modified file
  - query time
    - will invalidate the whole indexes as the current (immutable dataset) does
  - refresh time
    - Option1) should be considered as "deleted" file and also handled as newly "appended" file
    - Option2) or full rebuild?
  - option 1 **requires to identify which file is modified by using captured timestamp and size**
- For hybrid scan, the modified file
  - query time
    - considered as "deleted" file and also "appended" file
  - this also requires to identify which file is modified 

**Describe your proposed solution**
As FilebaseSignatureProvider is using file name + size + modification time, we could capture size and modification time additionally.

```
// Current
case class Content(root: String, directories: Seq[Content.Directory])                       
object Content {                                                                            
  case class Directory(path: String, files: Seq[String], fingerprint: NoOpFingerprint)      
}                                                                                           

// Suggested
case class File(path: String, size: Int, timestamp: Timestamp) // or timestamp: String
case class Content(root: String, directories: Seq[Content.Directory])                       
object Content {                                                                            
  case class Directory(path: String, files: Seq[File], fingerprint: NoOpFingerprint)      
}    
```

Then we could compare the captured values to identify each file is modified or not.

**Describe alternatives you've considered**

**Additional context**

apoorvedave1:
Fixes #120 and #107 

Store file name, size and last modified time in metadata. After this, index and data file contents will look like : 
```
  "content" : {
    "root" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0",
    "directories" : [ {
      "path" : "",
      "files" : [ {
        "name" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0/part-00007-5384c4fd-cb04-4d7c-ac13-ac7220ee17f3_00007.c000.snappy.parquet",
        "size" : 926,
        "modifiedTime" : 1597351292026
      }, {
        "name" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0/part-00011-5384c4fd-cb04-4d7c-ac13-ac7220ee17f3_00011.c000.snappy.parquet",
        "size" : 926,
        "modifiedTime" : 1597351291550
      }
      ....
```
Field Descriptions:
`directories.path`: path to be prepended to all file names within this directory
`directories.files`: List of files within this directory
`directories.files[i].name`: file name or path of the ith file. Prepend `directories.path` to this to get full file path.
`directories.files[i].size`: file size in bytes
`directories.files[i].modifiedTime`: Last modified time of this file. It's unit is epochs in **millisecond**. In the above example, 1597351291550 = GMT: Thursday, August 13, 2020 8:41:31.550 PM. https://en.wikipedia.org/wiki/Unix_time

For support of new features like append, delete

YES, This PR introduces a user facing change. It's a breaking change which will make older indexes (created until v0.2.0) unusable.

log files will have additional content for index files.: file names, size and modified time.

```
case class Directory(path: String, files: Seq[FileInfo], fingerprint: NoOpFingerprint)
```
Here `files` was changed from Seq[String] to Seq[FileInfo] where FileInfo contains filename, size and modified time.

unit tests

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
@@ -56,7 +56,7 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
   // TODO: Revisit if access modifier can be changed to protected.
   def op(): Unit
 
-  protected def end(): Unit = {
+  private def end(): Unit = {
     val newId = baseId + 2
     val entry = logEntry
     entry.state = finalState
@@ -73,7 +73,7 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
     }
   }
 
-  final protected def saveEntry(id: Int, entry: LogEntry): Unit = {
+  private def saveEntry(id: Int, entry: LogEntry): Unit = {
     entry.timestamp = System.currentTimeMillis()
     if (!logManager.writeLog(id, entry)) {
       throw HyperspaceException("Could not acquire proper state")
--- a/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
@@ -56,7 +56,7 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
   // TODO: Revisit if access modifier can be changed to protected.
   def op(): Unit
 
-  private def end(): Unit = {
+  protected def end(): Unit = {
     val newId = baseId + 2
     val entry = logEntry
     entry.state = finalState
@@ -73,7 +73,7 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
     }
   }
 
-  private def saveEntry(id: Int, entry: LogEntry): Unit = {
+  final protected def saveEntry(id: Int, entry: LogEntry): Unit = {
     entry.timestamp = System.currentTimeMillis()
     if (!logManager.writeLog(id, entry)) {
       throw HyperspaceException("Could not acquire proper state")
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -18,8 +18,6 @@ package com.microsoft.hyperspace.actions
 
 import java.io.FileNotFoundException
 
-import scala.util.{Failure, Try}
-
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.{DataFrame, SparkSession}
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -161,7 +161,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
     } catch {
       // FileNotFoundException is an expected exception before index gets created.
       case _: FileNotFoundException => Seq()
-      case e => throw e
+      case e : Throwable => throw e
     }
   }
 
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -85,7 +85,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
                 .Columns(resolvedIndexedColumns, resolvedIncludedColumns),
               IndexLogEntry.schemaString(schema),
               numBuckets)),
-          Content(path.toString, Seq(Directory("", leafFileInfo(path), NoOpFingerprint()))),
+          Content(path.toString, Seq(Directory("", indexFilesInfo(path), NoOpFingerprint()))),
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
@@ -149,7 +149,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
         resolvedIndexedColumns)
   }
 
-  private def leafFileInfo(path: Path): Seq[FileInfo] = {
+  private def indexFilesInfo(path: Path): Seq[FileInfo] = {
     try {
       val fs = path.getFileSystem(new Configuration)
       val statuses = fs.listStatus(path)
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -16,6 +16,9 @@
 
 package com.microsoft.hyperspace.actions
 
+import scala.util.{Success, Try}
+
+import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
@@ -23,6 +26,8 @@ import org.apache.spark.sql.sources.DataSourceRegister
 
 import com.microsoft.hyperspace.HyperspaceException
 import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.index.Content.Directory
+import com.microsoft.hyperspace.index.Content.Directory.FileInfo
 import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
 import com.microsoft.hyperspace.util.ResolverUtils
 
@@ -78,7 +83,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
                 .Columns(resolvedIndexedColumns, resolvedIncludedColumns),
               IndexLogEntry.schemaString(schema),
               numBuckets)),
-          Content(path.toString, Seq()),
+          Content(path.toString, Seq(Directory("", leafFileInfo(path), NoOpFingerprint()))),
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
@@ -99,7 +104,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
           _,
           _,
           _) =>
-        val files = location.allFiles.map(_.getPath.toString)
+        val files = location.allFiles.map(FileInfo(_))
         // Note that source files are currently fingerprinted when the optimized plan is
         // fingerprinted by LogicalPlanFingerprint.
         val sourceDataProperties =
@@ -140,6 +145,14 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
         indexDataPath.toString,
         numBuckets,
         resolvedIndexedColumns)
+  }
+
+  final def leafFileInfo(path: Path): Seq[FileInfo] = {
+    val fs = path.getFileSystem(new Configuration)
+    Try(fs.listStatus(path).map(FileInfo(_))) match {
+      case Success(value) => value
+      case _ => Seq()
+    }
   }
 
   private def resolveConfig(
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -16,7 +16,9 @@
 
 package com.microsoft.hyperspace.actions
 
-import scala.util.Try
+import java.io.FileNotFoundException
+
+import scala.util.{Failure, Try}
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
@@ -148,9 +150,19 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
   }
 
   private def leafFileInfo(path: Path): Seq[FileInfo] = {
-    // Assuming index directories don't contain nested directories. Only Leaf files.
-    val fs = path.getFileSystem(new Configuration)
-    Try { fs.listStatus(path).map(FileInfo(_)).toSeq }.getOrElse(Seq())
+    try {
+      val fs = path.getFileSystem(new Configuration)
+      val statuses = fs.listStatus(path)
+
+      // Assuming index directories don't contain nested directories. Only Leaf files.
+      assert(statuses.forall(!_.isDirectory))
+
+      statuses.map(FileInfo(_)).toSeq
+    } catch {
+      // FileNotFoundException is an expected exception before index gets created.
+      case _: FileNotFoundException => Seq()
+      case e => throw e
+    }
   }
 
   private def resolveConfig(
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -16,7 +16,7 @@
 
 package com.microsoft.hyperspace.actions
 
-import scala.util.{Success, Try}
+import scala.util.Try
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
@@ -147,12 +147,10 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
         resolvedIndexedColumns)
   }
 
-  final def leafFileInfo(path: Path): Seq[FileInfo] = {
+  private def leafFileInfo(path: Path): Seq[FileInfo] = {
+    // Assuming index directories don't contain nested directories. Only Leaf files.
     val fs = path.getFileSystem(new Configuration)
-    Try(fs.listStatus(path).map(FileInfo(_))) match {
-      case Success(value) => value
-      case _ => Seq()
-    }
+    Try { fs.listStatus(path).map(FileInfo(_)).toSeq }.getOrElse(Seq())
   }
 
   private def resolveConfig(
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
@@ -33,6 +33,7 @@ case class Content(root: String, directories: Seq[Content.Directory])
 object Content {
   case class Directory(path: String, files: Seq[FileInfo], fingerprint: NoOpFingerprint)
   object Directory {
+    // modifiedTime is an Epoch time in milliseconds. (ms `since 1970-01-01T00:00:00.000 UTC).
     case class FileInfo(name: String, size: Long, modifiedTime: Long)
     object FileInfo {
       def apply(s: FileStatus): FileInfo =
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
@@ -16,9 +16,11 @@
 
 package com.microsoft.hyperspace.index
 
+import org.apache.hadoop.fs.FileStatus
 import org.apache.spark.sql.types.{DataType, StructType}
 
 import com.microsoft.hyperspace.actions.Constants
+import com.microsoft.hyperspace.index.Content.Directory.FileInfo
 
 // IndexLogEntry-specific fingerprint to be temporarily used where fingerprint is not defined.
 case class NoOpFingerprint() {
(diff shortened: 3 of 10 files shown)


src/main/scala/com/microsoft/hyperspace/actions/Action.scala
```diff
@@ -73,7 +73,7 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
     }
   }
 
-  private def saveEntry(id: Int, entry: LogEntry): Unit = {
+  final protected def saveEntry(id: Int, entry: LogEntry): Unit = {
```
pirz, line 76: Just a question: Why are these functions changed from private to protected?

apoorvedave1: oh thanks for this. Looks like I missed some cleanup. This function doesn't need changing. I will revert.

imback82:
The approach looks fine to me. I assume `modifiedTime` is an epoch time? (If so, could you update the PR description?)

sezruby:
LGTM + seems this PR also closes #120. Thanks!

apoorvedave1:
> The approach looks fine to me. I assume `modifiedTime` is an epoch time? (If so, could you update the PR description?)

thanks @imback82 , updated description

apoorvedave1:
> LGTM + seems this PR also closes #120. Thanks!

thanks @sezruby for pointing that out. I have updated the description with this info

imback82:
Actually, I missed the `root`:
```
    "root" : "src/test/resources/indexLocation/index1/v__=0",
    "directories" : [ {
      "path" : "",
      "files" : [ {
        "name" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0/part-00007-5384c4fd-cb04-4d7c-ac13-ac7220ee17f3_00007.c000.snappy.parquet",
        "size" : 926,
        "modifiedTime" : 1597351292026
      },
```

For the above example, shouldn't we have the following if we are populating `root`:
```
    "root" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0/",
    "directories" : [ {
      "path" : "",
      "files" : [ {
        "name" : "part-00007-5384c4fd-cb04-4d7c-ac13-ac7220ee17f3_00007.c000.snappy.parquet",
        "size" : 926,
        "modifiedTime" : 1597351292026
      },
```
Did I misunderstand?

apoorvedave1:
> Actually, I missed the `root`:
> 
> ```
>     "root" : "src/test/resources/indexLocation/index1/v__=0",
>     "directories" : [ {
>       "path" : "",
>       "files" : [ {
>         "name" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0/part-00007-5384c4fd-cb04-4d7c-ac13-ac7220ee17f3_00007.c000.snappy.parquet",
>         "size" : 926,
>         "modifiedTime" : 1597351292026
>       },
> ```
> 
> For the above example, shouldn't we have the following if we are populating `root`:
> 
> ```
>     "root" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0/",
>     "directories" : [ {
>       "path" : "",
>       "files" : [ 
>         "name" : "part-00007-5384c4fd-cb04-4d7c-ac13-ac7220ee17f3_00007.c000.snappy.parquet",
>         "size" : 926,
>         "modifiedTime" : 1597351292026
>       },
> ```
> 
> Did I misunderstand?

oh good point. Thanks. How about we remove the root, and just use `directories` and `files`? I am not sure if we need `root` anymore

imback82:
I believe `root` was introduced to address scearios like
```
    "root" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/",
    "directories" : [ {
      "path" : "v__=0",
      "files" : [ {
        "name" : "1.parquet",
        "size" : 926,
        "modifiedTime" : 1597351292026
      },
     {
      "path" : "v__=1",
      "files" : [ {
        "name" : "1.parquet",
        "size" : 926,
        "modifiedTime" : 1597351292026
      },
```

So each file's full path will be `root / path / name`. But since the number of directories would be smaller, having `root` may not be too beneficial. We can think about removing this later since `root` can be empty in `root / path / name`. cc @rapoth 

Btw, we are not addressing this in this PR anyway, right?

apoorvedave1:
> I believe `root` was introduced to address scearios like
> 
> ```
>     "root" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/",
>     "directories" : [ {
>       "path" : "v__=0",
>       "files" : [ {
>         "name" : "1.parquet",
>         "size" : 926,
>         "modifiedTime" : 1597351292026
>       },
>      {
>       "path" : "v__=1",
>       "files" : [ {
>         "name" : "1.parquet",
>         "size" : 926,
>         "modifiedTime" : 1597351292026
>       },
> ```
> 
> So each file's full path will be `root / path / name`. But since the number of directories would be smaller, having `root` may not be too beneficial. We can think about removing this later since `root` can be empty in `root / path / name`. cc @rapoth
> 
> Btw, we are not addressing this in this PR anyway, right?

@imback82 , that's correct. This PR only _adds_ files (`FileInfo` json). It doesn't modify `root` and `directories[i].path` fields.

Should I update this pr to make `root` empty? I have not yet touched that section yet. Currently there is some duplication in root and file name as below:
```
  "content" : {
    "root" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0",
    "directories" : [ {
      "path" : "",
      "files" : [ {
        "name" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1/v__=0/part-00007-5384c4fd-cb04-4d7c-ac13-ac7220ee17f3_00007.c000.snappy.parquet",
        "size" : ...
      ....
```
(also I just realized, root may store relative path?

imback82:
Are we utilizing `root` in our code on the read path? Or do we always use files?

imback82:
Ah, I see that it's being used. We can address this as a separated PR.

imback82:
One thing to check is we have the following:
```
case class Relation(
    rootPaths: Seq[String],
    data: Hdfs,
```
where `data.properties.content` will have one `root`. How are we generating `root` if there are multiple `rootPaths`?

apoorvedave1:
> One thing to check is we have the following:
> 
> ```
> case class Relation(
>     rootPaths: Seq[String],
>     data: Hdfs,
> ```
> 
> where `data.properties.content` will have one `root`. How are we generating `root` if there are multiple `rootPaths`?

Since this `Relation` only represents the `data` and not the `index` (as of now at least), I don't think we have targeted this problem yet.

imback82:
> Since this `Relation` only represents the `data` and not the `index` (as of now at least), I don't think we have targeted this problem yet.

My question was how we are currently generating `root` when there are multiple `rootPaths`.

imback82:
Also, I think we should always store "full path" for the `root`. When I run an unit test, I see the following:
```
"content" : {
    "root" : "src/test/resources/e2eTests/indexLocation/leftIndex/v__=0",
    "directories" : [ ]
  },
```
Let me capture these in an issue so we can track. It shouldn't affect this PR.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -29,7 +31,14 @@ case class NoOpFingerprint() {
 // IndexLogEntry-specific Content that uses IndexLogEntry-specific fingerprint.
 case class Content(root: String, directories: Seq[Content.Directory])
 object Content {
-  case class Directory(path: String, files: Seq[String], fingerprint: NoOpFingerprint)
+  case class Directory(path: String, files: Seq[FileInfo], fingerprint: NoOpFingerprint)
```
imback82, line 34: This will be a breaking change right? We cannot consume v0.2.0 index?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -29,7 +31,14 @@ case class NoOpFingerprint() {
...
+  object Directory {
+    case class FileInfo(name: String, size: Long, modifiedTime: Long)
```
imback82, line 36: `// modifiedTime is an Epoch time`

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -142,6 +147,12 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
         resolvedIndexedColumns)
   }
 
+  private def leafFileInfo(path: Path): Seq[FileInfo] = {
+    // Assuming index directories don't contain nested directories. Only Leaf files.
```
imback82, line 151: Can you assert your assumption?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -142,6 +147,12 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    val fs = path.getFileSystem(new Configuration)
+    Try { fs.listStatus(path).map(FileInfo(_)).toSeq }.getOrElse(Seq())
```
imback82, line 153: When would this throw an exception and would it be ok to silently ignore?

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -106,8 +115,10 @@ class IndexLogEntryTest extends SparkFunSuite {
       Seq(
         Relation(
           Seq("rootpath"),
-          Hdfs(Hdfs.Properties(
-            Content("", Seq(Content.Directory("", Seq("f1", "f2"), NoOpFingerprint()))))),
+          Hdfs(
+            Hdfs.Properties(Content(
+              "",
+              Seq(Content.Directory("", Seq("f1", "f2").map(toFileInfo), NoOpFingerprint()))))),
```
imback82, line 121: Can you create `FileInfo` by hand here? `Seq(FileInfo(...), )` with non zero values?

imback82: To me, just handcrafting the `FileInfo` isn't too bad.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
apoorvedave1: This will throw exception when `path` does not exist. This will happen **before** the index got created. This method is used in the `def logEntry` which is called multiple times during the index creation.
During validation phase for example, the `path` doesn't exist. At this time, we expect an exception when listing index files.

Once index has been created, the `path` exists and we don't expect this exception. I will update this section to identify file not found exception and other exceptions in the next iteration
Thanks @imback82

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
apoorvedave1: yes this will be breaking, it won't consume v020 indexes

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
apoorvedave1: Thanks @imback82 , updated

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
apoorvedave1: Thanks @imback82 , updated

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
apoorvedave1: Thanks @imback82 , updated

apoorvedave1:
> > Since this `Relation` only represents the `data` and not the `index` (as of now at least), I don't think we have targeted this problem yet.
> 
> My question was how we are currently generating `root` when there are multiple `rootPaths`.
Oh I misunderstood. [Here's how it's done](https://github.com/microsoft/hyperspace/blob/8329071d61b745008976a7e46042da275104616a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala#L113):
```
Relation(
          location.rootPaths.map(_.toString),
          Hdfs(sourceDataProperties),
          dataSchema.json,
          fileFormatName,
          opts)
```

imback82:
> Oh I misunderstood. [Here's how it's done](https://github.com/microsoft/hyperspace/blob/8329071d61b745008976a7e46042da275104616a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala#L113):

Sorry, I wasn't clear. Can you paste the example index log entry generated when there are multiple root paths?

apoorvedave1:
> > Oh I misunderstood. [Here's how it's done](https://github.com/microsoft/hyperspace/blob/8329071d61b745008976a7e46042da275104616a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala#L113):
> 
> Sorry, I wasn't clear. Can you paste the example index log entry generated when there are multiple root paths?

oh understood. For this example with multiple root paths taken from 
https://github.com/microsoft/hyperspace/blob/8329071d61b745008976a7e46042da275104616a/src/test/scala/com/microsoft/hyperspace/actions/CreateActionTest.scala#L144

The output of `println(relations.rootPaths.toString)` is 
```
List(
"file:/C:/Users/apdave/AppData/Local/Temp/spark-28655027-435d-421a-8eb0-f9d3b382db0d/t1", 
"file:/C:/Users/apdave/AppData/Local/Temp/spark-28655027-435d-421a-8eb0-f9d3b382db0d/t2"
)
```
So I guess in summary, it's whatever is passed in the `spark.read.load(path)` method. It can be directory or individual file.

imback82:
Could you please paste the output of the "index log entry" (latestStable.json)? I am interested in the output of `content.root` vs. `rootPaths`.

Also, it would be interesting to check the case where multiple root paths are the same (`spark.read.parquet(path, path)`).

apoorvedave1:
> Could you please paste the output of the "index log entry" (latestStable.json)? I am interested in the output of `content.root` vs. `rootPaths`.
> 
> Also, it would be interesting to check the case where multiple root paths are the same (`spark.read.parquet(path, path)`). (most likely it's OK to have one `content.root`.)
Attaching latest stable.json.

[latestStable.txt](https://github.com/microsoft/hyperspace/files/5087281/latestStable.txt)

Here's the comparison between `relations[i].rootPaths` and `content.root`
```
  "source" : {
    "plan" : {
      "properties" : {
        "relations" : [ {
          "rootPaths" : [ "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/sampleparquet" ],
          "data" : {
            "properties" : {
              "content" : {
                "root" : "",
                "directories" : [ {
                  "path" : "",
                  "files" : [ {
                    "name" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/sampleparquet/part-00000-03cf3dd4-57d1-473b-85f3-ff96a045eb59-c000.snappy.parquet",
                    "size" : 1473,
                    "modifiedTime" : 1597706820607
                  },
```

imback82:
Thanks @apoorvedave1

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -142,6 +149,22 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
         resolvedIndexedColumns)
   }
 
+  private def leafFileInfo(path: Path): Seq[FileInfo] = {
```
imback82, line 152: nit: `leafFilesInfo`? or even `indexFilesInfo` is fine since there are some logics related to index data structure in this function.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: Can you update "Does this PR introduce any user-facing change?"?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -29,7 +31,15 @@ case class NoOpFingerprint() {
 // IndexLogEntry-specific Content that uses IndexLogEntry-specific fingerprint.
 case class Content(root: String, directories: Seq[Content.Directory])
 object Content {
-  case class Directory(path: String, files: Seq[String], fingerprint: NoOpFingerprint)
+  case class Directory(path: String, files: Seq[FileInfo], fingerprint: NoOpFingerprint)
+  object Directory {
+    // modifiedTime is an Epoch time in milliseconds. (ms `since 1970-01-01T00:00:00.000 UTC).
```
imback82, line 36: nit: ` is a typo?

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -42,17 +43,21 @@ class IndexLogManagerImplTest
...
           Hdfs(properties = Hdfs.Properties(content = Content(
             "/root/data",
             Seq(
-              Content.Directory("dir1", Seq("1.json", "2.json"), NoOpFingerprint()),
-              Content.Directory("dir2", Seq("1.json", "2.json"), NoOpFingerprint()))))),
+              Content
+                .Directory("dir1", Seq("1.json", "2.json").map(toFileInfo), NoOpFingerprint()),
+              Content.Directory(
+                "dir2",
+                Seq("1.json", "2.json").map(toFileInfo),
```
imback82, line 59: Can these also be handcrafted?

apoorvedave1: done, thanks @imback82

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
apoorvedave1: done, thanks @imback82

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
apoorvedave1: done, thanks @imback82

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
apoorvedave1: done, thanks @imback82

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -16,13 +16,20 @@
 
 package com.microsoft.hyperspace.actions
 
+import java.io.FileNotFoundException
+
+import scala.util.{Failure, Try}
```
imback82, line 21: Do you need this?

apoorvedave1: removed. Thanks @imback82

apoorvedave1:
should I create another PR first with JoinRUle and FilterRule utilizing index `files` rather than index `root` folder only? before opening the PR for #110

imback82:
> should I create another PR first with JoinRUle and FilterRule utilizing index `files` rather than index `root` folder only? before opening the PR for #110 , cc @imback82

Is that a prerequisite? If so, yes. Can we also fix #126 first?

Also, if root is non-empty, files should not contain the root as prefix because we need to construct the file as "root" / "dir" / "file". Is this what you are trying to address?

apoorvedave1:
> > should I create another PR first with JoinRUle and FilterRule utilizing index `files` rather than index `root` folder only? before opening the PR for #110 , cc @imback82
> 
> Is that a prerequisite? If so, yes. Can we also fix #126 first?
> 
> Also, if root is non-empty, files should not contain the root as prefix because we need to construct the file as "root" / "dir" / "file". Is this what you are trying to address?

yeah I will start with that.

PR #127, opened 2020-08-18
Index metadata should always store absolute paths

### LINKED ISSUE #126: The root paths captured in index log entry should always be absolute paths
### Describe the issue

The `root` path for the index is sometimes recorded as a relative path:
```
"content" : {
    "root" : "src/test/resources/e2eTests/indexLocation/leftIndex/v__=0",
    "directories" : [ ]
  },
```
This needs to be fixed to always contain absolute paths (or URIs).

### To Reproduce

Run any e2e tests and look at the generated index metadata.

### Expected behavior

`root` should be a absolute path.

apoorvedave1:
fixes #126 

currently metadata stores relative paths if system path is relative

Yes, now all paths stored in metadata will be fully qualified. This is not a breaking change. Previously created indexes (v0.2.0) will still be readable.

unit tests

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
@@ -44,10 +44,10 @@ trait IndexDataManager {
 }
 
 class IndexDataManagerImpl(indexPath: Path) extends IndexDataManager {
+  assert(indexPath.isAbsolute)
+
   // TODO: Investigate whether FileContext should be used instead of FileSystem for atomic renames.
   private lazy val fs: FileSystem = indexPath.getFileSystem(new Configuration)
-
-  private lazy val absoluteIndexPath: Path = fs.makeQualified(indexPath)
 
   /**
    * This method relies on the naming convention that directory name will be similar to hive
@@ -55,11 +55,11 @@ class IndexDataManagerImpl(indexPath: Path) extends IndexDataManager {
    * version id of the data.
    */
   override def getLatestVersionId(): Option[Int] = {
-    if (!fs.exists(absoluteIndexPath)) {
+    if (!fs.exists(indexPath)) {
       return None
     }
     val prefixLength = IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX.length + 1
-    val ids = fs.listStatus(absoluteIndexPath).collect {
+    val ids = fs.listStatus(indexPath).collect {
       case status
           if status.getPath.getName.startsWith(IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX) =>
         status.getPath.getName.drop(prefixLength).toInt
@@ -68,9 +68,7 @@ class IndexDataManagerImpl(indexPath: Path) extends IndexDataManager {
   }
 
   override def getPath(id: Int): Path = {
-    new Path(
-      absoluteIndexPath,
-      s"${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=${id.toString}")
+    new Path(indexPath, s"${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=${id.toString}")
   }
 
   override def delete(id: Int): Unit = FileUtils.delete(getPath(id))
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
@@ -47,17 +47,19 @@ class IndexDataManagerImpl(indexPath: Path) extends IndexDataManager {
   // TODO: Investigate whether FileContext should be used instead of FileSystem for atomic renames.
   private lazy val fs: FileSystem = indexPath.getFileSystem(new Configuration)
 
+  private lazy val absoluteIndexPath: Path = fs.makeQualified(indexPath)
+
   /**
    * This method relies on the naming convention that directory name will be similar to hive
    * partitioning scheme, i.e. "root/v__=value/f1.parquet" etc. Here the value represents the
    * version id of the data.
    */
   override def getLatestVersionId(): Option[Int] = {
-    if (!fs.exists(indexPath)) {
+    if (!fs.exists(absoluteIndexPath)) {
       return None
     }
     val prefixLength = IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX.length + 1
-    val ids = fs.listStatus(indexPath).collect {
+    val ids = fs.listStatus(absoluteIndexPath).collect {
       case status
           if status.getPath.getName.startsWith(IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX) =>
         status.getPath.getName.drop(prefixLength).toInt
@@ -66,7 +68,9 @@ class IndexDataManagerImpl(indexPath: Path) extends IndexDataManager {
   }
 
   override def getPath(id: Int): Path = {
-    new Path(indexPath, s"${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=${id.toString}")
+    new Path(
+      absoluteIndexPath,
+      s"${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=${id.toString}")
   }
 
   override def delete(id: Int): Unit = FileUtils.delete(getPath(id))
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -59,6 +59,7 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
 
   // Use FileContext instead of FileSystem for atomic renames?
   private lazy val fs: FileSystem = indexPath.getFileSystem(new Configuration)
+
   private lazy val hyperspaceLogPath: Path = new Path(indexPath, IndexConstants.HYPERSPACE_LOG)
 
   private val pathFromId: Int => Path = id => new Path(hyperspaceLogPath, id.toString)
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -57,8 +57,10 @@ trait IndexLogManager {
 class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging {
   // Use FileContext instead of FileSystem for atomic renames?
   private lazy val fs: FileSystem = indexPath.getFileSystem(new Configuration)
+  private lazy val absoluteIndexPath: Path = fs.makeQualified(indexPath)
 
-  private lazy val hyperspaceLogPath: Path = new Path(indexPath, IndexConstants.HYPERSPACE_LOG)
+  private lazy val hyperspaceLogPath: Path =
+    new Path(absoluteIndexPath, IndexConstants.HYPERSPACE_LOG)
 
   private val pathFromId: Int => Path = id => new Path(hyperspaceLogPath, id.toString)
 
@@ -113,8 +115,9 @@ class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging 
   override def createLatestStableLog(id: Int): Boolean = {
     getLog(id) match {
       case Some(logEntry) if Constants.STABLE_STATES.contains(logEntry.state) =>
-        Try(FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration))
-        match {
+        Try {
+          FileUtil.copy(fs, pathFromId(id), fs, latestStablePath, false, new Configuration)
+        } match {
           case Success(v) => v
           case Failure(e) =>
             logError(s"Failed to create the latest stable log with id = '$id'", e)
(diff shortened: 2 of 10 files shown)


src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
```diff
@@ -47,17 +47,19 @@ class IndexDataManagerImpl(indexPath: Path) extends IndexDataManager {
   // TODO: Investigate whether FileContext should be used instead of FileSystem for atomic renames.
   private lazy val fs: FileSystem = indexPath.getFileSystem(new Configuration)
 
+  private lazy val absoluteIndexPath: Path = fs.makeQualified(indexPath)
```
imback82, line 50: How about we move this to the caller side (`IndexLogManagerFactoryImpl` and `IndexDataManagerFactoryImpl`) and just do `assert(indexPath.isAbsolute)` in this class?

src/test/scala/com/microsoft/hyperspace/index/plananalysis/ExplainTest.scala
```diff
@@ -44,7 +44,8 @@ class ExplainTest extends SparkFunSuite with HyperspaceSuite {
     val dfFromSample = sampleData.toDF("Col1", "Col2")
     dfFromSample.write.parquet(sampleParquetDataLocation)
     sampleParquetDataFullPath =
-      fileSystem.getFileStatus(new Path(sampleParquetDataLocation)).getPath.toString
+      fileSystem.makeQualified(new Path(sampleParquetDataLocation)).toString
+    // fileSystem.getFileStatus(new Path(sampleParquetDataLocation)).getPath.toString
```
imback82, line 48: remove?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -407,7 +407,9 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
   }
 
   private def getIndexFilesPath(indexName: String): Path = {
-    new Path(systemPath, s"$indexName/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=0")
+    val fs = systemPath.getFileSystem(new Configuration)
```
imback82, line 410: Can we create a utils function that takes in `Path` and returns absolute path if it's not already absolute path?

src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
apoorvedave1: thanks @imback82 , done

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , done

src/test/scala/com/microsoft/hyperspace/index/plananalysis/ExplainTest.scala
apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -55,9 +55,10 @@ trait IndexLogManager {
 }
 
 class IndexLogManagerImpl(indexPath: Path) extends IndexLogManager with Logging {
+  assert(indexPath.isAbsolute)
+
   // Use FileContext instead of FileSystem for atomic renames?
   private lazy val fs: FileSystem = indexPath.getFileSystem(new Configuration)
-
```
imback82, line 60: nit: revert?

src/test/scala/com/microsoft/hyperspace/TestUtils.scala
```diff
@@ -24,4 +27,11 @@ object TestUtils {
     result.state = state
     result
   }
+
+  def makeAbsolute(location: String): Path = makeAbsolute(new Path(location))
+
+  def makeAbsolute(path: Path): Path = {
```
imback82, line 33: Can we create `PathUtils` so that it can be used in `factories.scala` as well?

src/test/scala/com/microsoft/hyperspace/TestUtils.scala
```diff
@@ -24,4 +27,11 @@ object TestUtils {
...
+    val fs: FileSystem = path.getFileSystem(new Configuration)
```
imback82, line 34: no type info needed for local val.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -32,6 +32,8 @@ import com.microsoft.hyperspace.util.FileUtils
 class IndexManagerTests extends SparkFunSuite with SparkInvolvedSuite {
   private val sampleParquetDataLocation = "src/test/resources/sampleparquet"
   private val indexStorageLocation = "src/test/resources/indexLocation"
+  private val absoluteIndexLocation = TestUtils.makeAbsolute(indexStorageLocation)
```
imback82, line 35: Can't we just make `indexStorageLocation` as absolute path instead of introducing a new var?

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -92,7 +95,7 @@ class IndexLogManagerImplTest
 
   test("testGetLog returns None if log not found") {
     val path = new Path(testRoot, "testPath")
-    assert(new IndexLogManagerImpl(path).getLog(0).isEmpty)
+    assert(logManager(path).getLog(0).isEmpty)
```
imback82, line 98: nit: you could have just made `path` as absolute?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
imback82: question: would `makeAbsolute` work even if the path doesn't exist?

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
apoorvedave1: thanks @imback82, done

src/test/scala/com/microsoft/hyperspace/TestUtils.scala
apoorvedave1: thanks @imback82, done

src/test/scala/com/microsoft/hyperspace/TestUtils.scala
apoorvedave1: thanks @imback82, done

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
apoorvedave1: thanks @imback82, done

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82, done. Yeah it works (e.g. index log manager and index data manager are created before the folders exist)

src/main/scala/com/microsoft/hyperspace/util/PathUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.util
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.Path
+
+object PathUtils {
+
```
imback82, line 23: nit: no empty line.

src/test/scala/com/microsoft/hyperspace/TestUtils.scala
```diff
@@ -16,6 +16,9 @@
 
 package com.microsoft.hyperspace
 
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
```
imback82, line 20: nit: revert these?

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -187,7 +187,7 @@ class IndexLogManagerImplTest
   }
 
   test("testUpdateLatestStableLog fails with exception if unable to find a valid log entry") {
-    val path = new Path(testRoot, UUID.randomUUID().toString)
+    val path = PathUtils.makeAbsolute(new Path(testRoot, UUID.randomUUID().toString))
```
imback82, line 190: can we just make `testRoot` an absolute path?

src/test/scala/com/microsoft/hyperspace/index/plananalysis/ExplainTest.scala
```diff
@@ -22,8 +22,9 @@ import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkFunSuite
 import org.apache.spark.sql.DataFrame
 
-import com.microsoft.hyperspace.{Hyperspace, Implicits}
+import com.microsoft.hyperspace.{Hyperspace, Implicits, TestUtils}
```
imback82, line 25: nit: revert?

src/main/scala/com/microsoft/hyperspace/util/PathUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+  def makeAbsolute(location: String): Path = makeAbsolute(new Path(location))
+
+  def makeAbsolute(path: Path): Path = {
+    val fs = path.getFileSystem(new Configuration)
+    fs.makeQualified(path)
```
imback82, line 28: I guess already absolute path is a no-op here, right?

imback82 (dismissed):
LGTM except for few nit comments. Thanks @apoorvedave1!

Looks like your initial approach would have saved churns in unit tests. Let me think about it more, but this PR itself looks fine to me.

src/main/scala/com/microsoft/hyperspace/util/PathUtils.scala
apoorvedave1: yes that's correct. I could add the check (if not absolute then update) but I thought this looks cleaner

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
apoorvedave1: thanks @imback82 , done

src/test/scala/com/microsoft/hyperspace/index/plananalysis/ExplainTest.scala
apoorvedave1: thanks @imback82 , done

src/test/scala/com/microsoft/hyperspace/TestUtils.scala
apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/util/PathUtils.scala
apoorvedave1: thanks @imback82 , done

apoorvedave1:
> LGTM except for few nit comments. Thanks @apoorvedave1!
> 
> Looks like your initial approach would have saved churns in unit tests. Let me think about it more, but this PR itself looks fine to me.

thanks @imback82 , I slightly prefer the initial approach because I was thinking it's OK to allow the caller to make mistakes and send relative paths, and we will make it qualified before using it in log files. This would save users some time in calling with relative path, failing and then calling it again with absolute path.
But I am ok with this approach as well.

imback82:
> I slightly prefer the initial approach because I was thinking it's OK to allow the caller to make mistakes and send relative paths, and we will make it qualified before using it in log files. This would save users some time in calling with relative path, failing and then calling it again with absolute path.
> But I am ok with this approach as well.

Hmm, I just checked the code again, and I think we just need to update `CreateActionBase.getIndexLogEntry` to handle the `path`, no? 

https://github.com/microsoft/hyperspace/blob/4cc3f74c27e6ee8600257997751bc3f4eba7c05f/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala#L86

Then, we don't need to worry about IndexDataManager, at which level to introduce, etc.

apoorvedave1:
> > I slightly prefer the initial approach because I was thinking it's OK to allow the caller to make mistakes and send relative paths, and we will make it qualified before using it in log files. This would save users some time in calling with relative path, failing and then calling it again with absolute path.
> > But I am ok with this approach as well.
> 
> Hmm, I just checked the code again, and I think we just need to update `CreateActionBase.getIndexLogEntry` to handle the `path`, no?
> 
> https://github.com/microsoft/hyperspace/blob/4cc3f74c27e6ee8600257997751bc3f4eba7c05f/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala#L86
> 
> Then, we don't need to worry about IndexDataManager, at which level to introduce, etc.

yeah this is also fine with me. I can start a fresh PR with this change to keep things clean.

imback82:
Closed by #128

Issue #60, opened 2020-06-24
Optimize Indexes support


apoorvedave1:
### Describe the issue
Blocked by #29 
Once hyperspace has an incremental indexing support, add support for optimizing indexes. This would pick parts of indexes which are made of large number of small files, and combine/compact them into small number of large files. This would improve the performance of the index since the number of index files would be greatly reduced

### To Reproduce

NA
### Expected behavior

`hyperspace.optimize(indexName)` would update indexes to improve their performance
### Environment 

NA

apoorvedave1:
duplicate of #111

PR #123, opened 2020-08-13
HybridScan for Append-only dataset

sezruby:
### What changes were proposed in this pull request?

Hybrid Scan for append-only dataset; indexes can still be utilized even though there are newly appended files in source relations.

- A new config
  - `spark.hyperspace.index.hybridscan.enabled`, default=`false` 
- Signature validation
  - calculate signature value only with the file list of a candidate index
- On-the-fly shuffle of appended-data
  - perform by`RepartitionByExpression` + IndexConfig
- Merge index data + shuffled appended data
  - Spark's Union doesn't retain the partition info after union operation
  - newly introduced `BucketUnion`
    - can retain the bucket info
    - cannot retain the sort info - additional `Sort`before SortMergeJoin
- Join Index Rule w/ hybrid scan
  - on-the-fly shuffle for appended data
    - BucketUnion => Project => Filter => Relation
- Filter Index Rule w/ hybrid scan
  - parquet source files: add the list of appended file paths to the location of Relation node for index
  - other format: couldn't just read the appended files along with index data files (parquet)
      - will be optimized with #146 
      - on-the-fly shuffle + BucketUnion + (use BucketSpec)
- rank algorithm update
  - if hybrid scan is enabled, pick the index which covers largest number of source files to minimize the amount of on-the-fly shuffle

BucketUnion of hyperspace won't cover
- getPreferredLocation
- outputOrdering 
- with CodegenSupport

### Why are the changes needed?

Indexes can be still utilized for appended dataset without refresh

### Does this PR introduce _any_ user-facing change?

Yes

Join plan example
```
=============================================================
Plan with indexes:
=============================================================
Project [id#0, name#1]
+- SortMergeJoin [id#0], [id#9], Inner
   :- *(3) Sort [id#0 ASC NULLS FIRST], false, 0
   <----:  +- BucketUnionExecutor 203 buckets, bucket columns: [id]---->
   <----:     :- *(1) Project [name#1, id#0]---->
   <----:     :  +- *(1) Filter (isnotnull(id#0) && (id#0 >= 1))---->
[...]

PR #129, opened 2020-08-20
[WIP] Deduplicate metadata path structure to use file path  = "root/directory/file"

apoorvedave1:
Changes in this PR:
- Index metadata will store index file contents as below (note the `<root>/<directory>/<file>` structure:
```
  "content" : {
    "root" : "file:/C:/Users/apdave/github/hyperspace-1/src/test/resources/indexLocation/index1",
    "directories" : [ {
      "path" : "v__=0",
      "files" : [ {
        "name" : "part-00007-ec24eb01-d061-4cd7-b10f-9ba9c3fdc972_00007.c000.snappy.parquet",
        "size" : 926,
        "modifiedTime" : 1597960540839
      }, {
        "name" : "part-00011-ec24eb01-d061-4cd7-b10f-9ba9c3fdc972_00011.c000.snappy.parquet",
        "size" : 926,
        "modifiedTime" : 1597960540730
      }, ...
```
- JoinIndexRule, FilterIndexRule (and future rules) will read from a list of index **files** which comprise of the latest snapshot of the index. This was not true in previous versions. Earlier, the index was represented by just its root directory.

For support of future features like #110 and #60 , it's now required to read a subset of index files from a list of index directories. For this, we are updating the metadata to store a complete list of files which represent the current state of the index.

yes, the metadata layout is changed. Previously created indexes will be unreadable.

unit tests

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -87,7 +87,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
           Content(
             root = absolutePath.getParent.toString,
             directories = Seq(
-              Directory("", indexFilesInfo(absolutePath), NoOpFingerprint()))),
+              Directory(absolutePath.getName, indexFilesInfo(absolutePath), NoOpFingerprint()))),
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -83,7 +83,8 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
                 .Columns(resolvedIndexedColumns, resolvedIncludedColumns),
               IndexLogEntry.schemaString(schema),
               numBuckets)),
-          Content(path.toString, Seq(Directory("", indexFilesInfo(path), NoOpFingerprint()))),
+          Content(root = path.getParent.toString,
+            directories = Seq(Directory(path.getName, indexFilesInfo(path), NoOpFingerprint()))),
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -87,7 +87,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
           Content(
             root = absolutePath.getParent.toString,
             directories = Seq(
-              Directory(absolutePath.getName, indexFilesInfo(absolutePath), NoOpFingerprint()))),
+              Directory("", indexFilesInfo(absolutePath), NoOpFingerprint()))),
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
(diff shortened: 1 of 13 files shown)


src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -85,8 +85,9 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
               IndexLogEntry.schemaString(schema),
               numBuckets)),
           Content(
-            absolutePath.toString,
-            Seq(Directory("", indexFilesInfo(absolutePath), NoOpFingerprint()))),
+            root = absolutePath.getParent.toString,
+            directories = Seq(
+              Directory(absolutePath.getName, indexFilesInfo(absolutePath), NoOpFingerprint()))),
```
imback82, line 90: I am actually fine leaving the `Directory.path` as "" to make it simple for now.

For example, if you have
```
/root/p=1/f1.txt
/root/p=2/f2.txt
```
We should theoretically store them as
```
root = "/root"
directories = Seq(Directory("p=1", Seq(FileInfo("f1"), _), Directory("p=2", Seq(FileInfo("f2"), _))
```
but it could get complicated to optimize when you have multiple partitions, so the following is also fine:
```
root = "/root"
directories = Seq(Directory("", Seq(FileInfo("p=1/f1"), _), Directory("", Seq(FileInfo("p=2/f2"), _))
```
as long as `full file path = "root/dir/filename"` holds true.

apoorvedave1: yes that works for me. Let me change the code to this format.
Question @imback82 : is it fine to switch from directories to files while using indexes (e.g. in this [line](https://github.com/microsoft/hyperspace/blob/3a1dfa74ca369eedf8f19330f131dc21085a1c84/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala#L107) use individual files in `InMemoryFileIndex.rootPaths` field.

apoorvedave1: we will eventually have to move here for `optimize`. cc @sezruby , @pirz

imback82: > yes that works for me. Let me change the code to this format.

Btw, it was just a suggestion. If you can handle it in this PR, please go ahead


> Question @imback82 : is it fine to switch from directories to files while using indexes

I think it will be fine, but I would measure it once at least. Btw, we can also utilize `PathFilter` instead if needed.

apoorvedave1: > > yes that works for me. Let me change the code to this format.
> 
> Btw, it was just a suggestion. If you can handle it in this PR, please go ahead
Oh thanks for clarification @imback82 . In that case, let's do the fix in this PR itself. I will ask @pirz and @sezruby if it works for them.
> 
> > Question @imback82 : is it fine to switch from directories to files while using indexes
> 
> I think it will be fine, but I would measure it once at least. Btw, we can also utilize `PathFilter` instead if needed (I am not sure which one will be better since we would already have the diff).
Sure, I will start some benchmarks before final merging of this. Thanks.

apoorvedave1: cc @pirz , @sezruby , could you please have a general look at the code if it works with your changes?
(not looking for a full review, just a quick glance should suffice)

imback82: Actually @apoorvedave1 has a better plan to revamp the `Content` structure. So we can close this PR for now. Thanks!

imback82:
@apoorvedave1 will come up with a better structure in index metadata to capture file structures. I am closing this for now. Thanks!

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
sezruby: @apoorvedave1 I think changing index directory structure won't affect #123.

> Question @imback82 : is it fine to switch from directories to files while using indexes

Could you let me know why do you need this?

BTW I think you could also use BucketUnion(#123) to read multiple root paths though it doesn't support outOrdering. What do you think?

Issue #150, opened 2020-09-11
Hybrid Scan for File/Partition Mutable Datasets


sezruby:
**Describe the problem**

After the original dataset got appended new files/partitions and/or some files/partitions got deleted, the index cannot be used until a user calls refresh API. 

**Describe your proposed solution**

Hybrid Scan is introduced to support these dataset seamlessly and also avoid frequent refreshment of indexes. This feature allows to utilize the outdated index by analyzing the difference and modifying the plan properly. The simple ideas for both are:

* appended files - on-the-fly index build for the appended data & merge into the index data
* deleted files - injecting pushdown filter conditions of ***lineage column*** to the index relation, so that the rows from the deleted files are not included in the results. (The lineage column is already introduced by #104)

Hybrid Scan can handle file/partition mutable dataset, for example:
```
// original dataset
["/path/to/part-0000.parquet", "/path/to/part-0001.parquet", "/path/to/part-0002.parquet"]

// got appended & deleted
[                              "/path/to/part-0001.parquet", "/path/to/part-0002.parquet", "/path/to/part-0003.parquet"]
```

##### Append-only
#165 addresses the plan modification for append-only dataset.

Basically there are 3 cases for now:

- **_Case 1_)** Filter Index Rule & parquet source format
  - In this case, we can just add the file list of appended files to the file list of the index relation (i.e. index data). It's because: 
    - it's guaranteed that newly appended source files always have the all columns in the index data (except for lineage).
    - Filter Index Rule does not utilize bucketing information for now; able to read both index data & newly appended data with 1 FileScan node. See below `InMemoryFileIndex`
```
...
<----:- *(1) Project [name#1, id#0]---->
<----:  +- *(1) Filter (isnotnull(id#0) && (id#0 >= 1))---->
<----:     +- *(1) FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, 
                     Location: InMemoryFileIndex[<list of index data files> , <list of newly appended files>], PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,1)], ReadSchema: struct<id:int,name:string>---->
...
```
- **_Case 2_)** Filter Index Rule  & non-parquet source format
  - Other than Parquet, we cannot simply read the data along with index data files in Parquet format with 1 FileScan node. So we need to merge both data in a different way. We could use the below Case3 for this, but Filter Index Rule doesn't use the bucketing information; it doesn't need to shuffle newly appended data.
[...]

PR #151, opened 2020-09-11
Hybrid scan operator for leveraging index alongside newly appended data - BucketUnion

sezruby:

For full context on the "why" for this PR, please see the main issue: https://github.com/microsoft/hyperspace/issues/150

Introducing a new `BucketUnion` operator that is useful for implementing hybrid scan, a technique that we propose that can leverage the index *and* appended data without the need to re-shuffle the index, thus preserving the benefits of the index.

As part of this, the following classes are being implemented:

  1. `BucketUnion`: Logical Plan operator; Used during logical plan optimization when the newly appended data needs to be union-ed with the data being read from the index.
  2. `BucketUnionExec`: SparkPlan (Physical operator); Calls into `BucketUnionRDD`
  3. `BucketUnionRDD`: RDD operator
  4. `BucketUnionRDDPartition`: Partition
  5. `BucketUnionStrategy`: SparkStrategy that is used when the Logical Plan is being converted to SparkPlan; More specifically, converts `BucketUnion` to `BucketUnionExec` 

> Note: You can find more detailed information about Bucketing optimization in [Bucketing 2.0: Improve Spark SQL Performance by Removing Shuffle](https://youtu.be/7cvaH33S7uc)

Spark does not support Union using `PartitionSpecification`, but just `PartitionerAwareUnionRDD` operation which does not retain outputPartitioning of result. Therefore, we define a new Union operation (being called the `BucketUnion`) which works when the following conditions are satisfied:

- input RDDs must have the same number of partitions.
- input RDDs must have the same partitioning keys.
- input RDDs must have the same column schema.

Unfortunately, since there is no explicit API to check Partitioning keys in RDD, we have to assure that on the caller side. Therefore, `BucketUnionRDD` is Hyperspace internal use only.

`BucketUnion` can be used to merge index data & newly appended data without losing the bucketing specification (`outputPartitioning`). 

**Existing experience**:
When the underlying data changes, Hyperspace decides to not use the index anymore.

**New experience**:
When the underlying data changes *and* hybrid scan is enabled, Hyperspace utilizes the index to the extent possible and performs a linear scan on the new data.

BucketUnionTest

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/plans/logical/BucketUnion.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/plans/logical/BucketUnion.scala
@@ -1,0 +1,68 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.index.plans.logical
+
+import org.apache.spark.sql.catalyst.catalog.BucketSpec
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+
+/**
+ * [[BucketUnion]] is logical plan for Bucket-aware Union operation which retains
+ * outputPartitioning of result RDDs so as to avoid performing unnecessary Shuffle after
+ * Union operation per bucket.
+ *
+ * @param children Child plans.
+ * @param bucketSpec Bucket Specification.
+ */
+private[hyperspace] case class BucketUnion(children: Seq[LogicalPlan], bucketSpec: BucketSpec)
+    extends LogicalPlan {
+  require(resolved)
+  override def output: Seq[Attribute] = children.head.output
+
+  // copy from org.apache.spark.sql.catalyst.plans.logical.Union
+  override def maxRows: Option[Long] = {
+    if (children.exists(_.maxRows.isEmpty)) {
+      None
+    } else {
+      Some(children.flatMap(_.maxRows).sum)
+    }
+  }
+
+  // copy from org.apache.spark.sql.catalyst.plans.logical.Union
+  override def maxRowsPerPartition: Option[Long] = {
+    if (children.exists(_.maxRowsPerPartition.isEmpty)) {
+      None
+    } else {
+      Some(children.flatMap(_.maxRowsPerPartition).sum)
+    }
+  }
+
+  // copy from org.apache.spark.sql.catalyst.plans.logical.Union
+  override lazy val resolved: Boolean = {
+    // allChildrenCompatible needs to be evaluated after childrenResolved
+    def allChildrenCompatible: Boolean =
+      children.tail.forall(
+        child =>
+          // compare the attribute number with the first child
+          child.output.length == children.head.output.length &&
+            // compare the data types with the first child
+            child.output.zip(children.head.output).forall {
+              case (l, r) => l.dataType.equals(r.dataType)
+          })
+    children.length > 1 && childrenResolved && allChildrenCompatible
+  }
+}
--- a/src/main/scala/com/microsoft/hyperspace/index/plans/logical/BucketUnion.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/plans/logical/BucketUnion.scala
@@ -22,7 +22,7 @@ import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 
 /**
  * [[BucketUnion]] is logical plan for Bucket-aware Union operation which retains
- * outputPartitioning of result RDDs so as to avoid performing unnecessary Shuffle after
+ * outputPartitioning of result RDDs so as to avoid performing unnecessary shuffle after
  * Union operation per bucket.
  *
  * @param children Child plans.
--- a/src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
@@ -1,0 +1,134 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.index
+
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.catalyst.catalog.BucketSpec
+
+import com.microsoft.hyperspace.index.execution.{BucketUnionExec, BucketUnionRDD, BucketUnionRDDPartition, BucketUnionStrategy}
+import com.microsoft.hyperspace.index.plans.logical.BucketUnion
+
+class BucketUnionTest extends HyperspaceSuite {
+  override val systemPath = new Path("src/test/resources/bucketUnionTest")
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+  }
+
+  override def afterAll(): Unit = {
+    super.afterAll()
+  }
+
+  test("BucketUnion require test") {
+    import spark.implicits._
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val df1_2 = Seq(("name1", 1), ("name2", 2)).toDF("name", "id")
+    val df2 = Seq((1, "name1", 20), (2, "name2", 10)).toDF("id", "name", "age")
+
+    intercept[IllegalArgumentException] {
+      BucketUnion(
+        Seq(df1.queryExecution.optimizedPlan, df2.queryExecution.optimizedPlan),
+        BucketSpec(1, Seq(), Seq()))
+      fail("shouldn't be reached")
+    }
+    intercept[IllegalArgumentException] {
+      BucketUnion(
+        Seq(df1.queryExecution.optimizedPlan, df1_2.queryExecution.optimizedPlan),
+        BucketSpec(1, Seq(), Seq()))
+      fail("shouldn't be reached")
+    }
+    BucketUnion(
+      Seq(df1.queryExecution.optimizedPlan, df1_1.queryExecution.optimizedPlan),
+      BucketSpec(1, Seq(), Seq()))
+  }
+
+  test("BucketUnionStrategy test") {
+    import spark.implicits._
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val bucket = BucketUnion(
+      Seq(df1.queryExecution.optimizedPlan, df1_1.queryExecution.optimizedPlan),
+      BucketSpec(1, Seq(), Seq()))
+
+    var bucketUnionExecSeen = false
+    BucketUnionStrategy(bucket).collect {
+      case BucketUnionExec(_, _) => bucketUnionExecSeen = true
+    }
+    assert(bucketUnionExecSeen)
+
+    bucketUnionExecSeen = false
+    BucketUnionStrategy(df1.queryExecution.optimizedPlan).collect {
+      case BucketUnionExec(_, _) => bucketUnionExecSeen = true
+    }
+    assert(!bucketUnionExecSeen)
+  }
+
+  test("BucketUnionExec test") {
+    import spark.implicits._
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val p1 = df1.repartition(10)
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val p1_1 = df1_1.repartition(9)
+    val p1_2 = df1_1.repartition(10)
+
+    intercept[AssertionError] {
+      val bucket = BucketUnion(
+        Seq(p1.queryExecution.optimizedPlan, p1_1.queryExecution.optimizedPlan),
+        BucketSpec(10, Seq(), Seq()))
+      spark.sessionState.executePlan(bucket).sparkPlan
+      fail("shouldn't be reached")
+    }
+
+    val bucket = BucketUnion(
+      Seq(p1.queryExecution.optimizedPlan, p1_2.queryExecution.optimizedPlan),
+      BucketSpec(10, Seq(), Seq()))
+
+    val exec = BucketUnionStrategy(bucket)
+
+    val plan = exec.collect {
+      case p: BucketUnionExec =>
+        assert(p.bucketSpec.numBuckets == 10)
+        assert(p.children.length == 2)
+        assert(p.output.length == p1.schema.fields.length)
+    }
+  }
+
+  test("BucketUnionRDD test") {
+    import spark.implicits._
+    val df1 = Seq((2, "name1"), (3, "name2")).toDF("id", "name")
+    val p1 = df1.repartition(10, $"id")
+    val df1_1 = Seq((2, "name1"), (3, "name2")).toDF("id", "name")
+    val p1_1 = df1_1.repartition(10, $"id")
+    val bucketSpec = BucketSpec(10, Seq("id"), Seq())
+
+    val rdd = new BucketUnionRDD[Row](spark.sparkContext, Seq(p1.rdd, p1_1.rdd), bucketSpec)
+    assert(rdd.getPartitions.length == 10)
+    assert(rdd.collect.length == 4)
+    rdd.partitions.head.asInstanceOf[BucketUnionRDDPartition]
+
+    val partitionSum = rdd
+      .mapPartitionsWithIndex {
+        case (partitionNum, it) => Iterator.single(partitionNum -> it.map(r => r.getInt(0)).sum)
+      }
+      .collect()
+      .toSeq
+
+    val availableSum = Set(0, 4, 6, 10)
+    assert(partitionSum.forall(p => availableSum.contains(p._2)))
+  }
+}
--- a/src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
@@ -25,13 +25,6 @@ import com.microsoft.hyperspace.index.plans.logical.BucketUnion
 
 class BucketUnionTest extends HyperspaceSuite {
   override val systemPath = new Path("src/test/resources/bucketUnionTest")
-  override def beforeAll(): Unit = {
-    super.beforeAll()
-  }
-
-  override def afterAll(): Unit = {
-    super.afterAll()
-  }
 
   test("BucketUnion require test") {
     import spark.implicits._
@@ -40,18 +33,20 @@ class BucketUnionTest extends HyperspaceSuite {
     val df1_2 = Seq(("name1", 1), ("name2", 2)).toDF("name", "id")
     val df2 = Seq((1, "name1", 20), (2, "name2", 10)).toDF("id", "name", "age")
 
+    // different column schema
     intercept[IllegalArgumentException] {
       BucketUnion(
         Seq(df1.queryExecution.optimizedPlan, df2.queryExecution.optimizedPlan),
         BucketSpec(1, Seq(), Seq()))
-      fail("shouldn't be reached")
     }
+
+    // different order of columns
     intercept[IllegalArgumentException] {
       BucketUnion(
         Seq(df1.queryExecution.optimizedPlan, df1_2.queryExecution.optimizedPlan),
         BucketSpec(1, Seq(), Seq()))
-      fail("shouldn't be reached")
     }
+
     BucketUnion(
       Seq(df1.queryExecution.optimizedPlan, df1_1.queryExecution.optimizedPlan),
       BucketSpec(1, Seq(), Seq()))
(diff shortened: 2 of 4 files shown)


src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val df1_2 = Seq(("name1", 1), ("name2", 2)).toDF("name", "id")
+    val df2 = Seq((1, "name1", 20), (2, "name2", 10)).toDF("id", "name", "age")
+
+    intercept[IllegalArgumentException] {
+      BucketUnion(
+        Seq(df1.queryExecution.optimizedPlan, df2.queryExecution.optimizedPlan),
+        BucketSpec(1, Seq(), Seq()))
+      fail("shouldn't be reached")
```
imback82, line 47: Not needed. I think if exception is not thrown `intercept` will fail.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+    }
+    intercept[IllegalArgumentException] {
```
imback82, line 49: Can you add a short comment why you expect this to throw an exception?

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val bucket = BucketUnion(
+      Seq(df1.queryExecution.optimizedPlan, df1_1.queryExecution.optimizedPlan),
+      BucketSpec(1, Seq(), Seq()))
+
+    var bucketUnionExecSeen = false
+    BucketUnionStrategy(bucket).collect {
+      case BucketUnionExec(_, _) => bucketUnionExecSeen = true
+    }
+    assert(bucketUnionExecSeen)
```
imback82, line 72:
```scala
    assert(BucketUnionStrategy(bucket).collect {
      case _: BucketUnionExec => true
    }.length == 1)
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+
+    bucketUnionExecSeen = false
+    BucketUnionStrategy(df1.queryExecution.optimizedPlan).collect {
+      case BucketUnionExec(_, _) => bucketUnionExecSeen = true
+    }
+    assert(!bucketUnionExecSeen)
```
imback82, line 78:
```scala
    assert(BucketUnionStrategy(df1.queryExecution.optimizedPlan).collect {
      case _: BucketUnionExec => true
    }.isEmpty)
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val p1_1 = df1_1.repartition(9)
+    val p1_2 = df1_1.repartition(10)
+
+    intercept[AssertionError] {
+      val bucket = BucketUnion(
+        Seq(p1.queryExecution.optimizedPlan, p1_1.queryExecution.optimizedPlan),
+        BucketSpec(10, Seq(), Seq()))
+      spark.sessionState.executePlan(bucket).sparkPlan
+      fail("shouldn't be reached")
```
imback82, line 94: ditto

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+      Seq(p1.queryExecution.optimizedPlan, p1_2.queryExecution.optimizedPlan),
+      BucketSpec(10, Seq(), Seq()))
+
+    val exec = BucketUnionStrategy(bucket)
+
+    val plan = exec.collect {
+      case p: BucketUnionExec =>
+        assert(p.bucketSpec.numBuckets == 10)
+        assert(p.children.length == 2)
+        assert(p.output.length == p1.schema.fields.length)
```
imback82, line 107: You need to make sure `plan.length == 1`, otherwise, these asserts may not even hit and the test can pass.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+      fail("shouldn't be reached")
+    }
+
+    val bucket = BucketUnion(
+      Seq(p1.queryExecution.optimizedPlan, p1_2.queryExecution.optimizedPlan),
+      BucketSpec(10, Seq(), Seq()))
+
+    val exec = BucketUnionStrategy(bucket)
+
+    val plan = exec.collect {
```
imback82, line 103: nit: `val plan = BucketUnionStrategy(bucket).collect {` to be consistent with other places.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+        assert(p.children.length == 2)
+        assert(p.output.length == p1.schema.fields.length)
+    }
+  }
+
+  test("BucketUnionRDD test") {
+    import spark.implicits._
+    val df1 = Seq((2, "name1"), (3, "name2")).toDF("id", "name")
+    val p1 = df1.repartition(10, $"id")
+    val df1_1 = Seq((2, "name1"), (3, "name2")).toDF("id", "name")
```
imback82, line 115: nit: Can we just name these as `df2` and `p2`? I don't see the importance of "_" here.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+    val p1_1 = df1_1.repartition(10, $"id")
+    val bucketSpec = BucketSpec(10, Seq("id"), Seq())
+
+    val rdd = new BucketUnionRDD[Row](spark.sparkContext, Seq(p1.rdd, p1_1.rdd), bucketSpec)
+    assert(rdd.getPartitions.length == 10)
+    assert(rdd.collect.length == 4)
+    rdd.partitions.head.asInstanceOf[BucketUnionRDDPartition]
```
imback82, line 122: what is this line?

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,134 @@
...
+
+    val partitionSum = rdd
+      .mapPartitionsWithIndex {
+        case (partitionNum, it) => Iterator.single(partitionNum -> it.map(r => r.getInt(0)).sum)
+      }
+      .collect()
+      .toSeq
+
+    val availableSum = Set(0, 4, 6, 10)
+    assert(partitionSum.forall(p => availableSum.contains(p._2)))
```
imback82, line 132: Can you add comments why tests are written this way? Sorry, I didn't really get this.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
imback82: did you mean `assert(rdd.partitions.head.isInstanceOf[BucketUnionRDDPartition])`?

sezruby: Yes asInstanceOf throws an exception for a failure but I'll change to isInstanceOf.
Thanks!

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
sezruby: Sorry I added some comment for this.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+package com.microsoft.hyperspace.index
+
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.catalyst.catalog.BucketSpec
+
+import com.microsoft.hyperspace.index.execution.{BucketUnionExec, BucketUnionRDD, BucketUnionRDDPartition, BucketUnionStrategy}
+import com.microsoft.hyperspace.index.plans.logical.BucketUnion
+
+class BucketUnionTest extends HyperspaceSuite {
```
imback82, line 26: nit: Can we just do `extends SparkFunSuite with SparkInvolvedSuite` and remove `systemPath`?

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+  override val systemPath = new Path("src/test/resources/bucketUnionTest")
+
+  test("BucketUnion require test") {
+    import spark.implicits._
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val df1_2 = Seq(("name1", 1), ("name2", 2)).toDF("name", "id")
+    val df2 = Seq((1, "name1", 20), (2, "name2", 10)).toDF("id", "name", "age")
```
imback82, line 34: nit: can we just do `df2`, `df3`, and `df4`?

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+
+    BucketUnion(
+      Seq(df1.queryExecution.optimizedPlan, df1_1.queryExecution.optimizedPlan),
+      BucketSpec(1, Seq(), Seq()))
+  }
+
+  test("BucketUnionStrategy test") {
+    import spark.implicits._
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
```
imback82, line 58: nit: `df2`

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    }.isEmpty)
+  }
+
+  test("BucketUnionExec test") {
+    import spark.implicits._
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val p1 = df1.repartition(10)
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val p1_1 = df1_1.repartition(9)
+    val p1_2 = df1_1.repartition(10)
```
imback82, line 78: nit:
```scala
val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
val p1 = df1.repartition(10)
val df2 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
val p2_1 = df2.repartition(9)
val p2_2 = df2.repartition(10)
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    val partitionSum = rdd
+      .mapPartitionsWithIndex {
+        case (partitionNum, it) => Iterator.single(partitionNum -> it.map(r => r.getInt(0)).sum)
+      }
+      .collect()
+      .toSeq
+
+    // since rows with id=2 assigned to the same partition and likewise rows with id=3,
+    // summation of each partition should be one of [0, 4, 6, 10]
+    val availableSum = Set(0, 4, 6, 10)
```
imback82, line 123: Shouldn't the test pass without `10`? I guess that's where my confusion came from.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    assert(partitionSum.forall(p => availableSum.contains(p._2)))
```
imback82, line 124: `assert(partitionSum.map(_._2).toSet == availableSum)`? `partitionSum` may not contain all of `availableSum` and can still pass with the current approach.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    assert(rdd.partitions.head.isInstanceOf[BucketUnionRDDPartition])
+
+    val partitionSum = rdd
+      .mapPartitionsWithIndex {
+        case (partitionNum, it) => Iterator.single(partitionNum -> it.map(r => r.getInt(0)).sum)
+      }
+      .collect()
+      .toSeq
+
+    // since rows with id=2 assigned to the same partition and likewise rows with id=3,
```
imback82, line 121: nit: Can we start a comment with a capital letter and end with "."?

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    import spark.implicits._
+    val df1 = Seq((2, "name1"), (3, "name2")).toDF("id", "name")
+    val p1 = df1.repartition(10, $"id")
+    val df2 = Seq((2, "name1"), (3, "name2")).toDF("id", "name")
+    val p2 = df2.repartition(10, $"id")
+    val bucketSpec = BucketSpec(10, Seq("id"), Seq())
+
+    val rdd = new BucketUnionRDD[Row](spark.sparkContext, Seq(p1.rdd, p2.rdd), bucketSpec)
+    assert(rdd.getPartitions.length == 10)
+    assert(rdd.collect.length == 4)
```
imback82, line 111: Shall we check the content as well?

imback82 (commented):
Few minor comments, but LGTM, thanks @sezruby!

imback82:
@apoorvedave1 @rapoth @pirz Can you review as well?

imback82:
@sezruby Could you update the title to be a bit more descriptive? (it's used as a commit message).

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
apoorvedave1, line 124: nit: we can avoid partition id:
```suggestion
    val partitionSum: Seq[Int] = rdd
      .mapPartitions(it => Iterator.single(it.map(r => r.getInt(0)).sum))
      .collect()
      .toSeq

    // since rows with id=2 assigned to the same partition and likewise rows with id=3,
    // summation of each partition should be one of [0, 4, 6, 10]
    val availableSum = Set(0, 4, 6, 10)
    assert(partitionSum.forall(p => availableSum.contains(p)))
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
apoorvedave1, line 123: How about hardcoding what each partition sum will be and just comparing like the following:
```scala
assert(partitionSum.equals(Seq(0, 6, 0, 0, 4, 0, 0, 0, 0, 0)))
```
assuming that the hashing function math doesn't change in spark.

apoorvedave1: I think that's because 10 is a _possible_ sum if all '2's and '3's fall in the same partition.

apoorvedave1: Also could you please add a comment here that
```
// Check if all partitioned keys with the same value fall in the same partition.
```
(unless my understanding of this test case is incorrect)

apoorvedave1 (approved):
a few minor comments, otherwise LGTM, thanks @sezruby

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
imback82: Well, this test should be deterministic.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+        Seq(df1.queryExecution.optimizedPlan, df1_2.queryExecution.optimizedPlan),
+        BucketSpec(1, Seq(), Seq()))
+    }
+
+    BucketUnion(
+      Seq(df1.queryExecution.optimizedPlan, df1_1.queryExecution.optimizedPlan),
+      BucketSpec(1, Seq(), Seq()))
+  }
+
+  test("BucketUnionStrategy test") {
```
pirz, line 55: nit: Can the test case names be a bit more descriptive, specifying very briefly what the test is doing? (similar to other test classes we have s.a. `CreateIndexTests.scala`). 
It would be helpful both for understanding/extending the code and also any potential test failure investigation in future.

src/main/scala/com/microsoft/hyperspace/index/plans/logical/BucketUnion.scala
```diff
@@ -0,0 +1,68 @@
...
+
+package com.microsoft.hyperspace.index.plans.logical
+
+import org.apache.spark.sql.catalyst.catalog.BucketSpec
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+
+/**
+ * [[BucketUnion]] is logical plan for Bucket-aware Union operation which retains
+ * outputPartitioning of result RDDs so as to avoid performing unnecessary Shuffle after
```
rapoth, line 25: nit: 
```suggestion
 * outputPartitioning of result RDDs so as to avoid performing unnecessary shuffle after
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.catalyst.catalog.BucketSpec
+
+import com.microsoft.hyperspace.index.execution.{BucketUnionExec, BucketUnionRDD, BucketUnionRDDPartition, BucketUnionStrategy}
+import com.microsoft.hyperspace.index.plans.logical.BucketUnion
+
+class BucketUnionTest extends HyperspaceSuite {
+  override val systemPath = new Path("src/test/resources/bucketUnionTest")
+
+  test("BucketUnion require test") {
```
rapoth, line 29: Is this accurate?

```suggestion
  test("BucketUnion test for operator pre-requisites") {
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
rapoth, line 55: Is this description accurate?

```suggestion
  test("BucketUnionStrategy test if strategy introduces BucketUnionExec in the Spark Plan") {
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
rapoth, line 78: nit: 

```suggestion
    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
    val p1 = df1.repartition(10)
    val df2 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
    val p2_1 = df2.repartition(9)
    val p2_2 = df2.repartition(10)
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val p1 = df1.repartition(10)
+    val df1_1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
+    val p1_1 = df1_1.repartition(9)
+    val p1_2 = df1_1.repartition(10)
+
+    // different number of partition
+    intercept[AssertionError] {
+      val bucket = BucketUnion(
+        Seq(p1.queryExecution.optimizedPlan, p1_1.queryExecution.optimizedPlan),
```
rapoth, line 83: nit: only if you make the change above
```suggestion
        Seq(p1.queryExecution.optimizedPlan, p2_1.queryExecution.optimizedPlan),
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+        BucketSpec(10, Seq(), Seq()))
+      spark.sessionState.executePlan(bucket).sparkPlan
+    }
+
+    val bucket = BucketUnion(
+      Seq(p1.queryExecution.optimizedPlan, p1_2.queryExecution.optimizedPlan),
```
rapoth, line 89: nit: only if you make the change above
```suggestion
      Seq(p1.queryExecution.optimizedPlan, p2_2.queryExecution.optimizedPlan),
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    assert(BucketUnionStrategy(bucket).collect {
+      case BucketUnionExec(_, _) => true
+    }.length == 1)
+
+    assert(BucketUnionStrategy(df1.queryExecution.optimizedPlan).collect {
+      case BucketUnionExec(_, _) => true
+    }.isEmpty)
+  }
+
+  test("BucketUnionExec test") {
```
rapoth, line 72:
```suggestion
  test("BucketUnionExec test that partition count matches on both sides") {
```

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+
+import com.microsoft.hyperspace.index.execution.{BucketUnionExec, BucketUnionRDD, BucketUnionRDDPartition, BucketUnionStrategy}
+import com.microsoft.hyperspace.index.plans.logical.BucketUnion
+
+class BucketUnionTest extends HyperspaceSuite {
+  override val systemPath = new Path("src/test/resources/bucketUnionTest")
+
+  test("BucketUnion require test") {
+    import spark.implicits._
+    val df1 = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
```
pirz, line 31: Some of these DF definitions are repeated per test; Do you think it is possible to define them once at the class level and initialize  them in beforeAll? (similar to `partitionedDataDF` and `nonPartitionedDataDF` in `CreateIndexTests.scala`).

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    assert(BucketUnionStrategy(bucket).collect {
+      case p: BucketUnionExec =>
+        assert(p.bucketSpec.numBuckets == 10)
+        assert(p.children.length == 2)
+        assert(p.output.length == p1.schema.fields.length)
+      true
+    }.length == 1)
+  }
+
+  test("BucketUnionRDD test") {
```
rapoth, line 101: Not sure if I got this right :) 

```suggestion
  test("BucketUnionRDD test that partition columns with same value fall in the same partition") {
```

rapoth (commented):
Thanks a lot for opening a short and consise PR! I really appreciate it!

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
imback82: I actually prefer the current way. The dfs are simple enough and it makes it easier to read in this scope; I don't have to go back and forth to remember how it was defined.

src/test/scala/com/microsoft/hyperspace/index/BucketUnionTest.scala
imback82: No worries. I just checked your latest comment, but let's add "why" instead of "what this does" as suggested here: https://github.com/microsoft/hyperspace/pull/151#discussion_r487190211

apoorvedave1 (approved):
LGTM 👍 , thanks @sezruby

imback82 (approved):
LGTM, thanks @sezruby!

PR #152, opened 2020-09-11
Add HyperspaceConf

sezruby:

Add utility object for getting config values from a given spark session with default value & data type.

Without this, we need to specify the config property & data type & default value every getConf which causes redundant and duplicate code.

No

Tested with another PR

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
@@ -27,6 +27,9 @@ object IndexConstants {
   val INDEX_SEARCH_PATHS = "spark.hyperspace.index.search.paths"
   val INDEX_NUM_BUCKETS = "spark.hyperspace.index.num.buckets"
 
+  val INDEX_HYBRID_SCAN_ENABLED = "spark.hyperspace.index.hybridscan.enabled"
+  val INDEX_HYBRID_SCAN_ENABLED_DEFAULT = false
+
   // Default number of buckets is set the default value of "spark.sql.shuffle.partitions".
   val INDEX_NUM_BUCKETS_DEFAULT: Int = SQLConf.SHUFFLE_PARTITIONS.defaultValue.get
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/microsoft/hyperspace/util/HyperspaceConf.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.microsoft.hyperspace.util
+
+import org.apache.spark.sql.SparkSession
+
+import com.microsoft.hyperspace.index.IndexConstants
+
+/**
+ * Helper class to extract Hyperspace config value
```
imback82, line 24:
```suggestion
 * Helper class to extract Hyperspace-related configs from SparkSession.
```

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -27,6 +27,9 @@ object IndexConstants {
   val INDEX_SEARCH_PATHS = "spark.hyperspace.index.search.paths"
   val INDEX_NUM_BUCKETS = "spark.hyperspace.index.num.buckets"
 
+  val INDEX_HYBRID_SCAN_ENABLED = "spark.hyperspace.index.hybridscan.enabled"
+  val INDEX_HYBRID_SCAN_ENABLED_DEFAULT = false
```
imback82, line 31: `val INDEX_HYBRID_SCAN_ENABLED_DEFAULT = "false"`?

src/main/scala/com/microsoft/hyperspace/util/HyperspaceConf.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+object HyperspaceConf {
+  def hybridScanEnabled(spark: SparkSession): Boolean = {
+    spark.sessionState.conf
+      .getConfString(
```
imback82, line 29: Can we just do `spark.conf.get(`?

imback82 (approved):
LGTM, thanks @sezruby!

Issue #65, opened 2020-06-24
Update optimization rules to consider indexes in transitioning states


pirz:
### Describe the issue
Currently, [`FilterIndexRule`](https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala) and [`JoinIndexRule`](https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala) only consider indexes in the ACTIVE state for optimization. Update the code to consider indexes in transitioning states as well.

### To Reproduce
NA

### Expected behavior
Optimization rules should consider indexes in transitioning states as candidates.

### Environment 
NA

PR #153, opened 2020-09-11
Modify getCandidateIndex for hybrid scan

sezruby:

- `lazy val allSourceFileInfos: Set[FileInfo]`
  - `lazy val` because this source file list of an index referred several times when applying Rules.
  - `Set` instead of `Seq` because
    - comparison of 2 Set[FileInfo] is the only use case of allSourceFileInfos now.
    - performance issue
- `isHybridScanCandidate()`
  1. get the current file list of the given relation
  2. check if there's any common files with index.allSourceFileInfos
    - TODO: plan signature comparison #158 
    - TODO: number of common files(or size) threshold #159 

The current signature design is not appropriate for hybrid scan.
For delete support, we need to figure out the list of deleted source files and check the other source files still have the same metadata when index build time or not.

No

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -31,7 +31,7 @@ object RuleUtils {
    *
    * @param indexManager indexManager
    * @param plan logical plan
-   * @param hybridScanEnabled HybridScan config.
+   * @param hybridScanEnabled Flag that checks if hybrid scan is enabled or disabled.
    * @return indexes built for this plan
    */
   def getCandidateIndexes(
@@ -56,11 +56,11 @@ object RuleUtils {
     }
 
     def isHybridScanCandidate(entry: IndexLogEntry, files: Seq[FileInfo]): Boolean = {
-      // TODO Some threshold about the similarity of source data files - number of common
-      //  files or total size of common files.
+      // TODO: Some threshold about the similarity of source data files - number of common files or
+      //  total size of common files.
       //  See https://github.com/microsoft/hyperspace/issues/159
-      // TODO As in [[PlanSignatureProvider]], Source plan signature comparison is required
-      //  to support arbitrary source plans at index creation.
+      // TODO: As in [[PlanSignatureProvider]], Source plan signature comparison is required to
+      //  support arbitrary source plans at index creation.
       //  See https://github.com/microsoft/hyperspace/issues/158
 
       // Find a common file between the input relation & index source files.
@@ -84,8 +84,8 @@ object RuleUtils {
             location.allFiles.map(f =>
               FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
         }
-        .flatten
-      allIndexes.filter(index => index.created && isHybridScanCandidate(index, files))
+      assert(files.length == 1)
+      allIndexes.filter(index => index.created && isHybridScanCandidate(index, files.flatten))
     } else {
       allIndexes.filter(index => index.created && signatureValid(index))
     }
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -55,7 +55,7 @@ object RuleUtils {
       }
     }
 
-    def isHybridScanCandidate(entry: IndexLogEntry, files: Seq[FileInfo]): Boolean = {
+    def isHybridScanCandidate(entry: IndexLogEntry, filesByRelations: Seq[FileInfo]): Boolean = {
       // TODO: Some threshold about the similarity of source data files - number of common files or
       //  total size of common files.
       //  See https://github.com/microsoft/hyperspace/issues/159
@@ -65,7 +65,7 @@ object RuleUtils {
 
       // Find a common file between the input relation & index source files.
       // Without the threshold described above, we can utilize exists & contain functions here.
-      files.exists(entry.allSourceFileInfos.contains)
+      filesByRelations.exists(entry.allSourceFileInfos.contains)
     }
 
     // TODO: the following check only considers indexes in ACTIVE state for usage. Update
@@ -74,7 +74,7 @@ object RuleUtils {
     val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
 
     if (hybridScanEnabled) {
-      val files = plan
+      val filesByRelations = plan
         .collect {
           case LogicalRelation(
               HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
@@ -84,8 +84,9 @@ object RuleUtils {
             location.allFiles.map(f =>
               FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
         }
-      assert(files.length == 1)
-      allIndexes.filter(index => index.created && isHybridScanCandidate(index, files.flatten))
+      assert(filesByRelations.length == 1)
+      allIndexes.filter(index =>
+        index.created && isHybridScanCandidate(index, filesByRelations.flatten))
     } else {
       allIndexes.filter(index => index.created && signatureValid(index))
     }
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -19,10 +19,10 @@ package com.microsoft.hyperspace.index.rules
 import scala.collection.mutable
 
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
-import org.apache.spark.sql.execution.datasources.LogicalRelation
+import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
 
 import com.microsoft.hyperspace.actions.Constants
-import com.microsoft.hyperspace.index.{IndexLogEntry, IndexManager, LogicalPlanSignatureProvider}
+import com.microsoft.hyperspace.index.{FileInfo, IndexLogEntry, IndexManager, LogicalPlanSignatureProvider}
 
 object RuleUtils {
 
@@ -31,9 +31,13 @@ object RuleUtils {
    *
    * @param indexManager indexManager
    * @param plan logical plan
+   * @param hybridScanEnabled HybridScan config.
    * @return indexes built for this plan
    */
-  def getCandidateIndexes(indexManager: IndexManager, plan: LogicalPlan): Seq[IndexLogEntry] = {
+  def getCandidateIndexes(
+      indexManager: IndexManager,
+      plan: LogicalPlan,
+      hybridScanEnabled: Boolean = false): Seq[IndexLogEntry] = {
     // Map of a signature provider to a signature generated for the given plan.
     val signatureMap = mutable.Map[String, Option[String]]()
 
@@ -51,11 +55,36 @@ object RuleUtils {
       }
     }
 
+    def isHybridScanCandidate(entry: IndexLogEntry, files: Set[FileInfo]): Boolean = {
+      // compare all the metadata of source files
+      assert(entry.relations.length == 1)
+      val commonFiles = files.intersect(entry.allSourceFileInfos)
+      // TODO threshold for the number of CommonFiles
+      // TODO plan signature
+      commonFiles.nonEmpty
+    }
+
     // TODO: the following check only considers indexes in ACTIVE state for usage. Update
     //  the code to support indexes in transitioning states as well.
     val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
 
-    allIndexes.filter(index => index.created && signatureValid(index))
+    if (hybridScanEnabled) {
+      val files = plan
+        .collect {
+          case LogicalRelation(
+          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+          _,
+          _,
+          _) =>
+            location.allFiles.map(f =>
+              FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
+        }
+        .flatten
+        .toSet
+      allIndexes.filter(index => index.created && isHybridScanCandidate(index, files))
+    } else {
+      allIndexes.filter(index => index.created && signatureValid(index))
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -51,11 +55,35 @@ object RuleUtils {
...
 
+    def isHybridScanCandidate(entry: IndexLogEntry, files: Set[FileInfo]): Boolean = {
+      // compare all the metadata of source files
+      assert(entry.relations.length == 1)
+      // TODO threshold for the number of CommonFiles
+      // TODO plan signature
+      files.exists(entry.allSourceFileInfos.contains)
+    }
+
     // TODO: the following check only considers indexes in ACTIVE state for usage. Update
```
rapoth, line 66: I remember there is already an issue for this. If so, can we just link to the issue containing the description. @apoorvedave1 to confirm if there is an issue.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -51,11 +55,35 @@ object RuleUtils {
       }
     }
 
+    def isHybridScanCandidate(entry: IndexLogEntry, files: Set[FileInfo]): Boolean = {
+      // compare all the metadata of source files
+      assert(entry.relations.length == 1)
+      // TODO threshold for the number of CommonFiles
+      // TODO plan signature
```
pirz, line 62: nit: Can you please add a bit more info for TODO to help code readability until the changes get in ultimately. Ideally, opening/linking an issue would be great.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: It seems #65; I'll add the link here.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Done, thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -51,11 +55,40 @@ object RuleUtils {
       }
     }
 
+    def isHybridScanCandidate(entry: IndexLogEntry, files: Seq[FileInfo]): Boolean = {
+      // TODO Some threshold about the similarity of source data files - number of common
```
imback82, line 59: nit: `TODO: Some ...`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -51,11 +55,40 @@ object RuleUtils {
...
+      //  files or total size of common files.
+      //  See https://github.com/microsoft/hyperspace/issues/159
+      // TODO As in [[PlanSignatureProvider]], Source plan signature comparison is required
```
imback82, line 62: ditto

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -31,9 +31,13 @@ object RuleUtils {
...
    * @param indexManager indexManager
    * @param plan logical plan
+   * @param hybridScanEnabled HybridScan config.
    * @return indexes built for this plan
    */
-  def getCandidateIndexes(indexManager: IndexManager, plan: LogicalPlan): Seq[IndexLogEntry] = {
+  def getCandidateIndexes(
+      indexManager: IndexManager,
+      plan: LogicalPlan,
+      hybridScanEnabled: Boolean = false): Seq[IndexLogEntry] = {
```
imback82, line 40: nit: can we remove default value? Rationale: this is an user-facing config, which is set to false by default. If we later change the default to true, we need to remember to update this to true as well.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -31,9 +31,13 @@ object RuleUtils {
    *
    * @param indexManager indexManager
    * @param plan logical plan
+   * @param hybridScanEnabled HybridScan config.
```
imback82, line 34: nit: `Flag that checks if hybrid scan is enabled or disabled`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -31,9 +31,13 @@ object RuleUtils {
...
    * @return indexes built for this plan
    */
-  def getCandidateIndexes(indexManager: IndexManager, plan: LogicalPlan): Seq[IndexLogEntry] = {
+  def getCandidateIndexes(
+      indexManager: IndexManager,
+      plan: LogicalPlan,
```
imback82, line 39: Currently this `plan` is always the result of `RuleUtils.getLogicalRelation` meaning this `plan` is expected to be in a fixed shape (linear, etc.). What would be the best way to address this? cc: @apoorvedave1 / @pirz

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -51,11 +55,40 @@ object RuleUtils {
...
+      val files = plan
+        .collect {
+          case LogicalRelation(
+              HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+              _,
+              _,
+              _) =>
+            location.allFiles.map(f =>
+              FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
+        }
```
imback82, line 86: Can we assert `collect` returns only one `LogicalRelation`?

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleTestSuite.scala
```diff
@@ -39,7 +40,13 @@ trait HyperspaceRuleTestSuite extends HyperspaceSuite {
...
       case Some(s) =>
         val sourcePlanProperties = SparkPlan.Properties(
-          Seq(),
+          Seq(
+            Relation(
+              Seq("dummy"),
+              Hdfs(Properties(Content(Directory("/")))),
+              "schema",
+              "format",
+              Map())),
```
imback82, line 49: Suggestion: https://github.com/microsoft/hyperspace/pull/142#discussion_r488171512

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -110,6 +111,71 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+
+    {
+      val readDf = spark.read.parquet(dataPath)
+      assert(
+        RuleUtils
+          .getCandidateIndexes(
+            indexManager,
+            readDf.queryExecution.optimizedPlan,
+            hybridScanEnabled = false)
+          .length === 0)
```
imback82, line 134: nit: prefer `isEmpty`

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -110,6 +111,71 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+
+    {
+      val readDf = spark.read.parquet(dataPath)
+      assert(
+        RuleUtils
+          .getCandidateIndexes(
+            indexManager,
+            readDf.queryExecution.optimizedPlan,
+            hybridScanEnabled = true)
+          .length === 0)
```
imback82, line 175: ditto

imback82 (commented):
few minor comments, but looking fine to me.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: It's temporary for this PR - I'll remove the default value assignment & use the config value instead in the next pr.

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleTestSuite.scala
sezruby: This is required - empty root directory throws exception during the test.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: OK. In that case, please leave a review comment on your intention. That helps reviewers.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -51,11 +55,40 @@ object RuleUtils {
...
+      assert(files.length == 1)
```
imback82, line 87: Is there a better name than `files`? maybe `filesByRelations`, `filesGroupedByRelation`, etc?

sezruby: `filesByRelations` looks good :)

sezruby:
@apoorvedave1 @pirz Could you review this PR? Thanks!

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -110,6 +111,71 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+            readDf.queryExecution.optimizedPlan,
+            hybridScanEnabled = true)
+          .length === 1)
+    }
+
+    // Delete 1 file.
+    readDf.queryExecution.optimizedPlan foreach {
+      case LogicalRelation(
+      HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _), _, _, _) =>
+        systemPath.getFileSystem(new Configuration).delete(location.allFiles.head.getPath, false)
```
imback82, line 149: indentation looks off, can you try auto-format this file?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -110,6 +111,71 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
   }
 
+  test("Verify getCandidateIndex for hybrid scan") {
+    val indexManager = IndexCollectionManager(spark)
+    val df = spark.range(1, 5).toDF("id")
+    val dataPath = systemPath.toString + "/hbtable"
+    df.write.parquet(dataPath)
+
+    val readDf = spark.read.parquet(dataPath)
+    indexManager.create(readDf, IndexConfig("index1", Seq("id")))
```
imback82, line 121: Can we use `withIndex` to clean up indexes created? (btw, I am thinking of a better way to to handle data path, so let's not worry about it for now).

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -110,6 +111,71 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+
```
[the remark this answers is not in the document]
imback82: Also, I think it's good to check both cases (hybridScanEnabled  on and off) for each scenario. So how about refactoring it as:
```scala
    val indexManager = IndexCollectionManager(spark)
    val df = spark.range(1, 5).toDF("id")
    val dataPath = systemPath.toString + "/hbtable"
    df.write.parquet(dataPath)

    val readDf = spark.read.parquet(dataPath)
    indexManager.create(readDf, IndexConfig("index1", Seq("id")))

    def verify(
        plan: LogicalPlan,
        hybridScanEnabled: Boolean,
        expectCandidateIndex: Boolean): Unit = {
      val indexes = RuleUtils
        .getCandidateIndexes(indexManager, plan, hybridScanEnabled)
      if (expectCandidateIndex) {
        assert(indexes.length == 1)
        assert(indexes.head.name == "index1")
      } else {
        assert(indexes.isEmpty)
      }
    }

    // Verify that a candidate index is returned with the unmodified data files whether
    // hybrid scan is enabled or not.
    {
      verify(optimizedPlan, hybridScanEnabled = false, expectCandidateIndex = true)
      verify(optimizedPlan, hybridScanEnabled = true, expectCandidateIndex = true)
    }

    // Scanerio #1: Append new files.
    df.write.mode("append").parquet(dataPath)

    {
      verify(optimizedPlan, hybridScanEnabled = false, expectCandidateIndex = false)
      verify(optimizedPlan, hybridScanEnabled = true, expectCandidateIndex = true)
    }

    // Scenario #2: Delete 1 file.
    readDf.queryExecution.optimizedPlan foreach {
      case LogicalRelation(
          _,
          _,
          _) =>
        systemPath.getFileSystem(new Configuration).delete(location.allFiles.head.getPath, false)
      case _ =>
    }

    {
      verify(optimizedPlan, hybridScanEnabled = false, expectCandidateIndex = false)
      verify(optimizedPlan, hybridScanEnabled = true, expectCandidateIndex = true)
    }

    // Scenario #3: Replace all files.
    df.write.mode("overwrite").parquet(dataPath)

    {
      verify(optimizedPlan, hybridScanEnabled = false, expectCandidateIndex = false)
      verify(optimizedPlan, hybridScanEnabled = true, expectCandidateIndex = false)
    }
```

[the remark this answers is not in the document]
sezruby: Thanks for the suggestion! updated :)

imback82 (approved):
LGTM, thanks @sezruby!

imback82:
Merged to master. @apoorvedave1 / @pirz Feel free to leave any comments if I missed something. Also, please take a look at https://github.com/microsoft/hyperspace/pull/153#discussion_r488208115. I think we need to follow up on it.

rapoth:
Thanks @sezruby! Can you create a follow-up item for https://github.com/microsoft/hyperspace/pull/153#discussion_r488208115 (and put in the necessary documentation) if it does not already exist?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: cc @rapoth I modified #160 a bit for this, but I think we need a new uber-issue for "arbitrary source plan" though it will be handled later.

rapoth: Good idea! Can you also open an uber issue for that when you get a chance please? (even if it is a work-in-progress issue, let's just capture all the work as we know so far)

sezruby: Ok I'll 👍

PR #155, opened 2020-09-12
[DESIGN DOC]: Hybrid Scan for File/Partition Mutable Datasets

rapoth:
Creating an initial PR we can use to finish the design doc.

This PR is a proposal for #150 

Hyperspace supports indexing immutable datasets. When the underlying data 
changes, users have to invoke `refresh` which will either fully or 
incrementally rebuild the index. While these approaches are simple and clean, 
for relatively small operations on the underlying datasets, it is cumbersome
for the user to remember to invoke index maintenance operations (failing which,
Hyperspace disables index usage).

Yes. Users can continue leveraging an index without invoking index maintenance operation, once
they set the hybrid scan configuration.

Github markdown rendering.


imback82 (approved):
LGTM as a starting doc, thanks @rapoth!

PR #162, opened 2020-09-14
Add merge functionality on Directory object

apoorvedave1:
### What changes were proposed in this pull request?
Add support for `merge`ing two `com.microsoft.hyperspace.index.Directory` objects. Pre-requisite for #29,  #163.

#### Feature Description
We need a functionality to merge two directory trees in order to support future functionalities like `Append` and `Optimize`. When we create index only on partially unindexed data, we need to make sure to log the latest snapshot of the index as the combination of previously created index files and newly created index files. This can be done through the suggested `merge` api on index `Directory`. 

#### Example Usecase: Merge contents of partially created indexes in case of append:
1. User creates index on original data. This index is stored in index directory (say) `index/v__=0`. This will be stored in metadata as
```
{
  "root" : {
    "name" : "file:/C:/",
    "files" : [ ],
    "subDirs" : [ {
      "name" : "Users",
      "files" : [ ],
      "subDirs" : [ {
        "name" : "apdave",
        "files" : [ ],
        "subDirs" : [ {
          "name" : "repo2",
          "files" : [ ],
          "subDirs" : [ {
            "name" : "testdata",
            "files" : [ ],
            "subDirs" : [ {
              "name" : "v__=0",
              "files" : [ {
                "name" : "part-00000-5782bdd3-4729-44e6-b54b-51b557f66792-c000.snappy.parquet",
                "size" : 1473,
                "modifiedTime" : 1585280921507
              }, {
                "name" : "part-00000-e8c8821f-a1b2-4b3b-be9e-687b6fa6d057-c000.snappy.parquet",
                "size" : 1878,
                "modifiedTime" : 1585280853559
              } ],
              "subDirs" : [ ]
            } ]
          } ]
        } ]
      } ]
    } ]
  }
}
```

2. User adds new data and calls `refreshIndex(mode = "quick")`. This will create another intermediate `Content` object similar to above, with `name = "v__=1"`.

3. We can now call `content.root.merge(content2.root)` to deduplicate the directory structure.
```
{
  "root" : {
    "name" : "file:/C:/",
    "files" : [ ],
    "subDirs" : [ {
      "name" : "Users",
      "files" : [ ],
      "subDirs" : [ {
        "name" : "apdave",
        "files" : [ ],
        "subDirs" : [ {
          "name" : "repo2",
          "files" : [ ],
          "subDirs" : [ {
            "name" : "testdata",
            "files" : [ ],
            "subDirs" : [ {
              "name" : "v__=0",
              "files" : [ {
[...]

Issue #136, opened 2020-08-24
Incremental Index Maintenance for File/Partition Mutable Datasets


apoorvedave1:
Please use the template below (default for Proposal Issues)

**Describe the problem**
Support for Index Maintenance for mutable datasets.

Mutable data comprises of appended data (new data added to existing data) and deleted data (files or partitions removed from existing data).

Problem: How do we keep the index up to date with updated data (either appended or deleted or both)?

**Describe your proposed solution**
We plan to expose the following apis to solve this problem:
1. `hyperspace.refreshIndex(indexName, mode = "quick")` : 
For Deleted Data: This api handles a "quick" delete version of index refresh by storing the list of deleted files in the metadata.
For Appended Data: This api creates additional index data on just the newly appended data files.

2. `hyperspace.refreshIndex(indexName, mode = "full")` : recreates full index from scratch. This doesn't need separate handling of deleted and appended data.

3. `hyperspace.optimizeIndex(indexName, mode = "quick")` : This optimization api works directly on the indexes by compacting large number of small index files into small number of larger files.

4. `hyperspace.optimizeIndex(indexName, mode = "full")` : This optimization api works directly on the indexes. Contrary to the "quick" mode optimize, this api takes all the index files, not just the smaller ones, and compacts them into larger files.

5. `hyperspace.refreshIndex(indexName, mode = "smart")` : 
For Deleted Data: This api removes index records from deleted source data files.
For Appended Data: This api creates additional index data with smart optimization of previously created small index files.

### Describe the issue
The whole process will be broken down into the following order, to allow parallel development of various functionalities

* [x] Delete Support: Add support for delete to index refresh #133, #142 @pirz
* [x] Append Support: Incremental Indexing: Support for incremental changes to index when new data is added #29 @apoorvedave1
* [x] Merge append and delete into seamless api: index refresh #105, #149, PR #187 @apoorvedave1 @pirz
* [ ] Enforce delete during read time #134 @pirz 
* [x] Quick/Full Optimize support for append-only data ISSUE: #111, PR: #166 @apoorvedave1
* [ ] Implement optimized smart refresh indexes of append-only data #112 @apoorvedave1

rapoth:
@apoorvedave1 Could you please explain why `incremental = "true"` is needed in the API? Can we do away with just the `mode` parameter? Thank you!

rapoth:
[...]

PR #163, opened 2020-09-14
Add RefreshIncrementalAction to support index creation on newly appended data

apoorvedave1:
Implement incremental indexing support for append-only data.

#### NOTE TO REVIEWERS
Uber Issue: Please follow: #136 for complete details.

This PR depends on PR #142 (Delete support on source data) and #162 (Merge support for directory objects) for major changes. Until the other PR goes in, This will be kept as WIP. It can still be viewed for general idea of how the classes will evolve.

Most files in the dependency PRs can be ignored in this PR, if the reviewer is familiar with them. The new, review-able files with new functionality are as follows:
**for Functionality: `RefreshIncremental.scala`,** 
**for Tests: `IndexManagerTests.scala`**

#### What this PR does
This feature allows hyperspace to create indexes on newly arrived data. If the user appends new data to existing, pre-indexed data, they can use `refresh` api to generate indexes only on the additional data.
This index creation will be faster than full refresh because it works only on additional data. This is different for a full refresh where the index is built from scratch on full data.

#### Algorithm Outline:
1. identify newly added data files
2. create new index version on these files
3. update metadata to reflect the latest snapshot of index. The latest snapshot points to the complete snapshot of the index, including all the old and the newly created index files. The same is true for data files as well.

A new `RefreshIncremental.scala` action class which is built based on `RefreshActionBase` class. Reviewers can start from this class to understand what data is being indexed, and how the new metadata is being generated to reflect the latest truth of the index.

To support incremental indexing on just the unindexed data

Yes, this feature introduces a support for creating (or 'updating') index on newly added data, by creating index only on the new data.
#### How to enable this feature
This feature is currently hidden behind a flag "spark.hyperspace.index.refresh.append.enabled" which defaults to **false**. It will be later on supported as the api `refreshIndex(indexName, mode="quick") along with support for other features (e.g. delete) within that api. Please follow #136 for complete details.


rapoth:
@apoorvedave1 Can you look at the suggestions I made for #142 as far as the section `Are there any user-facing changes?` and update your PR description as appropriate?

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.actions.Constants.States.{ACTIVE, REFRESHING}
+import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEvent, RefreshActionEvent}
+
+// TODO: This class depends directly on LogEntry. This should be updated such that
+//   it works with IndexLogEntry only. (for example, this class can take in
+//   derivedDataset specific logic for refreshing).
+private[actions] abstract class RefreshActionBase(
```
rapoth, line 30: Can you add a comment (not in code, just in GitHub PR) that this is refactored based on existing implementation? For instance, do you want another review?

rapoth: I think in general, it'll be good if you can mark classes that do not need additional reviews (since they are either refactored or are based on the delete PR)

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types.{DataType, StructType}
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
+import com.microsoft.hyperspace.util.ResolverUtils
+
+class RefreshIncremental(
```
rapoth, line 30: nit: Class documentation

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,94 @@
...
+    /*
+    // Option 1 to create indexable df:
+    // To remove pre-indexed files, add file name column, add file filter and remove file name
+    // column.
+    val predicate = indexedFiles.mkString("('", "','", "')")
+    val temporaryColumn = "_file_name"
+    return df.withColumn(temporaryColumn, input_file_name)
+      .where(s"$temporaryColumn not in $predicate")
+      .drop("_file_name")
+     */
```
rapoth, line 59: Are you planning on leaving this comment here? Or is it just for discussion in the PR?

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,94 @@
...
+  private lazy val indexableDf = {
+    // Currently we only support to create an index on a LogicalRelation.
+    assert(previousIndexLogEntry.relations.size == 1)
+    val relation = previousIndexLogEntry.relations.head
+
+    // TODO: improve this to take last modified time of files into account.
+    val indexedFiles = relation.data.properties.content.files.map(_.toString)
+
+    /*
+    // Option 1 to create indexable df:
```
rapoth, line 51: Btw, what's the difference you observed between Option 1 and 2? Did you do any perf benchmark?

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,110 @@
...
+      .options(relation.options)
+      .load(relation.rootPaths: _*)
+  }
+
+  private lazy val indexableDf = {
+    // Currently we only support to create an index on a LogicalRelation.
+    assert(previousIndexLogEntry.relations.size == 1)
+    val relation = previousIndexLogEntry.relations.head
+
+    // TODO: improve this to take last modified time of files into account.
```
rapoth, line 64: nit: Create an issue and add it into one of the uber issues we have floating around?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexDeleteTests.scala
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
```
rapoth, line 2: Can you add a comment that this file doesn't need review?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,67 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+        val newIndexLocation = s"$systemPath/index_$format"
+        indexCount = spark.read
+          .parquet(newIndexLocation +
+            s"/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=1")
+          .count()
+
+        // Check if index got updated
+        assert(indexCount == 3)
+
+        // Check if lastest log file is updated with newly created index files
```
rapoth, line 297: nit: 

```suggestion
        // Check if latest log file is updated with newly created index files
```

rapoth:
Doesn't this also depend on #162? If so, can you please update the PR description to say this (and also make a PR comment indicating that file doesn't need to be reviewed?)

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
apoorvedave1: this is up for discussion. I am fine with either option though

apoorvedave1: I will remove it if everyone is ok with the other implementation

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexDeleteTests.scala
apoorvedave1: yeah I added one comment in the beginning saying with NOTE TO REVIEWERS which was for this exact purpose. I am not sure how many more files I will pull from the dependent PR so I did not think of adding it on per file basis

apoorvedave1, line 2: Thanks @rapoth for the suggestion. , NOTE TO REVIEWERS: This file is picked from a dependent pr: #142 And it doesn't need review here.

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
apoorvedave1: no, both the options are compatible perf wise, neither is better than the other. It's only a question of which one looks cleaner, more readable/preferable

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
apoorvedave1: thanks @rapoth , I have updated the description showing exactly which files to review. I will also add the info saying where to review the remaining files.

sezruby:
>Algorithm Outline:
1. identify newly added data files
2. create new index version on these files
3. update metadata to reflect the latest snapshot of index. The latest snapshot points to the complete snapshot of the index, including all the old and the newly created index files. The same is true for data files as well.

I think "how to apply the refreshed index" should also be addressed in this PR. Merging the file paths of v1 and newly appended files (and loading them together) cannot retain the bucketing information & sort order. If it's not handled in this PR because of PR size, could you update PR description about this?

apoorvedave1:
> > Algorithm Outline:
> > 
> > 1. identify newly added data files
> > 2. create new index version on these files
> > 3. update metadata to reflect the latest snapshot of index. The latest snapshot points to the complete snapshot of the index, including all the old and the newly created index files. The same is true for data files as well.
> 
> I think "how to apply the refreshed index" should also be addressed in this PR. Merging the file paths of v1 and newly appended files (and loading them together) cannot retain the bucketing information & sort order. If it's not handled in this PR because of PR size, could you update PR description about this?

Thanks @sezruby , I am not sure what you mentioned is correct. As long as in the JoinRule and FilterRule I provide the bucket spec, it will work fine right? This means, I don't have to handle them at all in this PR and it will work fine.

sezruby:
IMHO at least the sort order won't work. Also is it guaranteed that FileScan node performs bucket-wise data loading from 2 different index directories? Could you check the actual bucket-wise data of DF from refreshed index?

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,94 @@
...
+    //   in the index log entry.
+    write(spark, indexableDf, indexConfig)
+  }
+
+  private lazy val indexableDf = {
+    // Currently we only support to create an index on a LogicalRelation.
+    assert(previousIndexLogEntry.relations.size == 1)
+    val relation = previousIndexLogEntry.relations.head
+
+    // TODO: improve this to take last modified time of files into account.
```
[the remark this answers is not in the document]
apoorvedave1: thanks @rapoth , I have kept this comment unresolved for now. I will close it after I fix it.

apoorvedave1:
> IMHO at least the sort order won't work. Also is it guaranteed that FileScan node performs bucket-wise data loading from 2 different index directories? Could you check the actual bucket-wise data of DF from refreshed index?

Thanks @sezruby , I have added a test case for your benefit. https://github.com/microsoft/hyperspace/pull/163/commits/3cbfea51b85613298281514781a3d164ae4f649e
Please have a look. I will clean up/refactor the test case later when dependent prs are in.

rapoth:
> > IMHO at least the sort order won't work. Also is it guaranteed that FileScan node performs bucket-wise data loading from 2 different index directories? Could you check the actual bucket-wise data of DF from refreshed index?
> 
> Thanks @sezruby , I have added a test case for your benefit. [3cbfea5](https://github.com/microsoft/hyperspace/commit/3cbfea51b85613298281514781a3d164ae4f649e)
> Please have a look. I will clean up/refactor the test case later when dependent prs are in.

Sort order won't work. But @apoorvedave1 - didn't you verify that bucket-wise data loading from 2 different index directories works? (the question that @sezruby is asking above) I remember you already verified it so just confirming.

apoorvedave1:
> e) I remember you already verified it so just confirming

Thanks @rapoth , yeah I verified it and I have added this [test case ](https://github.com/microsoft/hyperspace/commit/3cbfea51b85613298281514781a3d164ae4f649e) for Proof of Concept in case others have similar doubts.

sezruby:
@apoorvedave1 Could you share the plan for Sort Merge Join? - I wonder it has "Sort" node or not.  And the test just checks the plan, not the actual data like https://github.com/microsoft/hyperspace/pull/151/files#diff-03cdb5857f6dc02176565ef7bbf77863R122

rapoth:
I remember the Sort node is present but I will let @apoorvedave1 confirm.

apoorvedave1:
yeah sort node will be there as long as there are > 1 files for any bucket

sezruby:
I just added a path to the index relation and test it simply but it doesn't show "Sort". I think it's because we use the "bucketSpec" of the original index which including the sort order. So Spark considers the dataset as sorted.

```
 val location =
          new InMemoryFileIndex(
            spark,
            Seq(new Path(
              "file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index55/v__=0"))
              ++ index.content.files,

//
scala> hs.explain(join)
Plan with indexes:
Project [id#0, name#1, name#40]
+- SortMergeJoin [id#0], [id#39], Inner
   :- *(1) Project [id#0, name#1]
   :  +- *(1) Filter ((isnotnull(id#0) && (id#0 = 3)) && (id#0 >= 2))
   :     +- *(1) FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index55/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,3), GreaterThanOrEqual(id,2)], ReadSchema: struct<id:int,name:string>, SelectedBucketsCount: 1 out of 200
   +- *(2) Project [id#39, name#40]
      +- *(2) Filter ((isnotnull(id#39) && (id#39 >= 2)) && (id#39 = 3))
         +- *(2) FileScan parquet [id#39,name#40] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index55/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,2), EqualTo(id,3)], ReadSchema: struct<id:int,name:string>, SelectedBucketsCount: 1 out of 200
```

apoorvedave1:
> I just added a different index path to the index relation and test it simply but it doesn't show "Sort". I think it's because we use the "bucketSpec" of the original index which including the sort order. So Spark considers the dataset as sorted.
> 
> ```
>  val location =
>           new InMemoryFileIndex(
>             spark,
>             Seq(new Path(
>               "file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index55/v__=0"))
>               ++ index.content.files,
> 
> //
> scala> hs.explain(join)
> =============================================================
> Plan with indexes:
> =============================================================
> Project [id#0, name#1, name#40]
> +- SortMergeJoin [id#0], [id#39], Inner
>    :- *(1) Project [id#0, name#1]
>    :  +- *(1) Filter ((isnotnull(id#0) && (id#0 = 3)) && (id#0 >= 2))
>    :     +- *(1) FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index55/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,3), GreaterThanOrEqual(id,2)], ReadSchema: struct<id:int,name:string>, SelectedBucketsCount: 1 out of 200
>    +- *(2) Project [id#39, name#40]
>       +- *(2) Filter ((isnotnull(id#39) && (id#39 >= 2)) && (id#39 = 3))
>          +- *(2) FileScan parquet [id#39,name#40] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index55/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,2), EqualTo(id,3)], ReadSchema: struct<id:int,name:string>, SelectedBucketsCount: 1 out of 200
> ```
> 
> And I confirmed that bucket-wise data loading works with additional index paths. Let's add E2E test for this. Thanks!

Thanks @sezruby ,
I have added this [test case ](https://github.com/microsoft/hyperspace/commit/3cbfea51b85613298281514781a3d164ae4f649e) for this. Do you mean any other test? I guess I did not understand the ask here.

I may have misunderstood here, but if you mean we need a test case to check whether or not there will be a sort node, that's a known spark bucketing feature (please visit https://www.youtube.com/watch?v=6BD-Vv-ViBw) and doesn't need testing. If _any_ bucket has more than 1 file, even if individual files are sorted, this will add a Sort node. It doesn't need to go with this PR.

On the other hand if you want this test added just for general knowledge of the reviewers, we can add such a test in another PR.

Let me know if I misunderstood anything. If that's the case, please also outline the e2e test needed. Thanks

sezruby:
If so, we don't need a test for bucket-wise data loading. I just wanted to make sure that thing :) Thanks!

For Sort, you can try SortMergeJoin which is using the refreshed index.

I created 2 indexes for different dataset(but same schema/spec - (id,name),bucket num = 200) and just added another index path to the index relation.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -396,6 +398,78 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+      .toDF("Date", "RGUID", "Query", "imprs", "clicks")
+      .limit(3)
+      .write
+      .mode("append")
+      .parquet(refreshTestLocation)
+
+    // Refresh Index.
+    hyperspace.refreshIndex(indexConfig.indexName)
+
+    // Create a Join Query
```
rapoth, line 430: Is this going to be:

```suggestion
    // Create a Join Query.
```

? :)

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -439,32 +439,163 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
     FileUtils.deleteDirectory(tempDir.toFile)
   }
 
-  test("Directory.merge Test: test if merge works as expected.") {
-    // File Structure
-    // testDir/a/f1
-    // testDir/a/f2
-    // testDir/a/b/f3
-    // testDir/a/b/f4
+  test("Directory.merge: test if merge works as expected.") {
```
rapoth: Per the discussion elsewhere in another PR, I think these tests should contain what they are testing rather than saying `test if merge works as expected`. Therefore, it is better if we clarify what the author had in mind (through plain-words in the test name) and how they are testing (through the code) so reviewers can better verify the intent.

For instance:

```suggestion
  test("Merging a directory (which contains files) with another child directory (containing files) should merge successfully.") {
```

or anything else you can come up with.

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -439,32 +439,163 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
...
     assert(directoryEquals(actual1, expected))
     assert(directoryEquals(actual2, expected))
+  }
 
-    FileUtils.deleteDirectory(tempDir.toFile)
+  test("Directory.merge: test if merge works as expected when directories overlap.") {
+    // directory1:
+    // a/f1
+    // a/f2
+    // a/b/f3
```
rapoth: Thank you for all these examples in the comments! Makes it very easy to understand and cross-check if the test is testing as expected :)

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -439,32 +439,163 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
...
+    )
+
+    val actual1 = directory1.merge(directory2)
+    val actual2 = directory2.merge(directory1)
+
+    assert(directoryEquals(actual1, expected))
+    assert(directoryEquals(actual2, expected))
+  }
+
+  test("Directory.merge: test if exception is thrown when directory names are not equal.") {
```
rapoth: This is a great test name!

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -439,32 +439,163 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
...
+    // b/f4
+    val directory2 = Directory(
+      name = "b",
+      files = Seq(FileInfo("f3", 100L, 100L), FileInfo("f4", 100L, 100L))
+    )
+
+    val ex1 = intercept[HyperspaceException] (directory1.merge(directory2))
+    val ex2 = intercept[HyperspaceException] (directory2.merge(directory1))
+
+    assert(ex1.msg.contains("Merging directories with names a and b failed."))
```
rapoth: Does it make sense to parametrize `a` and `b` here instead of hard-coding? (for the case if someone changes the name of the directory and forgets to change here).

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -396,6 +398,78 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
       sortedRowsWithHyperspaceDisabled.sameElements(getSortedRows(dfAfterHyperspaceDisabled)))
   }
 
+  test("Verify Join Index Rule utilizes indexes correctly after incremental refresh.") {
```
rapoth, line 401: nit: Ok, so this was my question elsewhere in another PR. Your tests in Directory merge have `Directory.merge: ....` but here you just write `Verify ...` without the class and method name. Can we pick one way of doing this so it is consistent? I have no preference.

Please leave a comment on which one you are picking and why; we'll request others to do the same. Thank you!

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexDeleteTests.scala
rapoth: Thank you!

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
rapoth: Sorry, my bad. I reviewed something that got merged (my PR view was pointing to an outdated commit :()

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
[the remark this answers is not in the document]
rapoth: Sounds good, thank you!

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,94 @@
...
+    extends RefreshActionBase(spark, logManager, dataManager) {
+  final override def op(): Unit = {
+    // TODO: The current implementation picks the number of buckets from session config.
+    //   This should be user-configurable to allow maintain the existing bucket numbers
+    //   in the index log entry.
+    write(spark, indexableDf, indexConfig)
+  }
+
+  private lazy val indexableDf = {
+    // Currently we only support to create an index on a LogicalRelation.
```
rapoth, line 43: nit: 

```suggestion
    // Currently we only support creating an index on a LogicalRelation.
```

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,67 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
     }
   }
 
+  test("Verify refresh-incremental rebuild of index.") {
+    Seq(("csv", Map("header" -> "true")), ("parquet", Map()), ("json", Map())).foreach {
+      case (format, option: Map[String, String]) =>
+        spark.conf.set(IndexConstants.REFRESH_APPEND_ENABLED, true)
```
rapoth, line 257: naive question: didn't we decide in an earlier PR to use `withSQLConf`?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,67 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+        val refreshTestLocation = sampleParquetDataLocation + "refresh_" + format
```
rapoth, line 258: nit: for consistency:

```suggestion
        // Setup. Create sample data and index.
        val refreshTestLocation = sampleParquetDataLocation + "refresh_" + format
```

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,67 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+        val df = spark.read.format(format).options(option).load(refreshTestLocation)
+        hyperspace.createIndex(df, indexConfig)
+        var indexCount =
+          spark.read
+            .parquet(s"$systemPath/index_$format" +
+              s"/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=0")
+            .count()
+        assert(indexCount == 10)
+
+        // Change Original Data
```
rapoth, line 278: nit:

```suggestion
        // Change original data.
```

Since we now agree on what this should look like, please change it throughout.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,67 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
     }
   }
 
+  test("Verify refresh-incremental rebuild of index.") {
```
rapoth, line 254: nit:

```suggestion
  test("Verify refresh-incremental (append-only) should index only newly appended data.") {
```

Does this make sense? I wrote append-only to accommodate the merge phase between you and @pirz.

rapoth (commented):
Minor comments, thank you!

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: I am not sure if this comment applies, please correct me if I am wrong. for `Directory.merge()`, we are only testing a small part of the code, just a method, and whether it works correctly. Here in case of E2E tests, there are a lot of moving parts. It might not be good to merge these two different types of test names. It might hurt clarity in my opinion.
Please let me kknow if there is a better approach or I didn't undetstand correctly.

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * Create a logEntry with all data files, and index content merged with previous index content.
+   * @return Merged index log entry. This contains ALL data files (files which were indexed
+   *         previously, and files which are being indexed in this operation). Also contains ALL
+   *         index files (index files for previously indexed data as well as newly created files).
+   */
+  override def logEntry: LogEntry = {
+    val entry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
```
sezruby, line 72: `indexableDf` instead of `df`?

apoorvedave1: Thanks @sezruby , I think `df` is correct to get the ALL latest data files. I have added some comments in the code.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -407,6 +409,78 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+      query,
+      getIndexFilesPath(indexConfig.indexName, Seq(0, 1)) ++
+        getIndexFilesPath(indexConfig.indexName, Seq(0, 1)))
+
+    // Verify Bucketing works as expected. This is reflected in shuffle nodes being eliminated
+    // when hyperspace is enabled.
+    spark.disableHyperspace()
+    val dfWithHyperspaceDisabled = query()
+    var shuffleNodes = dfWithHyperspaceDisabled.queryExecution.executedPlan.collect {
+      case s : ShuffleExchangeExec => s
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @sezruby , removed spaces. I don't think true vs s makes any difference. Is `s` incorrect?

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,82 @@
...
+    spark.read
+      .schema(dataSchema)
+      .format(relation.fileFormat)
+      .options(relation.options)
+      .load(allFiles.diff(indexedFiles): _*)
+  }
+
+  /**
+   * Create a logEntry with all data files, and index content merged with previous index content.
+   * @return Merged index log entry. This contains ALL data files (files which were indexed
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @sezruby , fixed

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,82 @@
...
+  final override def op(): Unit = {
+    // TODO: The current implementation picks the number of buckets from session config.
+    //   This should be user-configurable to allow maintain the existing bucket numbers
+    //   in the index log entry.
+    write(spark, indexableDf, indexConfig)
+  }
+
+  private lazy val indexableDf = {
+    // Currently we only support creating an index on a LogicalRelation.
+    assert(previousIndexLogEntry.relations.size == 1)
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @sezruby , I think it will be wrong to rely on the assert in `relations`. Consider this scenario:

In future, say we start supporting multiple relations. But the current `RefreshIncremental` code depends on the fact that there is only one relation. If we remove the `assert` from here, we will forget to fix this class in the future.

If we leave `assert` in this class, it will fail in future when we start supporting multiple relations. This will make sure we don't forget  fixing this class.

Please let me know if this makes sense. Thanks.

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
```diff
@@ -0,0 +1,82 @@
...
+          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+          _,
+          _,
+          _) =>
+        location.allFiles().map(_.getPath.toString)
+    }.flatten
+
+    val dataSchema = DataType.fromJson(relation.dataSchemaJson).asInstanceOf[StructType]
+
+    // Create a df with only diff files from original list of files
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @sezruby ,

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: Thanks @rapoth ,

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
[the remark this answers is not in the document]
imback82: I am actually fine putting this assert in one place - `relations` instead of duplicating this in many places.

To me, we cannot blindly remove `assert` from `IndexLogEntry` without checking all the references. But if people think otherwise, I don't have a strong opinion on this.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -407,6 +409,78 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
       sortedRowsWithHyperspaceDisabled.sameElements(getSortedRows(dfAfterHyperspaceDisabled)))
   }
 
+  test("Verify Join Index Rule utilizes indexes correctly after incremental refresh.") {
+    // Setup. Create Data.
+    val sampleParquetDataLocation = "src/test/resources/sampleparquet"
+    spark.conf.set(IndexConstants.REFRESH_APPEND_ENABLED, true)
+    val refreshTestLocation = sampleParquetDataLocation + "refresh"
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @sezruby , added

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,68 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
     }
   }
 
+  test("Verify refresh-incremental (append-only) should index only newly appended data.") {
+    Seq(("csv", Map("header" -> "true")), ("parquet", Map()), ("json", Map())).foreach {
+      case (format, option: Map[String, String]) =>
+        spark.conf.set(IndexConstants.REFRESH_APPEND_ENABLED, true)
+        // Setup. Create sample data and index.
+        val refreshTestLocation = sampleParquetDataLocation + "refresh_" + format
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @sezruby , added

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncremental.scala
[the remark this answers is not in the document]
apoorvedave1: ok in that case I will move it. 
My only concern was it is prone to bugs and will depend on the original coders and memory to remember to fix this issue.

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
apoorvedave1: added #182

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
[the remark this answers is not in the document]
sezruby: We could remove `s: ` for `true`, but yes it's just a preference :)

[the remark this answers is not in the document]
apoorvedave1: oh i see, yea i will change this with next iteration. I will keep this comment unresolved until then. Thanks :)

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,101 @@
...
+    }.flatten
+
+    val dataSchema = DataType.fromJson(relation.dataSchemaJson).asInstanceOf[StructType]
+
+    // Create a df with only diff files from original list of files.
+    spark.read
+      .schema(dataSchema)
+      .format(relation.fileFormat)
+      .options(relation.options)
+      .load(allFiles.diff(indexedFiles): _*)
```
sezruby, line 74: I think we shouldn't refresh index if there's no appended files. Please refer:
https://github.com/microsoft/hyperspace/pull/142/files#diff-14be32100e14d26db8bc2ed7a662b1e8R66

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,61 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+
+        // Check if index got updated.
+        assert(indexCount == 3)
+
+        // Check if latest log file is updated with newly created index files
+        val indexPath = PathUtils.makeAbsolute(newIndexLocation)
+        val logManager = IndexLogManagerFactoryImpl.create(indexPath)
+        val latestLog = logManager.getLatestLog()
+        assert(latestLog.isDefined && latestLog.get.isInstanceOf[IndexLogEntry])
+        val indexLog = latestLog.get.asInstanceOf[IndexLogEntry]
```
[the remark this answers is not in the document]
apoorvedave1: thanks @sezruby , added

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,61 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+        val newIndexLocation = s"$systemPath/index"
+        indexCount = spark.read
+          .parquet(newIndexLocation +
+            s"/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=1")
+          .count()
+
+        // Check if index got updated.
+        assert(indexCount == 3)
+
+        // Check if latest log file is updated with newly created index files
```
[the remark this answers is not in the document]
apoorvedave1: thanks @sezruby , added

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,101 @@
...
+ * pre-indexed data, they can use refresh api to generate indexes only on the additional data.
+ *
+ * Algorithm Outline:
+ * - Identify newly added data files.
+ * - Create new index version on these files.
+ * - Update metadata to reflect the latest snapshot of index. This snapshot includes all the old
+ *   and the newly created index files. The source content points to the latest data files.
+ *
+ * @param spark SparkSession
+ * @param logManager Index LogManager for index being refreshed
```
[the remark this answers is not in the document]
apoorvedave1: thanks @sezruby , added

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -217,4 +217,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
     assert(partitionSchemas.length == 1)
     partitionSchemas.head.map(_.name)
   }
+
+  private val absolutePath: UserDefinedFunction = udf(
```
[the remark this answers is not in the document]
apoorvedave1: thanks @sezruby , removed it

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
apoorvedave1: thanks @sezruby , added

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -57,4 +57,7 @@ object IndexConstants {
 
   val REFRESH_DELETE_ENABLED = "spark.hyperspace.index.refresh.delete.enabled"
   val REFRESH_DELETE_ENABLED_DEFAULT = "false"
+
+  val REFRESH_APPEND_ENABLED = "spark.hyperspace.index.refresh.append.enabled"
```
imback82, line 61: https://github.com/microsoft/hyperspace/pull/171/files#r499691796

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -116,6 +116,17 @@ case class CancelActionEvent(appInfo: AppInfo, index: IndexLogEntry, message: St
 case class RefreshDeleteActionEvent(appInfo: AppInfo, index: IndexLogEntry, message: String)
     extends HyperspaceIndexCRUDEvent
 
+/**
+ * Index Refresh Event for appended source files. Emitted when refresh is called on an index
+ * with config flag set to remove index entries for appended source data files.
```
imback82, line 121: Is this a valid comment - `remove index entries`?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -407,6 +409,79 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
       sortedRowsWithHyperspaceDisabled.sameElements(getSortedRows(dfAfterHyperspaceDisabled)))
   }
 
+  test("Verify Join Index Rule utilizes indexes correctly after incremental refresh.") {
+    withTempDir { testDir =>
+      val refreshTestLocation = testDir + "/refresh"
+      withSQLConf(IndexConstants.REFRESH_APPEND_ENABLED -> "true") {
+        // Setup. Create Data.
+        val indexConfig = IndexConfig(s"index", Seq("c2"), Seq("c4"))
```
imback82, line 417:
```suggestion
        val indexConfig = IndexConfig("index", Seq("c2"), Seq("c4"))
```

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -407,6 +409,79 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+          .toDF("c1", "c2", "c3", "c4", "c5")
+          .limit(10)
+          .write
+          .parquet(refreshTestLocation)
+        val df = spark.read.load(refreshTestLocation)
+
+        // Create Index.
+        hyperspace.createIndex(df, indexConfig)
+
+        // Append to Original Data.
```
imback82, line 429: Can we respect casing in all of the comments in this file?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -407,6 +409,79 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+
+        // Verify Indexes are used, and all index files are picked.
+        verifyIndexUsage(
+          query,
+          getIndexFilesPath(indexConfig.indexName, Seq(0, 1)) ++
+            getIndexFilesPath(indexConfig.indexName, Seq(0, 1)))
+
+        // Verify Bucketing works as expected. This is reflected in shuffle nodes being eliminated
+        // when hyperspace is enabled. Also verify that sort nodes are not removed.
+        // Without Hyperspace.
```
imback82, line 458: `With Hyperspace disabled, verify blah`.
...
`With Hyperspace enabled, verify blah`..

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -407,6 +409,79 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+        assert(sortNodes.size == 2)
+
+        // With Hyperspace.
+        spark.enableHyperspace()
+        val dfWithHyperspaceEnabled = query()
+        shuffleNodes = dfWithHyperspaceEnabled.queryExecution.executedPlan.collect {
+          case s: ShuffleExchangeExec => s
+        }
+        assert(shuffleNodes.isEmpty)
+        sortNodes = dfWithHyperspaceEnabled.queryExecution.executedPlan.collect {
```
imback82, line 477: `// SortExec is expected to be present because there are multiple files per bucket.`

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -495,13 +570,12 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
   }
 
-  private def getIndexFilesPath(indexName: String, version: Int = 0): Seq[Path] = {
-    Content
-      .fromDirectory(
-        new Path(
-          systemPath,
-          s"$indexName/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=$version"))
-      .files
+  private def getIndexFilesPath(indexName: String, version: Seq[Int] = Seq(0)): Seq[Path] = {
```
imback82, line 573: `versions`

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -495,13 +570,12 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+    version.flatMap(v =>
```
imback82, line 574: nit: usually, `{` is preferred for multi lines:
```scala
    versions.flatMap { v =>
      Content
        .fromDirectory(
          new Path(systemPath, s"$indexName/${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=$v"))
        .files
    }
```

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,90 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
     }
   }
 
+  test("Verify refresh-incremental (append-only) throws exception if no new files found.") {
+    withTempDir { testDir =>
+      val refreshTestLocation = testDir + "/refresh"
```
imback82, line 256: I see many of these patterns now. Can you add the following to `HyperspaceSuite.scala`:
```scala
  protected def withTempPathAsString(f: String => Unit): Unit = {
    // The following is from SQLHelper.withTempPath with a modification to pass
    // String instead of File to "f". The reason this is copied instead of extending
    // SQLHelper is that some of the existing suites extend QueryTest and it causes
    // "inheriting conflicting members" issue.
    val path = Utils.createTempDir()
    path.delete()
    try f(path.toString) finally Utils.deleteRecursively(path)
  }
```
, and use it as
```scala
    withTempPathAsString { testPath =>
      withSQLConf(IndexConstants.REFRESH_APPEND_ENABLED -> "true") {
        // Setup. Create sample data and index.
        val indexConfig = IndexConfig(s"index", Seq("RGUID"), Seq("imprs"))
        import spark.implicits._
        SampleData.testData
          .toDF("Date", "RGUID", "Query", "imprs", "clicks")
          .limit(10)
          .write
          .parquet(testPath)
        val df = spark.read.parquet(testPath)
        hyperspace.createIndex(df, indexConfig)
        val ex = intercept[HyperspaceException] {
          hyperspace.refreshIndex(indexConfig.indexName)
        }
        assert(ex.msg.equals("Refresh aborted as no appended source data files found."))
      }
    }
```

```

File: src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala [hunk 41]
```diff
@@ -251,6 +251,90 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+      withSQLConf(IndexConstants.REFRESH_APPEND_ENABLED -> "true") {
+        // Setup. Create sample data and index.
+        FileUtils.delete(new Path(refreshTestLocation))
```
imback82, line 259: No need to do this now that we are using `withTemp*`.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,90 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+      }
+    }
+  }
+
+  test("Verify refresh-incremental (append-only) should index only newly appended data.") {
+    withTempDir { testDir =>
+      val refreshTestLocation = testDir + "/refresh"
+      withSQLConf(IndexConstants.REFRESH_APPEND_ENABLED -> "true") {
+        // Setup. Create sample data and index.
+        FileUtils.delete(new Path(refreshTestLocation))
```
imback82, line 282: ditto

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -251,6 +251,90 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+    val indexPath = PathUtils.makeAbsolute(newIndexLocation)
+    val logManager = IndexLogManagerFactoryImpl.create(indexPath)
+    val latestLog = logManager.getLatestLog()
+    assert(latestLog.isDefined && latestLog.get.isInstanceOf[IndexLogEntry])
+    val indexLog = latestLog.get.asInstanceOf[IndexLogEntry]
+    val indexFiles = indexLog.content.files
+    assert(indexFiles.nonEmpty)
+    assert(indexFiles.forall(_.getName.startsWith("part-0")))
+    assert(indexLog.state.equals("ACTIVE"))
+    // Check all files belong to v__=0.
```
imback82, line 334: Comment needs to be updated to handle multiple versions.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/HyperspaceSuite.scala
```diff
@@ -98,8 +98,20 @@ trait HyperspaceSuite extends SparkFunSuite with SparkInvolvedSuite {
    */
   protected def withTempDir(f: File => Unit): Unit = {
     val dir = Utils.createTempDir()
-    try f(dir) finally {
+    try f(dir)
+    finally {
```
apoorvedave1, line 102: this changed because of auto-format

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -66,8 +66,8 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
           var expectedSchema =
             StructType(Seq(StructField("RGUID", StringType), StructField("Date", StringType)))
           if (enableLineage) {
-            expectedSchema = expectedSchema.add(
-              StructField(IndexConstants.DATA_FILE_NAME_COLUMN, StringType))
+            expectedSchema =
+              expectedSchema.add(StructField(IndexConstants.DATA_FILE_NAME_COLUMN, StringType))
```
apoorvedave1, line 70: autoformat result

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,101 @@
...
+    val allFiles = df.queryExecution.optimizedPlan.collect {
+      case LogicalRelation(
+          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+          _,
+          _,
+          _) =>
+        location.allFiles().map(_.getPath.toString)
+    }.flatten
+
+    val dataSchema = DataType.fromJson(relation.dataSchemaJson).asInstanceOf[StructType]
```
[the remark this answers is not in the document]
apoorvedave1: thanks @imback82 , added

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,120 @@
...
+    if (appendedFiles.isEmpty) {
+      throw HyperspaceException("Refresh aborted as no appended source data files found.")
+    }
+  }
+
+  private lazy val appendedFiles = {
+    val relation = previousIndexLogEntry.relations.head
+
+    // TODO: improve this to take last modified time of files into account.
+    // https://github.com/microsoft/hyperspace/issues/182
```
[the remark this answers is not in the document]
apoorvedave1: thanks @imback82 , added

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,120 @@
...
+    val indexedFiles = relation.data.properties.content.files.map(_.toString)
+
+    val allFiles = df.queryExecution.optimizedPlan.collect {
+      case LogicalRelation(
+      HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+      _,
+      _,
+      _) =>
+        location.allFiles().map(_.getPath.toString)
```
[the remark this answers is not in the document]
apoorvedave1: thanks @imback82 , added

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,120 @@
...
+    }.flatten
+
+    allFiles.diff(indexedFiles)
+  }
+
+  private lazy val indexableDf = {
```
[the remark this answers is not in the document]
apoorvedave1: thanks @imback82 , added

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,120 @@
...
+    // Create a df with only diff files from original list of files.
+    spark.read
+      .schema(dataSchema)
+      .format(relation.fileFormat)
+      .options(relation.options)
+      .load(appendedFiles: _*)
+  }
+
+  /**
+   * Create a logEntry with all data files, and index content merged with previous index content.
```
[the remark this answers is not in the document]
apoorvedave1: thanks @imback82 , added

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,120 @@
...
+    val mergedContent = Content(
+      previousIndexLogEntry.content.root.merge(entry.content.root)
+    )
+
+    // New entry.
+    entry.copy(content = mergedContent)
+  }
+
+  override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
+    RefreshAppendActionEvent(appInfo, logEntry.asInstanceOf[IndexLogEntry], message)
```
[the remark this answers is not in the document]
apoorvedave1: thanks @imback82 , I made everything RefreshAppend for now. I will change it if required when merging append and delete apis.

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
apoorvedave1: thanks @imback82 , added

apoorvedave1: fixed teh comment

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , added

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , added

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , added

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 , this is nice!

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 , removed

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , fixed

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,120 @@
...
+
+    // Create a df with only diff files from original list of files.
+    spark.read
+      .schema(dataSchema)
+      .format(relation.fileFormat)
+      .options(relation.options)
+      .load(appendedFiles: _*)
+  }
+
+  /**
```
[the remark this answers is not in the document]
apoorvedave1: Thanks @imback82 , added https://github.com/microsoft/hyperspace/issues/183

imback82 (approved):
LGTM, thanks @apoorvedave1!

sezruby (approved):
LGTM, thanks @apoorvedave1!

PR #164, opened 2020-09-15
Fix rank algorithm for Hybrid Scan

sezruby:

This PR introduces new rank algorithm for each FilterIndex & JoinIndex to support Hybrid Scan properly. 
You can find more details about Hybrid scan in #150.

In order to support rank algorithm for Hybrid Scan efficiently, this PR introduces a new IndexLogEntryTag using #223 

```scala
  // COMMON_BYTES stores overlapping bytes of index source files and given relation.
  // This is set in getCandidateIndexes and utilized in rank functions.
  val COMMON_BYTES: IndexLogEntryTag[Long] = IndexLogEntryTag[Long]("commonBytes")
```

If hybrid scan is enabled, the outdated index might be chosen even if there are refresher ones.
Since Hybrid Scan can incur additional overhead such as on-the-fly shuffle for appended files or filtering for deleted files, it would be good to pick the indexes with less "diff" data.

For FilterIndexRule, the index with the largest common bytes will be chosen if Hybrid Scan is enabled.
For JoinIndexRule, the bucket number would be prior to common bytes as removing shuffles will outweigh the overhead from Hybrid Scan.

Yes. 
This PR changes the rank order of candidate indexes when hybrid scan is enabled.

Unit test


src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
```diff
@@ -0,0 +1,53 @@
...
+   * Otherwise, return the head of the given index list.
+   *
+   * @param candidates List of all indexes that fully cover logical plan.
+   * @param hybridScanEnabled HybridScan config.
+   * @return top-most index which is expected to maximize performance gain
+   *         according to ranking algorithm.
+   */
+  def rank(
+      candidates: Seq[IndexLogEntry],
+      hybridScanEnabled: Boolean = false): Option[IndexLogEntry] = {
```
sezruby, line 40: This default assignment will be removed later

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
```diff
@@ -0,0 +1,55 @@
...
+  test("testRank: temporal default behavior - return head") {
+    val ind1 = RankerTestHelper.index("ind1", Seq(t1c1), Seq(t1c2))
+    val ind2 = RankerTestHelper.index("ind2", Seq(t1c1), Seq(t1c2))
+    val ind3 = RankerTestHelper.index("ind3", Seq(t2c1), Seq(t2c2))
+
+    val indexes = Seq(ind1, ind2, ind3)
+    assert(FilterIndexRanker.rank(indexes, false).get.equals(ind1))
+  }
+
+  test("testRank: return the index with the largest source file list if HybridScan is enabled") {
```
rapoth, line 41: nit: Why `testRank` instead of `FilterIndexRankerTest`? Sorry for the naive question - I'm trying to understand how we should name these tests.

src/test/scala/com/microsoft/hyperspace/index/rankers/RankerTestHelper.scala
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.index.rankers
+
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference}
+import org.apache.spark.sql.types.{StructField, StructType}
+
+import com.microsoft.hyperspace.actions.Constants
+import com.microsoft.hyperspace.index.{Content, CoveringIndex, Directory, FileInfo, Hdfs, IndexLogEntry, LogicalPlanFingerprint, Relation, Signature, Source, SparkPlan}
```
rapoth, line 23: nit: Does the styling guide say anything about these kinds of long import lines? :)

src/test/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRankerTest.scala
```diff
@@ -31,57 +30,64 @@ class JoinIndexRankerTest extends SparkFunSuite {
...
+    val l_10 = RankerTestHelper.index("l1", Seq(t1c1), Seq(t1c2), 10)
+    val l_20 = RankerTestHelper.index("l2", Seq(t1c1), Seq(t1c2), 20)
+    val r_10 = RankerTestHelper.index("r1", Seq(t2c1), Seq(t2c2), 10)
+    val r_20 = RankerTestHelper.index("r2", Seq(t2c1), Seq(t2c2), 20)
 
     val indexPairs = Seq((l_10, r_10), (l_10, r_20), (l_20, r_20))
 
     val expectedOrder = Seq((l_20, r_20), (l_10, r_10), (l_10, r_20))
-    val actualOrder = JoinIndexRanker.rank(indexPairs)
+    val actualOrder = JoinIndexRanker.rank(indexPairs, false)
```
rapoth, line 53: nit: for readability

```suggestion
    val actualOrder = JoinIndexRanker.rank(indexPairs, hybridScanEnabled = false)
```

src/test/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRankerTest.scala
```diff
@@ -31,57 +30,64 @@ class JoinIndexRankerTest extends SparkFunSuite {
...
+    {
+      // test original algorithm
+      val indexPairs = Seq((l_10, r_10), (l_10, r_20), (l_20, r_20))
+      val expectedOrder = Seq((l_20, r_20), (l_10, r_10), (l_10, r_20))
+      val actualOrder = JoinIndexRanker.rank(indexPairs, hybridScanEnabled = false)
+      assert(actualOrder.equals(expectedOrder))
+    }
+
+    {
+      val indexPairs = Seq((l_10, r_10), (l_10, r_20), (l_20, r_20))
```
rapoth, line 74: nit: for consistency, please add a comment (similar to `test original algorithm` to this code block as well.

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
```diff
@@ -0,0 +1,55 @@
...
+import com.microsoft.hyperspace.index.FileInfo
+
+class FilterIndexRankerTest extends SparkFunSuite {
+
+  val t1c1 = AttributeReference("t1c1", IntegerType)()
+  val t1c2 = AttributeReference("t1c2", StringType)()
+  val t2c1 = AttributeReference("t2c1", IntegerType)()
+  val t2c2 = AttributeReference("t2c2", StringType)()
+
+  test("testRank: temporal default behavior - return head") {
```
rapoth, line 32: nit: What do you mean by `temporal default behavior`?

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
sezruby: I just used the previous one in JoinRankerTest. Yes we need some guideline for the naming 🤔

src/test/scala/com/microsoft/hyperspace/index/rankers/RankerTestHelper.scala
sezruby: import statement is the exception. :)

rapoth: Got it. I was hoping there's some additional guidelines but fine 🙂

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
rapoth: @imback82 Any recommendations?

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
sezruby: The default behavior of FilterIndexRanker is just temporarily returning the head of the list.

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
imback82: Yea, `testRank` seems weird. I would do something like `test("rank() should return the index with the largest number of source files if HybridScan is enabled)`. (Note that `FilterIndexRanker` is omitted since it's already under `FilterIndexRankerTest`).

rapoth: Oh... good you wrote a great name 🙂

I was also hoping: can we agree on a naming convention for tests now that the code base complexity is increasing? For instance:

  1. Tests do not need to contain the test method name
  2. Name of the test should clearly state the hypothesis and what call out what is being tested
  3. ...

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
rapoth: Got it! I'll make another comment. Thanks!

rapoth: Does this make sense?

```suggestion
  test("rank() should return the head of the list by default") {
```

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -27,22 +27,37 @@ object JoinIndexRanker {
...
       case ((left1, left2), (right1, right2)) =>
-        if (left1.numBuckets == left2.numBuckets && right1.numBuckets == right2.numBuckets) {
+        lazy val numSourceFilesLeft =
+          left1.allSourceFileInfos.size + left2.allSourceFileInfos.size
+        lazy val numSourceFilesRight =
+          right1.allSourceFileInfos.size + right2.allSourceFileInfos.size
+
+        if (hybridScanEnabled && numSourceFilesLeft != numSourceFilesRight) {
+          // For append-only dataset, the number of source files can indicate the amount of data
+          // that an index covers the input relation.
```
imback82, line 58: Is it b/c of perf that we are not actually calculating the sum of actual file sizes to cover delete scenario? Should we just check that this is actually an append scenario?

src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
```diff
@@ -0,0 +1,53 @@
...
+ * Ranker class for Filter rule indexes
+ */
+object FilterIndexRanker {
+
+  /**
+   * Pick the best index among all available index options according to their cost.
+   *
+   * If hybridScanEnabled is true, pick the index which covers the most source data
+   * so that we could minimize the amount of data for on-the-fly shuffle or merge.
+   * Otherwise, return the head of the given index list.
```
imback82, line 31: We are not really doing shuffle for filter index, right? You may need to document a different reasoning for getting the max number of source files.

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
sezruby: Yes it's because of the perf. And for delete, we could pick the index based on the creation time if not considering the follow corner cases:
1) create index A
2) delete most of the data
3) create a new index B
4) "undelete" of 2) which preserves the metadata of each files

In this case, we need to pick A rather than B. So I am thinking of introducing some "context" for each candidate index to reuse the diff file list calculated in getCandidateIndex (and verify the diff is append-only or not).
BTW why does join use sort instead of max/min?

src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
sezruby: Yes, for filter we don't shuffle the data (though the current version do shuffle in case non-parquet format). I'll try to rewrite the comment.

src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
```diff
@@ -0,0 +1,53 @@
...
+   *
+   * @param candidates List of all indexes that fully cover logical plan.
+   * @param hybridScanEnabled HybridScan config.
+   * @return Top-most index which is expected to maximize performance gain
+   *         according to ranking algorithm.
+   */
+  def rank(
+      candidates: Seq[IndexLogEntry],
+      hybridScanEnabled: Boolean): Option[IndexLogEntry] = {
```
pirz, line 40: should not it use `HyperspaceConf` you recently added, instead of a function argument?

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -19,30 +19,45 @@ package com.microsoft.hyperspace.index.rankers
...
+   * @param indexPairs Index pairs for left and right side of the join. All index pairs are
    *                   compatible with each other.
-   * @return rearranged index pairs according to their ranking. The first is the best.
+   * @param hybridScanEnabled HybridScan config.
+   * @return Rearranged index pairs according to their ranking. The first is the best.
    */
   def rank(
-      indexPairs: Seq[(IndexLogEntry, IndexLogEntry)]): Seq[(IndexLogEntry, IndexLogEntry)] = {
+      indexPairs: Seq[(IndexLogEntry, IndexLogEntry)],
+      hybridScanEnabled: Boolean): Seq[(IndexLogEntry, IndexLogEntry)] = {
```
pirz, line 47: ditto: drop and use `HyperspaceConf`?

sezruby:
Changed as draft since further investigation is required for this issue; we cannot guarantee that not using hybrid scan is always better than full indexes, especially for join index.

For example, it's might be more efficient to avoid full-shuffling of index data. So **an index pair with bucketing and hybrid scan** might be more efficient than **a full index pair without bucketing (no hybrid scan)**.

src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
sezruby: In that case, we need to pass "spark: SparkSession", but this config is only use of it. 
So I think it would be better to keep this version until `spark` is needed for other parts in this function later.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -449,8 +450,29 @@ case class IndexLogEntry(
   override def hashCode(): Int = {
     config.hashCode + signature.hashCode + numBuckets.hashCode + content.hashCode
   }
+
+  /**
+   * A mutable map for holding auxiliary information of this index log entry while applying rules.
+   */
+  @JsonIgnore
+  private val tags: mutable.Map[IndexLogEntryTag[_], Any] = mutable.Map.empty
```
sezruby, line 458: tag implementation from Spark 3.0 -  https://github.com/apache/spark/blob/eb9966b70055a67dd02451c78ec205d913a38a42/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala#L93

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -79,8 +79,21 @@ object RuleUtils {
...
-      val commonCnt = inputSourceFiles.count(entry.allSourceFileInfos.contains)
+      val (commonCnt, commonBytes) = inputSourceFiles.foldLeft(0L, 0L) { (res, f) =>
+        if (entry.allSourceFileInfos.contains(f)) {
+          (res._1 + 1, res._2 + f.size) // count, bytes
+        } else {
+          res
+        }
+      }
       val deletedCnt = entry.allSourceFileInfos.size - commonCnt
+      entry.setTagValue(IndexConstants.INDEX_COMMON_BYTES_TAG, commonBytes)
```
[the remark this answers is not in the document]
sezruby: I created #222 and #223 for this. I'll rebase this change after merging #223.
Please review them first. Thanks!

sezruby:
@pirz @apoorvedave1 Could you review this change and make any comment? Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -21,4 +21,8 @@ object IndexLogEntryTags {
   // This is set in getCandidateIndexes and utilized in transformPlanToUseIndex.
   val HYBRIDSCAN_REQUIRED: IndexLogEntryTag[Boolean] =
     IndexLogEntryTag[Boolean]("hybridScanRequired")
+
+  // COMMON_BYTES stores overlapping bytes of index source files and given relation.
+  // This is set in getCandidateIndexes and utilized in rank functions.
```
apoorvedave1, line 26: could you remove this line and just add a comment here for this line? if we later use it in more places, this comment will break.

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -16,40 +16,69 @@
...
   def rank(
-      indexPairs: Seq[(IndexLogEntry, IndexLogEntry)]): Seq[(IndexLogEntry, IndexLogEntry)] = {
-
+      leftPlan: LogicalPlan,
+      rightPlan: LogicalPlan,
+      indexPairs: Seq[(IndexLogEntry, IndexLogEntry)],
+      hybridScanEnabled: Boolean): Seq[(IndexLogEntry, IndexLogEntry)] = {
     indexPairs.sortWith {
       case ((left1, left2), (right1, right2)) =>
+        // These common bytes was calculated and tagged in getCandidateIndexes.
```
apoorvedave1, line 55:
```suggestion
        // These common bytes were calculated and tagged in getCandidateIndexes.
```

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -16,40 +16,69 @@
...
+        // The value is the summation of common source files of the given plan and each index.
+        lazy val leftCommonBytes = left1
+          .getTagValue(leftPlan, IndexLogEntryTags.COMMON_BYTES)
+          .get + left2.getTagValue(leftPlan, IndexLogEntryTags.COMMON_BYTES).get
```
apoorvedave1, line 59: should we add asserts / requires to make sure the tag exists?

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -16,40 +16,69 @@
...
+            left1.numBuckets > right1.numBuckets
+          } else {
+            // If both index pairs have the same number of buckets and Hybrid Scan is enabled,
+            // pick the pair with more common bytes with the given source plan, so as to
+            // reduce the overhead from handling appended and deleted files.
+            leftCommonBytes > rightCommonBytes
+          }
         } else if (left1.numBuckets == left2.numBuckets) {
+          // If one index pair has the same number of buckets, pick the one to avoid shuffling.
+          // This condition is prior to "more common bytes" condition for Hybrid Scan.
```
apoorvedave1, line 75: this comment don't add much info. Could we remove them? it's same as the code.

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -16,40 +16,69 @@
...
           true
         } else if (right1.numBuckets == right2.numBuckets) {
           false
         } else {
-          true
+          // Pick the pair with "more common bytes" if both pairs have different number of buckets.
+          !hybridScanEnabled || leftCommonBytes > rightCommonBytes
```
apoorvedave1, line 81: Could you please explain why hybridsScanEnabled matters here? From what I understand:

if hybridScanEnabled == true, choose right1, right2.
if hybridScanEnabled == false, choose the one with bigger common bytes.

apoorvedave1, line 81: Also in this scenario, one side is being shuffled right?
Consider this scenario:
we choose left1, left2 because left1.commonBytes + left2.commonBytes > right1.commonBytes + right2.commonBytes

It is possible that left1.commonBytes = 10, left2.commonBytes = 1000. right1.commonBytes = 500, right2.commonBytes = 500

But we choose left1 and left2. Now it is possible that left2 gets reshuffled because #buckets is different. But if we chose right1 and right2, we would get less shuffling right?

src/test/scala/com/microsoft/hyperspace/index/rankers/RankerTestHelper.scala
```diff
@@ -0,0 +1,72 @@
...
+    }
+    // This tag is originally set in getCandidateIndex, for Hybrid Scan.
+    // Set tag for testing rank algorithms.
+    entry.setTagValue(plan, IndexLogEntryTags.COMMON_BYTES, commonBytes)
+    entry.state = Constants.States.ACTIVE
+    entry
+  }
+
+  private def schemaFromAttributes(attributes: Attribute*): StructType =
+    StructType(attributes.map(a => StructField(a.name, a.dataType, a.nullable, a.metadata)))
```
apoorvedave1, line 71: HyperspaceRulesTestSuite contains this function. Can dedup

src/test/scala/com/microsoft/hyperspace/index/rankers/RankerTestHelper.scala
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference}
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.types.{StructField, StructType}
+
+import com.microsoft.hyperspace.actions.Constants
+import com.microsoft.hyperspace.index.{Content, CoveringIndex, Directory, FileInfo, Hdfs, IndexLogEntry, IndexLogEntryTags, LogicalPlanFingerprint, Relation, Signature, Source, SparkPlan}
+import com.microsoft.hyperspace.index.Hdfs.Properties
+
+object RankerTestHelper {
+  def createIndex(
```
apoorvedave1, line 28: can you directly use HyperspaceRulesTestSuite to get the index and then update the tags within the test?

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
sezruby: Oh, I thought both side will be shuffled with Spark's config value.
In that case, right1 + right2 should be prior. Then we could do:
`min(left1.commonBytes, left2.commonBytes) > min(rigtht1.commonBytes, right2.commonBytes)`, right?

src/test/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRankerTest.scala
```diff
@@ -16,73 +16,102 @@
...
+  test("rank() should prefer the largest common bytes if HybridScan is enabled") {
+    val fileList1 = Seq(FileInfo("a", 1, 1, 1), FileInfo("b", 2, 1, 2))
+    val fileList2 = Seq(FileInfo("c", 1, 1, 3), FileInfo("d", 1, 1, 4))
+    val l_10 = RankerTestHelper.createIndex("l1", Seq(t1c1), Seq(t1c2), 10, fileList1, dummy)
+    val l_20 = RankerTestHelper.createIndex("l2", Seq(t1c1), Seq(t1c2), 20, fileList2, dummy)
+    val r_10 = RankerTestHelper.createIndex("r1", Seq(t2c1), Seq(t2c2), 10, fileList1, dummy)
+    val r_20 = RankerTestHelper.createIndex("r2", Seq(t2c1), Seq(t2c2), 20, fileList2, dummy)
+
+    {
+      // Test original algorithm.
```
apoorvedave1, line 86: original meaning?

src/test/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRankerTest.scala
```diff
@@ -16,73 +16,102 @@
...
+      RankerTestHelper.createIndex("l2", Seq(t1c1), Seq(t1c2), numBuckets = 20, plan = dummy)
+    val r_20 =
+      RankerTestHelper.createIndex("r1", Seq(t2c1), Seq(t2c2), numBuckets = 20, plan = dummy)
 
     val indexPairs = Seq((l_10, r_20), (l_20, r_20))
 
     val expectedOrder = Seq((l_20, r_20), (l_10, r_20))
-    val actualOrder = JoinIndexRanker.rank(indexPairs)
+    val actualOrder =
+      JoinIndexRanker.rank(dummy, dummy, indexPairs, hybridScanEnabled = false)
```
apoorvedave1, line 59:
```suggestion
    val actualOrder = JoinIndexRanker.rank(dummy, dummy, indexPairs, hybridScanEnabled = false)
```

apoorvedave1: same for other places

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
apoorvedave1: actually it's slightly more different.

consider `t1 Join t2 where ...`
 - if t1.buckets == t2.buckets => no shuffling
 - if t1.buckets != t2.buckets != spark.sql.shuffle.partitions => both sides get shuffled
 - if t1.buckets != t2.buckets but t1.buckets == spark.sql.shuffle.partitions => only t2 gets shuffled
 - if t1.buckets != t2.buckets but t2.buckets == spark.sql.shuffle.partitions => only t1 gets shuffled

apoorvedave1: oh got it thanks!

sezruby:
@priz Could you review this change again? Thanks!

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -16,40 +16,80 @@
...
-   * no resource constraint.
+   * If hybridScanEnabled is false, pick the one with least amount of shuffling and most number
+   * of buckets. When two indices have same number of buckets, there is zero shuffling.
+   * If number of buckets differ, one of the indices gets reshuffled into the number of buckets
+   * equal to the other.
+   * Secondly, more the number of buckets, better the parallelism achieved during join in general
+   * assuming there is no resource constraint.
+   *
+   * If hybridScanEnabled is true, rank algorithm follows the algorithm above, but we prioritize
+   * the index with larger and usable index data for each index, so as to minimize the amount of
```
pirz, line 42: Is this sentence written correctly? "we prioritize the index with larger and usable index data __for each index__"
Should not the bold part be "for each relation"?

pirz, line 42: Is this sentence written correctly?
"we prioritize the index with larger and usable index data for __each index__"
Should not the bold part be "for each Join child"?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -79,8 +79,21 @@ object RuleUtils {
       //  See https://github.com/microsoft/hyperspace/issues/158
 
       // Find the number of common files between the source relations & index source files.
-      val commonCnt = inputSourceFiles.count(entry.sourceFileInfoSet.contains)
+      val (commonCnt, commonBytes) = inputSourceFiles.foldLeft(0L, 0L) { (res, f) =>
```
pirz, line 82: nit: Update the comment as the code now computes both common file count and bytes.

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
```diff
@@ -0,0 +1,90 @@
...
+    val ind2 = createIndexLogEntry("ind2", Seq(t1c1), Seq(t1c2), dummy, writeLog = false)
+    val ind3 = createIndexLogEntry("ind3", Seq(t2c1), Seq(t2c2), dummy, writeLog = false)
+
+    val indexes = Seq(ind1, ind2, ind3)
+    assert(FilterIndexRanker.rank(dummy, indexes, hybridScanEnabled = false).get.equals(ind1))
+  }
+
+  test(
+    "rank() should return the index with the largest common bytes of source files" +
+      "if HybridScan is enabled") {
```
pirz, line 59: nit: same as above missing dot.

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
```diff
@@ -0,0 +1,90 @@
...
+    val df = spark.read.text(systemPath.toString)
+    dummy = df.queryExecution.optimizedPlan
+  }
+
+  val t1c1 = AttributeReference("t1c1", IntegerType)()
+  val t1c2 = AttributeReference("t1c2", StringType)()
+  val t2c1 = AttributeReference("t2c1", IntegerType)()
+  val t2c2 = AttributeReference("t2c2", StringType)()
+
+  test("rank() should return the head of the list by default") {
```
pirz, line 48: nit: missing "."(dot) at the end of test name.

src/test/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRankerTest.scala
```diff
@@ -16,73 +16,104 @@
...
+import org.apache.spark.sql.types.{IntegerType, StringType}
 
-import com.microsoft.hyperspace.actions.Constants
-import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.index.{FileInfo, IndexConstants}
+import com.microsoft.hyperspace.index.rules.HyperspaceRuleTestSuite
+import com.microsoft.hyperspace.util.FileUtils
 
-class JoinIndexRankerTest extends SparkFunSuite {
+class JoinIndexRankerTest extends HyperspaceRuleTestSuite {
```
pirz, line 29: nit: Can you fix test names and add the missing dot to them?

pirz (commented):
Overall LGTM< just a few minor comments.

apoorvedave1:
LGTM, thanks @sezruby

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -277,7 +285,9 @@ object RuleUtils {
             _) =>
         val (filesDeleted, filesAppended) =
           if (!HyperspaceConf.hybridScanEnabled(spark) && index.hasSourceUpdate) {
-            // If the index contains the source update info, we need to handle
+            // If the index contains the source update info, it means the index was validated
+            // with the latest signature including appended files and deleted files, but
+            // index data is not updated with those files. Therefore, we need to handle
```
sezruby, line 290: Revised comment for quick refresh PR

sezruby:
@imback82  Could you review this change when you have the time? Thanks!

imback82:
Sorry for the delay. I will try to get to this this week. Btw, does this change any of the index selections for our TPCH/TPCDS benchmarks?

sezruby:
> Sorry for the delay. I will try to get to this this week. Btw, does this change any of the index selections for our TPCH/TPCDS benchmarks?

This change only affects Hybrid Scan case and currently all of the common bytes will be the same as we don't refresh any of indexes after modifying the dataset.

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleTestSuite.scala
```diff
@@ -62,15 +65,22 @@ trait HyperspaceRuleTestSuite extends HyperspaceSuite {
...
           Content.fromLeafFiles(indexFiles, new FileIdTracker).get,
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
         val logManager = new IndexLogManagerImpl(getIndexRootPath(name))
+        val commonBytes = inputFiles.foldLeft(0L) { (bytes, f) =>
+          bytes + f.size
+        }
+        // Set tag for testing rank algorithms.
+        indexLogEntry.setTagValue(plan, IndexLogEntryTags.COMMON_BYTES, commonBytes)
```
imback82, line 79: Can you set the tags inside test that require it?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -18,7 +18,9 @@ package com.microsoft.hyperspace.index
 
 object IndexLogEntryTags {
   // HYBRIDSCAN_REQUIRED indicates if Hybrid Scan is required for this index or not.
-  // This is set in getCandidateIndexes and utilized in transformPlanToUseIndex.
   val HYBRIDSCAN_REQUIRED: IndexLogEntryTag[Boolean] =
     IndexLogEntryTag[Boolean]("hybridScanRequired")
+
+  // COMMON_BYTES stores overlapping bytes of index source files and given relation.
+  val COMMON_BYTES: IndexLogEntryTag[Long] = IndexLogEntryTag[Long]("commonSourceSizeInBytes")
```
imback82, line 25: nit: Should this be `COMMON_SOURCE_SIZE_IN_BYTES`?

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -16,40 +16,78 @@
...
           true
-        } else if (right1.numBuckets == right2.numBuckets) {
+        } else if (left2.numBuckets == right2.numBuckets) {
           false
-        } else {
+        } else if (!hybridScanEnabled) {
           true
+        } else {
+          // Pick the pair with "more common bytes" if both pairs have different number of buckets.
+          leftCommonBytes > rightCommonBytes
```
imback82, line 90: How about:
```scala
} else {
  // At this point, both pairs have different number of buckets. If Hybrid Scan is enabled, pick the pair
  // with "more common bytes", otherwise pick the first pair.
  !hybridScanEnabled || leftCommonBytes > rightCommonBytes
}
```

src/test/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRankerTest.scala
```diff
@@ -16,73 +16,117 @@
...
+    leftPlan = spark.read.text(dummyDataPath1.toString).queryExecution.optimizedPlan
+    rightPlan = spark.read.text(dummyDataPath2.toString).queryExecution.optimizedPlan
+  }
+
+  before {
+    spark.conf.set(IndexConstants.INDEX_HYBRID_SCAN_ENABLED, "false")
+  }
+
+  after {
+    spark.conf.set(IndexConstants.INDEX_HYBRID_SCAN_ENABLED, "false")
```
imback82, line 51: Instead of this, can we use `withSQLConf` in tests that set this config?

imback82 (commented):

src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -16,40 +16,78 @@
...
-          left1.numBuckets > right1.numBuckets
-        } else if (left1.numBuckets == left2.numBuckets) {
+      case ((left1, right1), (left2, right2)) =>
+        // These common bytes were calculated and tagged in getCandidateIndexes.
+        // The value is the summation of common source files of the given plan and each index.
+        lazy val leftCommonBytes = left1
+          .getTagValue(leftChild, IndexLogEntryTags.COMMON_BYTES)
+          .getOrElse(0L) + right1
+          .getTagValue(rightChild, IndexLogEntryTags.COMMON_BYTES)
+          .getOrElse(0L)
```
imback82, line 66: can we create a local def to simplify this? `lazy val leftCommonBytes = getCommonBytes(left1, leftChild) + getCommonBytes(right1, rightChild)`, for example

imback82 (approved):

Issue #145, opened 2020-09-03
Improvement of HybridScan for FilterIndexRule


sezruby:
### Describe the issue

In #123, FilterIndexRule with HybridScan provides 2 methods depends on the file format of source data relation
  - For Parquet, we could just add the list of appended file paths to the FileScan node for index data.
```
Project => Filter => Relation (index data path + appended source file paths)
```
  - Other than Parquet, we cannot simply read the data along with index data files in Parquet format with 1 FileScan node. In this case, we need to read & merge the appended data in a different way. In #123, `BucketUnion` is used to support this, but it involves `Shuffle` of appended data and requires `BucketSpec` to read index data. Though we can retain `BucketSpec` after `BucketUnion` and exploit it for subsequent operations, this can be covered by Join Rule v2. So we should find another way to support this only for FilterIndexRule + non parquet format of source data. One possible approach is to use `Union` operator:

For instance, original plan: 
```
Project => Filter => Relation
```
BucketUnion:
```
BucketUnion => Project => Filter => Relation (index)
            |=> Shuffle => Project => Filter => Relation (appended data)
```
Union:
```
Union => Project => Filter => Relation (index)
      |=> Project => Filter => Relation (appended data)
```

sezruby:
Fixed by #165

PR #165, opened 2020-09-15
Modify logical plan to merge newly appended files and index data

sezruby:

This PR allows users to use the hybrid scan for append-only dataset. 

In order to support Hybrid Scan for append-only dataset, we need to merge the newly appended files and index data properly. Currently we have the following cases:

- **_Case 1_)** Filter Index Rule & parquet source format
  - In this case, we can just add the file list of appended files to the file list of the index relation (i.e. index data). It's because: 
    - it's guaranteed that newly appended source files always have the all columns in the index data (except for lineage).
    - Filter Index Rule does not utilize bucketing information for now; able to read both index data & newly appended data with 1 FileScan node. See below `InMemoryFileIndex`
```
...
<----:- *(1) Project [name#1, id#0]---->
<----:  +- *(1) Filter (isnotnull(id#0) && (id#0 >= 1))---->
<----:     +- *(1) FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, 
                     Location: InMemoryFileIndex[<list of index data files> , <list of newly appended files>], PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,1)], ReadSchema: struct<id:int,name:string>---->
...
```
- **_Case 2_)** Filter Index Rule  & non-parquet source format
  - In this case, We could use Union instead of BucketUnion. Please refer #145.  
- **_Case 3_)** Join Index Rule
  - We can utilize BucketUnion (#151) to merge index data and appended data. In this way, we can retain the bucketing information of index, which enables to avoid unnecessary shuffling of index data.
```
 +- BucketUnion 200 buckets, bucket columns: [l_orderkey]      <===== merge both plans
            :- Project [l_orderkey#21L]                        <===== original index plan
            :  +- Filter ((isnotnull(l_commitdate#32) && isnotnull(l_receiptdate#33)) && (l_commitdate#32 < l_receiptdate#33))
            :     +- Relation[l_orderkey#21L,l_partkey#22L,l_suppkey#23L,l_quantity#25,l_extendedprice#26,l_discount#27,l_returnflag#29,l_shipdate#31,l_commitdate#32,l_receiptdate#33,l_shipmode#35] parquet
            +- RepartitionByExpression [l_orderkey#21L], 200   <===== on-the-fly shuffle with index spec
               +- Project [l_orderkey#21L]                     <===== newly appended data
                  +- Filter ((isnotnull(l_commitdate#32) && isnotnull(l_receiptdate#33)) && (l_commitdate#32 < l_receiptdate#33))
                     +- Relation[l_orderkey#21L,l_partkey#22L,l_suppkey#23L,l_quantity#25,l_extendedprice#26,l_discount#27,l_returnflag#29,l_shipdate#31,l_commitdate#32,l_receiptdate#33,l_shipmode#35] parquet
```

- `transformPlanToUseIndex` returns the transformed plan to utilize the given index.
- `transformPlanToUsePureIndex`:  if `hybridScanEnbaled=false`, it replaces the source location with the index data location same as before.
- `transformPlanToUseHybridIndexDataScan`: if `hybridScanEnabled=true`, it firstly creates the plan with the index location similar to  `transformPlanToUsePureIndex`
  - and if there is appended data,
    - if _Case1_, add appended file lists to the location along with index data files, and then return.
    - if _Case3_, create on-the-fly shuffle for the appended data and do BucketUnion.

To support Hybrid Scan for append-only dataset. (#150)

Yes, if a user turns on Hybrid Scan (`spark.hyperspace.index.hybridscan.enabled=true`), outdated indexes whose dataset got appended new files could be a candidate for both FilterIndexRule & JoinIndexRule.
The query plan is modified in optimizer to support this and the changed plan can be checked with `hs.explain()` API.

FilterRule - _Case1_
```
scala> hs.explain(query)
Plan with indexes:
Project [id#0, name#1]
+- Filter (isnotnull(id#0) && (id#0 = 1))
   <----+- FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index33/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1)], ReadSchema: struct<id:int,name:string>---->

Plan without indexes:
Project [id#0, name#1]
+- Filter (isnotnull(id#0) && (id#0 = 1))
   <----+- FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table], PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1)], ReadSchema: struct<id:int,name:string>---->

Indexes used:
index33:file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index33/v__=0
```
FilterRule - _Case2_
```
Plan with indexes:
<----Union---->
<----:- *(1) Project [id#75L, name#76]---->
<----:  +- *(1) Filter (isnotnull(id#75L) && (id#75L = 1))---->
<----:     +- *(1) FileScan parquet [id#75L,name#76] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/indexjj/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1)], ReadSchema: struct<id:bigint,name:string>, SelectedBucketsCount: 1 out of 200---->
<----+- *(2) Project [id#75L, name#76]---->
   <----+- *(2) Filter (isnotnull(id#75L) && (id#75L = 1))---->
      <----+- *(2) FileScan json [id#75L,name#76] Batched: false, Format: JSON, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/tablej/part-00000-b6853714-dc1b-450b..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1)], ReadSchema: struct<id:bigint,name:string>---->
```

JoinRule - BroadcastHashJoin
```
Plan with indexes:
Project [id#0, name#1, name#40]
+- BroadcastHashJoin [id#0], [id#39], Inner, BuildRight
   <----:- BucketUnion 200 buckets, bucket columns: [id]---->
   <----:  :- *(1) Project [id#0, name#1]---->
   <----:  :  +- *(1) Filter ((isnotnull(id#0) && (id#0 = 1)) && (id#0 >= 1))---->
   <----:  :     +- *(1) FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index33/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1), GreaterThanOrEqual(id,1)], ReadSchema: struct<id:int,name:string>, SelectedBucketsCount: 1 out of 200---->
   <----:  +- Exchange hashpartitioning(id#0, 200)---->
   <----:     +- *(2) Project [id#0, name#1]---->
   <----:        +- *(2) Filter ((isnotnull(id#0) && (id#0 = 1)) && (id#0 >= 1))---->
   <----:           +- *(2) FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table/part-00003-0fc50086-fd6e-4527-..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1), GreaterThanOrEqual(id,1)], ReadSchema: struct<id:int,name:string>---->
   +- BroadcastExchange HashedRelationBroadcastMode(List(cast(input[0, int, true] as bigint)))
      <----+- BucketUnion 200 buckets, bucket columns: [id]---->
         <----:- *(3) Project [id#39, name#40]---->
         <----:  +- *(3) Filter ((isnotnull(id#39) && (id#39 >= 1)) && (id#39 = 1))---->
         <----:     +- *(3) FileScan parquet [id#39,name#40] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index33/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,1), EqualTo(id,1)], ReadSchema: struct<id:int,name:string>, SelectedBucketsCount: 1 out of 200---->
         <----+- Exchange hashpartitioning(id#39, 200)---->
            <----+- *(4) Project [id#39, name#40]---->
               <----+- *(4) Filter ((isnotnull(id#39) && (id#39 >= 1)) && (id#39 = 1))---->
                  <----+- *(4) FileScan parquet [id#39,name#40] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table/part-00003-0fc50086-fd6e-4527-..., PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,1), EqualTo(id,1)], ReadSchema: struct<id:int,name:string>---->

Plan without indexes:
Project [id#0, name#1, name#40]
+- BroadcastHashJoin [id#0], [id#39], Inner, BuildRight
   <----:- Project [id#0, name#1]---->
   <----:  +- Filter ((isnotnull(id#0) && (id#0 = 1)) && (id#0 >= 1))---->
   <----:     +- FileScan parquet [id#0,name#1] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table], PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1), GreaterThanOrEqual(id,1)], ReadSchema: struct<id:int,name:string>---->
   +- BroadcastExchange HashedRelationBroadcastMode(List(cast(input[0, int, true] as bigint)))
      <----+- *(1) Project [id#39, name#40]---->
         <----+- *(1) Filter ((isnotnull(id#39) && (id#39 >= 1)) && (id#39 = 1))---->
            <----+- *(1) FileScan parquet [id#39,name#40] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table], PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,1), EqualTo(id,1)], ReadSchema: struct<id:int,name:string>---->

Indexes used:
index33:file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index33/v__=0
```
Join Rule - Sort Merge Join
```
Plan with indexes:
Project [id#69, name#70, name#103]
+- SortMergeJoin [id#69], [id#102], Inner
   :- *(3) Sort [id#69 ASC NULLS FIRST], false, 0
   <----:  +- BucketUnion 200 buckets, bucket columns: [id]---->
   <----:     :- *(1) Project [id#69, name#70]---->
   <----:     :  +- *(1) Filter ((isnotnull(id#69) && (id#69 = 1)) && (id#69 >= 1))---->
   <----:     :     +- *(1) FileScan parquet [id#69,name#70] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index33/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1), GreaterThanOrEqual(id,1)], ReadSchema: struct<id:int,name:string>, SelectedBucketsCount: 1 out of 200---->
   <----:     +- Exchange hashpartitioning(id#69, 200)---->
   <----:        +- *(2) Project [id#69, name#70]---->
   <----:           +- *(2) Filter ((isnotnull(id#69) && (id#69 = 1)) && (id#69 >= 1))---->
   <----:              +- *(2) FileScan parquet [id#69,name#70] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table/part-00003-0fc50086-fd6e-4527-..., PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1), GreaterThanOrEqual(id,1)], ReadSchema: struct<id:int,name:string>---->
   +- *(6) Sort [id#102 ASC NULLS FIRST], false, 0
      <----+- BucketUnion 200 buckets, bucket columns: [id]---->
         <----:- *(4) Project [id#102, name#103]---->
         <----:  +- *(4) Filter ((isnotnull(id#102) && (id#102 >= 1)) && (id#102 = 1))---->
         <----:     +- *(4) FileScan parquet [id#102,name#103] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index33/v__=..., PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,1), EqualTo(id,1)], ReadSchema: struct<id:int,name:string>, SelectedBucketsCount: 1 out of 200---->
         <----+- Exchange hashpartitioning(id#102, 200)---->
            <----+- *(5) Project [id#102, name#103]---->
               <----+- *(5) Filter ((isnotnull(id#102) && (id#102 >= 1)) && (id#102 = 1))---->
                  <----+- *(5) FileScan parquet [id#102,name#103] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table/part-00003-0fc50086-fd6e-4527-..., PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,1), EqualTo(id,1)], ReadSchema: struct<id:int,name:string>---->

Plan without indexes:
Project [id#69, name#70, name#103]
+- SortMergeJoin [id#69], [id#102], Inner
   :- *(2) Sort [id#69 ASC NULLS FIRST], false, 0
   <----:  +- Exchange hashpartitioning(id#69, 200)---->
   <----:     +- *(1) Project [id#69, name#70]---->
   <----:        +- *(1) Filter ((isnotnull(id#69) && (id#69 = 1)) && (id#69 >= 1))---->
   <----:           +- *(1) FileScan parquet [id#69,name#70] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table], PartitionFilters: [], PushedFilters: [IsNotNull(id), EqualTo(id,1), GreaterThanOrEqual(id,1)], ReadSchema: struct<id:int,name:string>---->
   +- *(4) Sort [id#102 ASC NULLS FIRST], false, 0
      <----+- Exchange hashpartitioning(id#102, 200)---->
         <----+- *(3) Project [id#102, name#103]---->
            <----+- *(3) Filter ((isnotnull(id#102) && (id#102 >= 1)) && (id#102 = 1))---->
               <----+- *(3) FileScan parquet [id#102,name#103] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/IdeaProjects/spark2/table], PartitionFilters: [], PushedFilters: [IsNotNull(id), GreaterThanOrEqual(id,1), EqualTo(id,1)], ReadSchema: struct<id:int,name:string>---->

Indexes used:
index33:file:/C:/Users/eunsong/IdeaProjects/spark2/spark-warehouse/indexes/index33/v__=0
```

Unit test & TPCH validation

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -20,8 +20,9 @@ import scala.collection.mutable
 
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.catalog.BucketSpec
 import org.apache.spark.sql.catalyst.expressions.Attribute
-import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan, Project, RepartitionByExpression}
+import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan, Project, RepartitionByExpression, Union}
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, InMemoryFileIndex, LogicalRelation, PartitioningAwareFileIndex}
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
 import org.apache.spark.sql.types.StructType
@@ -119,9 +120,9 @@ object RuleUtils {
   }
 
   /**
-   * Get replacement plan for current plan. The replacement plan reads data from indexes.
+   * Transform the current plan to utilize index. The transformed plan reads data from indexes.
    * If HybridScan is enabled, additional logical plans for the appended data would be
-   * generated and merged with index data plan. Refer [[getComplementIndexPlan]].
+   * generated and merged with index data plan. Refer [[transformPlanToUseHybridIndexDataScan]].
    *
    * Pre-requisites
    * - We know for sure the index which can be used to replace the plan.
@@ -135,30 +136,32 @@ object RuleUtils {
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
    * @return Replacement plan.
    */
-  def transformPlanWithIndex(
+  def transformPlanToUseIndex(
       spark: SparkSession,
       index: IndexLogEntry,
       plan: LogicalPlan,
       useBucketSpec: Boolean): LogicalPlan = {
     val transformed = if (HyperspaceConf.hybridScanEnabled(spark)) {
-      transformPlanUsingHybridScan(spark, index, plan, useBucketSpec)
+      transformPlanToUseHybridIndexDataScan(spark, index, plan, useBucketSpec)
     } else {
-      transformPlan(spark, index, plan, useBucketSpec)
+      transformPlanToUsePureIndexScan(spark, index, plan, useBucketSpec)
     }
     assert(!transformed.equals(plan))
     transformed
   }
 
   /**
-   * Get alternative logical plan of the current plan using the given index.
+   * Transform the current plan to utilize index.
+   * The transformed plan reads data from indexes instead of the source relations.
+   * Bucketing information of the index is retained if useBucketSpec is true.
    *
    * @param spark Spark session.
    * @param index Index used in replacement plan.
    * @param plan Current logical plan.
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
-   * @return Alternative logical plan.
+   * @return Transformed logical plan.
    */
-  private def transformPlan(
+  private def transformPlanToUsePureIndexScan(
       spark: SparkSession,
       index: IndexLogEntry,
       plan: LogicalPlan,
@@ -187,33 +190,26 @@ object RuleUtils {
   }
 
   /**
-   * Get alternative logical plan of the current plan using the given index.
+   * Transform the current plan to utilize index even with newly appended source files.
    * With HybridScan, indexes with newly appended files to its source relation are also
-   * eligible and we reconstruct new plans for the appended files so as to merge into
-   * the index data.
+   * eligible and we reconstruct new plans for the appended files so as to merge with
+   * bucketed index data correctly.
    *
    * @param spark Spark session.
    * @param index Index used in replacement plan.
    * @param plan Current logical plan.
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
-   * @return Alternative logical plan.
+   * @return Transformed logical plan.
    */
-  private def transformPlanUsingHybridScan(
+  private def transformPlanToUseHybridIndexDataScan(
       spark: SparkSession,
       index: IndexLogEntry,
       plan: LogicalPlan,
       useBucketSpec: Boolean): LogicalPlan = {
-    // Since the index data is in "parquet" format, we cannot read source files in
-    // formats other than "parquet" using 1 FileScan node (as the operator requires
-    // files in one homogenous format). To address this, we need to read the appended
-    // source files using another FileScan node injected into the plan and subsequently
-    // merge the data into the index data.
-    //
-    // Though BucketUnion (using BucketSpec and on-the-fly Shuffle) is used to merge
-    // them for now, these plans will be optimized with a Union operator at a later time
-    // (see #145).
-    val useBucketUnion = useBucketSpec || !index.relations.head.fileFormat.equals("parquet")
+    val isParquetSourceFormat = index.relations.head.fileFormat.equals("parquet")
     var filesAppended: Seq[Path] = Nil
+
+    // Get replaced plan with index data and appended files if applicable.
     val replacedPlan = plan transformDown {
       case baseRelation @ LogicalRelation(
             _ @HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
@@ -225,19 +221,26 @@ object RuleUtils {
         val curFileSet = location.allFiles
           .map(f => FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
           .toSet
-        filesAppended =
-          (curFileSet -- index.allSourceFileInfos).map(f => new Path(f.name)).toSeq
-
+        filesAppended = (curFileSet -- index.allSourceFileInfos).map(f => new Path(f.name)).toSeq
         // TODO: Hybrid Scan delete support.
 
         // If BucketSpec of index data isn't used (e.g., in the case of FilterIndex currently),
         // or the source format is not parquet. We could read appended data from source files
         // along with the index data.
         val readPaths = {
-          if (useBucketUnion) {
+          if (useBucketSpec || !isParquetSourceFormat) {
+            // FilterIndexRule: since the index data is in "parquet" format, we cannot read
+            // source files in formats other than "parquet" using 1 FileScan node as the
+            // operator requires files in one homogenous format. To address this, we need
+            // to read the appended source files using another FileScan node injected into
+            // the plan and subsequently merge the data into the index data. Please refer below
+            // [[Union]] operation.
             index.content.files
           } else {
-            index.content.files ++ filesAppended
+            val files = index.content.files ++ filesAppended
+            // Reset filesAppended as it's handled along with index data.
+            filesAppended = Nil
+            files
           }
         }
 
@@ -246,7 +249,7 @@ object RuleUtils {
           newLocation,
           new StructType(),
           StructType(index.schema.filter(baseRelation.schema.contains(_))),
-          if (useBucketUnion) Some(index.bucketSpec) else None,
+          if (useBucketSpec) Some(index.bucketSpec) else None,
           new ParquetFileFormat,
           Map())(spark)
 
@@ -254,88 +257,99 @@ object RuleUtils {
           baseOutput.filter(attr => relation.schema.fieldNames.contains(attr.name))
         baseRelation.copy(relation = relation, output = updatedOutput)
     }
-    if (useBucketUnion && filesAppended.nonEmpty) {
-      // If BucketSpec of the index is used to read the index data or the source format
-      // is not parquet, we need to shuffle the appended data in the same way to correctly
-      // merge with bucked index data.
-      getComplementIndexPlan(spark, index, plan, replacedPlan, filesAppended)
+
+    if (filesAppended.nonEmpty) {
+      // If there are unhandled appended files, we need to create additional plans:
+      //   1) Read the appended files with the same push-down filters.
+      //   2) If Shuffle is required, perform shuffle for the plan.
+      //   3) Merge both index data plan and the additional plan for appended files
+      //      by using [[BucketUnion]] or [[Union]].
+      // For more details, please refer https://github.com/microsoft/hyperspace/issues/150.
+
+      val planForAppended = transformPlanWithAppendedFiles(spark, index, plan, filesAppended)
+      if (useBucketSpec) {
+        // If Bucketing information of the index is used to read the index data, we need to
+        // shuffle the appended data in the same way to correctly merge with bucketed index data.
+
+        // Reset sortColumnNames as BucketUnion does not keep the sort order within a bucket.
+        val bucketSpec = index.bucketSpec.copy(sortColumnNames = Seq())
+
+        // Merge index plan & newly shuffled plan by using bucket-aware union.
+        BucketUnion(
+          Seq(replacedPlan, shuffleUsingIndexSpec(bucketSpec, planForAppended)),
+          bucketSpec)
+      } else {
+        // If bucketing is not necessary (e.g. FilterIndexRule), we use [[Union]] to merge
+        // the appended data without additional shuffle.
+        Union(replacedPlan, planForAppended)
+      }
     } else {
       replacedPlan
     }
   }
 
   /**
-   * Get complement plan for an index with appended data.
-   *
-   * This method consists of the following steps
-   * 1) Get a plan from originalPlan by replacing data location with appended data
-   * 2) On-the-fly shuffle for the appended data, using indexedColumns & numBuckets.
-   *   - Shuffle is located before Project to utilize Push-down Filters
-   *     - Shuffle => Project => Filter => Relation
-   *   - if Project filters indexedColumns, then Shuffle should be located after the node
-   *     - Project => Shuffle => Filter => Relation
-   * 3) Bucket-aware union both indexPlan and complementPlan to avoid repartitioning index data
+   * Transform the current plan to utilize index. The transformed plan reads data from indexes.
+   * Transform tGet modified plan by replacing data location with appended files.
+   * The result plan will be merged with the plan with index by [[BucketUnion]] or [[Union]].
    *
    * NOTE: This method currently only supports replacement of Scan Nodes i.e. Logical relations
    *
    * @param spark Spark session.
    * @param index Index used in replacement plan.
    * @param originalPlan Original plan.
-   * @param planWithIndex Replaced plan with index.
    * @param filesAppended Appended files to the source relation.
-   * @return Consolidated plan of planWithIndex and a complement plan for filesAppended
+   * @return Linear logical plan for appended files.
    */
-  private def getComplementIndexPlan(
+  private def transformPlanWithAppendedFiles(
       spark: SparkSession,
       index: IndexLogEntry,
       originalPlan: LogicalPlan,
-      planWithIndex: LogicalPlan,
       filesAppended: Seq[Path]): LogicalPlan = {
-    // 1) Replace the location of LogicalRelation with appended files
-    val complementIndexPlan = originalPlan transformDown {
-      case baseRelation @ LogicalRelation(
-            fsRelation @ HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
-            baseOutput,
-            _,
-            _) =>
-        // Set the same output schema with the index plan to merge them using BucketUnion
+    // Replace the location of LogicalRelation with appended files.
+    val planForAppended = originalPlan transformDown {
+      case baseRelation @ LogicalRelation(fsRelation: HadoopFsRelation, baseOutput, _, _) =>
+        // Set the same output schema with the index plan to merge them using BucketUnion.
         val updatedOutput =
           baseOutput.filter(attr => index.schema.fieldNames.contains(attr.name))
-
-        if (filesAppended.nonEmpty) {
-          val newLocation =
-            new InMemoryFileIndex(spark, filesAppended, Map(), None)
-          val newRelation =
-            fsRelation.copy(
-              location = newLocation,
-              dataSchema = StructType(index.schema.filter(baseRelation.schema.contains(_))))(
-              spark)
-          baseRelation.copy(relation = newRelation, output = updatedOutput)
-        } else {
-          baseRelation
-        }
+        val newLocation =
+          new InMemoryFileIndex(spark, filesAppended, Map(), None)
+        val newRelation =
+          fsRelation.copy(
+            location = newLocation,
+            dataSchema = StructType(index.schema.filter(baseRelation.schema.contains(_))))(spark)
+        baseRelation.copy(relation = newRelation, output = updatedOutput)
     }
-    assert(!originalPlan.equals(complementIndexPlan))
+    assert(!originalPlan.equals(planForAppended))
+    planForAppended
+  }
 
+  /**
+   * Get modified plan performing on-the-fly shuffle the source data using the index specification.
+   *
+   * @param bucketSpec BucketSpec of the index used in the current replacement plan.
+   * @param plan Plan to be shuffled.
+   * @return Modified plan by injecting on-the-fly shuffle with given index specification.
+   */
+  private[rules] def shuffleUsingIndexSpec(
+      bucketSpec: BucketSpec,
+      plan: LogicalPlan): LogicalPlan = {
     // Extract top level plan including all required columns for shuffle in its output.
+    // This is located inside this function because of bucketSpec.bucketColumnNames
     object ExtractTopLevelPlanForShuffle {
       type returnType = (LogicalPlan, Seq[Option[Attribute]], Boolean)
       def unapply(plan: LogicalPlan): Option[returnType] = plan match {
-        case project @ Project(
-              _,
-              Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _))) =>
-          val indexedAttrs = getIndexedAttrs(project, index.indexedColumns)
+        case project @ Project(_, Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _))) =>
+          val indexedAttrs = getIndexedAttrs(project, bucketSpec.bucketColumnNames)
           Some(project, indexedAttrs, true)
-        case project @ Project(
-              _,
-              LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
-          val indexedAttrs = getIndexedAttrs(project, index.indexedColumns)
+        case project @ Project(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
+          val indexedAttrs = getIndexedAttrs(project, bucketSpec.bucketColumnNames)
           Some(project, indexedAttrs, true)
         case filter @ Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
-          val indexedAttrs = getIndexedAttrs(filter, index.indexedColumns)
+          val indexedAttrs = getIndexedAttrs(filter, bucketSpec.bucketColumnNames)
           Some(filter, indexedAttrs, false)
         case relation @ LogicalRelation(_: HadoopFsRelation, _, _, _) =>
-          val indexedAttrs = getIndexedAttrs(relation, index.indexedColumns)
+          val indexedAttrs = getIndexedAttrs(relation, bucketSpec.bucketColumnNames)
           Some(relation, indexedAttrs, false)
       }
       def getIndexedAttrs(
@@ -346,10 +360,10 @@ object RuleUtils {
       }
     }
 
-    // Step 2) During merge of the index data with the newly appended data,
-    // perform on-the-fly shuffle of the appended-data with the same partition
-    // structure of index so that we could avoid incurring shuffle of the index
-    // data (which is assumed to be the larger sized dataset).
+    // During merge of the index data with the newly appended data, perform
+    // on-the-fly shuffle of the appended-data with the same partition structure
+    // of index so that we could avoid incurring shuffle of the index data
+    // (which is assumed to be the larger sized dataset).
     //
     // To do this optimally, we would have to push-down filters/projects before
     // shuffling the newly appended data to reduce the amount of data that gets
@@ -364,20 +378,21 @@ object RuleUtils {
     // else:
     //    Case 2: Shuffle node should come **after** Project and/or Filter node
     //                Plan: Shuffle => Project => Filter => Relation
+    //
+    // Currently, we only perform on-the-fly shuffle when applying JoinIndexRule.
+    // Therefore, it's always guaranteed that the children nodes have all indexed columns
+    // in their output; Case 1 won't be shown in use cases.
 
     var shuffleInjected = false
     // Remove sort order because we cannot guarantee the ordering of source files.
-    val bucketSpec = index.bucketSpec.copy(sortColumnNames = Seq())
-    val shuffled = complementIndexPlan transformDown {
+    val shuffled = plan transformDown {
       case p if shuffleInjected => p
-      case ExtractTopLevelPlanForShuffle(plan, indexedAttr, isProject)
+      case ExtractTopLevelPlanForShuffle(p, indexedAttr, isProject)
           if !isProject || indexedAttr.forall(_.isDefined) =>
         shuffleInjected = true
-        RepartitionByExpression(indexedAttr.flatten, plan, index.numBuckets)
+        RepartitionByExpression(indexedAttr.flatten, p, bucketSpec.numBuckets)
     }
-
-    // 3) Merge index plan & newly shuffled plan by using bucket-aware union.
-    // Currently, BucketUnion does not keep the sort order within a bucket.
-    BucketUnion(Seq(planWithIndex, shuffled), bucketSpec)
+    assert(shuffleInjected)
+    shuffled
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -63,7 +63,7 @@ object RuleUtils {
       }
     }
 
-    def isHybridScanCandidate(entry: IndexLogEntry, filesByRelations: Seq[FileInfo]): Boolean = {
+    def isHybridScanCandidate(entry: IndexLogEntry, inputSourceFiles: Seq[FileInfo]): Boolean = {
       // TODO: Some threshold about the similarity of source data files - number of common files or
       //  total size of common files.
       //  See https://github.com/microsoft/hyperspace/issues/159
@@ -71,9 +71,9 @@ object RuleUtils {
       //  support arbitrary source plans at index creation.
       //  See https://github.com/microsoft/hyperspace/issues/158
 
-      // Find a common file between the input relation & index source files.
-      // Without the threshold described above, we can utilize exists & contain functions here.
-      val commonCnt = filesByRelations.count(entry.allSourceFileInfos.contains)
+      // Find the number of common files and deleted files between the source relations
+      // & index source files.
+      val commonCnt = inputSourceFiles.count(entry.allSourceFileInfos.contains)
       val deletedCnt = entry.allSourceFileInfos.size - commonCnt
 
       // Currently, Hybrid Scan only support for append-only dataset.
@@ -87,7 +87,7 @@ object RuleUtils {
 
     if (hybridScanEnabled) {
       // TODO: Duplicate listing files for the given relation as in
-      //  [[transformPlanToUseHybridIndexDataScan]]
+      //  [[transformPlanToUseHybridScan]]
       //  See https://github.com/microsoft/hyperspace/issues/160
       val filesByRelations = plan
         .collect {
@@ -130,14 +130,14 @@ object RuleUtils {
    * constructed and merged with the transformed plan with the index.
    *
    * Pre-requisites
-   * - We know for sure the index which can be used to replace the plan.
+   * - We know for sure the index which can be used to transform the plan.
    * - The plan should be linear and include 1 LogicalRelation.
    *
    * @param spark Spark session.
-   * @param index Index used in replacement plan.
+   * @param index Index used in transformation of plan.
    * @param plan Current logical plan.
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
-   * @return Replacement plan.
+   * @return Transformed plan.
    */
   def transformPlanToUseIndex(
       spark: SparkSession,
@@ -145,9 +145,9 @@ object RuleUtils {
       plan: LogicalPlan,
       useBucketSpec: Boolean): LogicalPlan = {
     val transformed = if (HyperspaceConf.hybridScanEnabled(spark)) {
-      transformPlanToUseHybridIndexDataScan(spark, index, plan, useBucketSpec)
+      transformPlanToUseHybridScan(spark, index, plan, useBucketSpec)
     } else {
-      transformPlanToUsePureIndexScan(spark, index, plan, useBucketSpec)
+      transformPlanToUseIndexOnlyScan(spark, index, plan, useBucketSpec)
     }
     assert(!transformed.equals(plan))
     transformed
@@ -158,20 +158,20 @@ object RuleUtils {
    * The transformed plan reads data from indexes instead of the source relations.
    * Bucketing information of the index is retained if useBucketSpec is true.
    *
-   * NOTE: This method currently only supports replacement of Scan Nodes i.e. Logical relations.
+   * NOTE: This method currently only supports transformation of Scan Nodes i.e. Logical relations.
    *
    * @param spark Spark session.
-   * @param index Index used in replacement plan.
+   * @param index Index used in transformation of plan.
    * @param plan Current logical plan.
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
    * @return Transformed logical plan that leverages an index.
    */
-  private def transformPlanToUsePureIndexScan(
+  private def transformPlanToUseIndexOnlyScan(
       spark: SparkSession,
       index: IndexLogEntry,
       plan: LogicalPlan,
       useBucketSpec: Boolean): LogicalPlan = {
-    // Note that we replace *only* the base relation and not other portions of the plan
+    // Note that we transform *only* the base relation and not other portions of the plan
     // (e.g., filters). For instance, given the following input plan:
     //        Project(A,B) -> Filter(C = 10) -> Scan (A,B,C,D,E)
     // in the presence of a suitable index, the getIndexPlan() method will emit:
@@ -202,20 +202,20 @@ object RuleUtils {
    * bucketed index data correctly.
    *
    * @param spark Spark session.
-   * @param index Index used in replacement plan.
+   * @param index Index used in transformation of plan.
    * @param plan Current logical plan.
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
    * @return Transformed logical plan that leverages an index and merges appended data.
    */
-  private def transformPlanToUseHybridIndexDataScan(
+  private def transformPlanToUseHybridScan(
       spark: SparkSession,
       index: IndexLogEntry,
       plan: LogicalPlan,
       useBucketSpec: Boolean): LogicalPlan = {
     val isParquetSourceFormat = index.relations.head.fileFormat.equals("parquet")
-    var filesAppended: Seq[Path] = Nil
+    var unhandledAppendedFiles: Seq[Path] = Nil
 
-    // Get replaced plan with index data and appended files if applicable.
+    // Get transformed plan with index data and appended files if applicable.
     val indexPlan = plan transformDown {
       case baseRelation @ LogicalRelation(
             _ @HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
@@ -224,31 +224,29 @@ object RuleUtils {
             _) =>
         val curFileSet = location.allFiles
           .map(f => FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
-        filesAppended =
+        val filesAppended =
           curFileSet.filterNot(index.allSourceFileInfos.contains).map(f => new Path(f.name))
         // TODO: Hybrid Scan delete support.
 
-        // If BucketSpec of index data isn't used (e.g., in the case of FilterIndex currently),
-        // or the source format is not parquet. We could read appended data from source files
-        // along with the index data.
-        val readPaths = {
+        val filesToRead = {
           if (useBucketSpec || !isParquetSourceFormat) {
-            // FilterIndexRule: Since the index data is in "parquet" format, we cannot read
-            // source files in formats other than "parquet" using 1 FileScan node as the
-            // operator requires files in one homogenous format. To address this, we need
-            // to read the appended source files using another FileScan node injected into
-            // the plan and subsequently merge the data into the index data. Please refer below
-            // [[Union]] operation.
+            // Since the index data is in "parquet" format, we cannot read source files
+            // in formats other than "parquet" using 1 FileScan node as the operator requires
+            // files in one homogenous format. To address this, we need to read the appended
+            // source files using another FileScan node injected into the plan and subsequently
+            // merge the data into the index data. Please refer below [[Union]] operation.
+            unhandledAppendedFiles = filesAppended
             index.content.files
           } else {
+            // If BucketSpec of index data isn't used (e.g., in the case of FilterIndex currently)
+            // aud the source format is parquet, we could read the appended files along
+            // with the index data.
             val files = index.content.files ++ filesAppended
-            // Reset filesAppended as it's handled with index data.
-            filesAppended = Nil
             files
           }
         }
 
-        val newLocation = new InMemoryFileIndex(spark, readPaths, Map(), None)
+        val newLocation = new InMemoryFileIndex(spark, filesToRead, Map(), None)
         val relation = HadoopFsRelation(
           newLocation,
           new StructType(),
@@ -262,27 +260,27 @@ object RuleUtils {
         baseRelation.copy(relation = relation, output = updatedOutput)
     }
 
-    if (filesAppended.nonEmpty) {
+    if (unhandledAppendedFiles.nonEmpty) {
       // If there are unhandled appended files, we need to create additional plans
       // by the following steps:
-      // Step 1) Generate a plan(planForAppended) from the current plan to read
+      // Step 1) Generate a plan (planForAppended) from the current plan to read
       //   the appended files similar to indexPlan.
       // Step 2) If Shuffle is required, perform shuffle for the plan.
       // Step 3) Merge both indexPlan and planForAppended by using [[BucketUnion]] or [[Union]].
       // For more details, see https://github.com/microsoft/hyperspace/issues/150.
 
       val planForAppended =
-        transformPlanToReadFromAppendedFiles(spark, index.schema, plan, filesAppended)
+        transformPlanToReadAppendedFiles(spark, index.schema, plan, unhandledAppendedFiles)
       if (useBucketSpec) {
         // If Bucketing information of the index is used to read the index data, we need to
         // shuffle the appended data in the same way to correctly merge with bucketed index data.
 
         // Clear sortColumnNames as BucketUnion does not keep the sort order within a bucket.
-        val bucketSpec = index.bucketSpec.copy(sortColumnNames = Seq())
+        val bucketSpec = index.bucketSpec.copy(sortColumnNames = Nil)
 
         // Merge index plan & newly shuffled plan by using bucket-aware union.
         BucketUnion(
-          Seq(indexPlan, transformPlanToShuffleUsingIndexSpec(bucketSpec, planForAppended)),
+          Seq(indexPlan, transformPlanToShuffleUsingBucketSpec(bucketSpec, planForAppended)),
           bucketSpec)
       } else {
         // If bucketing is not necessary (e.g. FilterIndexRule), we use [[Union]] to merge
@@ -306,12 +304,12 @@ object RuleUtils {
    * @param filesAppended Appended files to the source relation.
    * @return Transformed linear logical plan for appended files.
    */
-  private def transformPlanToReadFromAppendedFiles(
+  private def transformPlanToReadAppendedFiles(
       spark: SparkSession,
       indexSchema: StructType,
       originalPlan: LogicalPlan,
       filesAppended: Seq[Path]): LogicalPlan = {
-    // Replace the location of LogicalRelation with appended files.
+    // Transform the location of LogicalRelation with appended files.
     val planForAppended = originalPlan transformDown {
       case baseRelation @ LogicalRelation(fsRelation: HadoopFsRelation, baseOutput, _, _) =>
         // Set the same output schema with the index plan to merge them using BucketUnion.
@@ -339,28 +337,24 @@ object RuleUtils {
    * @param plan Plan to be shuffled.
    * @return Transformed plan by injecting on-the-fly shuffle with given bucket specification.
    */
-  private[rules] def transformPlanToShuffleUsingIndexSpec(
+  private[rules] def transformPlanToShuffleUsingBucketSpec(
       bucketSpec: BucketSpec,
       plan: LogicalPlan): LogicalPlan = {
     // Extract top level plan including all required columns for shuffle in its output.
-    // This is located inside this function because of bucketSpec.bucketColumnNames.
     object ExtractTopLevelPlanForShuffle {
       type returnType = (LogicalPlan, Seq[Option[Attribute]], Boolean)
       def unapply(plan: LogicalPlan): Option[returnType] = plan match {
-        case project @ Project(_, Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _))) =>
-          val indexedAttrs = getIndexedAttrs(project, bucketSpec.bucketColumnNames)
-          Some(project, indexedAttrs, true)
-        case project @ Project(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
-          val indexedAttrs = getIndexedAttrs(project, bucketSpec.bucketColumnNames)
-          Some(project, indexedAttrs, true)
-        case filter @ Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
-          val indexedAttrs = getIndexedAttrs(filter, bucketSpec.bucketColumnNames)
-          Some(filter, indexedAttrs, false)
-        case relation @ LogicalRelation(_: HadoopFsRelation, _, _, _) =>
-          val indexedAttrs = getIndexedAttrs(relation, bucketSpec.bucketColumnNames)
-          Some(relation, indexedAttrs, false)
+        case p @ Project(_, Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _))) =>
+          Some(p, getIndexedAttrs(p, bucketSpec.bucketColumnNames), true)
+        case p @ Project(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
+          Some(p, getIndexedAttrs(p, bucketSpec.bucketColumnNames), true)
+        case f @ Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
+          Some(f, getIndexedAttrs(f, bucketSpec.bucketColumnNames), false)
+        case r @ LogicalRelation(_: HadoopFsRelation, _, _, _) =>
+          Some(r, getIndexedAttrs(r, bucketSpec.bucketColumnNames), false)
       }
-      def getIndexedAttrs(
+
+      private def getIndexedAttrs(
           plan: LogicalPlan,
           indexedColumns: Seq[String]): Seq[Option[Attribute]] = {
         val attrMap = plan.output.attrs.map(attr => (attr.name, attr)).toMap
@@ -393,7 +387,6 @@ object RuleUtils {
     // for future use cases.
 
     var shuffleInjected = false
-    // Remove sort order because we cannot guarantee the ordering of source files.
     val shuffled = plan transformDown {
       case p if shuffleInjected => p
       case ExtractTopLevelPlanForShuffle(p, indexedAttr, isProject)
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -281,15 +281,15 @@ object RuleUtils {
    * @param spark Spark session.
    * @param index Index used in replacement plan.
    * @param originalPlan Original plan.
-   * @param indexPlan Replaced plan with index.
+   * @param planWithIndex Replaced plan with index.
    * @param filesAppended Appended files to the source relation.
-   * @return complementIndexPlan integrated plan of indexPlan and complementPlan.
+   * @return Consolidated plan of planWithIndex and a complement plan for filesAppended
    */
   private def getComplementIndexPlan(
       spark: SparkSession,
       index: IndexLogEntry,
       originalPlan: LogicalPlan,
-      indexPlan: LogicalPlan,
+      planWithIndex: LogicalPlan,
       filesAppended: Seq[Path]): LogicalPlan = {
     // 1) Replace the location of LogicalRelation with appended files
     val complementIndexPlan = originalPlan transformDown {
@@ -378,6 +378,6 @@ object RuleUtils {
 
     // 3) Merge index plan & newly shuffled plan by using bucket-aware union.
     // Currently, BucketUnion does not keep the sort order within a bucket.
-    BucketUnion(Seq(indexPlan, shuffled), bucketSpec)
+    BucketUnion(Seq(planWithIndex, shuffled), bucketSpec)
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -164,7 +164,7 @@ object RuleUtils {
    * @param index Index used in replacement plan.
    * @param plan Current logical plan.
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
-   * @return Transformed logical plan.
+   * @return Transformed logical plan that leverages an index.
    */
   private def transformPlanToUsePureIndexScan(
       spark: SparkSession,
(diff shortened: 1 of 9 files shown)


rapoth:
> The query plan is modified in optimizer to support this and the changed plan can be checked with hs.explain() API.

Can you add an example of how `hs.explain` will look like for a simple scenario?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,246 @@ object RuleUtils {
...
+   */
+  private def getIndexPlan(
+      spark: SparkSession,
+      index: IndexLogEntry,
+      plan: LogicalPlan,
+      useBucketSpec: Boolean): LogicalPlan = {
+    // Here we can't replace the plan completely with the index. This will create problems.
+    // For e.g. if Project(A,B) -> Filter(C = 10) -> Scan (A,B,C,D,E)
+    // if we replace this plan with index Scan (A,B,C), we lose the Filter(C=10) and it will
+    // lead to wrong results. So we only replace the base relation.
```
rapoth, line 162:
```suggestion
    // Note that we replace *only* the base relation and not any sub plans (e.g., filters).
    // For instance, given the following input plan:
    //        Project(A,B) -> Filter(C = 10) -> Scan (A,B,C,D,E)
    // in the presence of a suitable index, this method will emit:
    //        Project(A,B) -> Filter(C = 10) -> Index Scan (A,B,C)
    // In other words, the Filter is untouched to preserve correctness. 
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,246 @@ object RuleUtils {
...
+  private def getHybridScanIndexPlan(
+      spark: SparkSession,
+      index: IndexLogEntry,
+      plan: LogicalPlan,
+      useBucketSpec: Boolean): LogicalPlan = {
+    // Other than "parquet" format, we cannot read source files along with index data
+    // files in parquet format with 1 FileScan node. In this case, we need to read the
+    // appended source files by another FileScan node and merge into the index data.
+    // Though BucketUnion (using BucketSpec and on-the-fly Shuffle) is used to merge
+    // them for now, we will try to optimize these plans with Union operator.
```
rapoth, line 202:
```suggestion
    // Since the index data is in "parquet" format, we cannot read source files in 
    // formats other than "parquet" using 1 FileScan node (as the operator requires
    // files in one homogenous format). 
    // 
    // To address this, we need to read the appended source files using another 
    // FileScan node injected into the plan and subsequently merge the data into 
    // the index data.
    //
    // Though BucketUnion (using BucketSpec and on-the-fly Shuffle) is used to merge
    // them for now, these plans will be optimized with a Union operator at a later time
    // (see #145).
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,246 @@ object RuleUtils {
...
+            _ @HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+            baseOutput,
+            _,
+            _) =>
+        val curFileSet = location.allFiles
+          .map(f => FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
+          .toSet
+
+        // if BucketSpec of index data isn't used, we could read appended data from
+        // source files directly.
```
rapoth, line 215:
```suggestion
        // if BucketSpec of index data isn't used (e.g., in the case of FilterIndex currently), we could 
        // read appended data from source files directly.
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: One question about the line length limit. It seems our line length limit is `maxColumn = 98` (in scalafmt.conf) though the below coding style guides recommend 80 and 100. Is there any reason for using 98?

https://docs.scala-lang.org/style/indentation.html#line-wrapping => 80
https://github.com/databricks/scala-style-guide#linelength => 100

rapoth: Good question. :) It was inspired from [Spark's scalafmt.conf](https://github.com/apache/spark/blob/master/dev/.scalafmt.conf)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,252 @@ object RuleUtils {
...
+              location = newLocation,
+              dataSchema = StructType(index.schema.filter(baseRelation.schema.contains(_))))(
+              spark)
+          baseRelation.copy(relation = newRelation, output = updatedOutput)
+        } else {
+          baseRelation
+        }
+    }
+
+    if (!originalPlan.equals(complementIndexPlan)) {
```
imback82, line 308: So, it looks like if `filesAppended` computed in `getHybridScanIndexPlan ` is empty, we don't even need to call `getComplementIndexPlan`, no?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,252 @@ object RuleUtils {
...
+      // Remove sort order because we cannot guarantee the ordering of source files
+      val bucketSpec = index.bucketSpec.copy(sortColumnNames = Seq())
+
+      object ExtractTopLevelPlanForShuffle {
```
imback82, line 312: Can we define this outside this function with proper comment, etc.?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,252 @@ object RuleUtils {
...
+
+      // 2) Perform on-the-fly Shuffle with the same partition structure of index
+      // so that we could avoid incurring Shuffle of whole index data at merge stage.
+      // In order to utilize push-down filters, we would locate Shuffle node after
+      // Project or Filter node. (Case 1)
+      // However, if Project node excludes any of indexedColumns, Shuffle will be
+      // converted to RoundRobinPartitioning which can cause wrong result issues.
+      // So Shuffle should be located before Project node in that case. (Case 2)
+      // Case 1) Shuffle => Project => Filter => Relation (Project&Filter will be pushed down)
+      // Case 2) Project => Shuffle => Filter => Relation (Filter will be pushed down)
```
imback82, line 348: Can you point me to the tests that are handling these two cases?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,252 @@ object RuleUtils {
...
+   * @param index index used in replacement plan
+   * @param originalPlan original plan
+   * @param indexPlan replaced plan with index
+   * @return complementIndexPlan integrated plan of indexPlan and complementPlan
+   */
+  private def getComplementIndexPlan(
+      spark: SparkSession,
+      index: IndexLogEntry,
+      originalPlan: LogicalPlan,
+      indexPlan: LogicalPlan): LogicalPlan = {
```
imback82, line 278: Can we name this with `planWithIndex`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,252 @@ object RuleUtils {
...
+              spark)
+          baseRelation.copy(relation = newRelation, output = updatedOutput)
+        } else {
+          baseRelation
+        }
+    }
+
+    if (!originalPlan.equals(complementIndexPlan)) {
+      // Remove sort order because we cannot guarantee the ordering of source files
+      val bucketSpec = index.bucketSpec.copy(sortColumnNames = Seq())
```
imback82, line 310: Can we define this right before it's used below.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,252 @@ object RuleUtils {
...
+
+      object ExtractTopLevelPlanForShuffle {
+        type returnType = (LogicalPlan, Seq[Option[Attribute]], Boolean)
+        def unapply(plan: LogicalPlan): Option[returnType] = plan match {
+          case project @ Project(
+                _,
+                Filter(_, LogicalRelation(HadoopFsRelation(_, _, _, _, _, _), _, _, _))) =>
```
imback82, line 317: `HadoopFsRelation(_, _, _, _, _, _)` => `_: HadoopFsRelation`

imback82 (commented):
I did one round of review (but not tests yet). It looks pretty cool!

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,242 @@
...
+        assert(transformed.equals(join.queryExecution.optimizedPlan), "No plan transformation.")
+      }
+
+      withSQLConf("spark.hyperspace.index.hybridscan.enabled" -> "true") {
+        val planWithHybridScan = JoinIndexRule(join.queryExecution.optimizedPlan)
+        assert(
+          !planWithHybridScan.equals(join.queryExecution.optimizedPlan),
+          "Plan should be transformed.")
+
+        // check appended file is added to relation node or not
```
imback82, line 140: Comments should start with upper case and end with "." (if it's a sentence). Can you go thru your comments?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -213,18 +213,18 @@ object RuleUtils {
...
             _,
             _) =>
-        val curFileSet = location.allFiles
+        lazy val curFileSet = location.allFiles
           .map(f => FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
           .toSet
+        lazy val filesAppended =
+          (curFileSet -- index.allSourceFileInfos).map(f => new Path(f.name)).toSeq
 
         // if BucketSpec of index data isn't used (e.g., in the case of FilterIndex currently),
```
rapoth, line 222:
```suggestion
        // If BucketSpec of index data isn't used (e.g., in the case of FilterIndex currently),
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: I realized that we don't need to check this condition if we use Union for filter index rule here. 
For join rule, the index cannot be a candidate if the plan doesn't have all indexed columns.

Can I let this code without test and removed with #168 or add Union in this PR? Actually it's quite small change.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: It's located here because of `index.indexedColumns`. I tried to define outside and pass the column names but I couldn't find a way.

Anyway it seems this extractor is not required for now..

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
rapoth, line 348: I am pretty sure I made this change but don't know what happened to that suggestion :) Sorry I made it a bit verbose (please feel free to modify as you see fit). I just thought this is such an important piece that it needs some explanation so the "future me" won't forget :D 

Also, note that I reversed Case 1 and 2 to write this properly. 

```suggestion
      // Step 2) During merge of the index data with the newly appended data, 
      // perform on-the-fly shuffle of the appended-data with the same partition 
      // structure of index so that we could avoid incurring shuffle of the index 
      // data (which is assumed to be the larger sized dataset).
      // 
      // To do this optimally, we would have to push-down filters/projects before
      // shuffling the newly appended data to reduce the amount of data that gets 
      // shuffled. However, if project excludes any of the [[indexedColumns]], Spark 
      // handles the shuffle through [[RoundRobinPartitioning]], which can cause 
      // wrong results.
      //
      // Therefore, we do the following:
      // if Project node excludes any of the [[indexedColumns]]:
      //    Case 1: Shuffle should be located **before** Project but **after** Filter
      //                Plan: Project => Shuffle => Filter => Relation 
      // else:
      //    Case 2: Shuffle node should come **after** Project and/or Filter node 
      //                Plan: Shuffle => Project => Filter => Relation
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +113,252 @@ object RuleUtils {
...
+   *
+   * NOTE: This method currently only supports replacement of Scan Nodes i.e. Logical relations
+   *
+   * @param spark Spark session
+   * @param index index used in replacement plan
+   * @param originalPlan original plan
+   * @param indexPlan replaced plan with index
+   * @return complementIndexPlan integrated plan of indexPlan and complementPlan
+   */
+  private def getComplementIndexPlan(
```
rapoth, line 281: This seems to be the "return" value of `getHybridScanIndexPlan` so why is this called `getComplementPlan`? What is it complementing? From what I can understand the use of complement doesn't seem appropriate. I don't have any good ideas but wanted to understand the semantic meaning here.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Yes but it's not always computed in getHybridScanIndexPlan. 
Actually I'd like to compute the current file list & appended files & deleted files for each index once and keep across Rules ..

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: I used "complement" as this additional plan (for appended files) complements the index plan. 
getCompleteIndexPlan / getComplemantaryIndexPlan / .. 🤔 

Any good suggestion would be welcomed. :)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: I updated the description but we might not need to handle Case1) because `Shuffle` is only applied for Join Index which requires all indexed columns in its join condition. (refer https://github.com/microsoft/hyperspace/pull/165#discussion_r490674500)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: Well, I see that `filesAppended` is computed `if (!useBucketUnion)` in `getHybridScanIndexPlan`, but this function is getting called `if (useBucketUnion)`, so it's always computed anyway?

I was thinking something like:
```scala
var appendedFiles = Nil // This can be updated to a tuple when deletedFiles are introduced.
val replacedPlan = plan transformDown {
  // Always update appendedFiles.
}

if (useBucketUnion && appendFiles.nonEmpty) {
  bucketUnion(spark, index, plan, replacedPlan)
} else {
  replacedPlan
}
```
I think this will make it much easier to understand the flow. (Sorry if I misunderstood the code.)

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,242 @@
...
+      val nodes = planWithHybridScan collect {
+        case p @ LogicalRelation(fsRelation: HadoopFsRelation, _, _, _) =>
+          val appendedFileCnt = fsRelation.location.inputFiles.count(_.contains(".copy"))
+          val indexFileCnt = fsRelation.location.inputFiles.count(_.contains("index2"))
+          assert(appendedFileCnt === 1 || indexFileCnt === 4)
+          assert(appendedFileCnt * indexFileCnt === 0)
+          p
+        case p @ BucketUnion(_, _) => p
+      }
+      assert(nodes.count(p => p.getClass.toString.contains("LogicalRelation")) === 2)
```
imback82, line 208: `_.isInstanceOf[LogicalRelation]`?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -167,6 +168,7 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
           case _ =>
         }
       }
+       */
 
       {
         val optimizedPlan = spark.read.parquet(dataPath).queryExecution.optimizedPlan
```
imback82, line 174: Don't you need to comment out the following line? ` verify(optimizedPlan, hybridScanEnabled = true, expectCandidateIndex = true)`? Is this also breaking when `hybridScanEnabled = false`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: In this way, if we don't pass the appended file list, it's computed twice in both functions.
In order to pass the appended file list to the getComplementIndexPlan, we might need a map <LogicalRelation, appendedfiles> (for multiple relations later).. is it okay to introduce these kind of map here?

imback82: Oops, yes, I meant to pass the `appendFiles` to the function.

Another reason to cache if we plan to support multiple relations. :) But, no need to introduce the mapping in this PR. Btw, `transformPlanWithIndex` supports only linear LogicalRelation at the moment right (meaning it will break)? I think I left a comment about this pre-requisite, but forgot about it.

sezruby: I updated pre-requisite comment and added `filesAppended`. It looks better to understand the flow 👍

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
rapoth: I saw your other renames. How about `complementIndexScanWithDataScan`? (since you are already using a verb like `transform`, I thought we could use `complement` as a word)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +117,267 @@ object RuleUtils {
...
+   * @param filesAppended Appended files to the source relation.
+   * @return Consolidated plan of planWithIndex and a complement plan for filesAppended
+   */
+  private def getComplementIndexPlan(
+      spark: SparkSession,
+      index: IndexLogEntry,
+      originalPlan: LogicalPlan,
+      planWithIndex: LogicalPlan,
+      filesAppended: Seq[Path]): LogicalPlan = {
+    // 1) Replace the location of LogicalRelation with appended files
```
rapoth, line 294: nit:
```suggestion
    // Step 1) Replace the location of LogicalRelation with appended files
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +117,267 @@ object RuleUtils {
...
+    val bucketSpec = index.bucketSpec.copy(sortColumnNames = Seq())
+    val shuffled = complementIndexPlan transformDown {
+      case p if shuffleInjected => p
+      case ExtractTopLevelPlanForShuffle(plan, indexedAttr, isProject)
+          if !isProject || indexedAttr.forall(_.isDefined) =>
+        shuffleInjected = true
+        RepartitionByExpression(indexedAttr.flatten, plan, index.numBuckets)
+    }
+
+    // 3) Merge index plan & newly shuffled plan by using bucket-aware union.
```
rapoth, line 379:
```suggestion
    // Step 3) Merge index plan & newly shuffled plan by using bucket-aware union.
```

rapoth (commented):
Minor opinionated comments but please feel free to ignore. Looking great, thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: While adding 'var files appended', I refactored getComplemetIndex and also applied the Union change as I thought the change is simple enough. 
Sorry for the big diff, but I think this looks better than before.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: I refactored the getComplemetPlan into 2 different functions - transformPlanWithAppendedFiles, shufflePlanWithIndexSpec :)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -119,46 +123,50 @@ object RuleUtils {
...
+   * Bucketing information of the index is retained if useBucketSpec is true.
+   *
+   * NOTE: This method currently only supports replacement of Scan Nodes i.e. Logical relations
    *
    * @param spark Spark session.
    * @param index Index used in replacement plan.
    * @param plan Current logical plan.
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
-   * @return Alternative logical plan.
+   * @return Transformed logical plan.
```
rapoth, line 167: nit:
```suggestion
   * @return Transformed logical plan that leverages an index.
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -187,57 +195,56 @@ object RuleUtils {
...
-   * the index data.
+   * eligible and we reconstruct new plans for the appended files so as to merge with
+   * bucketed index data correctly.
    *
    * @param spark Spark session.
    * @param index Index used in replacement plan.
    * @param plan Current logical plan.
    * @param useBucketSpec Option whether to use BucketSpec for reading index data.
-   * @return Alternative logical plan.
+   * @return Transformed logical plan.
```
rapoth, line 208: nit: consistency

```suggestion
   * @return Transformed logical plan that leverages an index and merges appended data.
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -187,57 +195,56 @@ object RuleUtils {
...
-        // TODO: Duplicate listing files for the given relation as in isHybridScanCandidate.
-        //  See https://github.com/microsoft/hyperspace/issues/160
         val curFileSet = location.allFiles
           .map(f => FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
-          .toSet
         filesAppended =
-          (curFileSet -- index.allSourceFileInfos).map(f => new Path(f.name)).toSeq
-
+          curFileSet.filterNot(index.allSourceFileInfos.contains).map(f => new Path(f.name))
         // TODO: Hybrid Scan delete support.
```
rapoth, line 229: nit: If you are going to immediately address this, it's ok. Otherwise, add an issue here so we can track.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -246,96 +253,111 @@ object RuleUtils {
...
+
+        // Clear sortColumnNames as BucketUnion does not keep the sort order within a bucket.
+        val bucketSpec = index.bucketSpec.copy(sortColumnNames = Seq())
+
+        // Merge index plan & newly shuffled plan by using bucket-aware union.
+        BucketUnion(
+          Seq(indexPlan, shuffleUsingIndexSpec(bucketSpec, planForAppended)),
+          bucketSpec)
+      } else {
+        // If bucketing is not necessary (e.g. FilterIndexRule), we use [[Union]] to merge
```
rapoth, line 288: +1 This is much clearer. Thank you!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -246,96 +253,111 @@ object RuleUtils {
...
+   * by using [[BucketUnion]] or [[Union]].
    *
    * @param spark Spark session.
-   * @param index Index used in replacement plan.
+   * @param indexSchema Index schema used for the output.
    * @param originalPlan Original plan.
-   * @param planWithIndex Replaced plan with index.
    * @param filesAppended Appended files to the source relation.
-   * @return Consolidated plan of planWithIndex and a complement plan for filesAppended
+   * @return Linear logical plan for appended files.
```
rapoth, line 307: nit: consistency 

```suggestion
   * @return Transformed linear logical plan for appended files.
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -246,96 +253,111 @@ object RuleUtils {
...
    */
-  private def getComplementIndexPlan(
+  private def transformPlanWithAppendedFiles(
```
rapoth, line 309:
```suggestion
  private def transformPlanToReadFromAppendedFiles(
```

What do you think? Might also be consistent with our use of `To` in the method names above.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -246,96 +253,111 @@ object RuleUtils {
...
+   * Transform the plan to perform on-the-fly Shuffle the data based on bucketSpec.
+   *
+   * Pre-requisites
+   * - The plan should be linear and include 1 LogicalRelation.
+   *
+   * @param bucketSpec Bucket specification used for Shuffle.
+   * @param plan Plan to be shuffled.
+   * @return Transformed plan by injecting on-the-fly shuffle with given bucket specification.
+   */
+  private[rules] def shuffleUsingIndexSpec(
```
rapoth, line 342: How about:

```suggestion
  private[rules] def transformPlanToShuffleUsingIndexSpec(
```

I'll leave it to you. Either looks good to me. This comment is only to stay consistent.

rapoth (commented):
Very minor comments but LGTM, thanks @sezruby!

sezruby:
@pirz @apoorvedave1 Could you do a review for this PR? Thanks!

apoorvedave1:
LGTM 👍 thanks @sezruby

apoorvedave1 (approved):

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,243 @@
...
+
+      // Check appended file is added to relation node or not.
+      val nodes = planWithHybridScan collect {
+        case p @ LogicalRelation(fsRelation: HadoopFsRelation, _, _, _) =>
+          val appendedFileCnt = fsRelation.location.inputFiles.count(_.contains(".copy"))
+          val indexFileCnt = fsRelation.location.inputFiles.count(_.contains("index2"))
+          assert(appendedFileCnt === 1 || indexFileCnt === 4)
+          assert(appendedFileCnt * indexFileCnt === 0)
+          p
+        case p @ Union(_) => p
```
imback82, line 210: `case p: Union => p`

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,243 @@
...
+      }
+      assert(nodes.count(p => p.isInstanceOf[LogicalRelation]) === 2)
+      assert(nodes.count(p => p.isInstanceOf[Union]) === 1)
```
imback82, line 213: nit: `_.isInstanceOf`

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,243 @@
...
+
+      spark.enableHyperspace()
+      val execPlan = spark.sessionState.executePlan(planWithHybridScan).executedPlan
+
+      val execNodes = execPlan collect {
+        case p @ UnionExec(children) =>
+          assert(children.size === 2)
+          // head is always the index plan.
+          assert(Try(children.head.asInstanceOf[WholeStageCodegenExec]).isSuccess)
+          assert(Try(children.last.asInstanceOf[WholeStageCodegenExec]).isSuccess)
```
imback82, line 225: I think you need to update the comment to differentiate the behavior from the previous test.

Btw, you can turn off whole stage code gen if it makes the validation easier (e.g, `BucketedReadSuite.scala` does that)

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,243 @@
...
+      }
+
+      // Make sure there is no shuffle.
+      execPlan.foreach(p => assert(Try(p.asInstanceOf[ShuffleExchangeExec]).isFailure))
+      assert(execNodes.count(p => p.isInstanceOf[UnionExec]) === 1)
+      // 1 of index, 1 of appended file
+      assert(execNodes.count(p => p.isInstanceOf[FileSourceScanExec]) === 2)
+
+      val query2 = df.filter(df("clicks") <= 2000).select(df("query"))
+      checkAnswer(query, query2)
```
imback82, line 240: same here.

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,243 @@
...
+  private val indexConfig2 = IndexConfig("index2", Seq("clicks"), Seq("query"))
+
+  private def copyFirstFile(df: DataFrame): Unit = {
+    val files = df.queryExecution.optimizedPlan.collect {
+      case LogicalRelation(
+          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+          _,
+          _,
+          _) =>
+        location.allFiles
```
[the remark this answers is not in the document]
imback82: I see that you are doing `df.inputFiles.head` in another test. Should you do the same here?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -185,6 +179,49 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+      .parquet(dataPath)
+
+    val df = spark.read.parquet(dataPath)
+    val query = df.filter(df("id") >= 3).select("id", "name")
+    val bucketSpec = BucketSpec(100, Seq("id"), Seq())
+    val shuffled = RuleUtils.transformPlanToShuffleUsingIndexSpec(
+      bucketSpec,
+      query.queryExecution.optimizedPlan)
+
+    assert((shuffled collect {
```
imback82, line 198: nit: `shuffled.collect` to be consistent.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -185,6 +179,49 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+      case RepartitionByExpression(attrs: Seq[Expression], p: Project, numBuckets: Int) =>
```
imback82, line 199: `case RepartitionByExpression(attrs, p: Project, numBuckets)`

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -185,6 +179,49 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+        assert(numBuckets == 100)
+        assert(attrs.size == 1)
+        assert(attrs.head.asInstanceOf[Attribute].name.contains("id"))
+        assert(p.projectList.exists(a => a.name.equals("id")) && p.projectList.exists(a =>
+          a.name.equals("name")))
```
imback82, line 204: nit: `a => a.name` -> `_.name`

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -185,6 +179,49 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+    val bucketSpec2 = BucketSpec(100, Seq("age"), Seq())
+    val query2 = df.filter(df("id") <= 3).select("id", "name")
+    val shuffled2 =
+      RuleUtils.transformPlanToShuffleUsingIndexSpec(
+        bucketSpec2,
+        query2.queryExecution.optimizedPlan)
+    assert((shuffled2 collect {
+      case Project(
+          _,
+          r @ RepartitionByExpression(attrs: Seq[Expression], _: Filter, numBuckets: Int)) =>
```
imback82, line 217: `r` is not used => `RepartitionByExpression(attrs, _: Filter, numBuckets))`

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -185,6 +179,49 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+      case RepartitionByExpression(attrs: Seq[Expression], p: Project, numBuckets: Int) =>
+        assert(numBuckets == 100)
+        assert(attrs.size == 1)
+        assert(attrs.head.asInstanceOf[Attribute].name.contains("id"))
+        assert(p.projectList.exists(a => a.name.equals("id")) && p.projectList.exists(a =>
+          a.name.equals("name")))
+        true
+    }).length == 1)
+
+    val bucketSpec2 = BucketSpec(100, Seq("age"), Seq())
```
imback82, line 208: Can you make a short comment on what you are testing for each scenario?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -185,6 +179,49 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+    val query2 = df.filter(df("id") <= 3).select("id", "name")
+    val shuffled2 =
+      RuleUtils.transformPlanToShuffleUsingIndexSpec(
+        bucketSpec2,
+        query2.queryExecution.optimizedPlan)
+    assert((shuffled2 collect {
```
imback82, line 214: ditto

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
sezruby: Done

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+
+        val updatedOutput =
+          baseOutput.filter(attr => relation.schema.fieldNames.contains(attr.name))
+        baseRelation.copy(relation = relation, output = updatedOutput)
+    }
+
+    if (filesAppended.nonEmpty) {
+      // If there are unhandled appended files, we need to create additional plans
+      // by the following steps:
+      // Step 1) Generate a plan(planForAppended) from the current plan to read
```
imback82, line 268:
```suggestion
      // Step 1) Generate a plan (planForAppended) from the current plan to read
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+      // For more details, see https://github.com/microsoft/hyperspace/issues/150.
+
+      val planForAppended =
+        transformPlanToReadFromAppendedFiles(spark, index.schema, plan, filesAppended)
+      if (useBucketSpec) {
+        // If Bucketing information of the index is used to read the index data, we need to
+        // shuffle the appended data in the same way to correctly merge with bucketed index data.
+
+        // Clear sortColumnNames as BucketUnion does not keep the sort order within a bucket.
+        val bucketSpec = index.bucketSpec.copy(sortColumnNames = Seq())
```
imback82, line 281: nit: we can use `Nil` here. (more consistent in this file)

imback82: Maybe we should just use `numBuckets` in `BucketUnion*`? Are we using any other fields from `BucketSpec`?

Also, can you add an assert in `BucketUnionExec.outputPartitioning` that the number of partitions is same as the number of buckets? I think that's our assumption right?
```
assert(children.head.outputPartitioning.asInstanceOf[HashPartitioning].numPartitions == bucketSpec.numBuckets)
```

imback82: Btw, do we need to check if the bucketed columns are the partitioning expressions?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+            _) =>
+        val curFileSet = location.allFiles
+          .map(f => FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
+        filesAppended =
+          curFileSet.filterNot(index.allSourceFileInfos.contains).map(f => new Path(f.name))
+        // TODO: Hybrid Scan delete support.
+
+        // If BucketSpec of index data isn't used (e.g., in the case of FilterIndex currently),
+        // or the source format is not parquet. We could read appended data from source files
+        // along with the index data.
```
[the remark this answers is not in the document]
imback82: Hmm.. the condition in the comment doesn't match the if and else condition in the code. The comment is saying `if (!useBucketSpec || !isParquetSourceFormat)`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+   * The result will be merged with the plan which is reading index data
+   * by using [[BucketUnion]] or [[Union]].
+   *
+   * @param spark Spark session.
+   * @param indexSchema Index schema used for the output.
+   * @param originalPlan Original plan.
+   * @param filesAppended Appended files to the source relation.
+   * @return Transformed linear logical plan for appended files.
+   */
+  private def transformPlanToReadFromAppendedFiles(
```
imback82, line 309: nit: `transformPlanToReadAppendedFiles`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+            location = newLocation,
+            dataSchema = StructType(indexSchema.filter(baseRelation.schema.contains(_))))(spark)
+        baseRelation.copy(relation = newRelation, output = updatedOutput)
+    }
+    assert(!originalPlan.equals(planForAppended))
+    planForAppended
+  }
+
+  /**
+   * Transform the plan to perform on-the-fly Shuffle the data based on bucketSpec.
```
imback82, line 333: I am a bit confused with "IndexSpec" in the function name. Did you mean `transformPlanToShuffleUsingBucketSpec`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+   *
+   * @param bucketSpec Bucket specification used for Shuffle.
+   * @param plan Plan to be shuffled.
+   * @return Transformed plan by injecting on-the-fly shuffle with given bucket specification.
+   */
+  private[rules] def transformPlanToShuffleUsingIndexSpec(
+      bucketSpec: BucketSpec,
+      plan: LogicalPlan): LogicalPlan = {
+    // Extract top level plan including all required columns for shuffle in its output.
+    // This is located inside this function because of bucketSpec.bucketColumnNames.
```
imback82, line 346: You can pass a tuple to `unapply` and update `transformDown` below accordingly, so this is not a hard requirement. But now that this function is refactored, it can live here (and maybe we can remove this comment as well).

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+          val indexedAttrs = getIndexedAttrs(project, bucketSpec.bucketColumnNames)
+          Some(project, indexedAttrs, true)
+        case filter @ Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
+          val indexedAttrs = getIndexedAttrs(filter, bucketSpec.bucketColumnNames)
+          Some(filter, indexedAttrs, false)
+        case relation @ LogicalRelation(_: HadoopFsRelation, _, _, _) =>
+          val indexedAttrs = getIndexedAttrs(relation, bucketSpec.bucketColumnNames)
+          Some(relation, indexedAttrs, false)
+      }
+      def getIndexedAttrs(
```
imback82, line 363:
```suggestion
      }

      private def getIndexedAttrs(
```

pirz:
LGTM, Thanks @sezruby

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+   */
+  private[rules] def transformPlanToShuffleUsingIndexSpec(
+      bucketSpec: BucketSpec,
+      plan: LogicalPlan): LogicalPlan = {
+    // Extract top level plan including all required columns for shuffle in its output.
+    // This is located inside this function because of bucketSpec.bucketColumnNames.
+    object ExtractTopLevelPlanForShuffle {
+      type returnType = (LogicalPlan, Seq[Option[Attribute]], Boolean)
+      def unapply(plan: LogicalPlan): Option[returnType] = plan match {
+        case project @ Project(_, Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _))) =>
```
imback82, line 350: nit: Shall we make this a little more concise?
```scala
def unapply(plan: LogicalPlan): Option[returnType] = plan match {
  case p @ Project(_, Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _))) =>
    Some(p, getIndexedAttrs(p, bucketSpec.bucketColumnNames), true)
  case p @ Project(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
    Some(p, getIndexedAttrs(p, bucketSpec.bucketColumnNames), true)
  case f @ Filter(_, LogicalRelation(_: HadoopFsRelation, _, _, _)) =>
    Some(f, getIndexedAttrs(f, bucketSpec.bucketColumnNames), false)
  case r @ LogicalRelation(_: HadoopFsRelation, _, _, _) =>
    Some(r, getIndexedAttrs(r, bucketSpec.bucketColumnNames), false)
}
```

You can use currying and bind `bucketSpec.bucketColumnNames` if you want to simplify further.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+    //    Case 2: Shuffle node should come **after** Project and/or Filter node
+    //                Plan: Shuffle => Project => Filter => Relation
+    //
+    // Currently, we only perform on-the-fly shuffle when applying JoinIndexRule.
+    // Therefore, it's always guaranteed that the children nodes have all indexed columns
+    // in their output; Case 1 won't be shown in use cases. The implementation is kept
+    // for future use cases.
+
+    var shuffleInjected = false
+    // Remove sort order because we cannot guarantee the ordering of source files.
```
imback82, line 396: Is this comment still valid?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +121,287 @@ object RuleUtils {
...
+    //    Case 1: Shuffle should be located **before** Project but **after** Filter
+    //                Plan: Project => Shuffle => Filter => Relation
+    // else:
+    //    Case 2: Shuffle node should come **after** Project and/or Filter node
+    //                Plan: Shuffle => Project => Filter => Relation
+    //
+    // Currently, we only perform on-the-fly shuffle when applying JoinIndexRule.
+    // Therefore, it's always guaranteed that the children nodes have all indexed columns
+    // in their output; Case 1 won't be shown in use cases. The implementation is kept
+    // for future use cases.
```
imback82, line 393: I couldn't comment on https://github.com/microsoft/hyperspace/pull/165#discussion_r490674500, but is this now being tested? If it's not covered, then I wouldn't add the case and just throw if this case is hit.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: This behavior is covered by below ` test("Verify the location of injected shuffle for Hybrid Scan.") {`

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -185,6 +179,63 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
     }
   }
 
+  test("Verify the location of injected shuffle for Hybrid Scan.") {
+    val dataPath = systemPath.toString + "/hbtable"
```
imback82, line 183: Can you check if you can utilize `withTempDir`? I just committed.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -185,6 +179,63 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+    }.length == 1)
+
+    // Check if the shuffle node should be injected where all bucket columns
+    // are available as its input.
+    // For example,
+    // Plan: Project ("id", "name") -> Filter ("id") -> Relation
+    // should be transformed:
+    //   Project ("id", "name") -> Shuffle ("age") -> Filter ("id") -> Relation
+    // , NOT:
+    //   Shuffle ("age") -> Project("id", "name") -> Filter ("id") -> Relation
```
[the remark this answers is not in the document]
sezruby: This is just an example for invalid case (`// , NOT:`). The scenario will be handled in `transformPlanToShuffleUsingBucketSpec`.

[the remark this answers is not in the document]
imback82: Ah OK. Misread the comment. Thanks!

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,293 @@
...
+      indexConfig: IndexConfig,
+      appendCnt: Int): Unit = {
+    hyperspace.createIndex(df, indexConfig)
+    val list = df.inputFiles
+    assert(appendCnt < list.length)
+    for (i <- 0 until appendCnt) {
+      val sourcePath = new Path(list(i))
+      val destPath = new Path(list(i) + ".copy")
+      systemPath.getFileSystem(new Configuration).copyToLocalFile(sourcePath, destPath)
+    }
```
imback82, line 56:
```scala
  // Creates an index with given 'df' and 'indexConfig' and copies the first 'appendCnt'
  // number of input files from 'df'.
  def setupIndexAndAppendData(df: DataFrame, indexConfig: IndexConfig, appendCnt: Int): Unit = {
    hyperspace.createIndex(df, indexConfig)
    val inputFiles = df.inputFiles
    assert(appendCnt < inputFiles.length)

    val fs = systemPath.getFileSystem(new Configuration)
    for (i <- 0 until appendCnt) {
      val sourcePath = new Path(inputFiles(i))
      val destPath = new Path(inputFiles(i) + ".copy")
      fs.copyToLocalFile(sourcePath, destPath)
    }
  }
```

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,293 @@
...
+      assert(baseQuery.queryExecution.optimizedPlan.equals(filter.queryExecution.optimizedPlan))
+    }
+
+    withSQLConf("spark.hyperspace.index.hybridscan.enabled" -> "true") {
+      val filter = filterQuery
+      val planWithHybridScan = filter.queryExecution.optimizedPlan
+      assert(!baseQuery.queryExecution.optimizedPlan.equals(planWithHybridScan))
+
+      // Check appended file is added to relation node or not.
+      val nodes = planWithHybridScan collect {
```
imback82, line 117: nit: can we use `.collect`? We use `.collect` in other files consistently.

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -0,0 +1,293 @@
...
+            assert(dataFilters.toString.contains(" <= 2000)"))
+            p
+        }
+        assert(execNodes.count(_.isInstanceOf[UnionExec]) === 1)
+        // 1 of index, 1 of appended file
+        assert(execNodes.count(_.isInstanceOf[FileSourceScanExec]) === 2)
+        // Make sure there is no shuffle.
+        execPlan.foreach(p => assert(!p.isInstanceOf[ShuffleExchangeExec]))
+
+        checkAnswer(baseQuery, filter)
```
imback82, line 289: Test looks much better! 👍

imback82 (commented):
Looks good to me! (pending minor comments)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: Also, as far as I know, bucket spec is meant to be used for the scan node. So it may be more confusing if we print out it in the `BucketUnion`. We can take this up as a separate PR, but I think we need to address this.

sezruby: Ok I revised the comment a bit. But I think the information might help to understand the behavior? Does this look confusing?
```
 +- *(5) Project [o_orderkey#61L, o_orderdate#65, o_shippriority#68]
:        +- *(5) SortMergeJoin [c_custkey#5L], [o_custkey#62L], Inner
:           :- *(1) Project [c_custkey#5L]
:           :  +- *(1) Filter ((isnotnull(c_mktsegment#11) && (c_mktsegment#11 = BUILDING)) && isnotnull(c_custkey#5L))
:           :     +- *(1) FileScan parquet [c_custkey#5L,c_mktsegment#11] Batched: true, Format: Parquet, Location: InMemoryFileIndex[wasb://eunjin-hyperspace-test@taruntestdiag.blob.core.windows.net/indexes/index..., PartitionFilters: [], PushedFilters: [IsNotNull(c_mktsegment), EqualTo(c_mktsegment,BUILDING), IsNotNull(c_custkey)], ReadSchema: struct<c_custkey:bigint,c_mktsegment:string>, SelectedBucketsCount: 200 out of 200
:           +- *(4) Sort [o_custkey#62L ASC NULLS FIRST], false, 0
:              +- BucketUnion 200 buckets, bucket columns: [o_custkey]             <===============
:                 :- *(2) Project [o_orderkey#61L, o_custkey#62L, o_orderdate#65, o_shippriority#68]
:                 :  +- *(2) Filter (((isnotnull(o_orderdate#65) && (o_orderdate#65 < 1995-03-15)) && isnotnull(o_custkey#62L)) && isnotnull(o_orderkey#61L))
:                 :     +- *(2) FileScan parquet [o_custkey#62L,o_orderkey#61L,o_orderdate#65,o_shippriority#68] Batched: true, Format: Parquet, Location: InMemoryFileIndex[wasb://eunjin-hyperspace-test@taruntestdiag.blob.core.windows.net/indexes/index..., PartitionFilters: [], PushedFilters: [IsNotNull(o_orderdate), LessThan(o_orderdate,1995-03-15), IsNotNull(o_custkey), IsNotNull(o_orde..., ReadSchema: struct<o_custkey:bigint,o_orderkey:bigint,o_orderdate:string,o_shippriority:int>, SelectedBucketsCount: 200 out of 200
:                 +- Exchange hashpartitioning(o_custkey#62L, 200)
:                    +- *(3) Project [o_orderkey#61L, o_custkey#62L, o_orderdate#65, o_shippriority#68]
:                       +- *(3) Filter (((isnotnull(o_orderdate#65) && (o_orderdate#65 < 1995-03-15)) && isnotnull(o_custkey#62L)) && isnotnull(o_orderkey#61L))
:                          +- *(3) FileScan parquet [o_custkey#62L,o_orderkey#61L,o_orderdate#65,o_shippriority#68] Batched: true, Format: Parquet, Location: InMemoryFileIndex[wasb://eunjin-hyperspace-test@taruntestdiag.blob.core.windows.net/data/tpch-par..., PartitionFilters: [], PushedFilters: [IsNotNull(o_orderdate), LessThan(o_orderdate,1995-03-15), IsNotNull(o_custkey), IsNotNull(o_orde..., ReadSchema: struct<o_custkey:bigint,o_orderkey:bigint,o_orderdate:string,o_shippriority:int>
```

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -112,79 +112,129 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite {
...
+          verify(optimizedPlan, hybridScanEnabled = false, expectCandidateIndex = false)
+          verify(optimizedPlan, hybridScanEnabled = true, expectCandidateIndex = false)
+        }
       }
     }
   }
 
+  test("Verify the location of injected shuffle for Hybrid Scan.") {
+    withTempDir { tempDir =>
+      val dataPath = tempDir + "/hbtable"
```
imback82, line 183: you can just do `withTempDir { dataPath =>` right? (No need for "/hbtable")

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: I think it may be better if we print out the output partitioning, similar to `Exchange` since we are unioning plans that have the same hash partitioning.

imback82: But new comment sounds better and let's take this up separately.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
sezruby: Yes but it still requires `toString`

imback82: Ah, it fails since the directory is already created. You need `withTempPath` (which is same implemenation as `withTempDir` except that it deletes the directory created). I pushed the changes.

Btw, `parquet()` works on a directory (if you do overwrite).

imback82 (approved):
LGTM, thanks @sezruby! :rocket::rocket::rocket:

imback82:
Merged to master. Great work @sezruby! cc: @rapoth

Issue #169, opened 2020-09-19
Add index metadata update in RefreshIndex for quick mode


pirz:
### Describe the issue
This issue tracks the work for adding index metadata update action in RefreshIndex quick mode.
Index metadata update action is invoked by user thru calling RefreshIndex API with `mode = "quick"`.
Index metadata update, modifies `IndexLogEntry` by:

1. Collecting list of appended and deleted source data files from last version of index and updating `appendedFiles` and `deletedFiles` lists under source content.
2. Updating list of source data files. (Related: Issue #198)
3. Updating index signature according to latest source data files.

This is needed as part of:

- Add support for deletes in RefreshIndex for quick mode. (issue #134)  
- Add support for appends in RefreshIndex for quick mode

### To Reproduce
1. User creates an index on some data.
2. User deletes some data files or partitions from source data files used in step (1) and appends some new data files.
3. User should be able to refresh the index metadata by updating index's source content and its signature according to the status of latest source data file from step (2). 

### Environment 
All.

sezruby:
Closed by mistake and #238 will close this issue.

Issue #134, opened 2020-08-21
Add support for deletes in RefreshIndex for quick mode


pirz:
### Describe the issue
This is the parent issue to track the work of adding support for delete in quick mode.

**User Scenario.**
1. User has a dataset 
2. User uses Hyperspace to create an index 
3. User deletes an existing partition OR removes files from an existing partition
4. **Metadata Update:** User calls refresh index API in quick mode to update Index Metadata
5. **Enforce Delete during read:** Index should become usable during query time by excluding the deleted records when index is used.

During index refresh in quick mode, we do not fix the index files, but we get the list of files/partitions that were deleted (from Step 3) and write a new log entry with this diff. 
When the user now queries the data, we would get the latest state from the log and add a filter node on top of index scan to exclude index entries coming from deleted source data files.

**Related issues.**
- [x] Add Lineage to index records (Issue #104)
- [ ] Add index metadata update in RefreshIndex for quick mode (Issue #169)
- [ ] Modify Hyperspace rules to leverage an index with deleted source data files after RefreshIndex in quick mode  (Issue #174)

### To Reproduce
N/A

### Expected behavior
The scenario described above should work correctly.

### Environment 
All.

PR #170, opened 2020-09-19
Update index log entry for enforce delete during read time

pirz:
This PR adds changes to index metadata for capturing list of source data files deleted or appended and updated index fingerprint accordingly.
List of deleted files are needed as part of adding support for enforcing delete during read time.

- **Tracking Issue**: Issue #169 
 - **Parent Issue**: Issue #134 

This PR adds changes for updating index metadata once some data files are deleted from or appended to an index source data files.
This is done by: 
1. Extending `IndexLogEntry` structure to save:
   - a list of deleted source data files, called `deleted`, and
   - a list of appended source data files, called `appended`.
2. Adding a new refresh action for creating a newer version of an existing index, by updating index metadata as:
     - Detect deleted and appended source data files.
     - Add deleted source data files to `deleted` and appended source data files to `appended` in index metadata. 
     - Update index fingerprint according to latest source data files. (This is required for Hyperspace to correctly matches index with a query written on latest source data files).

`deleted` property is used to enforce delete during query time. Once index is leveraged. Index records coming from already deleted source data files, listed under `deleted`, are excluded from contributing to query results. This is done via PR #175. 

Currently, this feature is protected under a Spark configuration flag: `spark.hyperspace.index.refresh.source.content.enabled` and is disabled by default.

Yes, this PR modifies index metadata and extends `IndexLogEntry` structure by adding a new fields `deleted: Seq[String]` and `appended: Seq[String]` under:
`IndexLogEntry.source.plan.properties.relations.data.properties` which captures a list of source data files which are deleted and added to index's source data files.

**Old experience:**

2. User enables Hyperspace and issues a query. Hyperspace is able to use the index.
3. User deletes some files from the original data and/or adds some new files under "/path/to/dataset/". 
4. User issues a query but Hyperspace detects data change and decides to disable index usage.
5. User invokes refresh to update the index.
6. Hyperspace does a full index rebuild.
7. Now, if a query is issued on latest source data files, Hyperspace can leverage index.

**New experience:**
Steps 1 - 4 remain the same. 
- If user disabled `spark.hyperspace.index.refresh.source.content.enabled` then Hyperspace experience remains similar to 5 and 6 above.
- If user enables `spark.hyperspace.index.refresh.source.content.enabled` and calls refresh then:
  1. Hyperspace detects the deleted and appended files and computes index signature according to latest dataset files. It updates index metadata by adding deleted and appended files and updating index signature.
  2. User can now issues queries and Hyperspace will use the index.

**Impact on `IndexLogEntry` content**
Example of IndexLogEntry before this change:
```
{
  "name" : "filterIndex",
  "derivedDataset" : {...},
  "content" : {...},
  "source" : {
    "plan" : {
      "properties" : {
        "relations" : [ {
          "rootPaths" : [ "file:/C:/..." ],
          "data" : {
            "properties" : {
              "content" : {
                "root" : {
                  "name" : "file:/C:/",
                  "files" : [ ],
                  "subDirs" : [ {...} ]
                },
                "fingerprint" : {...}
              }
            },
            "kind" : "HDFS"
          },
          "dataSchemaJson" : "...",
          "fileFormat" : "parquet",
          "options" : { }
        } ],
        "rawPlan" : null,
        "sql" : null,
        "fingerprint" : {...}
      },
      "kind" : "Spark"
    }
  },
  "extra" : { },
  "version" : "0.1",
  "id" : 3,
  "state" : "ACTIVE",
  "timestamp" : ...,
  "enabled" : true
}
```

New IndexLogEntry example (after this change):
(`deleted` and `appended` are added under `source.plan.properties.relations.data.properties`). 
```
{
  "name" : "filterIndex",
  "derivedDataset" : {...},
  "content" : {...},
  "source" : {
    "plan" : {
      "properties" : {
        "relations" : [ {
          "rootPaths" : [ "file:/C:/..." ],
          "data" : {
            "properties" : {
              "content" : {
                "root" : {
                  "name" : "file:/C:/",
                  "files" : [ ],
                  "subDirs" : [ {...} ]
                },
                "fingerprint" : {...}
              },
                "deleted" : [ "file:/C:/.../part-00000-8fc.parquet" ],
                "appended" : [ "file:/C:/.../part-00000-9ac.parquet" ]
            },
            "kind" : "HDFS"
          },
          "dataSchemaJson" : "...",
          "fileFormat" : "parquet",
          "options" : { }
        } ],
        "rawPlan" : null,
        "sql" : null,
        "fingerprint" : {...}
      },
      "kind" : "Spark"
    }
  },
  "extra" : { },
  "version" : "0.1",
  "id" : 3,
  "state" : "ACTIVE",
  "timestamp" : ...,
  "enabled" : true
}
```

New test cases added under `RefreshIndexTests.scala` and `E2EHyperspaceRulesTests.scala`.


src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
```diff
@@ -0,0 +1,80 @@
...
+
+package com.microsoft.hyperspace.actions
+
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.SparkSession
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index.{Content, IndexDataManager, IndexLogManager}
+
+private[actions] abstract class RefreshDeleteActionBase(
```
pirz, line 25: **To Reviewers:**
This class is a simple code refactor. `validate` and `deletedFiles` defs are copied from `RefreshDeleteAction` class here as they are shared between `RefreshDeleteAction` and `DeleteOnReadAction` classes.

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteAction.scala
```diff
@@ -44,30 +42,13 @@ class RefreshDeleteAction(
     spark: SparkSession,
     logManager: IndexLogManager,
     dataManager: IndexDataManager)
-    extends RefreshActionBase(spark, logManager, dataManager)
+    extends RefreshDeleteActionBase(spark, logManager, dataManager)
     with Logging {
 
   final override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
     RefreshDeleteActionEvent(appInfo, logEntry.asInstanceOf[IndexLogEntry], message)
   }
```
pirz, line 51: `validate` is moved to (new) class `RefreshDeleteActionBase`.

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteAction.scala
```diff
@@ -89,37 +70,4 @@ class RefreshDeleteAction(
       previousIndexLogEntry.numBuckets,
       indexConfig.indexedColumns)
   }
-
-  /**
-   * Compare list of source data files from previous IndexLogEntry to list
-   * of current source data files, validate fileInfo for existing files and
-   * identify deleted source data files.
-   */
-  private lazy val deletedFiles: Seq[String] = {
```
pirz, line 98: **To Reviewers:** `deletedFiles` is moved to (new) class `RefreshDeleteActionBase`.

imback82:
This PR is marked as WIP and I was requested to review. Usually, WIP means "not ready for review" yet.

I will mark this PR as draft, and please ping me back when you convert this back to PR.

pirz:
> This PR is marked as WIP and I was requested to review. Usually, WIP means "not ready for review" yet.
> 
> I will mark this PR as draft, and please ping me back when you convert this back to a regular PR (click "Ready for review").

@imback82 It is no longer a draft, ready to be reviewed.

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
```diff
@@ -0,0 +1,82 @@
...
+  override def validate(): Unit = {
+    super.validate()
+    if (!previousIndexLogEntry.hasLineageColumn(spark)) {
+      throw HyperspaceException(
+        "Index refresh (to handle deleted source data) is " +
+          "only supported on an index with lineage.")
+    }
+
+    if (deletedFiles.isEmpty || deletedFiles.toSet.equals(
+          previousIndexLogEntry.source.plan.properties.
```
pirz, line 44: **To Reviewers:** This part of validate is changed (it now compares deleted files to the list of already excluded files).

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
```diff
@@ -0,0 +1,82 @@
...
+            relations.head.data.properties.excluded.toSet)) {
+      throw HyperspaceException("Refresh aborted as no deleted source data file found.")
```
pirz, line 46: **To Reviewers:** This part of validate is changed (it now compares deleted files to the list of already excluded files).

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
```diff
@@ -0,0 +1,74 @@
...
+package com.microsoft.hyperspace.actions
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.SparkSession
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.telemetry.{AppInfo, DeleteOnReadActionEvent, HyperspaceEvent}
+
+class DeleteOnReadAction(
```
sezruby, line 26: Could you some comments for this class like RefreshDeleteAction?
Thanks!

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
sezruby, line 44: nit: `previousIndexLogEntry.relations.head.data.properties.excluded`?

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
sezruby: What if a user want "optimize" refresh after "quick"(OnRead) refresh and no data change?

pirz: The scenario you pointed to should be addressed by a call to `optimize`, not `refresh`.
`optimize` and `refresh` are two different APIs and they should have different `validate` checks (at least based on current design). The `validate` function here is used with `refresh` when dealing with deleted files and that is why at least one new deleted file should be found comparing to previous case.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
pirz: Sure, done.

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
pirz: yes, as it has been used in multiple places I added a new function `excludedFiles` to `IndexLogEntry` to further simplify the calls.

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
sezruby: Sorry for the lack of clarity. I meant RefreshDeleteAction after DeleteOnReadAction without data change.
In this case, `validate` of `RefreshDeleteAction ` will be aborted because of `deletedFiles.toSet.equals(previousIndexLogEntry.excludedFiles.toSet)`?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -450,7 +450,64 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
 
       // Verify index usage on latest version of index (v=1) after refresh.
       verifyIndexUsage(query2, getIndexFilesPath(indexConfig.indexName, 1))
-      FileUtils.delete(dataPath)
+      FileUtils.delete(new Path(location))
```
sezruby, line 453: Could you move this to `after()` or use a different path for "location" in the below test?
If this test failed, test data might affect the below test.

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
pirz: I see, that is a fair scenario. Thanks for bringing it up.
I changed the code by:
1. Moved `excluded` files check of `validate` to `DeleteOnRead` class.
2. Added a new test case under `RefreshIndexTests` to validate this scenario.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
pirz: Sure, I changed the path used in tests to avoid any side-effect due to failures.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.spark.sql.SparkSession
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.telemetry.{AppInfo, DeleteOnReadActionEvent, HyperspaceEvent}
+
+/**
+ * Refresh index by updating list of excluded source data files and index signature
+ * in index metadata.
+ * Note this Refresh Action only fixes an index metadata w.r.t deleted source data files
```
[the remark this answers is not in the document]
imback82: I also see log messages like "Refresh aborted", "Refresh index is updating index metadata", "during index refresh", etc.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
```diff
@@ -0,0 +1,105 @@
...
+    extends RefreshDeleteActionBase(spark, logManager, dataManager)
+    with Logging {
+
+  final override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
+    DeleteOnReadActionEvent(appInfo, logEntry.asInstanceOf[IndexLogEntry], message)
+  }
+
+  override def validate(): Unit = {
+    super.validate()
+    if (deletedFiles.toSet.equals(previousIndexLogEntry.excludedFiles.toSet)) {
```
imback82, line 55: You can create `private lazy val newExludedFiles: Seq[String] = deletedFiles diff previousIndexLogEntry.excludedFiles` and use it here (`if (newExludedFiles.isEmpty)`) and in line 103?

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteAction.scala
imback82: Cool, thanks for letting us know.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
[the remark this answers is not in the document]
imback82: After going thru the code once, it seems that this should be "RefreshSomethingAction" than "DeleteOnReadAction".

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -57,4 +57,7 @@ object IndexConstants {
 
   val REFRESH_DELETE_ENABLED = "spark.hyperspace.index.refresh.delete.enabled"
   val REFRESH_DELETE_ENABLED_DEFAULT = "false"
+
+  val ENFORCE_DELETE_ON_READ_ENABLED = "spark.hyperspace.index.enforce.delete.on.read.enabled"
```
imback82, line 61: btw, you don't have to have "." for each word. you can do "deleteOnRead.enabled", for example.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -324,7 +324,7 @@ case class Hdfs(properties: Hdfs.Properties) {
   val kind = "HDFS"
 }
 object Hdfs {
-  case class Properties(content: Content)
+  case class Properties(content: Content, excluded: Seq[String] = Seq[String]())
```
imback82, line 327: `= Nil`?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -403,6 +403,11 @@ case class IndexLogEntry(
     sourcePlanSignatures.head
   }
 
+  def excludedFiles: Seq[String] = {
+    // Only one relation is currently supported (See relations).
```
imback82, line 407: We can remove this comment. The `assert` has been moved to `relations` to remove duplicate comments like this in the previous PR - I forgot which one).

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -450,7 +450,64 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+    }
+  }
+
+  test(
+    "Validate index usage for enforce delete on read " +
+      "after some source data file is deleted.") {
```
imback82, line 459: nit: just put it in one line?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -450,7 +450,64 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+      spark.enableHyperspace()
+      val planRootPaths = getAllRootPaths(query2().queryExecution.optimizedPlan)
+      spark.disableHyperspace()
+      assert(planRootPaths.equals(Seq(PathUtils.makeAbsolute(location))))
+
+      // Refresh index to update its metadata.
+      hyperspace.refreshIndex(indexConfig.indexName)
+
+      // Verify index is used after refresh fixes fingerprint in index metadata.
+      // TODO Replace check with verifyIndexUsage() once EnforceDeleteOnRead covers query path.
```
imback82, line 504:
```suggestion
      // TODO: Replace check with verifyIndexUsage() once EnforceDeleteOnRead covers query path.
```

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -125,7 +125,8 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
         |                  "kind" : "NoOp",
         |                  "properties" : { }
         |                }
-        |              }
+        |              },
+        |              "excluded" : []
```
imback82, line 129: Can we test with a non empty value?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -208,6 +208,155 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
...
+        assert(originalIndex.length == 1)
+        assert(originalIndex.head.excludedFiles.isEmpty)
+
+        // Delete one source data file.
+        val deletedFile1 = deleteDataFile(nonPartitionedDataPath)
+
+        // Refresh index and validate updated IndexLogEntry.
+        hyperspace.refreshIndex(indexConfig.indexName)
+        val refreshedIndex1 = ixManager.getIndexes()
+        assert(refreshedIndex1.head.excludedFiles.equals(Seq(deletedFile1.toString)))
```
imback82, line 287: Can we remove "1" in the variable name?

imback82 (commented):
Generally looking fine to me. One major comment is around the naming `DeleteOnReadAction`, since it doesn't seem to be conveying what the action is doing (refreshing the index log entry only on deleted files). Any thoughts? @apoorvedave1 @rapoth @sezruby

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -450,7 +450,64 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
+      spark.enableHyperspace()
+      assert(
+        queryPlanHasExpectedRootPaths(
+          query2().queryExecution.optimizedPlan,
+          getIndexFilesPath(indexConfig.indexName)))
+      FileUtils.delete(new Path(location))
```
imback82, line 510: Can you utilize `withTempDir`?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -208,6 +208,155 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
     }
   }
 
+  test(
+    "Validate refresh index (to handle deletes from the source data) " +
+      "for enforce delete on read.") {
+    // Save test data non-partitioned.
+    SampleData.save(
+      spark,
+      nonPartitionedDataPath,
```
imback82, line 217: We should be able to utilize `withTempDir` here instead as well.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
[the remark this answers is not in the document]
pirz: I changed it to `RefreshForDeleteOnReadAction` to capture the fact that it is a Refresh action.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
```diff
@@ -0,0 +1,98 @@
...
+ * Refresh index by updating list of excluded source data files and index signature
+ * in index metadata.
+ * Note this Refresh Action only fixes an index metadata w.r.t deleted source data files
+ * and does not consider new source data files (if any).
+ * If some original source data file(s) are removed between previous version of index and
+ * now, this Action refreshes index as follows:
+ * 1. Deleted source data files are identified.
+ * 2. New index fingerprint is computed w.r.t latest source data files.
+ * 3. IndexLogEntry is updated by modifying list of excluded source data files and
+ * index fingerprint, computed in above steps.
```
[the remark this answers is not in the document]
pirz: Done.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
```diff
@@ -0,0 +1,98 @@
...
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.telemetry.{AppInfo, DeleteOnReadActionEvent, HyperspaceEvent}
+
+/**
+ * Refresh index by updating list of excluded source data files and index signature
+ * in index metadata.
+ * Note this Refresh Action only fixes an index metadata w.r.t deleted source data files
+ * and does not consider new source data files (if any).
```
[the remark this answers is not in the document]
pirz: Oh yes, good point, thnx, I fixed the comment.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
pirz: Sure, done.

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
pirz: Sure, I added a sample deleted file.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
pirz: Actually, we already had `withTempDir` under `HyperspaceSuite` and it served the purpose; I just switched the code to use that.

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
pirz: Actually, we already had `withTempDir` under `HyperspaceSuite` and it served the purpose; I just switched the code to use that.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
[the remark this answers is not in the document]
imback82: Hmm, should we just call this `RefreshLogEntryAction` and update both deleted and appended files? (I think appended can be leveraged during read with @sezruby's recent PR?)

To me, having an transaction log action that has a dependency on the "read" behavior does not sound right.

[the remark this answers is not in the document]
pirz: @imback82 Do you mean beside `excluded`, in this PR we also add a new property to `IndexLogEntry` as `appended` and populate it?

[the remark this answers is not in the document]
imback82: Yes, both `excluded` and `appended` can be utilized right?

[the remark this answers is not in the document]
pirz: Yes, `excluded` is utilized for delete - `appended` "could" be utilized for HybridScan, specifically for below code.
This should work as signature always makes sure that we pick an index if it can replace a table scan done on the current/latest set of files. (@sezruby can comment on this).

```
val curFileSet = location.allFiles
          .map(f => FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
val filesAppended =
          curFileSet.filterNot(index.allSourceFileInfos.contains).map(f => new Path(f.name))
```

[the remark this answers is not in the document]
pirz: I renamed `RefreshForDeleteOnReadAction` -> `RefreshLogEntryAction` and added `appended` property to IndexLogEntry.
However, we should now rename `RefreshDeleteActionBase` to sth else as one of its children: `RefreshLogEntryAction` is no longer just addressing deleted files. I could not come-up with a better name for `RefreshDeleteActionBase`, any suggestions?

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -57,4 +57,7 @@ object IndexConstants {
 
   val REFRESH_DELETE_ENABLED = "spark.hyperspace.index.refresh.delete.enabled"
   val REFRESH_DELETE_ENABLED_DEFAULT = "false"
+
+  val REFRESH_LOGENTRY_ACTION_ENABLED = "spark.hyperspace.index.refresh.logentry.enabled"
```
rapoth, line 61: @pirz Could you please explain what this config corresponds to?

Also, the PR description has a different config `spark.hyperspace.index.enforce.delete.on.read.enabled`. Could you update as appropriate? Thank you!

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -419,49 +419,104 @@ class E2EHyperspaceRulesTests extends HyperspaceSuite with SQLHelper {
...
-      // Refresh the index to remove deleted source data file records from index.
-      hyperspace.refreshIndex(indexConfig.indexName)
-
-      // Verify index usage on latest version of index (v=1) after refresh.
-      verifyIndexUsage(query2, getIndexFilesPath(indexConfig.indexName, 1))
-      FileUtils.delete(dataPath)
+  test("Validate index usage for enforce delete on read after some source data file is deleted.") {
+    withSQLConf(
+      IndexConstants.INDEX_LINEAGE_ENABLED -> "true",
+      IndexConstants.REFRESH_LOGENTRY_ACTION_ENABLED -> "true") {
```
rapoth, line 471: As I read this code, `REFRESH_LOGENTRY_ACTION_ENABLED` does not seem very informative. Was there any discussion around this?

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -125,7 +125,9 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
         |                  "kind" : "NoOp",
         |                  "properties" : { }
         |                }
-        |              }
+        |              },
+        |              "excluded" : ["file:/rootpath/f1"],
+        |              "appended" : ["file:/rootpath/f3"]
```
rapoth, line 130: nit: PR description talks about `excluded` but looks like you're handling `appended` as well? Can you update the text throughout for consistency?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -118,7 +118,7 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
 
       val ex = intercept[HyperspaceException](hyperspace.refreshIndex(indexConfig.indexName))
       assert(
-        ex.getMessage.contains(s"Index refresh (to handle deleted source data) is " +
+        ex.getMessage.contains(s"Index refresh (to handle deleted or appended source data) is " +
           "only supported on an index with lineage."))
```
rapoth, line 122: Sorry, I'm confused (again!). Why do we need lineage for handling appends?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -203,8 +202,186 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
...
+        // Append another new source data file.
+        val appendedFile2 = new Path(deletedFile2.getParent, "newFile2")
+        FileUtils.createFile(
+          appendedFile2.getFileSystem(new Configuration),
+          appendedFile2,
+          "I am some random content for yet another new file :).")
+
+        // Refresh index and validate updated IndexLogEntry.
+        // `excluded` files should contain both deleted source data files.
+        // `appened` files should contain both appended source data files.
```
rapoth, line 262: nit:

```suggestion
        // `appended` files should contain both the appended source data files.
```

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -203,8 +202,186 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
...
+
+        // Append another new source data file.
+        val appendedFile2 = new Path(deletedFile2.getParent, "newFile2")
+        FileUtils.createFile(
+          appendedFile2.getFileSystem(new Configuration),
+          appendedFile2,
+          "I am some random content for yet another new file :).")
+
+        // Refresh index and validate updated IndexLogEntry.
+        // `excluded` files should contain both deleted source data files.
```
rapoth, line 261: nit: 
```suggestion
        // `excluded` files should contain both the deleted source data files.
```

rapoth (commented):
Couple of minor comments. Thanks!

rapoth:
@pirz In your PR description, it'd be nice to also capture the details surrounding the user experience that this PR is enabling? For instance, user enables this flag and this is what happens.

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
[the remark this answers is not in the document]
sezruby: We could utilize the current code base - check if the index entry has `appended` or `excluded` files and use them instead filesAppended or filesDeleted in `transformPlanToUseHybridScan`(https://github.com/microsoft/hyperspace/pull/171/files#diff-1ac09a3b18348fa9bf89fd4c42f08390R237) - then #175 might be unnecessary?
To do this, we need to make sure that there is no diff in the source file list at query time. 

BTW I think, when hybrid scan config is enabled: 
1.  check `signatureValid` first in `getCandidateIndex`
2. if there's no candidate, try to get candidate index with `isHybridScanCandidate`

But I'm not sure this could be best - for example, it's might be more efficient to avoid full-shuffling of index data. So an index pair **with proper bucketing and hybrid scan** _might be more efficient_ than full index pair **without proper bucketing (no hybrid scan)**. I guess rank algorithm for hybrid scan might be more complex than I thought in #164 ..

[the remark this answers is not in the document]
imback82: > We could utilize the current code base - check if the index entry has `appended` or `excluded` files and use them instead filesAppended or filesDeleted in `transformPlanToUseHybridScan`(https://github.com/microsoft/hyperspace/pull/171/files#diff-1ac09a3b18348fa9bf89fd4c42f08390R237)

Yes, this was exactly what I was thinking.

> To do this, we need to make sure that there is no diff in the source file list at query time.

I think we can utilize the signature this PR computes; it included new files as well.

BTW, should we call these fields as `added` (maybe `appended`) and `deleted` instead? WDYT? cc: @rapoth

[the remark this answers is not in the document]
rapoth: +1 to both code reuse and appended/deleted.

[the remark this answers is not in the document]
rapoth: @pirz @imback82 @sezruby I thought more about this. It would be good to not have duplicate code paths for delete. For instance, hybrid scan is already taking care of applying the delete predicates appropriately so we should not have a secondary path do this. Same thing with appends (if applicable). This will save us from future maintenance hell :)

CC: @apoorvedave1

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
pirz: Yes, this is based on some other comment in this PR. @imback82 suggested to capture both excluded and appended source data files and change the action name to RefreshLogEntryAction. Here is the comment: https://github.com/microsoft/hyperspace/pull/170#discussion_r497039610

imback82: Looks like the comment is around the action name, not the config?

rapoth: Yes, not the config. To me `REFRESH_LOGENTRY_ACTION` is not really clear (but it could just be me).

imback82: btw, isn't this config temporary until we have a public API?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
pirz: We dont need lineage for appended. Lineage is needed for deleted source data files. However, both `excluded` and `appended` files are updated simultaneously and under same call to `RefreshLogEntryAction`. That is why lineage column check happens for such an action call and this test checks that.

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
pirz: PR description is updated.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
pirz: @imback82 So far we had meaningful mapping between configs and corresponding action they trigger. Do you suggest using a config name s.a. `REFRESH_FOR_DELETE_ON_READ` to trigger an Action that updates Index metadata and adds both `excluded` and `appended` files?

imback82: How about REFRESH_LOG_ENTRY_ONLY or something along that line without ACTION? I think the confusion from @rapoth may be the fact that action was included, which is internal?

imback82: > Yes, not the config. To me `REFRESH_LOGENTRY_ACTION` is not really clear (but it could just be me).

Oh wait, are you saying the action name is not clear? Any other suggestion then?

src/main/scala/com/microsoft/hyperspace/actions/DeleteOnReadAction.scala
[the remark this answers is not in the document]
pirz: @rapoth "support for delete on read" and "support for hybrid scan" are triggered via separate configs (at least for now). I assume if we combine them then one can only benefit from "delete" on read if "hybrid scan" is enabled for the index and they may bring some other stuff. I believe we originally had a discussion about it (not adding "delete on read" and relying on Hybridscan, but the decision was these are different actions and they are not necessarily targeting same customers).
I prefer to keep them separate and let the PRs move forward. If we wanna combine them, we can do that once we drop (temp) feature flags and finalize the API calls.  @rapoth  What do you think?

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
rapoth: I was thinking something like: `RefreshSourceFilesAction` or `RefreshSourceFilesListAction` or `RefreshSourceMetadataAction` (assuming I understood this right). I just feel `RefreshLogEntryAction` is too generic and not conveying what that thing is doing. Does any of this make sense?

pirz: what about this?
for action: `RefreshSourceContentAction`
for config: `REFRESH_SOURCE_CONTENT`

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
rapoth: Thanks, that's what I thought. In that case, is there a way this message can be improved? If let's say something failed when handling appended source data (hypothetically), saying that we do not support this over an index with no lineage doesn't seem proper.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
rapoth: @pirz Let me think. In the mean time, can you update the PR appropriately please? (assume these are the renames we will do)

rapoth:
> @pirz In your PR description, it'd be nice to also capture the details surrounding the user experience that this PR is enabling? For instance, user enables this flag and this is what happens.

@pirz This is the specific comment I was interested in seeing being reflected in the PR. Please refer to your previous PR where we wrote down step by step user actions and how this PR fits in.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
pirz: @rapoth - Sure, can we just finalize what we wanna rename `RefreshDeleteActionBase` (this is the "parent" class)?
 It would be great to come to a consensus as each rename could affect many comments and code pieces and I would rather do it once.

pirz:
> > @pirz In your PR description, it'd be nice to also capture the details surrounding the user experience that this PR is enabling? For instance, user enables this flag and this is what happens.
> 
> @pirz This is the specific comment I was interested in seeing being reflected in the PR. Please refer to your previous PR where we wrote down step by step user actions and how this PR fits in.

Sure, I just want to finalize the rename decisions before updating the text. I have had done it multiple times so far and then another name change came. Lets first make everybody happy with the name of properties, config and classes and then I will update/extend PR description and revise code comments. I am specifically blocked on what you wanna call `RefreshDeleteActionBase` (the parent class).

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteActionBase.scala
```diff
@@ -0,0 +1,75 @@
...
+
+package com.microsoft.hyperspace.actions
+
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.SparkSession
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index.{Content, IndexDataManager, IndexLogManager}
+
+private[actions] abstract class RefreshDeleteActionBase(
```
pirz, line 25: **To Reviewers** This class needs a rename, as its children are no longer dealing only with delete-related items. Any suggestion?

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
pirz: We have only one config and that is: "spark.hyperspace.index.refresh.source.content.enabled".
PR description is updated to reflect latest changes.

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
pirz: I changed the exception msg to:
`Index refresh to update source content in index metadata is only supported on an index with lineage.`.
This still has an implicit restriction that index has to have lineage so we one can trigger an action to update index metadata.
The reason for that is currently "delete on read" is the only scenario that is relying/using such an index metadata update.
We can lift this restriction and distinguish between a case where we only have "appended" files once the usage of that update is added.

pirz:
> > @pirz In your PR description, it'd be nice to also capture the details surrounding the user experience that this PR is enabling? For instance, user enables this flag and this is what happens.
> 
> @pirz This is the specific comment I was interested in seeing being reflected in the PR. Please refer to your previous PR where we wrote down step by step user actions and how this PR fits in.

PR description is now updated accordingly. Please kindly take a look.

rapoth:
@pirz I spent some time thinking about this. What are your thoughts on this?

  1. We change the PR title to say: `Update index meta-data only for enabling Smart Refresh` and disconnect this PR from the work pertaining to delete (we will still leverage this PR but this PR is more general IMHO). We handle capturing both deleted and appended files in the index meta-data as it's already implemented in this PR.
  2. Rename the classes appropriately to remove the occurrence of the word `Delete` in any of the Refresh Actions. We should call these along the lines of `RefreshMetadataOnlyAction` if need be. @imback82 can help propose the correct names.
  3. For your next PR on enforce-delete-on-read, there will be no special explicit flags. Let us aim on reusing the code paths from Hybrid scan (if we do it right, maybe we don't need another PR or we need a small PR to make this happen). Everything will be implicit and will be under `mode='smart'` or `mode='quick'`. Therefore, I propose the following:
     1. When the user calls `refreshIndex(mode='quick')`, we update the meta-data accordingly. When the user runs a query, we can then enforce-delete-on-read automatically. It follows that we should get rid of this flag `spark.hyperspace.index.refresh.source.content.enabled`. If the user did not call this `refreshIndex(mode='quick')`, we can disable index usage. 
     2. When the user calls `refreshIndex(mode='smart')`, we go rewrite portions of the index - basically, the previous PR. If the user did not call this `refreshIndex(mode='smart')`, we can disable index usage. 

Please let me know if this makes sense. To complete the loop, for append, here's what would happen:

   1. When the user calls `refreshIndex(mode='quick')`, we update the meta-data accordingly **AND** start an incremental index job which indexes the new data. When the user runs a query, we can reuse the new incrementally indexed data. If the user did not call this `refreshIndex(mode='quick')`, we can disable index usage. 
   2. When the user calls `refreshIndex(mode='smart')`, we update the meta-data accordingly **AND** start an incremental index job which indexes the new data + optimize existing small files. If the user did not call this `refreshIndex(mode='smart')`, we can disable index usage. 

CC: @imback82 @sezruby @apoorvedave1

pirz:
> * We change the PR title to say: `Update index meta-data only for enabling Smart Refresh` and disconnect this PR from the work pertaining to delete 

I assume you mean ".. for enabling Quick refresh" as metadata changes are gonna be used in quick mode.

> * Rename the classes appropriately to remove the occurrence of the word `Delete` in any of the Refresh Actions. We should call these along the lines of `RefreshMetadataOnlyAction` if need be. @imback82 can help propose the correct names.

I assume this does *not* include `RefreshDeleteAction` (which is already in master), as that is a standalone class and deals fully with delete (will be used in smart mode). 

> * For your next PR on enforce-delete-on-read, there will be no special explicit flags. Let us aim on reusing the code paths from Hybrid scan (if we do it right, maybe we don't need another PR or we need a small PR to make this happen). Everything will be implicit and will be under `mode='smart'` or `mode='quick'`. Therefore, I propose the following:

Sure, but can you clarify how that consolidated code path will be used? (Sorry, I am a bit confused for this part).
With no flags, should it rely on `mode`. I assume `mode` will not be introduced in this PR, then the next PR will have dependency on both this PR (for metadata changes) and some other PR which introduces `mode` and API changes.

sezruby:
@rapoth There is no use of "appended" file list in the description. 

If we always build the incremental index, it's not necessary to keep the appended file list. 

I think we could have another option (for better hybrid scan) to use both list - refreshIndex(mode='metadata')
 (or 'metadataOnly', 'signature', 'forHybrid'..?)
In this way, the user can clearly see which index could be reused at query time. (in case hybrid scan config is disabled)

And for naming, I prefer
"RefreshSourceFilesAction"
and
"MetadataUpdateActionBase"

rapoth:
> @rapoth There is no use of "appended" file list in the description.
> 
> If we always build the incremental index, it's not necessary to keep the appended file list.
> 
> I think we could have another option (for better hybrid scan) to use both list - refreshIndex(mode='metadata')
> (or 'metadataOnly', 'signature', 'forHybrid'..?)
> In this way, the user can clearly see which index could be reused at query time. (in case hybrid scan config is disabled)
> With this, we could suggest this until we optimize the rank algorithm properly. WDYT??
> 
> And for naming, I prefer
> "RefreshSourceFilesAction"
> and
> "MetadataUpdateActionBase"

Yes, I requested @pirz to update the description appropriately since the code is out-of-sync. Long story short, appended files is not going to be useful for incremental index for appends. It will only be useful for hybrid scan - is this correct?

If the list of appended files goes in as part of this PR, can Hybrid scan utilize it? I like `refreshIndex(mode='metadata')` - @imback82 / @pirz ?

@sezruby Will this `mode='metadata'` be in addition to `smart` and `quick`?

rapoth:
> Sure @rapoth - Thanks. Here are some questions:
> 
> > * We change the PR title to say: `Update index meta-data only for enabling Smart Refresh` and disconnect this PR from the work pertaining to delete 
> 
> I assume you mean ".. for enabling Quick refresh" as metadata changes are gonna be used in quick mode.
> 

Yes, sorry my bad.

> > * Rename the classes appropriately to remove the occurrence of the word `Delete` in any of the Refresh Actions. We should call these along the lines of `RefreshMetadataOnlyAction` if need be. @imback82 can help propose the correct names.
> 
> I assume this does *not* include `RefreshDeleteAction` (which is already in master), as that is a standalone class and deals fully with delete (will be used in smart mode). 

Yep - let's wait for @imback82 to make the final recommendation on the class names and hierarchies.

> 
> > * For your next PR on enforce-delete-on-read, there will be no special explicit flags. Let us aim on reusing the code paths from Hybrid scan (if we do it right, maybe we don't need another PR or we need a small PR to make this happen). Everything will be implicit and will be under `mode='smart'` or `mode='quick'`. Therefore, I propose the following:
> 
> Sure, but can you clarify how that consolidated code path will be used? (Sorry, I am a bit confused for this part).

I think you and @sezruby should sync up to see how we can reuse that portion of the hybrid scan code path. 

> With no flags, should it rely on `mode`. I assume `mode` will not be introduced in this PR, then the next PR will have dependency on both this PR (for metadata changes) and some other PR which introduces `mode` and API changes. 
> 

Yes, that's right.

rapoth:
> 
> And for naming, I prefer
> "RefreshSourceFilesAction"
> and
> "MetadataUpdateActionBase"

Sorry @sezruby, I missed this. @imback82 please consider this when you are thinking of other renames that may be needed.

pirz:
> > @rapoth There is no use of "appended" file list in the description.
> > If we always build the incremental index, it's not necessary to keep the appended file list.
> > I think we could have another option (for better hybrid scan) to use both list - refreshIndex(mode='metadata')
> > (or 'metadataOnly', 'signature', 'forHybrid'..?)
> > In this way, the user can clearly see which index could be reused at query time. (in case hybrid scan config is disabled)
> > With this, we could suggest this until we optimize the rank algorithm properly. WDYT??
> > And for naming, I prefer
> > "RefreshSourceFilesAction"
> > and
> > "MetadataUpdateActionBase"
> 
> Yes, I requested @pirz to update the description appropriately since the code is out-of-sync. Long story short, appended files is not going to be useful for incremental index for appends. It will only be useful for hybrid scan - is this correct?
> 
> If the list of appended files goes in as part of this PR, can Hybrid scan utilize it? I like `refreshIndex(mode='metadata')` - @imback82 / @pirz ?
> 
> @sezruby Will this `mode='metadata'` be in addition to `smart` and `quick`?

Sorry, but I am a bit confused here. PR description was updated and it was talking about "appended" and "deleted" files, before these comments. Can you please clarify if we are talking about this PR or some other PR?

rapoth:
> > > @rapoth There is no use of "appended" file list in the description.
> > > If we always build the incremental index, it's not necessary to keep the appended file list.
> > > I think we could have another option (for better hybrid scan) to use both list - refreshIndex(mode='metadata')
> > > (or 'metadataOnly', 'signature', 'forHybrid'..?)
> > > In this way, the user can clearly see which index could be reused at query time. (in case hybrid scan config is disabled)
> > > With this, we could suggest this until we optimize the rank algorithm properly. WDYT??
> > > And for naming, I prefer
> > > "RefreshSourceFilesAction"
> > > and
> > > "MetadataUpdateActionBase"
> > 
> > 
> > Yes, I requested @pirz to update the description appropriately since the code is out-of-sync. Long story short, appended files is not going to be useful for incremental index for appends. It will only be useful for hybrid scan - is this correct?
> > If the list of appended files goes in as part of this PR, can Hybrid scan utilize it? I like `refreshIndex(mode='metadata')` - @imback82 / @pirz ?
> > @sezruby Will this `mode='metadata'` be in addition to `smart` and `quick`?
> 
> Sorry, but I am a bit confused here. PR description was updated and it was talking about "appended" and "deleted" files, before these comments. Can you please clarify if we are talking about this PR or some other PR?

Thanks! I'm good from my side. Apparently, I was looking at an alternate version. @sezruby ?

sezruby:
@pirz @rapoth Sorry the description I mentioned is the one in comment https://github.com/microsoft/hyperspace/pull/170#issuecomment-701707985

Yes, 'metadata' is an additional refresh mode to use hybrid scan w/o query time detection of source file list. 
It will also allow selective hybrid scan for each index.

rapoth:
> @pirz @rapoth Sorry the description I mentioned is the one in comment #170 (comment)
> 

Ok, last question. I re-read my comment and I did mention appended files. 

> We handle capturing both deleted and appended files in the index meta-data as it's already implemented in this PR.

imback82:
> > And for naming, I prefer
> > "RefreshSourceFilesAction"
> > and
> > "MetadataUpdateActionBase"
> 
> Sorry @sezruby, I missed this. @imback82 please consider this when you are thinking of other renames that may be needed.

Sorry for the late response (busy at work 😄). The names sound reasonable, but how about `RefreshSourceMetadataAction` (`RefreshSourceFilesAction` sounds like we are actually refreshing source files - whatever that is).

Now that we are introducing metadata only action, I think we may need to revisit the class hierarchy for refresh. For example, making `MetadataUpdateActionBase` extend `RefreshActionBase` (which extends `CreateActionBase`) doesn't seem like a like good hierarchy. @pirz Can you give a shot at this? If not, I can take a look on Friday. I think we would know better what the name of `MetadataUpdateActionBase` would be when we rethink on the hierarchy (it could become `MetadataUpdateAction` as a `trait`, for example.)

rapoth:
> > > And for naming, I prefer
> > > "RefreshSourceFilesAction"
> > > and
> > > "MetadataUpdateActionBase"
> > 
> > 
> > Sorry @sezruby, I missed this. @imback82 please consider this when you are thinking of other renames that may be needed.
> 
> Sorry for the late response (busy at work 😄). The names sound reasonable, but how about `RefreshSourceMetadataAction` (`RefreshSourceFilesAction` sounds like we are actually refreshing source files - whatever that is).

Yes, being explicit about Metadata in the Action name is a good call. Thanks!

pirz:
#188 addressed parts of the changes in this PR. New metadata action will be added as a separate PR.

PR #171, opened 2020-09-21
Inject push-down filter to exclude indexed rows from deleted files

sezruby:

 - **Tracking Issue**: #150
 - **Parent Issue**: #150
 - **Dependencies**: #165

With lineage column in the index data, we could exclude the deleted rows from the index data at query time, by injecting `Filter(Not(In(<lineage column>, <deleted files>)))`.
I added additional config `spark.hyperspace.index.hybridscan.delete.enabled` for enabling/disabling hybrid scan for delete dataset until performance validation and optimization.
- `spark.hyperspace.index.hybridscan.enabled = true/ spark.hyperspace.index.hybridscan.delete.enabled = false` => hybrid scan only for appended dataset; if there are deleted files in source relation, the index won't be applied.
- `spark.hyperspace.index.hybridscan.enabled = false / spark.hyperspace.index.hybridscan.delete.enabled = true` won't work

To support this, additional project & filter node is injected to LogicalRelation plan of index data. Reading the index data with the lineage column and filter the deleted files, and then exclude the lineage column as it shouldn't be shown in the result.
```
   :  :- Project [clicks#456, query#454]
   :  :  +- Filter ((isnotnull(clicks#456) && (clicks#456 >= 2000)) && (clicks#456 <= 4000))
   :  :     +- Project [Query#454, clicks#456]
   :  :        +- Filter NOT _data_file_name#474 IN  (file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/part-00003-3b411cdd-be12-4dad-a6a0-23fd244bd6e4-c000.snappy.parquet)
```

Note that the injected Project & Fitler is optimized by Spark so that there's no extra stage for them:
```
   :     +- BucketUnion 200 buckets, bucket columns: [clicks]
   :        :- *(1) Project [clicks#456, query#454]
   :        :  +- *(1) Filter ((((isnotnull(_data_file_name#474) && NOT (_data_file_name#474 = file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/part-00003-3b411cdd-be12-4dad-a6a0-23fd244bd6e4-c000.snappy.parquet)) && isnotnull(clicks#456)) && (clicks#456 >= 2000)) && (clicks#456 <= 4000))
   :        :     +- *(1) FileScan parquet [clicks#456,Query#454,_data_file_name#474] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/hybridScanTest/index..., PartitionFilters: [], PushedFilters: [IsNotNull(_data_file_name), Not(EqualTo(_data_file_name,file:/C:/Users/eunsong/repo/hyperspace2/..., ReadSchema: struct<clicks:int,Query:string,_data_file_name:string>, Sele
ctedBucketsCount: 200 out of 200
```

Yes, if a user enabled Hybrid Scan delete support, indexes invalidated by "deleted files" from its source relation can be still utilized with modified plan.

The followings are the logical & physical plan for Join plan, with both appended and deleted files.

Logical plan
```
Project [clicks#456, query#454, query#468]
+- Join Inner, (clicks#456 = clicks#470)
   :- BucketUnion 200 buckets, bucket columns: [clicks]
   :  :- Project [clicks#456, query#454]
   :  :  +- Filter ((isnotnull(clicks#456) && (clicks#456 >= 2000)) && (clicks#456 <= 4000))
   :  :     +- Project [Query#454, clicks#456]
   :  :        +- Filter NOT _data_file_name#474 IN (file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/part-00003-3b411cdd-be12-4dad-a6a0-23fd244bd6e4-c000.snappy.parquet)
   :  :           +- Relation[Query#454,clicks#456,_data_file_name#474] parquet
   :  +- RepartitionByExpression [clicks#456], 200
   :     +- Project [clicks#456, query#454]
   :        +- Filter ((isnotnull(clicks#456) && (clicks#456 >= 2000)) && (clicks#456 <= 4000))
   :           +- Relation[Query#454,clicks#456] parquet
   +- BucketUnion 200 buckets, bucket columns: [clicks]
      :- Project [clicks#470, query#468]
      :  +- Filter ((isnotnull(clicks#470) && (clicks#470 <= 4000)) && (clicks#470 >= 2000))
      :     +- Project [Query#468, clicks#470]
      :        +- Filter NOT _data_file_name#475 IN (file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/part-00003-3b411cdd-be12-4dad-a6a0-23fd244bd6e4-c000.snappy.parquet)
      :           +- Relation[Query#468,clicks#470,_data_file_name#475] parquet
      +- RepartitionByExpression [clicks#470], 200
         +- Project [clicks#470, query#468]
            +- Filter ((isnotnull(clicks#470) && (clicks#470 <= 4000)) && (clicks#470 >= 2000))
               +- Relation[Query#468,clicks#470] parquet
```

Physical plan
```
*(7) Project [clicks#456, query#454, query#468]
+- *(7) SortMergeJoin [clicks#456], [clicks#470], Inner
   :- *(3) Sort [clicks#456 ASC NULLS FIRST], false, 0
   :  +- *(3) Filter isnotnull(clicks#456)
   :     +- BucketUnion 200 buckets, bucket columns: [clicks]
   :        :- *(1) Project [clicks#456, query#454]
   :        :  +- *(1) Filter ((((isnotnull(_data_file_name#474) && NOT (_data_file_name#474 = file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/part-00003-3b411cdd-be12-4dad-a6a0-23fd244bd6e4-c000.snappy.parquet)) && isnotnull(clicks#456)) && (clicks#456 >= 2000)) && (clicks#456 <= 4000))
   :        :     +- *(1) FileScan parquet [clicks#456,Query#454,_data_file_name#474] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/hybridScanTest/index..., PartitionFilters: [], PushedFilters: [IsNotNull(_data_file_name), Not(EqualTo(_data_file_name,file:/C:/Users/eunsong/repo/hyperspace2/..., ReadSchema: struct<clicks:int,Query:string,_data_file_name:string>, Sele
ctedBucketsCount: 200 out of 200
   :        +- Exchange hashpartitioning(clicks#456, 200)
   :           +- *(2) Project [clicks#456, query#454]
   :              +- *(2) Filter ((isnotnull(clicks#456) && (clicks#456 >= 2000)) && (clicks#456 <= 4000))
   :                 +- *(2) FileScan parquet [clicks#456,Query#454] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/..., PartitionFilters: [], PushedFilters: [IsNotNull(clicks), GreaterThanOrEqual(clicks,2000), LessThanOrEqual(clicks,4000)], ReadSchema: struct<clicks:int,Query:string>
   +- *(6) Sort [clicks#470 ASC NULLS FIRST], false, 0
      +- *(6) Filter isnotnull(clicks#470)
         +- BucketUnion 200 buckets, bucket columns: [clicks]
            :- *(4) Project [clicks#470, query#468]
            :  +- *(4) Filter ((((isnotnull(_data_file_name#475) && NOT (_data_file_name#475 = file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/part-00003-3b411cdd-be12-4dad-a6a0-23fd244bd6e4-c000.snappy.parquet)) && isnotnull(clicks#470)) && (clicks#470 <= 4000)) && (clicks#470 >= 2000))
            :     +- *(4) FileScan parquet [clicks#470,Query#468,_data_file_name#475] Batched: true, Format: Parquet, Location: InMemoryFileIndex[file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/hybridScanTest/index..., PartitionFilters: [], PushedFilters: [IsNotNull(_data_file_name), Not(EqualTo(_data_file_name,file:/C:/Users/eunsong/repo/hyperspace2/..., ReadSchema: struct<clicks:int,Query:string,_data_file_name:string>, Sele
ctedBucketsCount: 200 out of 200
            +- ReusedExchange [clicks#470, query#468], Exchange hashpartitioning(clicks#456, 200)

```

unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
@@ -27,8 +27,18 @@ object IndexConstants {
   val INDEX_SEARCH_PATHS = "spark.hyperspace.index.search.paths"
   val INDEX_NUM_BUCKETS = "spark.hyperspace.index.num.buckets"
 
+  // This config enables Hybrid scan on mutable dataset at query time.
+  // Currently, this config allows to perform Hybrid scan on append-only dataset.
+  // For delete dataset, "spark.hyperspace.index.hybridscan.delete.enabled" is
+  // also needed to be set.
   val INDEX_HYBRID_SCAN_ENABLED = "spark.hyperspace.index.hybridscan.enabled"
   val INDEX_HYBRID_SCAN_ENABLED_DEFAULT = "false"
+
+  // This is a temporary config to support Hybrid scan on both append & delete dataset.
+  // The config does not work without the Hybrid scan config
+  // "spark.hyperspace.index.hybridscan.enabled"
+  // and will be removed after performance validation and optimization.
+  // See https://github.com/microsoft/hyperspace/issues/184
   val INDEX_HYBRID_SCAN_DELETE_ENABLED = "spark.hyperspace.index.hybridscan.delete.enabled"
   val INDEX_HYBRID_SCAN_DELETE_ENABLED_DEFAULT = "false"
 
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
@@ -29,6 +29,8 @@ object IndexConstants {
 
   val INDEX_HYBRID_SCAN_ENABLED = "spark.hyperspace.index.hybridscan.enabled"
   val INDEX_HYBRID_SCAN_ENABLED_DEFAULT = "false"
+  val INDEX_HYBRID_SCAN_DELETE_ENABLED = "spark.hyperspace.index.hybridscan.delete.enabled"
+  val INDEX_HYBRID_SCAN_DELETE_ENABLED_DEFAULT = "false"
 
   // Default number of buckets is set the default value of "spark.sql.shuffle.partitions".
   val INDEX_NUM_BUCKETS_DEFAULT: Int = SQLConf.SHUFFLE_PARTITIONS.defaultValue.get
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
@@ -103,7 +103,7 @@ object FilterIndexRule
           .getContext(spark)
           .indexCollectionManager
         val candidateIndexes =
-          RuleUtils.getCandidateIndexes(indexManager, r, HyperspaceConf.hybridScanEnabled(spark))
+          RuleUtils.getCandidateIndexes(spark, indexManager, r)
 
         candidateIndexes.filter { index =>
           indexCoversPlan(
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -138,7 +138,7 @@ object RuleUtils {
    *
    * Pre-requisites
    * - We know for sure the index which can be used to transform the plan.
-   * - The plan should be linear and include 1 LogicalRelation.
+   * - The plan should be linear and include one LogicalRelation.
    *
    * @param spark Spark session.
    * @param index Index used in transformation of plan.
@@ -151,6 +151,9 @@ object RuleUtils {
       index: IndexLogEntry,
       plan: LogicalPlan,
       useBucketSpec: Boolean): LogicalPlan = {
+    // Check pre-requisite.
+    assert(getLogicalRelation(plan).isDefined)
+
     val transformed = if (HyperspaceConf.hybridScanEnabled(spark)) {
       transformPlanToUseHybridScan(spark, index, plan, useBucketSpec)
     } else {
@@ -224,8 +227,9 @@ object RuleUtils {
 
     // Get transformed plan with index data and appended files if applicable.
     val indexPlan = plan transformUp {
-      // Use transformUp here as currently 1 Logical Relation is allowed (pre-requisite).
-      // Otherwise, we need additional check not to process newly injected Filter-Not-In nodes.
+      // Use transformUp here as currently one Logical Relation is allowed (pre-requisite).
+      // With transformDown, the transformed plan will be matched and transformed recursively
+      // which causes Stack overflow exception.
       case baseRelation @ LogicalRelation(
             _ @HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
             baseOutput,
@@ -244,7 +248,9 @@ object RuleUtils {
               (Nil, filesAppended)
             }
           } else {
-            // Keep append-only implementation for efficiency.
+            // Append-only implementation of getting appended files for efficiency.
+            // It is guaranteed that there is no deleted files via the condition
+            // 'deletedCnt == 0 && commonCnt > 0' in isHybridScanCandidate function.
             (
               Nil,
               curFileSet.filterNot(index.allSourceFileInfos.contains).map(f => new Path(f.name)))
@@ -253,10 +259,13 @@ object RuleUtils {
         val filesToRead = {
           if (useBucketSpec || !isParquetSourceFormat || filesDeleted.nonEmpty) {
             // Since the index data is in "parquet" format, we cannot read source files
-            // in formats other than "parquet" using 1 FileScan node as the operator requires
+            // in formats other than "parquet" using one FileScan node as the operator requires
             // files in one homogenous format. To address this, we need to read the appended
             // source files using another FileScan node injected into the plan and subsequently
             // merge the data into the index data. Please refer below [[Union]] operation.
+            // In case there are both deleted and appended files, we cannot handle the appended
+            // files along with deleted files as source files do not have the lineage column which
+            // requires for excluding the index data from deleted files.
             unhandledAppendedFiles = filesAppended
             index.content.files
           } else {
@@ -268,17 +277,19 @@ object RuleUtils {
           }
         }
 
-        val readSchema = if (filesDeleted.isEmpty) {
-          StructType(index.schema.filter(baseRelation.schema.contains(_)))
-        } else {
-          StructType(index.schema)
-        }
+        // In order to handle deleted files, read index data with the lineage column so that
+        // we could inject Filter-Not-In conditions on the lineage column to exclude the indexed
+        // rows from the deleted files.
+        val newSchema = StructType(
+          index.schema.filter(s =>
+            baseRelation.schema.contains(s) || (filesDeleted.nonEmpty && s.name.equals(
+              IndexConstants.DATA_FILE_NAME_COLUMN))))
 
         val newLocation = new InMemoryFileIndex(spark, filesToRead, Map(), None)
         val relation = HadoopFsRelation(
           newLocation,
           new StructType(),
-          readSchema,
+          newSchema,
           if (useBucketSpec) Some(index.bucketSpec) else None,
           new ParquetFileFormat,
           Map())(spark)
@@ -289,11 +300,11 @@ object RuleUtils {
         if (filesDeleted.isEmpty) {
           baseRelation.copy(relation = relation, output = updatedOutput)
         } else {
-          val lAttr = AttributeReference(IndexConstants.DATA_FILE_NAME_COLUMN, StringType)(
-            NamedExpression.newExprId)
+          val lineageAttr = AttributeReference(IndexConstants.DATA_FILE_NAME_COLUMN, StringType)()
           val deletedFileNames = filesDeleted.map(f => Literal(f.name)).toArray
-          val rel = baseRelation.copy(relation = relation, output = updatedOutput ++ Seq(lAttr))
-          Project(updatedOutput, Filter(Not(In(lAttr, deletedFileNames)), rel))
+          val rel =
+            baseRelation.copy(relation = relation, output = updatedOutput ++ Seq(lineageAttr))
+          Project(updatedOutput, Filter(Not(In(lineageAttr, deletedFileNames)), rel))
         }
     }
 
@@ -370,7 +381,7 @@ object RuleUtils {
    * Transform the plan to perform on-the-fly Shuffle the data based on bucketSpec.
    *
    * Pre-requisite
-   * - The plan should be linear and include 1 LogicalRelation.
+   * - The plan should be linear and include one LogicalRelation.
    *
    * @param bucketSpec Bucket specification used for Shuffle.
    * @param plan Plan to be shuffled.
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -21,14 +21,14 @@ import scala.collection.mutable
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.catalog.BucketSpec
-import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, In, Literal, NamedExpression, Not}
 import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan, Project, RepartitionByExpression, Union}
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, InMemoryFileIndex, LogicalRelation, PartitioningAwareFileIndex}
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
-import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.types.{StringType, StructType}
 
 import com.microsoft.hyperspace.actions.Constants
-import com.microsoft.hyperspace.index.{FileInfo, IndexLogEntry, IndexManager, LogicalPlanSignatureProvider}
+import com.microsoft.hyperspace.index.{FileInfo, IndexConstants, IndexLogEntry, IndexManager, LogicalPlanSignatureProvider}
 import com.microsoft.hyperspace.index.plans.logical.BucketUnion
 import com.microsoft.hyperspace.util.HyperspaceConf
 
@@ -37,17 +37,20 @@ object RuleUtils {
   /**
    * Get active indexes for the given logical plan by matching signatures.
    *
+   * @param spark Spark Session.
    * @param indexManager IndexManager.
    * @param plan Logical plan.
-   * @param hybridScanEnabled Flag that checks if hybrid scan is enabled or disabled.
    * @return Indexes built for this plan.
    */
   def getCandidateIndexes(
+      spark: SparkSession,
       indexManager: IndexManager,
-      plan: LogicalPlan,
-      hybridScanEnabled: Boolean): Seq[IndexLogEntry] = {
+      plan: LogicalPlan): Seq[IndexLogEntry] = {
     // Map of a signature provider to a signature generated for the given plan.
     val signatureMap = mutable.Map[String, Option[String]]()
+
+    val hybridScanEnabled = HyperspaceConf.hybridScanEnabled(spark)
+    val hybridScanDeleteEnabled = HyperspaceConf.hybridScanDeleteEnabled(spark)
 
     def signatureValid(entry: IndexLogEntry): Boolean = {
       val sourcePlanSignatures = entry.source.plan.properties.fingerprint.properties.signatures
@@ -76,8 +79,12 @@ object RuleUtils {
       val commonCnt = inputSourceFiles.count(entry.allSourceFileInfos.contains)
       val deletedCnt = entry.allSourceFileInfos.size - commonCnt
 
-      // Currently, Hybrid Scan only support for append-only dataset.
-      deletedCnt == 0 && commonCnt > 0
+      if (hybridScanDeleteEnabled && entry.hasLineageColumn(spark)) {
+        commonCnt > 0
+      } else {
+        // For append-only dataset.
+        deletedCnt == 0 && commonCnt > 0
+      }
     }
 
     // TODO: the following check only considers indexes in ACTIVE state for usage. Update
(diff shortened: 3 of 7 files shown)


src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -147,12 +97,16 @@ object FilterIndexRule
       outputColumns: Seq[String],
       filterColumns: Seq[String],
       fsRelation: HadoopFsRelation): Seq[IndexLogEntry] = {
+    val indexManager = Hyperspace
+      .getContext(spark)
+      .indexCollectionManager
```
pirz, line 102: I am just curious, is there a reason to move this code snippet outside of `Some` block. We dont use IndexManager for the `None` case.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +128,318 @@ object RuleUtils {
...
+   *
+   * NOTE: This method currently only supports replacement of Scan Nodes i.e. Logical relations
+   *
+   * @param spark Spark session.
+   * @param index Index used in replacement plan.
+   * @param plan Current logical plan.
+   * @param useBucketSpec Option whether to use BucketSpec for reading index data.
+   * @return Transformed logical plan that leverages an index.
+   */
+  private def transformPlanToUsePureIndexScan(
```
pirz, line 176: (just a naming suggestion, feel free to ignore it if you disagree)
rename `transformPlanToUsePureIndexScan` -> `transformPlanToUseIndexOnlyScan`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -106,4 +128,318 @@ object RuleUtils {
...
+      spark: SparkSession,
+      index: IndexLogEntry,
+      plan: LogicalPlan,
+      useBucketSpec: Boolean): LogicalPlan = {
+    val transformed = if (HyperspaceConf.hybridScanEnabled(spark)) {
+      transformPlanToUseHybridIndexDataScan(spark, index, plan, useBucketSpec)
+    } else {
+      transformPlanToUsePureIndexScan(spark, index, plan, useBucketSpec)
+    }
+    assert(!transformed.equals(plan))
```
pirz, line 159: I think this `assert` should be replaced by an `exception` thrown.
Imagine the replacement fails and no transformation happens. I think, This assert basically makes the whole user query to break. One functionality aspect of Hyeprspace is that: "A user's query should not really break due to the "extra" optimization Hyperspace tries to add". That is why in the rules, we used `Exceptions` and in the `Catch` block of `transofrm` we catch any exception, log it and return the original plan. We dont really wanna fail a query if our attempt to (further) optimize fails.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
sezruby: I'll revert this in #165. Thanks :)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: The name was also suggested.  I think `PureIndexScan` is also good naming :)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: It's because there are pre-requite conditions and we previously check conditions for this like linear plan, candidate indexes, .. etc. I think some asserts are required for development & testing. If we only throw exceptions which are ignored in Catch block, it's hard to debug any issue or even unnoticed.
This assert was suggested by @imback82 in #165.

imback82: Yes, assert makes more sense than exception here since we are validating the "assumption" for this function.

imback82:
@sezruby Can you resolve the conflicts?

imback82:
> * `spark.hyperspace.index.hybridscan.enabled = false / spark.hyperspace.index.hybridscan.delete.enabled = true` => hybrid scan only for appended dataset; if there are deleted files in source relation, the index won't be applied.
> * `spark.hyperspace.index.hybridscan.enabled = false / spark.hyperspace.index.hybridscan.delete.enabled = true` won't work

I think you meant `true` and `true` in the first bullet?

Btw, why do we need to introduce `spark.hyperspace.index.hybridscan.delete.enabled` config? Is there a particular use case we are targeting?

sezruby:
@imback82 I added the config to make this delete support switchable until performance validation. However, we could keep both configs if needed.

Do you think we should also support delete-only case?

imback82:
Ah, OK. We can revisit the config names once we have the perf numbers. Thanks! (It would have been nice if this info was captured in the PR description).

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -216,20 +223,35 @@ object RuleUtils {
     var unhandledAppendedFiles: Seq[Path] = Nil
 
     // Get transformed plan with index data and appended files if applicable.
-    val indexPlan = plan transformDown {
+    val indexPlan = plan transformUp {
+      // Use transformUp here as currently 1 Logical Relation is allowed (pre-requisite).
```
imback82, line 227: nit: `1` -> `one`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -216,20 +223,35 @@ object RuleUtils {
...
+      // Otherwise, we need additional check not to process newly injected Filter-Not-In nodes.
```
imback82, line 228: Can we test the prerequisite as assert?
```scala
def verifyPrerequisite(plan: LogicalPlan) {
  getLogicalRelation(plan).isDefined
  // We can extend this as needed later.
}
```
And I think we can call `verifyPrerequisite` from `transformPlanToUseIndex` since its comment also has this as a pre-requisite.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -216,20 +223,35 @@ object RuleUtils {
...
+          if (HyperspaceConf.hybridScanDeleteEnabled(spark) && index.hasLineageColumn(spark)) {
+            val (exist, nonExist) = curFileSet.partition(index.allSourceFileInfos.contains)
+            val filesAppended = nonExist.map(f => new Path(f.name))
+            if (exist.length < index.allSourceFileInfos.size) {
+              (index.allSourceFileInfos -- exist, filesAppended)
+            } else {
+              (Nil, filesAppended)
+            }
+          } else {
+            // Keep append-only implementation for efficiency.
```
imback82, line 247: Let's remove `Keep`:

`// Append-only implementation of getting appended files for efficiency.`

Also, can you add comment that in `else`, there should be no deleted files, which is checked in `isHybridScanCandidate`? I think a better way is to keep a context to pass around. (we should do this as a follow up).

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -216,20 +223,35 @@ object RuleUtils {
...
+            (
+              Nil,
+              curFileSet.filterNot(index.allSourceFileInfos.contains).map(f => new Path(f.name)))
+          }
 
         val filesToRead = {
-          if (useBucketSpec || !isParquetSourceFormat) {
+          if (useBucketSpec || !isParquetSourceFormat || filesDeleted.nonEmpty) {
```
imback82, line 254: Can you update the comment with `filesDeleted.nonEmpty` condition (and why)?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -246,18 +268,33 @@ object RuleUtils {
           }
         }
 
+        val readSchema = if (filesDeleted.isEmpty) {
```
imback82, line 271: `newSchema`? Also, can you add a comment why you have a different branch for delete (e.g., need to utilized DATA_FILE_NAME_COLUMN, etc.)? BTW, can you just append this data file name column to the `StructType(index.schema.filter(baseRelation.schema.contains(_)))` or no?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -246,18 +268,33 @@ object RuleUtils {
...
           Map())(spark)
 
         val updatedOutput =
           baseOutput.filter(attr => relation.schema.fieldNames.contains(attr.name))
-        baseRelation.copy(relation = relation, output = updatedOutput)
+
+        if (filesDeleted.isEmpty) {
+          baseRelation.copy(relation = relation, output = updatedOutput)
+        } else {
+          val lAttr = AttributeReference(IndexConstants.DATA_FILE_NAME_COLUMN, StringType)(
```
imback82, line 292: `AttributeReference(IndexConstants.DATA_FILE_NAME_COLUMN, StringType)()` (`NamedExpression.newExprId` is default).

also what is `l` in `lAttr`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: Also, we only match `LogicalRelation` so does it matter if we do transform up or down? If I missed something can you elaborate in the comment?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -120,59 +120,94 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
...
         // Scenario #2: Delete 1 file.
-        {
-          FileUtils.delete(new Path(indexFile), isRecursive = false)
-        }
+        FileUtils.delete(new Path(sourceFile))
 
         {
           val optimizedPlan = spark.read.parquet(dataPath).queryExecution.optimizedPlan
-          verify(optimizedPlan, hybridScanEnabled = false, expectCandidateIndex = false)
-          // TODO: expectedCandidateIndex = true once delete dataset is supported.
```
imback82, line 165: Should we add `hybridScanDeleteEnabled = true` case?

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -39,22 +39,34 @@ class HybridScanTest extends QueryTest with HyperspaceSuite {
...
-    assert(appendCnt < inputFiles.length)
+    assert(appendCnt + deleteCnt < inputFiles.length)
 
     val fs = systemPath.getFileSystem(new Configuration)
     for (i <- 0 until appendCnt) {
       val sourcePath = new Path(inputFiles(i))
       val destPath = new Path(inputFiles(i) + ".copy")
       fs.copyToLocalFile(sourcePath, destPath)
     }
+    for (i <- 1 to deleteCnt) {
```
imback82, line 67: can we be consistent of using `until` or `to` (just check above)

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -291,4 +340,339 @@ class HybridScanTest extends QueryTest with HyperspaceSuite {
...
+    }
+  }
+
+  test(
+    "Delete-only: filter index & parquet, json format, " +
+      "index relation should have additional filter for deleted files") {
+
+    Seq(
+      (sampleParquetDataLocationDelete, "index2", "parquet"),
+      (sampleJsonDataLocationDelete, "index5", "json")) foreach {
```
imback82, line 373: There are so many indexes created for this suite that it's very hard to follow.

Can you at least describe what `index2` and `index5` are at the beginning of test so that I don't have to go back and forth to understand these indexes?

I am also fine creating the `index5` in this test since it's not being shared anyway (unless you are testing a scenario with multiple indexes in place).

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: The transformed logical relation is matched again with transformDown which causes stack overflow. I'll update the comment.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: `l` from lineage. How about `lineageAttr` ??

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: I moved the condition in filter:
```
val newSchema = StructType(index.schema.filter(s =>
            baseRelation.schema.contains(s) || (filesDeleted.nonEmpty && s.name.equals(
              IndexConstants.DATA_FILE_NAME_COLUMN))))
```

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
sezruby: These are intended not to use `i - 1` or `deletedCnt +1`. Which do you think is better here?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: Ah OK, makes sense. Thanks.

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
imback82: Ah OK, then it should be fine. Let's add the delete behavior to the `def` comment.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -29,6 +29,8 @@ object IndexConstants {
 
   val INDEX_HYBRID_SCAN_ENABLED = "spark.hyperspace.index.hybridscan.enabled"
   val INDEX_HYBRID_SCAN_ENABLED_DEFAULT = "false"
+  val INDEX_HYBRID_SCAN_DELETE_ENABLED = "spark.hyperspace.index.hybridscan.delete.enabled"
```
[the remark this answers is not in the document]
sezruby: I think it's better to move these config related vals to HyperspaceConf and make some comments there. What do you think?
Can we do this with a separate PR?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: > And I think we can call verifyPrerequisite from transformPlanToUseIndex since its comment also has this as a pre-requisite.

I added the assert stmt not function in `transformPlanToUseIndex`. Let's create the validate function later when we have another condition.

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
sezruby: I renamed the indexes more descriptively. Let me know if they are still unclear. Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -76,8 +79,12 @@ object RuleUtils {
       val commonCnt = inputSourceFiles.count(entry.allSourceFileInfos.contains)
       val deletedCnt = entry.allSourceFileInfos.size - commonCnt
```
apoorvedave1, line 80: can move this in the else block below, so it's only calculated when delete enabled

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -216,25 +226,46 @@ object RuleUtils {
...
+
+        val (filesDeleted, filesAppended) =
+          if (HyperspaceConf.hybridScanDeleteEnabled(spark) && index.hasLineageColumn(spark)) {
+            val (exist, nonExist) = curFileSet.partition(index.allSourceFileInfos.contains)
+            val filesAppended = nonExist.map(f => new Path(f.name))
+            if (exist.length < index.allSourceFileInfos.size) {
+              (index.allSourceFileInfos -- exist, filesAppended)
+            } else {
+              (Nil, filesAppended)
+            }
```
apoorvedave1, line 249: can we replace this section with just 
```scala
(index.allSourceFileInfos -- exist, filesAppended)
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -216,25 +226,46 @@ object RuleUtils {
     var unhandledAppendedFiles: Seq[Path] = Nil
```
apoorvedave1, line 226: could we please move this variable closer to its usage? Also please add a description for this variable next to declaration?

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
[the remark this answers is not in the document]
imback82: We can do the move as a separate PR, but let's add some comments in this PR.

imback82:
@sezruby Can you resolve the conflict?

imback82 (commented):
Few minor comments, but LGTM!

imback82 (approved):

imback82:
Thanks for working on this @sezruby!

Issue #160, opened 2020-09-14
Optimization of index selection process


sezruby:
### Describe the issue

Currently, this `getCandidateIndex` function is done first among the index selection & application like the following:
1) getCandidateIndex - get list of indexes whose source files match with the input relation
2) column schema check - all indexed columns & all required columns (using resolver)
3) other conditions (for join)
4) rank algorithm

However, getCandidateIndex could be much heavier than the other steps in case the input relation has numerous files.

We could optimize these steps by 
- validating index schema or root directory before comparing all the input files (PR #173)
- not trying to apply rules for plans which is already modified by Hyperspace (PR #178) 
- reading & calculating diff files only once and caching it across the application of all Rules

sezruby:
> * reading & calculating diff files only once and caching it across the application of all Rules

Let's handle this with #222.

PR #173, opened 2020-09-23
Check indexed and included columns before signature calculation

sezruby:

 - **Tracking Issue**: #160 
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

This PR changes the order of index validation.
Currently, we gets the candidate index in the following order:

1. getCandidateIndex => signature calculation & comparison
2. index schema check => all required columns & all indexed columns
3. other conditions (e.g. compatible check in Join Index)
4. rank algorithm

However, signature calculation might take longer in case there are numerus source files. In addition, if there are a lot of indexes for different locations, file-level check without any pruning might be very inefficient.

With this PR, the index schema is firstly validated before the signature check.
1. index schema check => each candidate index should include all required columns & all indexed columns
2. getCandidateIndex => signature calculation & comparison
3. other conditions (e.g. compatible left & right indexes for Join Index)
4. rank algorithm

No

Tested by FilterIndexRuleTest, JoinIndexRuleTest and E2EHyperspaceRulesTests.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
@@ -118,6 +118,9 @@ object FilterIndexRule
             fsRelation.fileFormat)
         }
 
+        // Get candidate via file-level metadata validation. This is performed after pruning
+        // by column schema, as this might be expensive when there are numerous files in the
+        // relation or many indexes to be checked.
         RuleUtils.getCandidateIndexes(candidateIndexes, r, HyperspaceConf.hybridScanEnabled(spark))
 
       case None => Nil // There is zero or more than one LogicalRelation nodes in Filter's subplan
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
@@ -24,6 +24,7 @@ import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.datasources._
 
 import com.microsoft.hyperspace.{ActiveSparkSession, Hyperspace}
+import com.microsoft.hyperspace.actions.Constants
 import com.microsoft.hyperspace.index.IndexLogEntry
 import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEventLogging, HyperspaceIndexUsageEvent}
 import com.microsoft.hyperspace.util.{HyperspaceConf, ResolverUtils}
@@ -102,10 +103,13 @@ object FilterIndexRule
         val indexManager = Hyperspace
           .getContext(spark)
           .indexCollectionManager
-        val candidateIndexes =
-          RuleUtils.getCandidateIndexes(indexManager, r, HyperspaceConf.hybridScanEnabled(spark))
 
-        candidateIndexes.filter { index =>
+        // TODO: the following check only considers indexes in ACTIVE state for usage. Update
+        //  the code to support indexes in transitioning states as well.
+        //  See https://github.com/microsoft/hyperspace/issues/65
+        val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
+
+        val candidateIndexes = allIndexes.filter { index =>
           indexCoversPlan(
             outputColumns,
             filterColumns,
@@ -113,6 +117,8 @@ object FilterIndexRule
             index.includedColumns,
             fsRelation.fileFormat)
         }
+
+        RuleUtils.getCandidateIndexes(candidateIndexes, r, HyperspaceConf.hybridScanEnabled(spark))
 
       case None => Nil // There is zero or more than one LogicalRelation nodes in Filter's subplan
     }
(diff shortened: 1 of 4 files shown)


src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -152,9 +153,13 @@ object FilterIndexRule
...
           .indexCollectionManager
-        val candidateIndexes = RuleUtils.getCandidateIndexes(indexManager, r)
 
-        candidateIndexes.filter { index =>
+        // TODO: the following check only considers indexes in ACTIVE state for usage. Update
+        //  the code to support indexes in transitioning states as well.
+        //  See https://github.com/microsoft/hyperspace/issues/65
+        val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
+
+        val candidateIndexes = allIndexes.filter { index =>
```
pirz, line 162: nit: Can you add a brief comment here as why the order of code blocks matter here (first index schema check then signature computation/comparison - similar to what you already explained in the PR description)? This is to make sure nobody in a future PR moves the code blocks around and causes negative perf impact.

pirz (commented):
LGTM - I added some minor points for adding comments.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -358,9 +328,13 @@ object JoinIndexRule
...
     require(resolve(spark, rRequiredIndexedCols, rRequiredAllCols).isDefined)
 
-    val lUsable = getUsableIndexes(lIndexes, lRequiredIndexedCols, lRequiredAllCols)
-    val rUsable = getUsableIndexes(rIndexes, rRequiredIndexedCols, rRequiredAllCols)
-    val compatibleIndexPairs = getCompatibleIndexPairs(lUsable, rUsable, lRMap)
+    val lUsable = getUsableIndexes(allIndexes, lRequiredIndexedCols, lRequiredAllCols)
+    val rUsable = getUsableIndexes(allIndexes, rRequiredIndexedCols, rRequiredAllCols)
+
+    val lIndexes = RuleUtils.getCandidateIndexes(lUsable, RuleUtils.getLogicalRelation(left).get)
+    val rIndexes = RuleUtils.getCandidateIndexes(rUsable, RuleUtils.getLogicalRelation(right).get)
```
pirz, line 335: ditto: (Similar to comment on FilterIndex rule) Please add a brief comment on why schema check should be done before signature computation/comparison.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
sezruby: Thanks added some comment below. :)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -37,13 +36,13 @@ object RuleUtils {
   /**
    * Get active indexes for the given logical plan by matching signatures.
```
apoorvedave1, line 37: could you please update the description

apoorvedave1: also the pre-requisite of ACTIVE indexes

imback82:
@sezruby Can you resolve the conflict?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Thanks! I updated the comment. BTW seems active indexes are not pre-requisite for this function as we check the status with  `index => index.created` (`def created`: status == "ACTIVE")

sezruby:
Actually there will be inevitable conflict with #171, around `getCandidateIndexes` 
I'll rebase it as soon as one of them is merged.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -341,9 +307,22 @@ object JoinIndexRule
...
-    val compatibleIndexPairs = getCompatibleIndexPairs(lUsable, rUsable, lRMap)
+    val lUsable = getUsableIndexes(allIndexes, lRequiredIndexedCols, lRequiredAllCols)
+    val rUsable = getUsableIndexes(allIndexes, rRequiredIndexedCols, rRequiredAllCols)
+
+    // Get candidate via file-level metadata validation. This is performed after pruning
+    // by column schema, as this might be expensive when there are numerous files in the
+    // relation or many indexes to be checked.
+    val lIndexes = RuleUtils.getCandidateIndexes(
+      lUsable,
+      RuleUtils.getLogicalRelation(left).get,
```
imback82, line 318: is it guaranteed that `getLogicalRelation` would always return `Some` at this point?

sezruby: Thanks! I added the condition in `isApplicable`

imback82 (approved):
LGTM, thanks @sezruby!

Issue #161, opened 2020-09-14
[PROPOSAL]: Introduce BloomFilter index


sezruby:
**This is WIP, but please feel free to add any comment on this 😃**

Support BloomFilter index utilizing Spark's feature (https://spark.apache.org/docs/latest/api/java/org/apache/spark/util/sketch/BloomFilter.html)

**This is the second type of index in Hyperspace, so it may require additional works on the overall codebase to support different types of indexes. For example, getCandidateIndex needs to be fixed so that the current FilterIndexRule and JoinIndexRule only use the covering indexes.**

### How can we derive benefit from the index

We could use BloomFilter in 2 ways:
- Bloom Filter for each file
  - where: "equal" condition pushed down
  - how: reducing the number of files to read
- Bloom Filter for Join 
  - where: Join plan
  - how: reducing the rows of counterpart of join

#### Is the overhead during query time reasonable?
TODO

#### API extension
- IndexConfig for BF index => BFIndexConfig 

```
hs.createIndex(df, BFIndexConfig("indexed_column", expectedNumItems, fpp))
```

#### Metadata change
```
case class BloomFilterIndex(properties: BloomFilterIndex.Properties) {
  val kind = "BloomFilterIndex"
}
object BloomFilterIndex {
  case class Properties(indexColumn: String, expectedNumItems: Int, fpp: float, globalBF: Binary)
}
```
* indexColumn, expectedNumItems, fpp : for BloomFilter construction
* `globalBF`: BloomFilter binary data for whole dataset.

#### At creation

We firstly build the bloom filter for each source file and then merge them into 1 `globalBF` for BF join purpose.
The file-level BF index can be stored as parquet format with 2 columns - "FilePath", "BloomFilterData".
* in parquet format
     * FilePath: file path string
     * BloomFilterData: Binary (`writeTo` function from `BloomFilter` API)

For globalBF, the location can be in IndexLogEntry or a separate file, but it should be managed together.

#### At refresh

* Full refresh
  1. filter file-level index data with deleted file paths => now it has only valid rows
  2. create a DF by generating new entries for appended files
  3. union both and write as a new version
  4. update globalBF using the updated file-level BF
 
* quick refresh - (if full refresh is not much slow, we might not need to implement this.)
  * For appended files
    * add new BF entries and "append" to the index content directory. => like incremental index.
    * merge newly created BFs into global BF.
  * For deleted files
    * basically it's okay not to update BF indexes for deleted files because it's false-positive matches.
[...]

PR #176, opened 2020-09-24
[PROPOSAL] Introduce Bloom Filter Index

sezruby:

 - **Tracking Issue**: #161 
 - **Parent Issue**: N/A
 - **Dependencies**: N/A
   - Issue 1
   - Issue 2

A draft design document for #161 

No

Intellij markdown util


sezruby:
@imback82 Can we merge this PR? So that @thugsatbay could update the document.

imback82:
@rapoth What's the plan for proposals moving forward? Are we checking in as a doc or have it as an issue similar to #263?

imback82:
Another way is to create a "Proposal" category under Discussions.

imback82:
In summary, we can pick one of the three to write a proposal:
1) Check in as a doc: current PR
2) Create as an issue and label is as "proposal": #263 
3) Create a proposal under Discussions.

@rapoth @sezruby @pirz @apoorvedave1 @thugsatbay WDYT?

imback82:
My vote is `2)` since `1)` doesn't seem to be going well: check https://github.com/microsoft/hyperspace/tree/master/docs/design. Since design docs can/need to be iterated quite a few times, requiring a check in to the repo doesn't seem like a good idea for productivity.

rapoth:
I agree. It looks like we're having more luck with writing proposals as Issues. The second form was only meant for very large items but we might as well evolve the template inside Issues itself. The only rationale behind a check-in was to encourage PR reviews (a.k.a. Design reviews) but let's do whatever scales for the community.

I am neutral as far as using _Disussions_ is concerned. There is a category _Idea_ that @thugsatbay pointed me to yesterday that we could use for initial discussions (before proposals are written). It looks like we can create a new category as well that we could leverage for _Proposals_.  However, do note that since discussions are not conventional issues, they do not show up in our planning system so it might be hard to incorporate planning. Therefore, unless someone strongly objects, we could have initial ideas as discussions but Proposals should remain as issues.

sezruby:
Yep 2) would be good as it's easy to edit & track.

pirz:
Using (2) is Ok as long as people do check them there. The main issue is that adding an item under issues which is really a design doc and needs interactive comment/reply does not fit with the "issues" category. What if we have separate category just for design proposals? Is that doable in Github?

thugsatbay:
Even if we end up discussing on `Discussion platform` on Github we would end up creating an # Issue for tracking for planning which will just make the flow ugly. So I guess I feel option 2 is good enough. By adding correct/differentiating labels while creating # Issues for proposal and design docs we can do a quick filter that allows us to categorize our content and should work for what @pirz asked in his last question ?!.

PR #177, opened 2020-09-26
Expose HyperspaceSuite.withTempDir() to be used in tests involving temporary directories.

imback82:
I see many tests where we need to be careful about deleting a directory used after a test is run.

This PR exposes `HyperspaceSuite.withTempDir()` where you can use it like the following:
```scala
withTempDir { dir =>
  foo(dir)
}
```
This is a copy of `SparkFunSuite.withTempDir`, which is available in Spark 3.0. When we move to Spark 3.0, we can just remove this function and any existing code should just work since `HyperspaceSuite` extends `SparkFunSuite`.

### DIFF
--- a/src/main/scala/org/apache/spark/util/hyperspace/Utils.scala
+++ b/src/main/scala/org/apache/spark/util/hyperspace/Utils.scala
@@ -16,11 +16,21 @@
 
 package org.apache.spark.util.hyperspace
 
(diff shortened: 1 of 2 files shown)


imback82:
@sezruby @apoorvedave1 @pirz Let's try to utilize this where applicable.

PR #178, opened 2020-09-29
Skip applying rules for modified plans by Hyperspace

sezruby:

 - **Tracking Issue**: #160 
 - **Parent Issue**: N/A.
 - **Dependencies**: N/A

To avoid applying rules for already modified plans by Hyperspace, this PR introduces a new tag to the modified plan.
Spark provides `TreeNodeTag` feature, but it's only available from [Spark 3.0](https://github.com/apache/spark/blob/branch-3.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala)

This PR creates a new identifier to `options` field in `HadoopFsRelation` of each modified plan which location is replaced by index data path (or appended file paths for hybrid scan).

```
// IndexConstants.scala 
 val INDEX_RELATION_IDENTIFIER: (String, String) = ("indexRelation" -> "true")
```

With this, in `FilterIndexRule` and `JoinIndexRule`, we can check whether each candidate plan has the identifier or not. If the plan is modified, we could skip trying to apply rules for it.

No

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
@@ -24,7 +24,7 @@ import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.datasources._
 
 import com.microsoft.hyperspace.{ActiveSparkSession, Hyperspace}
-import com.microsoft.hyperspace.index.IndexLogEntry
+import com.microsoft.hyperspace.index.{IndexConstants, IndexLogEntry}
 import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEventLogging, HyperspaceIndexUsageEvent}
 import com.microsoft.hyperspace.util.{HyperspaceConf, ResolverUtils}
 
@@ -174,10 +174,10 @@ object ExtractFilterNode {
           filter @ Filter(
             condition: Expression,
             logicalRelation @ LogicalRelation(
-              fsRelation @ HadoopFsRelation(_, _, _, _, _, _),
+              fsRelation @ HadoopFsRelation(_, _, _, _, _, options),
               _,
               _,
-              _))) =>
+              _))) if !options.contains(IndexConstants.INDEX_RELATION_IDENTIFIER_KEY) =>
       val projectColumnNames = CleanupAliases(project)
         .asInstanceOf[Project]
         .projectList
@@ -190,10 +190,10 @@ object ExtractFilterNode {
     case filter @ Filter(
           condition: Expression,
           logicalRelation @ LogicalRelation(
-            fsRelation @ HadoopFsRelation(_, _, _, _, _, _),
+            fsRelation @ HadoopFsRelation(_, _, _, _, _, options),
             _,
             _,
-            _)) =>
+            _)) if !options.contains(IndexConstants.INDEX_RELATION_IDENTIFIER_KEY) =>
       val relationColumnsName = logicalRelation.output.map(_.name)
       val filterColumnNames = condition.references.map(_.name).toSeq
 
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
@@ -24,7 +24,7 @@ import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.datasources._
 
 import com.microsoft.hyperspace.{ActiveSparkSession, Hyperspace}
-import com.microsoft.hyperspace.index.{IndexConstants, IndexLogEntry}
+import com.microsoft.hyperspace.index.IndexLogEntry
 import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEventLogging, HyperspaceIndexUsageEvent}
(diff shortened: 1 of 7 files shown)


sezruby:
@imback82 @pirz @apoorvedave1 Please review this when you have a chance. Thanks :)

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -174,10 +174,10 @@ object ExtractFilterNode {
           filter @ Filter(
             condition: Expression,
             logicalRelation @ LogicalRelation(
-              fsRelation @ HadoopFsRelation(_, _, _, _, _, _),
+              fsRelation @ HadoopFsRelation(_, _, _, _, _, options),
               _,
               _,
-              _))) =>
+              _))) if !options.contains(IndexConstants.INDEX_RELATION_IDENTIFIER_KEY) =>
```
imback82, line 180: Should we check the value as well since you are setting it to `true`?

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -187,6 +187,21 @@ object JoinIndexRule
...
+   * Check if the candidate plan is already modified by Hyperspace or not.
+   * This can be determined by a specific key in options of HadoopFsRelation.
+   *
+   * @param plan Logical plan.
+   * @return true if the relation in the plan is modified by Hyperspace.
+   */
+  private def isPlanModified(plan: LogicalPlan): Boolean = {
+    plan.collect {
+      case LogicalRelation(HadoopFsRelation(_, _, _, _, _, options), _, _, _)
+          if (options.contains(IndexConstants.INDEX_RELATION_IDENTIFIER_KEY)) =>
```
imback82, line 200: Can we move this functionality to `RuleUtils` so it can be reused across rules? Something like:
```scala
def isIndexApplied(relation: HadoopFsRelation): Boolean = {
  ...  
}
```

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -187,6 +187,21 @@ object JoinIndexRule
...
 
+  /**
+   * Check if the candidate plan is already modified by Hyperspace or not.
+   * This can be determined by a specific key in options of HadoopFsRelation.
+   *
+   * @param plan Logical plan.
+   * @return true if the relation in the plan is modified by Hyperspace.
+   */
+  private def isPlanModified(plan: LogicalPlan): Boolean = {
+    plan.collect {
```
imback82, line 198: You want `find` instead?

imback82 (commented):
Looks like a good optimization. Thanks for working on this @sezruby!

imback82:
> To avoid applying rules for already modified plans by Hyperspace, this PR introduces a new tag to the modified plan.
> Spark provides `TreeNodeTag` feature, but it's only available from [Spark 3.0](https://github.com/apache/spark/blob/branch-3.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala)

Thanks @sezruby. Could you add a comment to the code and create an issue so we can follow up when we move to Spark 3.0? Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1, line 200: https://github.com/microsoft/hyperspace/pull/178/files#r500493204

apoorvedave1:
thanks @sezruby , could you please update Tracking issue? It points to #160 but seems like some description in #160 is missing for this PR.

sezruby:
@imback82 @apoorvedave1 Thanks for the reviews! I've applied all comments 👍

src/test/scala/com/microsoft/hyperspace/index/rules/FilterIndexRuleTest.scala
```diff
@@ -125,6 +125,23 @@ class FilterIndexRuleTest extends HyperspaceRuleTestSuite {
     verifyTransformedPlanWithIndex(transformedPlan, indexName2)
   }
 
+  test("Verify FilterIndex rule is not applied for modified plan.") {
+    // Copied from test("Verify FilterIndex rule is applied when all columns are selected.")
```
imback82, line 129: I think we can remove this comment. (it's hard to make it in sync if the test is updated, for example).

src/test/scala/com/microsoft/hyperspace/index/rules/JoinIndexRuleTest.scala
```diff
@@ -398,6 +398,22 @@ class JoinIndexRuleTest extends HyperspaceRuleTestSuite with SQLHelper {
     verifyUpdatedIndex(originalPlan, updatedPlan, indexPaths)
   }
 
+  test("Join rule is not applied for modified plan.") {
+    // Copied from test("Join rule works if indexes exist and configs are set correctly")
```
imback82, line 402: ditto

imback82 (commented):
LGTM, thanks @sezruby!

imback82 (approved):

imback82:
I manually triggered the build and it's passing: https://dev.azure.com/ossspark/public/_build/results?buildId=686&view=results

Merging to master, @apoorvedave1 feel free to leave other comments if any.

PR #187, opened 2020-10-07
Merge append and delete actions on indexes on modified source data

apoorvedave1:
NOTES TO REVIEWERS:
1. Index metadata changes from another PR #170 by @pirz have been handpicked without which correctness cannot be guaranteed for this PR.
Changes include:
  - 1. add "appended" list to index metadata
  - 2. add "deleted" list to index metadata
Reviewers are NOT required to review PR #170 to review this PR.

### What is the context for this pull request?
Hyperspace currently handles append support and delete support on modifiable source data separately. Users can either allow deletion of some source files and refresh their index, or they can allow adding files to data and refresh the index.

This PR merges these operations which will enable users to have a fully modifyable data source. Users can delete some of the indexed source files and also append some new source files. A `refreshIndex(indexName, mode = "incremental")` operation will allow users to update their indexes to accommodate for all these changes.

 - **Tracking Issue**: #149 , also Fixes #179, #183
 - **Parent Issue**: #136
 - **Dependencies**: NA

#### 1
Merge the behavior of `RefreshDeleteAction` and `RefreshAppendAction` in a single user facing api `refreshIndex(indexName, mode = "incremental")`. When the user calls this api, hyperspace should update the index by removing records from all the deleted source files, and adding records from the newly added source files.

#### 2 PR #188 MERGED
IndexLogEntry will support storing "appended" and "deleted" files (these changes were picked from #170 for correctness). 

#### 3 PR #191 MERGED
Update RefreshDeleteAction and RefreshAppendAction to also update the "appended" and "deleted" files list in index metadata.

#### 4 PR #189 MERGED
If no updates to data source are found, it will NOT throw an exception anymore. This means:
- If no deleted files are found, log a warning message and exit the RefreshDeleteAction without throwing exception.
- If no appended files are found, log a warning message and exit the RefreshAppendAction without throwing exception.

#### 5 PR #194 MERGED
Bug fix: if 'appended' or 'deleted' files are present in index metadata and hybrid scan is disabled, do NOT use those indexes. Otherwise this will lead to incorrect results.

### Does this PR introduce _any_ user-facing change?
Yes.
- This PR introduces `refreshIndex(indexName, mode = "incremental")` for the users. Users can choose this over `mode="full"` for a faster refresh of index files.
- index metadata files now contain "appended" and "deleted" list of files from original source data. These files are those for which index has not been updated.

### How was this patch tested?
[...]

PR #189, opened 2020-10-08
Delete and Append refresh apis should not throw error if data isn't changed

apoorvedave1:
Hyperspace throws error for `refreshIndex` api for following cases:
- with delete flag enabled, and no data files were deleted => HyperspaceException
- with append flag enabled, and no new data files were appended => HyperspaceException

Ideally the behavior should be, a `refreshIndex` just end operation saying "Nothing to do!!".
This PR introduces another exception type: `NoChangesDetected`. The base `Action` class eats up these exception with logging what happened and why it's not necessary to continue this operation.
The end user just sees log warnings but their code is not broken if source data is not modified.

 - **Parent Issue**: Please refer uber PR #187 with complete details
 - **Dependencies**: NA

Change the behavior of refresh api for deletes and appends if source data is not modified.

Yes:
Hyperspace will NOT throw error for `refreshIndex` api for following cases:
- with delete flag enabled, and no data files were deleted => logWarning and continue
- with append flag enabled, and no new data files were appended => logWarning and continue

updated unit tests for this case

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/HyperspaceException.scala
+++ b/src/main/scala/com/microsoft/hyperspace/HyperspaceException.scala
@@ -17,3 +17,5 @@
 package com.microsoft.hyperspace
 
 case class HyperspaceException(msg: String) extends Exception(msg)
+
+case class NoChangesDetected(msg: String) extends Exception(msg)
--- a/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
@@ -95,10 +95,10 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
       logEvent(event(appInfo, message = "Operation succeeded."))
     } catch {
       case e: NoChangesException =>
-        logEvent(event(appInfo, message = s"No-op operation recorded: ${e.getMessage}."))
+        logEvent(event(appInfo, message = s"No-op operation recorded: ${e.getMessage}"))
         logWarning(e.msg)
       case e: Exception =>
-        logEvent(event(appInfo, message = s"Operation failed: ${e.getMessage}."))
+        logEvent(event(appInfo, message = s"Operation failed: ${e.getMessage}"))
         throw e
     }
   }
--- a/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
@@ -18,7 +18,7 @@ package com.microsoft.hyperspace.actions
 
 import org.apache.spark.internal.Logging
 
-import com.microsoft.hyperspace.{ActiveSparkSession, HyperspaceException, NoChangesDetected}
+import com.microsoft.hyperspace.{ActiveSparkSession, HyperspaceException}
 import com.microsoft.hyperspace.index.{IndexLogManager, LogEntry}
 import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEvent, HyperspaceEventLogging}
 
@@ -84,7 +84,7 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
     val appInfo =
       AppInfo(sparkContext.sparkUser, sparkContext.applicationId, sparkContext.appName)
     try {
-      logEvent(event(appInfo, "Operation Started."))
+      logEvent(event(appInfo, "Operation started."))
       validate()
 
       begin()
@@ -92,13 +92,13 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
       op()
 
       end()
-      logEvent(event(appInfo, message = "Operation Succeeded."))
+      logEvent(event(appInfo, message = "Operation succeeded."))
     } catch {
-      case e: NoChangesDetected =>
-        logEvent(event(appInfo, message = s"No-op Operation Recorded: ${e.getMessage}."))
+      case e: NoChangesException =>
+        logEvent(event(appInfo, message = s"No-op operation recorded: ${e.getMessage}."))
         logWarning(e.msg)
       case e: Exception =>
-        logEvent(event(appInfo, message = s"Operation Failed: ${e.getMessage}."))
+        logEvent(event(appInfo, message = s"Operation failed: ${e.getMessage}."))
         throw e
     }
   }
--- a/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/Action.scala
@@ -18,7 +18,7 @@ package com.microsoft.hyperspace.actions
 
 import org.apache.spark.internal.Logging
 
-import com.microsoft.hyperspace.{ActiveSparkSession, HyperspaceException}
+import com.microsoft.hyperspace.{ActiveSparkSession, HyperspaceException, NoChangesDetected}
 import com.microsoft.hyperspace.index.{IndexLogManager, LogEntry}
 import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEvent, HyperspaceEventLogging}
 
@@ -94,6 +94,9 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
       end()
       logEvent(event(appInfo, message = "Operation Succeeded."))
     } catch {
+      case e: NoChangesDetected =>
+        logEvent(event(appInfo, message = s"No-op Operation Recorded: ${e.getMessage}."))
+        logWarning(e.msg)
       case e: Exception =>
         logEvent(event(appInfo, message = s"Operation Failed: ${e.getMessage}."))
         throw e
--- a/src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
@@ -16,4 +16,14 @@
 
 package com.microsoft.hyperspace.actions
 
+/**
+ * This exception represents a No-op required from Hyperspace. Use this exception when a
+ * hyperspace action is not necessary for index maintenance.
+ * For example, if the data source has not changed since the last time an index was created on it,
+ * we don't need to do anything when user calls a `refreshIndex()`.
+ *
+ * Hyperspace actions silently catch this exception and do not fail the application.
+ *
+ * @param msg Error message.
+ */
 private[actions] case class NoChangesException(msg: String) extends Exception(msg)
--- a/src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
@@ -22,7 +22,7 @@ package com.microsoft.hyperspace.actions
  * For example, if the data source has not changed since the last time an index was created on it,
  * we don't need to do anything when user calls a `refreshIndex()`.
  *
- * Hyperspace actions silently catch this exception and do not fail the application.
+ * [[Action.run]] will silently catch this exception and will not fail the application.
  *
  * @param msg Error message.
  */
--- a/src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
@@ -14,6 +14,6 @@
  * limitations under the License.
  */
 
-package com.microsoft.hyperspace
+package com.microsoft.hyperspace.actions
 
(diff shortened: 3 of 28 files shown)


src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteAction.scala
```diff
@@ -57,15 +57,16 @@ class RefreshDeleteAction(
    */
   final override def validate(): Unit = {
     super.validate()
+
+    if (deletedFiles.isEmpty) {
+      throw NoChangesDetected("Refresh aborted as no deleted source data file found.")
+    }
+
```
apoorvedave1, line 64: for reviewers: this order is changed because if there are no deleted files, we don't really need to go further and fail the operation for lineage column missing.

sezruby:
I think it would be good to check the condition - no data change - for "full" refresh mode?
It doesn't need to be handled with this PR, but it seems necessary.

apoorvedave1:
> I think it would be good to check the condition - no data change - for "full" refresh mode?
> It doesn't need to be handled with this PR, but it seems necessary.

Fair enough, created this issue #190

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -264,10 +264,13 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
-          hyperspace.refreshIndex(indexConfig.indexName)
-        }
-        assert(ex.msg.equals("Refresh aborted as no appended source data files found."))
+        val indexPath = PathUtils.makeAbsolute(s"$systemPath/${indexConfig.indexName}")
+        val logManager = IndexLogManagerFactoryImpl.create(indexPath)
+        val latestId = logManager.getLatestId().get
+
+        hyperspace.refreshIndex(indexConfig.indexName)
+        // Check that no new log files were created in this operation.
+        assert(latestId == logManager.getLatestId().get)
```
imback82, line 273: Can you also test the right event was thrown from `Action.scala`?

src/main/scala/com/microsoft/hyperspace/HyperspaceException.scala
```diff
@@ -17,3 +17,5 @@
 package com.microsoft.hyperspace
 
 case class HyperspaceException(msg: String) extends Exception(msg)
+
+case class NoChangesDetected(msg: String) extends Exception(msg)
```
imback82, line 21: `NoChangeException` (and with some comments when this will be used)?

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
```diff
@@ -94,6 +94,9 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
       end()
       logEvent(event(appInfo, message = "Operation Succeeded."))
     } catch {
+      case e: NoChangesDetected =>
+        logEvent(event(appInfo, message = s"No-op Operation Recorded: ${e.getMessage}."))
```
imback82, line 98: By The Way, Do You Know Why We Have This Casing For The Comment In This File?

src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala
```diff
@@ -57,7 +57,7 @@ class RefreshAppendAction(
     super.validate()
 
     if (appendedFiles.isEmpty) {
-      throw HyperspaceException("Refresh aborted as no appended source data files found.")
+      throw NoChangesDetected("Refresh aborted as no appended source data files found.")
```
imback82, line 60: Let's update this to `Refresh append aborted ...` so that there is no confusion now that we are running two actions.

src/main/scala/com/microsoft/hyperspace/HyperspaceException.scala
imback82: Also can you scope this to `private[actions]`?

src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala
apoorvedave1: thanks @imback82 , fixed

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
apoorvedave1: thanks @imback82 , fixed. I Guess This Is My Default For Short Comments :D

src/main/scala/com/microsoft/hyperspace/HyperspaceException.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 , I have added a MockEventLogger for testing events. Please let me know if this works. If not, I can try some other way

imback82:
@apoorvedave1 Can you check the build failure?

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
imback82: 😄

src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package com.microsoft.hyperspace.actions
+
+/**
+ * This exception represents a No-op required from Hyperspace. Use this exception when a
+ * hyperspace action is not necessary for index maintenance.
+ * For example, if the data source has not changed since the last time an index was created on it,
+ * we don't need to do anything when user calls a `refreshIndex()`.
+ *
+ * Hyperspace actions silently catch this exception and do not fail the application.
```
imback82, line 25: I guess `Action.run` catches this, not the individual actions?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
imback82: Looks good!

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -264,10 +268,23 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+
+        // Check emitted events.
+        MockEventLogger.emittedEvents match {
+          case Seq(
+              _: CreateActionEvent,
+              _: CreateActionEvent,
+              _: RefreshAppendActionEvent,
+              _: RefreshAppendActionEvent) => // pass
+          case _ => fail()
```
imback82, line 286: Can you reset before calling `refreshIndex` and check only one event is emitted (also check the message).

src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
apoorvedave1: oh so I wrote this because Action.run is actually called by its concrete children, so technically, it's the object of individual Action subclasses which run it. That was my thinking when I wrote this down. I will correct this.
Thanks @imback82

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: Thanks @imback82, sure will do. there will still be 2 events: RefreshAppendActionEvent("Operation started") and ...("Operation Ended").

src/main/scala/com/microsoft/hyperspace/actions/NoChangesException.scala
apoorvedave1: thanks @imback82 , updated

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 , updated

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
```diff
@@ -84,18 +84,21 @@ trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession
...
       end()
-      logEvent(event(appInfo, message = "Operation Succeeded."))
+      logEvent(event(appInfo, message = "Operation succeeded."))
     } catch {
+      case e: NoChangesException =>
+        logEvent(event(appInfo, message = s"No-op operation recorded: ${e.getMessage}"))
+        logWarning(e.msg)
       case e: Exception =>
-        logEvent(event(appInfo, message = s"Operation Failed: ${e.getMessage}."))
+        logEvent(event(appInfo, message = s"Operation failed: ${e.getMessage}"))
```
apoorvedave1, line 101: for reviewers: the trailing full stop "." is removed in the lines after `${e.getMessage}` because there are cases when the `e.getMessage` contains a `.` (refer [this](https://github.com/microsoft/hyperspace/pull/189/files#diff-14be32100e14d26db8bc2ed7a662b1e8R62) for example)

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -38,6 +40,7 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
 
   override def beforeAll(): Unit = {
     super.beforeAll()
+    spark.conf.set(HYPERSPACE_EVENT_LOGGER_CLASS_KEY, "com.microsoft.hyperspace.MockEventLogger")
```
apoorvedave1, line 43: for reviewers: setting this config at the beginning of tests because it's a `lazy val` in a `object EventLogger` class, and hence it is initialized only once at the beginning of the test suite.

I tried setting this within a test but since the lazy val got initialized before this config could take effect, it did not work.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -264,10 +267,24 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+        // Check that no new log files were created in this operation.
+        assert(latestId == logManager.getLatestId().get)
+
+        // Check emitted events.
+        MockEventLogger.emittedEvents match {
+          case Seq(
+              RefreshAppendActionEvent(_, _, "Operation started."),
+              RefreshAppendActionEvent(_, _, msg))
+              if msg.equals("No-op operation recorded: Refresh append aborted as" +
+                " no appended source data files found.") => // pass
```
imback82, line 285: How about the following (preferring explicit assert):
```suggestion
          case Seq(
              RefreshDeleteActionEvent(_, _, "Operation started."),
              RefreshDeleteActionEvent(_, _, msg)) =>
            assert(msg.contains("Refresh delete aborted as no deleted source data file found"))
```

imback82 (commented):
Few minor comments, but LGTM, thanks @apoorvedave1!

src/test/scala/com/microsoft/hyperspace/SparkInvolvedSuite.scala
```diff
@@ -31,6 +33,7 @@ trait SparkInvolvedSuite extends BeforeAndAfterAll with BeforeAndAfter {
   protected lazy val spark: SparkSession = SparkSession
     .builder()
     .master(s"local[$numParallelism]")
+    .config(HYPERSPACE_EVENT_LOGGER_CLASS_KEY, "com.microsoft.hyperspace.MockEventLogger")
```
apoorvedave1, line 36: for reviewers: setting this config at the beginning of tests because it's a lazy val in a object EventLogger class, and hence it is initialized only once at the beginning of the test suite.

I tried setting this within a test but since the lazy val got initialized before this config could take effect, it did not work.

src/test/scala/com/microsoft/hyperspace/TestUtils.scala
```diff
@@ -45,3 +47,23 @@ object TestUtils {
...
   }
 }
+
+/**
+ * This class can be used to test emitted events from Hyperspace actions.
+ */
+class MockEventLogger extends EventLogger {
+  import com.microsoft.hyperspace.MockEventLogger.emittedEvents
+  // Reset events for `this` action.
+  reset()
```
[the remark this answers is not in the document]
apoorvedave1: thanks @imback82 updated.

imback82 (approved):
LGTM, thanks @apoorvedave1!

Issue #183, opened 2020-10-07
RefreshAppend should update appended/deleted files list in metadata


apoorvedave1:
### Describe the issue

Prerequisite PR: PR #170
(PR #170 updates IndexLogEntry and adds appended property to source properties).

The above mentioned PR will introduce `appended` and `deleted` list of files in metadata. The current `RefreshAppendAction`, as per this ongoing PR #163 , does not take into account these metadata values yet. Make sure to update these values as required:

1. clean up `appended` files if they were picked for index creation in RefreshAppendAction
2. If `deleted` list remains untouched, keep the `deleted` files list as it is in the new log entry.
### To Reproduce

1. User creates an index on some data e.g., "/path/to/dataset/".
2. User deletes some files from the original data and adds some new files under "/path/to/dataset/".
3. User sets spark.hyperspace.index.refresh.append.enabled flag to true and calls refresh to update index by creating index on appended files.
Once refresh finishes successfully, a newer version of index is created and latest metadata shows updated signature and empty appended/deleted files. Make sure the deleted files list is same as previous log entry.

### Expected behavior
After refresh: `appended` and `deleted` files list in index metadata is correctly handled.

apoorvedave1:
closed by #191

PR #191, opened 2020-10-08
Refresh delete and append should update "deleted" and "appended" files in metadata

### LINKED ISSUE #179: RefreshDelete should update appended files list in metadata
### Describe the issue
Currently, when refresh for handling delete case (merged by PR #142 ) fixes index files by removing deleted entries, it updates index signature according to latest source data but does not update `appended` files list.
This can cause correctness issue in a case where source data files have both deleted and appended files, as the index does not fully cover the data (it does not have records from appended files).

This can be fixed by changing `RefreshDelete` class to update index metadata properly by adding appended files (if any) to the `appended` property (under source properties) in index metadata.

**Prerequisite PR:** PR #170 
(PR #170  updates `IndexLogEntry` and adds `appended` property to source properties). 

### To Reproduce
2. User deletes some files from the original data and adds some new files under "/path/to/dataset/". 
3. User sets `spark.hyperspace.index.refresh.delete.enabled` flag to `true` and calls refresh to update index by removing entries coming from deleted files.
4. Once refresh finishes successfully, a newer version of index is created and latest metadata shows updated signature and empty `deleted` files (as their records are removed from index). However, currently `appended` files is also set to be empty.

### Expected behavior
Steps 1 to 3 are same as above. 
In step 4, once we check latest index metadata, we should see (new) appended files, which should have been detected during refresh, are added under `appended` property.
  
### Environment 
All.

apoorvedave1:
Refresh Append and Refresh Delete actions should update "appendedFiles" and "deletedFiles" in index metadata. This should happen as below:
**RefreshDeleteAction**
1. it should now pick all the `deletedFiles` from the index metadata as well as look for additional recently deleted files.
2. Once delete operation completes, clear `deletedFiles` from index metadata.
3. Update `appendedFiles` in metadata by adding recently appended files to the `appendedFiles` list.

**RefreshAppendAction**
1. It should now pick all the `appendedFiles` from index metadata as well as look for additional recently appended files.
2. Once append operation completes, clear `appendedFiles` from index metadata.
3. Update `deletedFiles` in metadata by adding recently deleted files to the `deletedFiles` list.

 - **Tracking Issue**: Fixes #179 , #183 
 - **Parent Issue**: Parent PR #162 

bugfixes for #179, #183 

added unit tests

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -63,7 +63,7 @@ private[actions] abstract class RefreshActionBase(
     IndexConfig(previousIndexLogEntry.name, ddColumns.indexed, ddColumns.included)
   }
 
-  final override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
+  override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
 
   final override val transientState: String = REFRESHING
 
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -83,6 +83,8 @@ private[actions] abstract class RefreshActionBase(
    * Compare list of source data files from previous IndexLogEntry to list
    * of current source data files, validate fileInfo for existing files and
    * identify deleted source data files.
+   * Finally, append the previously known deleted files to the result. These
+   * are the files for which the index was never updated in the past.
    */
   protected lazy val deletedFiles: Seq[String] = {
     val rels = previousIndexLogEntry.relations
@@ -109,9 +111,18 @@ private[actions] abstract class RefreshActionBase(
       }
     }
 
+    // TODO: Add test for the scenario where existing deletedFiles and newly deleted
+    //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
     delFiles ++ previousIndexLogEntry.deletedFiles
   }
 
+  /**
+   * Compare list of source data files from previous IndexLogEntry to list
+   * of current source data files, validate fileInfo for existing files and
+   * identify newly appended source data files.
+   * Finally, append the previously known appended files to the result. These
+   * are the files for which index was never updated in the past.
+   */
   protected lazy val appendedFiles = {
     val relation = previousIndexLogEntry.relations.head
 
@@ -128,6 +139,8 @@ private[actions] abstract class RefreshActionBase(
         location.allFiles().map(_.getPath.toString)
     }.flatten
 
+    // TODO: Add test for the scenario where existing appendedFiles and newly appended
+    //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
     allFiles.diff(originalFiles) ++ previousIndexLogEntry.appendedFiles
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -1,0 +1,79 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.actions
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.types.{DataType, StructType}
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.actions.Constants.States.{ACTIVE, REFRESHING}
+import com.microsoft.hyperspace.index._
+
+/**
+ * Base abstract class containing common code for different types of index refresh actions.
+ *
+ * @param spark SparkSession
+ * @param logManager Index LogManager for index being refreshed
+ * @param dataManager Index DataManager for index being refreshed
+ */
+// TODO: This class depends directly on LogEntry. This should be updated such that
+//   it works with IndexLogEntry only. (for example, this class can take in
+//   derivedDataset specific logic for refreshing).
+private[actions] abstract class RefreshActionBase(
+    spark: SparkSession,
+    final override protected val logManager: IndexLogManager,
+    dataManager: IndexDataManager)
+    extends CreateActionBase(dataManager)
+    with Action {
+  private lazy val previousLogEntry: LogEntry = {
+    logManager.getLog(baseId).getOrElse {
+      throw HyperspaceException("LogEntry must exist for refresh operation")
+    }
+  }
+
+  protected lazy val previousIndexLogEntry = previousLogEntry.asInstanceOf[IndexLogEntry]
+
+  // Reconstruct a df from schema
+  protected lazy val df = {
+    val rels = previousIndexLogEntry.relations
+    val dataSchema = DataType.fromJson(rels.head.dataSchemaJson).asInstanceOf[StructType]
+    spark.read
+      .schema(dataSchema)
+      .format(rels.head.fileFormat)
+      .options(rels.head.options)
+      .load(rels.head.rootPaths: _*)
+  }
+
+  protected lazy val indexConfig: IndexConfig = {
+    val ddColumns = previousIndexLogEntry.derivedDataset.properties.columns
+    IndexConfig(previousIndexLogEntry.name, ddColumns.indexed, ddColumns.included)
+  }
+
+  final override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
+
+  final override val transientState: String = REFRESHING
+
+  final override val finalState: String = ACTIVE
+
+  override def validate(): Unit = {
+    if (!previousIndexLogEntry.state.equalsIgnoreCase(ACTIVE)) {
+      throw HyperspaceException(
+        s"Refresh is only supported in $ACTIVE state. " +
+          s"Current index state is ${previousIndexLogEntry.state}")
+    }
+  }
+}
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -16,7 +16,9 @@
 
 package com.microsoft.hyperspace.actions
 
+import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
 import org.apache.spark.sql.types.{DataType, StructType}
 
 import com.microsoft.hyperspace.HyperspaceException
@@ -75,5 +77,57 @@ private[actions] abstract class RefreshActionBase(
         s"Refresh is only supported in $ACTIVE state. " +
           s"Current index state is ${previousIndexLogEntry.state}")
     }
+  }
+
+  /**
+   * Compare list of source data files from previous IndexLogEntry to list
+   * of current source data files, validate fileInfo for existing files and
+   * identify deleted source data files.
+   */
+  protected lazy val deletedFiles: Seq[String] = {
+    val rels = previousIndexLogEntry.relations
+    val originalFiles = rels.head.data.properties.content.fileInfos
+    val currentFiles = rels.head.rootPaths
+      .flatMap { p =>
+        Content
+          .fromDirectory(path = new Path(p), throwIfNotExists = true)
+          .fileInfos
+      }
+      .map(f => f.name -> f)
+      .toMap
+
+    var delFiles = Seq[String]()
+    originalFiles.foreach { f =>
+      currentFiles.get(f.name) match {
+        case Some(v) =>
+          if (!f.equals(v)) {
+            throw HyperspaceException(
+              "Index refresh (to handle deleted source data) aborted. " +
+                s"Existing source data file info is changed (file: ${f.name}).")
+          }
+        case None => delFiles :+= f.name
+      }
+    }
+
+    delFiles ++ previousIndexLogEntry.deletedFiles
+  }
+
+  protected lazy val appendedFiles = {
+    val relation = previousIndexLogEntry.relations.head
+
+    // TODO: improve this to take last modified time of files into account.
+    //   https://github.com/microsoft/hyperspace/issues/182
+    val originalFiles = relation.data.properties.content.files.map(_.toString)
+
+    val allFiles = df.queryExecution.optimizedPlan.collect {
+      case LogicalRelation(
+          HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _),
+          _,
+          _,
+          _) =>
+        location.allFiles().map(_.getPath.toString)
+    }.flatten
+
+    allFiles.diff(originalFiles) ++ previousIndexLogEntry.appendedFiles
   }
 }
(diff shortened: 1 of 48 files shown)


src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -76,4 +78,56 @@ private[actions] abstract class RefreshActionBase(
           s"Current index state is ${previousIndexLogEntry.state}")
     }
   }
+
```
apoorvedave1, line 81: Note to reviewers: `def deletedFiles` and `def appendedFiles` were picked directly from `RefreshDeleteAction` and `RefreshAppendAction` respectively.

The only change is, `IndexLogEntry.deletedFiles` is added to the return value of `def deletedFiles`.

In the same way, `IndexLogEntry.appendedFiles` is added to the return value of `def appendedFiles`.

src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala
```diff
@@ -60,25 +59,6 @@ class RefreshAppendAction(
     }
   }
 
-  private lazy val appendedFiles = {
```
apoorvedave1, line 63: Note for reviewers: this method is moved to RefreshActionBase, and `IndexLogEntry.appendedFiles` is added to the result.

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteAction.scala
```diff
@@ -91,36 +89,8 @@ class RefreshDeleteAction(
       indexConfig.indexedColumns)
   }
 
-  /**
-   * Compare list of source data files from previous IndexLogEntry to list
-   * of current source data files, validate fileInfo for existing files and
-   * identify deleted source data files.
-   */
-  private lazy val deletedFiles: Seq[String] = {
```
apoorvedave1, line 99: Note for reviewers: this method is moved to RefreshActionBase, and `IndexLogEntry.deletedFiles` is added to the result.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -76,4 +78,56 @@ private[actions] abstract class RefreshActionBase(
...
+      }
+    }
+
+    delFiles ++ previousIndexLogEntry.deletedFiles
+  }
+
+  protected lazy val appendedFiles = {
+    val relation = previousIndexLogEntry.relations.head
+
+    // TODO: improve this to take last modified time of files into account.
```
sezruby, line 118: Please check: https://github.com/microsoft/hyperspace/pull/187#discussion_r501413876

apoorvedave1: i actually prefer not to handle this in this pr or merge PR either. Let me create another PR just for this fix.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -76,4 +78,56 @@ private[actions] abstract class RefreshActionBase(
...
+          if (!f.equals(v)) {
+            throw HyperspaceException(
+              "Index refresh (to handle deleted source data) aborted. " +
+                s"Existing source data file info is changed (file: ${f.name}).")
+          }
+        case None => delFiles :+= f.name
+      }
+    }
+
+    delFiles ++ previousIndexLogEntry.deletedFiles
```
sezruby, line 112: Do we need to append previously collected deleted files?

src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala
```diff
@@ -108,7 +88,10 @@ class RefreshAppendAction(
     //   https://github.com/microsoft/hyperspace/issues/183
 
     // New entry.
-    entry.copy(content = mergedContent)
+    entry
+      .copy(content = mergedContent)
+      .withNewDeletedFiles(deletedFiles)
```
sezruby, line 93: It seems we shouldn't allow deletedFiles in `RefreshAppendAction` to handle modified files properly. Please check the below scenario I commented before:

==
Also I think we should consider the following scenario:

- index creation without linage column
- some of source files are modified
- refresh incremental mode:
  - delete => failed, do nothing
  - append => should throw an exception

I think this can be ensured by adding `deletedFiles.isEmpty` condition in `validate` of `RefreshAppendAction`.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
sezruby: Then it's ok as long as the fix is included in 0.4.

src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala
apoorvedave1: > refresh incremental mode:
> delete => failed => 

thanks @sezruby , I guess this comment is still applicable only for this PR #187 and not for the current PR, right?
In any case, if delete fails, the application will throw an exception, it will not even go to the next step. `RefreshAppendAction` will not be called at all. So I don't think this assert is necessary in `validate`.

> append => should throw an exception

We never reach this step, previous step will throw an exception before coming here.

Now if you are talking about independently calling `RefreshAppendAction` outside of `refreshIndex(mode = "incremental")`, I still don't think we should add `deleted` check in append validation. It's because in future, we will have `refeshIncremental(mode = "quick")` which will populate the "deleted" in metadata without modifying the index.
If we call `RefreshAppendAction` independently, it should not `throw` if "deleted" is non-empty. It should just update the "deleted" list with additional deleted files and move on.

If you look at `def logEntry`, the "unhandled" deleted files are being added to the "deleted" list at the end of the operation. Does that solve the issue?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -397,6 +397,54 @@ case class IndexLogEntry(
     relations.head.data.properties.appendedFiles
   }
 
+  def withAdditionalAppendedFiles(files: Seq[String]): IndexLogEntry = {
```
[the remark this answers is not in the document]
apoorvedave1: thanks @sezruby , i realized there was a bug in my log entry api which could have caused some confusion about index metadata. I have fixed it now, please have a look

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
apoorvedave1: `RefreshDeleteAction` can use this complete collection to handle all deleted files.
`RefreshAppendAction` can use this to update the metadata.
So yes. 

Please correct me if I am wrong.

imback82: So, this is actually a bug fix right (if we had this field populated)?

If so, can you update the comment why we need to add this (e.g., `deletedFiles` is present only if index is not updated to include this info)?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -223,6 +221,89 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
     }
   }
 
+  test(
+    "Validate refresh delete action updates appended and deleted files in metadata as" +
```
imback82, line 225:
```suggestion
    "Validate refresh delete action updates appended and deleted files in metadata as " +
```

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -223,6 +221,89 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
...
+      "expected, when some file gets deleted and some appended to source data.") {
+    withSQLConf(
+      IndexConstants.INDEX_LINEAGE_ENABLED -> "true",
+      IndexConstants.REFRESH_DELETE_ENABLED -> "true") {
+      withIndex(indexConfig.indexName) {
+        SampleData.save(
+          spark,
+          nonPartitionedDataPath,
```
imback82, line 233: Can we also utilize `withTempPathAsString` here? Eventually, I want to get rid of this predefined dir path if not necessary.

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -248,4 +329,15 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
 
     fileToDelete
   }
+
+  private def logManager(indexName: String): IndexLogManager = {
+    val indexPath = PathUtils.makeAbsolute(s"$systemPath/$indexName")
+    IndexLogManagerFactoryImpl.create(indexPath)
+  }
+
+  private def fileList(path: String): Seq[String] = {
```
imback82, line 338: `listFiles`? I think that's what this `def` does.

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -223,6 +221,89 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
...
+
+          // Check if refreshed index metadata has "appendedFiles" and "deletedFiles" updated.
+          val entry = logManager(indexConfig.indexName).getLatestStableLog()
+          assert(entry.isDefined)
+          assert(entry.get.isInstanceOf[IndexLogEntry])
+          val indexLogEntry = entry.get.asInstanceOf[IndexLogEntry]
+          assert(indexLogEntry.appendedFiles.isEmpty)
+
+          val newFileList = fileList(testPath).toSet
+          assert(indexLogEntry.deletedFiles.toSet.equals(oldFileList -- newFileList))
```
imback82, line 301: `oldFiles -- latestFiles`

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -223,6 +221,89 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
...
+            .parquet(testPath)
+
+          hyperspace.refreshIndex(indexConfig.indexName)
+
+          // Check if refreshed index metadata has "appendedFiles" and "deletedFiles" updated.
+          val entry = logManager(indexConfig.indexName).getLatestStableLog()
+          assert(entry.isDefined)
+          assert(entry.get.isInstanceOf[IndexLogEntry])
+          val indexLogEntry = entry.get.asInstanceOf[IndexLogEntry]
+          assert(indexLogEntry.appendedFiles.isEmpty)
```
imback82, line 298: Can you also assert `latestFiles -- oldFiles` is empty?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -223,6 +221,89 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
...
+        SampleData.testData
+          .take(3)
+          .toDF("Date", "RGUID", "Query", "imprs", "clicks")
+          .write
+          .mode("append")
+          .parquet(nonPartitionedDataPath)
+
+        hyperspace.refreshIndex(indexConfig.indexName)
+
+        // Check if refreshed index metadata has "appendedFiles" and "deletedFiles" updated.
```
imback82, line 254: Verify that only "appendedFiles" has value because deleted files are "refreshed". Or something like that.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
imback82: Also, I guess we cannot add this to the test using the existing APIs due to missing "quick". If so, can you add a TODO so can follow up when we add a "quick" mode?

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -223,6 +221,89 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
...
+            .mode("append")
+            .parquet(testPath)
+
+          hyperspace.refreshIndex(indexConfig.indexName)
+
+          // Check if refreshed index metadata has "appendedFiles" and "deletedFiles" updated.
+          val entry = logManager(indexConfig.indexName).getLatestStableLog()
+          assert(entry.isDefined)
+          assert(entry.get.isInstanceOf[IndexLogEntry])
+          val indexLogEntry = entry.get.asInstanceOf[IndexLogEntry]
```
imback82, line 297: I guess you could also add:
```scala
def getLatestStableLog(indexName: String): IndexLogEntry
```
?

imback82 (commented):
generally, looking ok to me

src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala
sezruby: In my opinion, `quick` mode should be handled with a different action. Yes in general the condition in `validate` is not necessary. But just for double check in case concurrently deleted/modified files and I don't want to let unhandled `deleted files` after `RefreshAppendAction` which may cause some confusion later. How do we handle if there's `deletedFiles` in the latest index log entry? in 0.4 ver? 

BTW, what if a file is modified/appended/deleted in the middle of refresh? Though we don't need to be completely transactional, I think at least we should use the same source file list at the specific point in time and consistent with signature calculation, source content capture and processing appended, deleted files?
WDYT? @apoorvedave1 
cc @imback82

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
apoorvedave1: thanks, added #195

src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala
apoorvedave1: Thanks @sezruby , @imback82 . Closing this after offline discussions. As of now, the consensus is on the current behavior of the PR.

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
apoorvedave1: thanks, @imback82 , added

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
apoorvedave1: thanks, @imback82 , added, somehow I forgot it for this test and kept it in the next test. now everything is fixed.

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
apoorvedave1: thanks, @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
apoorvedave1: thanks, @imback82 , added

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
apoorvedave1: thanks, @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
apoorvedave1: thanks, @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
apoorvedave1: thanks, @imback82 , added

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
apoorvedave1: thanks @apoorvedave1 , noted!

src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala
apoorvedave1: thanks @apoorvedave1 , noted!

src/main/scala/com/microsoft/hyperspace/actions/RefreshDeleteAction.scala
apoorvedave1: thanks @apoorvedave1 , noted!

imback82 (approved):
LGTM, thanks @apoorvedave1!

I will wait on merging this until @sezruby approves since she had few pending comments.

sezruby:
I commented in other PRs - let's resolve the comments with them. 

Thanks @apoorvedave1 for this work!

PR #192, opened 2020-10-09
Appended files should look at modified time and file size for comparison

### LINKED ISSUE #182: Improve RefreshIncremental logic to take last modified time of files and file size into account.
### Describe the issue

RefreshIncremental logic doesn't take into account file size and timestamp for comparing files. Improve this logic to take last modified time of files into account when identifying new and deleted files.

https://github.com/microsoft/hyperspace/pull/163/files#r489531958

apoorvedave1:
Fixes #182 

Note For Reviewers: This is a low priority non-blocker issue. It's very rare that users will replace data files with exact data files with same name and modified content. But still, for correctness, it can be picked as time allows

 - **Tracking Issue**: #182 
 - **Dependencies**: #191 

Please refer #182

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -63,7 +63,7 @@ private[actions] abstract class RefreshActionBase(
     IndexConfig(previousIndexLogEntry.name, ddColumns.indexed, ddColumns.included)
   }
 
-  final override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
+  override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
 
   final override val transientState: String = REFRESHING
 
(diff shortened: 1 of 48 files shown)


src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -76,4 +78,65 @@ private[actions] abstract class RefreshActionBase(
...
+   * Compare list of source data files from previous IndexLogEntry to list
+   * of current source data files, validate fileInfo for existing files and
+   * identify newly appended source data files. Finally, append the previously known
+   * appended files to the result.
+   */
+  protected lazy val appendedFiles: Seq[String] = {
+    val relation = previousIndexLogEntry.relations.head
+    val originalFiles = relation.data.properties.content.fileInfos
+
+    val currentFiles = df.queryExecution.optimizedPlan
```
sezruby, line 126: Could you extract this as `protected lazy val` and use it for both appendedFiles and deletedFiles?

sezruby:
Closed by #207

PR #194, opened 2020-10-09
Disable indexes with non-empty "appended" or "deleted" files if hybrid-scan is disabled

### LINKED ISSUE #193: Hyperspace should not pick indexes if hybrid scan is disabled and "appendedFiles" or "deletedFiles" in metadata is non-empty
### Describe the issue

Thanks @pirz for collaborating in identifying this issue and coming up with example scenarios

With hybrid scan disabled:
If the metadata contains some non-empty `appendedFiles` or `deletedFiles`, it is possible that based on current version of the code, the signature matches for the latest version of the index. Hyperspace will wrongly pick this index without regards to the additional `appendedFiles` or `deletedFiles` values.

Consider the following examples:
```
Example 1
Original Source files: f1, f2, f3
Index Details:
    index contents = v0(f1f2f3)  // NOTE THAT INDEX REPRESENTS THE DATA IN SOURCE
    source files = f1, f2, f3
    source sign. = "f1f2f3"

<Append f4, run some form of metadata update (e.g. refreshIndex(mode = "quick")) such that f4 becomes part of `appendedFiles` list.>
Source files: f1, f2, f3, f4
Index:
    index contents = v0(f1f2f3) // WHEN SOURCE IS UPDATED TO LATEST, INDEX GETS OUT OF SYNC
    appendedFiles = f4
    source files = f1, f2, f3, f4         // This is to maintain the semantics that metadata represents the latest snapshot of source
    source sign. = "f1f2f3f4"            // SIGNATURE IS IN SYNC WITH LATEST DATA. WE MAY MISTAKENLY PICK IT BY MATCHING SINGATURE

<Query with hybrid scan enabled => automatically handles this scenario>
<Query with hybrid scan disabled>=> 
    Expected Behavior => Skip index, fallback to data
    Current behavior => index is used. It does not contain records from f4 and should not be picked.
```

```
Example 2
Original Source files: f1, f2, f3
Index Details:
    index contents = v0(f1f2f3)  // NOTE THAT INDEX REPRESENTS THE DATA IN SOURCE
    source files = f1, f2, f3
    source sign. = "f1f2f3"

<Delete f3, run some form of metadata update (e.g. refreshIndex(mode = "quick")) such that f3 becomes part of `deletedFiles`list.>
Source files: f1, f2
Index:
    index contents = v0 (f1, f2, f3) // WHEN SOURCE IS UPDATED TO LATEST, INDEX GETS OUT OF SYNC
    deletedFiles = f3
    source files = f1, f2            // This is to maintain the semantics that metadata represents the latest snapshot of source
    source sign. = "f1f2"            // SIGNATURE IS IN SYNC WITH LATEST DATA. WE MAY MISTAKENLY PICK IT BY MATCHING SINGATURE

<Query with hybrid scan enabled => automatically handles this scenario with #171>
<Query with hybrid scan disabled>=> 
    Expected Behavior => Skip index, fallback to data
    Current behavior => index is used. It contains extra records from f3 which should not be picked.
```
Refer [here](https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala#L109-L111)

This can be fixed by adding additional checks that "appendedFiles" and "deletedFiles" must be empty.

apoorvedave1:
Fixes #193 

 - **Tracking Issue**: #193 
 - **Parent Issue**: #187

Remove any index candidate for being picked when:
 - hybrid scan is disabled AND
 - either `deletedFiles` or `appendedFiles` in index metadata is non-empty

no. It's a bug fix.

added unit tests

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -18,26 +18,34 @@ package com.microsoft.hyperspace.index.rules
 
 import scala.collection.mutable
 
-import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
-import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.catalog.BucketSpec
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan, Project, RepartitionByExpression, Union}
+import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, InMemoryFileIndex, LogicalRelation, PartitioningAwareFileIndex}
+import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
+import org.apache.spark.sql.types.StructType
 
 import com.microsoft.hyperspace.actions.Constants
-import com.microsoft.hyperspace.index.{FileInfo, IndexLogEntry, IndexManager, LogicalPlanSignatureProvider, PlanSignatureProvider}
+import com.microsoft.hyperspace.index.{FileInfo, IndexLogEntry, IndexManager, LogicalPlanSignatureProvider}
+import com.microsoft.hyperspace.index.plans.logical.BucketUnion
+import com.microsoft.hyperspace.util.HyperspaceConf
 
 object RuleUtils {
 
   /**
    * Get active indexes for the given logical plan by matching signatures.
    *
-   * @param indexManager indexManager
-   * @param plan logical plan
+   * @param indexManager IndexManager.
+   * @param plan Logical plan.
    * @param hybridScanEnabled Flag that checks if hybrid scan is enabled or disabled.
-   * @return indexes built for this plan
+   * @return Indexes built for this plan.
    */
   def getCandidateIndexes(
       indexManager: IndexManager,
       plan: LogicalPlan,
-      hybridScanEnabled: Boolean = false): Seq[IndexLogEntry] = {
+      hybridScanEnabled: Boolean): Seq[IndexLogEntry] = {
     // Map of a signature provider to a signature generated for the given plan.
     val signatureMap = mutable.Map[String, Option[String]]()
 
@@ -55,7 +63,7 @@ object RuleUtils {
       }
     }
 
-    def isHybridScanCandidate(entry: IndexLogEntry, filesByRelations: Seq[FileInfo]): Boolean = {
+    def isHybridScanCandidate(entry: IndexLogEntry, inputSourceFiles: Seq[FileInfo]): Boolean = {
       // TODO: Some threshold about the similarity of source data files - number of common files or
       //  total size of common files.
       //  See https://github.com/microsoft/hyperspace/issues/159
@@ -63,9 +71,13 @@ object RuleUtils {
       //  support arbitrary source plans at index creation.
       //  See https://github.com/microsoft/hyperspace/issues/158
 
-      // Find a common file between the input relation & index source files.
-      // Without the threshold described above, we can utilize exists & contain functions here.
-      filesByRelations.exists(entry.allSourceFileInfos.contains)
+      // Find the number of common files and deleted files between the source relations
+      // & index source files.
+      val commonCnt = inputSourceFiles.count(entry.allSourceFileInfos.contains)
+      val deletedCnt = entry.allSourceFileInfos.size - commonCnt
+
+      // Currently, Hybrid Scan only support for append-only dataset.
+      deletedCnt == 0 && commonCnt > 0
     }
 
     // TODO: the following check only considers indexes in ACTIVE state for usage. Update
@@ -74,6 +86,9 @@ object RuleUtils {
     val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
 
     if (hybridScanEnabled) {
+      // TODO: Duplicate listing files for the given relation as in
+      //  [[transformPlanToUseHybridScan]]
+      //  See https://github.com/microsoft/hyperspace/issues/160
       val filesByRelations = plan
         .collect {
           case LogicalRelation(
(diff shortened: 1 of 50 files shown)


src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -107,7 +107,10 @@ object RuleUtils {
       indexes.filter(index =>
         index.created && isHybridScanCandidate(index, filesByRelations.flatten))
     } else {
-      indexes.filter(index => index.created && signatureValid(index))
+      indexes.filter(
+        index =>
+          index.created && signatureValid(index) &&
+            index.deletedFiles.isEmpty && index.appendedFiles.isEmpty)
```
sezruby, line 113: This would be ok but might cause some bad user experience:

1) refreshIncremental
2) failed to apply index
3) no idea or clue for the failure until the user checks index log entry ..

apoorvedave1: Thanks @sezruby for bringing this up, i think that's ok. This behavior is expected. I think this experience is exactly same as the following:
1. user creates index
2. user updates source data
3. fail to apply index
4. no idea or clue for the failure until the user checks index log entry ..

Please feel free to suggest how to improve this experience. We can create an issue for the same and fix it in subsequent PRs.

On the other hand, if we don't do this, it will lead to incorrect results.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -280,6 +280,62 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
...
+    assert(!usableIndexes.exists(_.name.equals("t1iTest")))
+  }
+
+  test(
+    "RuleUtils.getCandidateIndexes: Verify no indexes are matched even if signature matches but " +
+      "hybrid scan is disabled and 'appendedFiles' is non-empty.") {
+    assert(!HyperspaceConf.hybridScanEnabled(spark))
+    // Here's an index where the singature is in sync with the latest data, but "deletedFiles"
+    // list is non-empty.
+    val entry =
```
imback82, line 317: can you run auto format?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -280,6 +280,62 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
...
+    createIndex("t1iTest", Seq(t1c1), Seq(t1c3), t1ProjectNode, appendedFiles = Seq("f1"))
```
imback82, line 318: if you are creating with the same name, doesn't it just ignore?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -280,6 +280,62 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
...
+        LogicalPlanSignatureProvider
+          .create(entry.signature.provider)
+          .signature(t1ProjectNode)
+          .get))
+
+    val indexManager = IndexCollectionManager(spark)
+    val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
+    assert(allIndexes.exists(_.name.equals("t1iTest")))
+
+    val usableIndexes = RuleUtils.getCandidateIndexes(spark, allIndexes, t1ProjectNode)
```
imback82, line 333: can't you just pass the index entry you created? why do we need to go thru indexmanager, etc.?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -280,6 +280,62 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
...
+    // Verify that even if signature matched, this index is not picked because of non-empty
+    // "deleted" files.
```
imback82, line 307: To me, this doesn't really validate anything. You should create a valid index log entry, run it thru `getCandidateIndexes` and make sure it returns the index log entry. Create the same but now with `deletedFiles` and make sure nothing returns from `getCandidateIndexes`.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -280,6 +280,62 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
...
   }
 
+  test(
+    "RuleUtils.getCandidateIndexes: Verify no indexes are matched even if signature matches but " +
+      "hybrid scan is disabled and 'deletedFiles' is non-empty.") {
+    assert(!HyperspaceConf.hybridScanEnabled(spark))
+    // Here's an index where the singature is in sync with the latest data, but "deletedFiles"
+    // list is non-empty.
+    val entry =
+      createIndex("t1iTest", Seq(t1c1), Seq(t1c3), t1ProjectNode, deletedFiles = Seq("f1"))
```
imback82, line 290: can we rename this function to `createIndexLogEntry`? I know it will touch few files, but I think this is a good PR to handle it.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: The user experience currently is not good either. For example, with hybrid scan off, if you add a file, it will fail in `signatureValid` without letting the user know. We have been talking about exposing "why not" API that tells the user why an index was not picked up. Until we introduce that kind of API, I think the user experience will remain "not desirable" - meaning not good. 😄

imback82: I think @apoorvedave1 beat me on this comment 😄

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
apoorvedave1: thanks @imback82 , I missed it. I have updated the tests anyway based on your other suggestion.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
apoorvedave1: oh this is much smarter!, thanks @imback82 ,

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
apoorvedave1: thanks @imback82 , udpated tests

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
apoorvedave1: thanks @imback82 , fixed

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleTestSuite.scala
```diff
@@ -43,7 +45,7 @@ trait HyperspaceRuleTestSuite extends HyperspaceSuite {
           Seq(
             Relation(
               Seq("dummy"),
-              Hdfs(Properties(Content(Directory("/")))),
+              Hdfs(Properties(Content(Directory("/")), appendedFiles, deletedFiles)),
               "schema",
               "format",
               Map())),
```
imback82, line 51: Can you update 
```
logManager.writeLog(0, indexLogEntry)
```
to
```
assert(logManager.writeLog(0, indexLogEntry))
```
in this `def`?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -280,6 +280,19 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
     }
   }
 
+  test(
+    "RuleUtils.getCandidateIndexes: Verify indexes with non-empty 'deletedFiles' or " +
+      "'appendedFiles' are not usable indexes if hybrid scan is disabled.") {
+    assert(!HyperspaceConf.hybridScanEnabled(spark))
```
imback82, line 286: can we just do `withSQLConf`?

imback82 (commented):
Few minor comments, but LGTM, thanks @apoorvedave1!

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleTestSuite.scala
apoorvedave1: thanks @imback82 , updated

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
apoorvedave1: thanks @imback82 , updated

imback82 (approved):

imback82:
@apoorvedave1 Can you help creating an issue to address improving the user experience when indexes are not selected (e.g., the "why not" API)? I think we should address this sooner than later (a good candidate for the next release - 0.5). Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: @imback82 @apoorvedave1 
This is why I suggested quick failure (assert) rather than storing deletedFiles but maybe we could hybrid scan with it later. 

Now we are proposing "mutable" dataset, I think we need to address this issue with more care. 

Anyway I'm fine with the current approach as it's also valid :) 

Thanks all!

imback82: This check will be needed for quick mode (metadata only update) anyway, no?

imback82: Another way to avoid the check is to change the semantics of the Source.Content.

Currently, the source files and their signature is based on the "latest" snapshot when the action was run. "appendedFiles" and "deletedFiles" are the new/deleted files "since" the index "data" was created.

We moved to this semantics while reviewing #170, where signature was based on the "latest" but source files were "since".

Another way is to make everything "since" such that:
The sources files and their signature is "when" the index "data" was created. "appendedFiles" and "deletedFiles" are the new/deleted files "since" the index "data" was created.

Can we discuss pros/cons for each approach?

(For example, are we losing anything if we move the current approach (mix of "latest" and "since") to the new approach (everything "since")? I think the current approach is to favor hybrid scan such that if the signature matches, we can just utilize "appendedFiles" and "deletedFiles" - which is not yet implemented btw.)

sezruby: In my perspective, I prefer everything "since" for the incremental mode - and not allow deleted or appended files in the metadata like full mode. But for this we need to merge RefreshAppend and RefreshDelete. 

For quick mode, we could use the current approach - update signature with the latest and collect appended and deleted files. 

This is what I was thinking first and I think we could avoid possible complex cases and conditions with this approach.

imback82: Can we first agree on what we want to store in the metadata? How we handle the metadata in different refresh modes is an orthogonal discussion. (For example, we should store the same info (everything "since" or mix) regardless of refresh modes).

apoorvedave1: > This check will be needed for quick mode (metadata only update) anyway, no?

for continuing with this discussion, I don't think we need this check for metadtata only update either. Could you please give me a scenario?
I was thinking metadata only update is much simpler. with `RefreshMetadataAction` (or something I read in another PR), we can just update the 'deleted' and 'appended' files in metadata. Isn't that enough? Why do we need the assert(deleted.nonEmpty) I don't really understand.

imback82: I meant "This check" = ` index.deletedFiles.isEmpty && index.appendedFiles.isEmpty)` in this file, not the assert. What I mean is that with the current semantics, we need this check regardless of the refresh modes (which are orthogonal).

apoorvedave1: yeah these checks are required with the current semantics regardless of the refresh mode. So here's my understanding:

If hybrid scan is enabled, we know the data-source-changes (either adds or deletes) will be handled by hybrid scan.
If it is not, the right way to handle this is DO NOT USE the index because it's not up to date with the data.

So I am currently fine with this semantic.

sezruby: Sorry I misread. I thought "since" means "latest". So "latest"-`source.Content` indicates the source files of the index data in that version. Then I think "mixed" approach is better as there is no special use of "since" "appendedFiles" and "deletedFiles" during query time or refresh action?

imback82: Oh OK. The use of "latest" could be confusing. Just to make sure everyone is on the same page, let's assume the following events in sequence:
- There are source files: f1, f2, f3
- An index is created.
- A source file is added: f4
- A source file is deleted: f1
- Refresh "quick" (metadata only operation)

With the current semantics (mix of "latest" and "since"), we will have the following in the latest index log entry:
- source files / signature: f2, f3, f4 / sig(f2, f3, f4)
- appendedFiles: f4 
- deletedFiles: f1

With the semantics of always using "since":
- source files / signature: f1, f2, f3 / sig(f1, f2, f3)
- appendedFiles: f4 
- deletedFiles: f1

And I am hearing that the current semantics (mix of "latest" and "since") will be a better use. Is that correct?

sezruby: That's correct. The benefit of `quick` is skipping the comparison of file list for hybrid scan and using the signature instead. ( + and it also allows selective hybrid scan - disable hybrid scan config and do quick refresh only for the indexes which are candidates of hybrid scan op. )
On the other hand, with always using "since" approach, we need to check if the source file list in the given plan could be the candidate of each index in some how at query execution time.

imback82: What if we capture the latest source signature as separate field? For example, we introduce `Update` case class that captures "appendedFiles", "deletedFiles", and "latestSourceFilesSignature"' for the `Source.Content`? (I am just exploring the best option for both semantics / usage before the next release, so please bear with me 😄 ). It seems that if the signature doesn't match, we need to fallback to iterating files for hybrid scan anyway.


> ( + and it also allows selective hybrid scan - disable hybrid scan config and do quick refresh only for the indexes which are candidates of hybrid scan op. )

Can you also elaborate on this scenario? I didn't get the part where you disable the hybrid scan config and do quick refresh since the paths are different (one for query and another for non-query).

sezruby: > What if we capture the latest source signature as separate field? For example, we introduce Update case class that captures "appendedFiles", "deletedFiles", and "latestSourceFilesSignature"' for the Source.Content? 

This could be an option though the original signature value won't be used anymore.
But in this way, we could keep the same semantic for Source.Content. 
And in the case of "full" and "incremental" refresh, `Source.Content` contains the source files at refresh (both full and incremental) the same as now?

> Can you also elaborate on this scenario? I didn't get the part where you disable the hybrid scan config and do quick refresh since the paths are different (one for query and another for non-query).

For example, there're 5 indexes but one of them has many "diff" files in its source files since the index build. But a user want to use hybrid scan for an index with very little diff files. If global hybrid scan config is enabled, Hyperspace tries to apply hybrid scan for all candidate indexes - if there's no proper hybrid scan threshold in candidate selection. 
So `quick` refresh can be an option for this situation:
- disable hybrid scan config
- quick refresh for some indexes with little diff
- hybrid scan is applied only for `quick` refreshed indexes

Although this is different context, I just wanted to describe the possible use case of `quick` refresh.

rapoth: @imback82 @sezruby Should we move this discussion to an issue so we can follow-up properly through a PR, if necessary? If you agree, @sezruby - can you please create an issue with all the approaches + pros/cons so everyone is on the same page? You can paste the link to that issue here as a final reply.

rapoth:
> @apoorvedave1 Can you help creating an issue to address improving the user experience when indexes are not selected (e.g., the "why not" API)? I think we should address this sooner than later (a good candidate for the next release - 0.5). Thanks!

@apoorvedave1 Can you take care of creating this issue with the relevant details?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: > This could be an option though the original signature value won't be used anymore.

Yes, it will be used only if `Update` is `isEmpty == true` (assuming we can capture `Update` as `Option[]`).

> But in this way, we could keep the same semantic for Source.Content.

Yes, I think the code logic will be simpler as well.

> And in the case of "full" and "incremental" refresh, `Source.Content` contains the source files at refresh (both full and incremental) the same as now?

Well, it will store the source files that are indexed, so it could be different if there are deleted files before `RefreshAppend` runs.

> Although this is different context, I just wanted to describe the possible use case of `quick` refresh.

Oh, you mean refresh "on-the-fly" during query time.

sezruby: Created #198 to track this issue. 
BTW
>>Although this is different context, I just wanted to describe the possible use case of quick refresh.
>
>Oh, you mean refresh "on-the-fly" during query time.

it's possible but I meant a user can choose the candidate indexes for hybrid scan - with disabled global config.

imback82: OK, in this case, I still don't understand 😄. @rapoth do you understand this scenario?

sezruby: Ok let me try again 😁 
1. disable "**global**" hybrid scan config
2. user can choose some "beneficial - with high perf improvement" indexes with little diff and do `quick` refresh for the indexes
3.  hybrid scan can be performed for `quick` refreshed indexes even if the global config is disabled. The other unrefreshed indexes with diff won't be the candidates for hybrid scan.

imback82: Thanks 👍 . So in a nutshell, the selective hybrid scan is "we can apply the hybrid scan even if the user didn't enable it only when the signature matches", right?

sezruby: Right. That's the point 👍

imback82: Thanks!

Issue #148, opened 2020-09-08
Hyperspace Index on Delta Lake


sezruby:
### Describe the issue

Support Hyperspace Index on Delta Lake Table
https://delta.io/

#### Retrieve source file list from Relation
* [TahoeFileIndex.scala](https://github.com/delta-io/delta/blob/master/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala)
* TahoeFileIndex provides `def inputFiles` API; we could use the API as it handles all version & delta log related things.
```
// Both logical & phisycal plan have this "delta" information
// TahoeLogFileIndex from LogicalRelation.HaddopFsRelation.Location

> rel.location.asInstanceOf[TahoeLogFileIndex]
org.apache.spark.sql.delta.files.TahoeLogFileIndex = Delta[version=2, file:/tmp/delta-table]

> val e = rel.location.asInstanceOf[TahoeLogFileIndex]
> e.inputFiles
res19: Array[String] = Array(file:/tmp/delta-table/part-00003-e628f8cd-1609-4adb-ad46-7399bb67f666-c000.snappy.parquet, file:/tmp/delta-table/part-00003-8504e0b9-b6de-4eb8-96d1-9a0b5598faee-c000.snappy.parquet, file:/tmp/delta-table/part-00004-ca1393a7-05e8-4960-ab3f-c0cfaab9a4b7-c000.snappy.parquet, file:/tmp/delta-table/part-00004-2f5baa84-2bce-4e30-a1e9-08b36f28b8fc-c000.snappy.parquet, file:/tmp/delta-table/part-00006-52a5ce8c-e222-49ca-a8d4-e18fcd540fc9-c000.snappy.parquet, file:/tmp/delta-table/part-00007-9907856f-c231-4e17-81e9-3e65b3ec58b0-c000.snappy.parquet, file:/tmp/delta-table/part-00001-c20d1b6d-4d33-40cd-a587-204f1368726b-c000.snappy.parquet, file:/tmp/delta-table/part-00000-f0ed3f09-e749-4f6b-9855-b070175f010e-c000.snappy.parquet, file:/tmp/delt…
```

#### Create & refresh index
We could generate an index based on the given DF and path.
```
val df = spark.read.format("delta").load("/tmp/delta-table")
```

#### Issues
* [update / merge into / delete / overwrite] causes to rewrite whole data => full refresh required
* Delta lake jar dependency
* delta log retention: default 7 days

#### Things to check
[...]

PR #197, opened 2020-10-12
Support Delta lake relation

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: #148 
 - **Dependencies**: #227 

Support Delta lake relation for "immutable" dataset.
This PR covers index creation/refresh(full mode)/application using FilterIndexRule and JoinIndexRule.

As we can rely on Delta Lake for data change and management, the version info and path of Delta Lake relation can be used in FileBasedSignatureProvider - instead of hash computation of all source files:
```scala
...
 case LogicalRelation(
          _,
          _,
          _) =>
        fingerprint ++= location.allFiles.foldLeft("")(
          (accumulate: String, fileStatus: FileStatus) =>
            HashingUtils.md5Hex(accumulate + getFingerprint(fileStatus)))
 case LogicalRelation(
          HadoopFsRelation(location: TahoeLogFileIndex, _, _, _, _, _),
          _,
          _,
          _) =>
        fingerprint ++= location.tableVersion + location.path.toString
...
```

Yes, a user can create/refresh/apply indexes on Delta Lake relation.

Unit test


build.sbt
```diff
@@ -27,6 +27,7 @@ scalaVersion := scala212
 crossScalaVersions := supportedScalaVersions
 
 libraryDependencies ++= Seq(
+  "io.delta" %% "delta-core" % "0.6.1",
```
sezruby, line 30: I met the following error at `sbt assembly` and no idea how to fix it.
```
[info] Merging files...
[warn] Merging 'META-INF/MANIFEST.MF' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.abego.treelayout/org.abego.treelayout.core/pom.properties' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.abego.treelayout/org.abego.treelayout.core/pom.xml' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.antlr/ST4/pom.properties' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.antlr/ST4/pom.xml' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.antlr/antlr-runtime/pom.properties' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.antlr/antlr-runtime/pom.xml' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.antlr/antlr4-runtime/pom.properties' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.antlr/antlr4-runtime/pom.xml' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.antlr/antlr4/pom.properties' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.antlr/antlr4/pom.xml' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.glassfish/javax.json/pom.properties' with strategy 'discard'
[warn] Merging 'META-INF/maven/org.glassfish/javax.json/pom.xml' with strategy 'discard'
scala.MatchError: /Users/eunjin/IdeaProjects/hyperspace_ej/python/run-tests.py (of class java.lang.String)
    at sbtassembly.AssemblyUtils$.sourceOfFileForMerge(AssemblyUtils.scala:23)
    at sbtassembly.MergeStrategy$$anonfun$sbtassembly$MergeStrategy$$filenames$1.apply(MergeStrategy.scala:28)
    at sbtassembly.MergeStrategy$$anonfun$sbtassembly$MergeStrategy$$filenames$1.apply(MergeStrategy.scala:27)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
    at scala.collection.Iterator$class.foreach(Iterator.scala:727)
    at scala.collection.AbstractIterator.foreach(Iterator.scala:1157)
    at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
    at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
    at scala.collection.AbstractTraversable.map(Traversable.scala:105)
    at sbtassembly.MergeStrategy$.sbtassembly$MergeStrategy$$filenames(MergeStrategy.scala:27)
    at sbtassembly.MergeStrategy$$anon$6.apply(MergeStrategy.scala:93)
    at sbtassembly.MergeStrategy.apply(MergeStrategy.scala:21)
    at sbtassembly.Assembly$.sbtassembly$Assembly$$applyStrategy$1(Assembly.scala:107)
    at sbtassembly.Assembly$$anonfun$14.apply(Assembly.scala:132)
    at sbtassembly.Assembly$$anonfun$14.apply(Assembly.scala:129)
[error] (*:assembly) scala.MatchError: /<PATH>/hyperspace_ej/python/run-tests.py (of class java.lang.String)
```
and below `assemblyMergeStrategy in assembly` didn't work.

build.sbt
```diff
@@ -27,6 +27,11 @@ scalaVersion := scala212
 crossScalaVersions := supportedScalaVersions
 
 libraryDependencies ++= Seq(
+  if (scalaVersion.value == scala211 || sparkVersion.value == "2.4.2") {
```
AFFogarty, line 30: I wonder if instead of `sparkVersion.value == "2.4.2"` we should have some higher-level variable like `val sparkMajorVersion = 2` or something like this.  Then we don't have to worry about things breaking if we change `sparVersion` to `2.4.4` or some other value later on.

build.sbt
```diff
@@ -27,6 +27,11 @@ scalaVersion := scala212
...
+    "io.delta" %% "delta-core" % "0.6.1"
+  } else {
+    "io.delta" %% "delta-core" % "0.7.0"
```
imback82, line 33: I think we should do "provided" since few cloud providers come with delta JARs built in.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -119,6 +120,37 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
           dataSchema.json,
           fileFormatName,
           opts)
+      case LogicalRelation(
+          HadoopFsRelation(location: TahoeLogFileIndex, _, dataSchema, _, fileFormat, options),
```
imback82, line 124: `fileFormat` not used.

build.sbt
imback82: Also, let's not worry about 0.7.0 since it is for Spark 3.0

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -119,6 +120,37 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+          _,
+          _,
+          _) =>
+        val files = location
+          .getSnapshot(false)
```
imback82, line 129: `.getSnapshot(false) // stalenessAcceptable to false ...`

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -119,6 +120,37 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+          .filesForScan(Seq(), Seq(), false)
+          .files
```
imback82, line 131: Can we use `matchingFiles` or not?

Can you add comment why `Seq()` is fine for partition/data filters?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -119,6 +120,37 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+          .map { f =>
+            new FileStatus(
+              /* length */ f.size,
```
imback82, line 134: `f.size, // length`?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -119,6 +120,37 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
           dataSchema.json,
           fileFormatName,
           opts)
+      case LogicalRelation(
```
imback82, line 123: Question: Is it possible that the location is `TahoeBatchFileIndex`?

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -66,11 +66,19 @@ private[actions] abstract class RefreshActionBase(
...
-    spark.read
-      .schema(dataSchema)
-      .format(rels.head.fileFormat)
-      .options(rels.head.options)
-      .load(rels.head.rootPaths: _*)
+    if (rels.head.fileFormat.equals("delta")) {
+      spark.read
+        .schema(dataSchema)
+        .format(rels.head.fileFormat)
+        .options(rels.head.options - "versionAsOf" - "timestampAsOf")
```
imback82, line 73: Can you explain why we need to remove these options?

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -66,11 +66,19 @@ private[actions] abstract class RefreshActionBase(
   protected lazy val df = {
     val rels = previousIndexLogEntry.relations
```
imback82, line 67: nit (existing code): `val rel = previousIndexLogEntry.relations.head` and use it instead of repeating `rels.head`.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
sezruby: `matchingFiles` seems to be used when there are additional data filters or partition filters. So I think we don't need to use it for now. 

And I added partitionFilter.

build.sbt
sezruby: Thanks @imback82 for the fix 👍

build.sbt
sezruby: Thanks! I'll refer this comment later for Spark 3.0 :)

imback82:
Btw, I am working on the extension model so that you can plug in new source formats. I will have the PR up soon so that we can test it using Delta first.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -64,13 +64,23 @@ private[actions] abstract class RefreshActionBase(
...
+    val rel = previousIndexLogEntry.relations.head
+    val dataSchema = DataType.fromJson(rel.dataSchemaJson).asInstanceOf[StructType]
+    if (rel.fileFormat.equals("delta")) {
+      // Exclude "versionAsOf" and "timestampAsOf" options, so that this DataFrame for refresh
+      // can see the latest snapshot.
+      spark.read
+        .schema(dataSchema)
+        .format(rel.fileFormat)
+        .options(rel.options - "versionAsOf" - "timestampAsOf")
+        .load(rel.rootPaths.head)
```
apoorvedave1, line 76: you can probably use just this section because versionAsOf and timestampAsOf are only used in delta. the same code would work for all supported formats right?

apoorvedave1:
the approach looks fine to me. Thanks @sezruby

apoorvedave1 (commented):
the approach looks fine to me. Thanks @sezruby

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -0,0 +1,99 @@
...
+
+      // The index should be applied for the updated version.
+      assert(verifyIndexUse(query().queryExecution.optimizedPlan, "deltaIndex"))
+
+      // The index should not be applied for the version at index creation.
+      assert(!verifyIndexUse(query(Some(0)).queryExecution.optimizedPlan, "deltaIndex"))
+    }
+  }
+
+  def verifyIndexUse(plan: LogicalPlan, indexName: String): Boolean = {
```
pirz, line 88: nit rename: verifyIndexUsage ?
BTW should not be private?

sezruby:
This PR will be delivered by #265.

sezruby:
Closed by #265

PR #199, opened 2020-10-12
Change lineage UDF to fix CreateIndex performance issue

pirz:
This PR changes the UDF used for normalizing source data file paths to fill lineage column values.
We normalize path values in the lineage column by removing extra preceding `/` characters in the full file path for each and every index entry.

Currently, file path normalization happens by calling: `PathUtils.makeAbsolute()`. However, according to some simple performance benchmark, we realized this adds a significant overhead to createIndex operation. This PR modifies the UDF to use  String's `replace()` function. Given that `input_file()` function returns full path to source data file for each row, this change of UDF is safe and produces the same output as the original UDF.
  
No.

Existing tests cases under `RefreshIndexTests` and `E2EHyperspaceRulesTests` already cover this code path.


imback82 (approved):
LGTM (assuming perf improvement is verified), thanks @pirz!

apoorvedave1 (approved):
LGTM thanks @pirz

Issue #159, opened 2020-09-14
[FEATURE REQUEST]: Introduce a similarity threshold for Hybrid Scan


sezruby:
**Feature requested**

As a developer, I want to introduce some similarity threshold about dataset for hybrid scan in order to avoid performance regression.

Currently, Hybrid scan considers an index as a candidate if an input relation includes one or more same source files of the index. However, in case the "diff" dataset is much larger than the common dataset, shuffling overhead for the "diff" dataset might outweigh the benefit from the index.

In order to avoid the regression by doing Hybrid scan, we could introduce some threshold so that the indexes which covers less amount data of the given relation would not be considered as a candidate index.

The threshold might be
- the number of common files; only works for append-only dataset
- the total size of common files

And we could adjust the threshold - "N % of similarity " - with a new Hyperspace config and set a proper default value based on some experimental results.

**Acceptance criteria** 

- [x] The threshold should be configurable by `HyperspaceConf`.
- [x] The default value should be determined based on experimental results using large dataset.
- [x] Calculating threshold should be finished in reasonable time.

**Success criteria**

- [ ] Calculating threshold overhead.
- [ ] The experimental result for the default value.

**Additional context**

It would be good to print some log why hybrid scan is not applied.

rapoth:
Out of curiosity, in case of issues like this, does it make sense to provide `To Reproduce` and `Expected behavior`?

sezruby:
I think this is more like "User story" in JIRA, not bug or issue report. As this project does not have JIRA-like thing, it would be good to have a new template(and tag) here for relatively small features or tasks for tacking development.

rapoth:
Good point. I created a [new template for feature request](https://github.com/microsoft/hyperspace/issues/new?assignees=&labels=untriaged%2C+enhancement&template=feature-request-template.md&title=%5BFEATURE+REQUEST%5D%3A+). If you agree, can you change this issue to follow that template?

sezruby:
Thanks @rapoth :) I updated the document.

PR #202, opened 2020-10-12
Add hybrid scan config for the max number of deleted files

sezruby:

 - **Tracking Issue**: #159 
 - **Parent Issue**: #150 
 - **Dependencies**: n/a

With the current performance limitation of Hybrid scan operation for delete dataset, we limit the number of  deleted files to avoid regression from Hybrid scan. If the number of deleted files is larger than this config, the index is disabled and cannot be a candidate for Hybrid Scan.

```
  val INDEX_HYBRID_SCAN_DELETE_MAX_NUM_FILES =
    "spark.hyperspace.index.hybridscan.delete.maxNumDeletedFiles"
  val INDEX_HYBRID_SCAN_DELETE_MAX_NUM_FILES_DEFAULT = "10"
```

Yes, a new config is added for Hybrid scan delete dataset.

unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
@@ -42,11 +42,19 @@ object IndexConstants {
   val INDEX_HYBRID_SCAN_DELETE_ENABLED = "spark.hyperspace.index.hybridscan.delete.enabled"
   val INDEX_HYBRID_SCAN_DELETE_ENABLED_DEFAULT = "false"
 
+  // With the current performance limitation of Hybrid scan for delete files, we limit
+  // the number of deleted files to avoid regression from Hybrid scan.
+  // If the number of deleted files is larger than this config, the index is disabled and
+  // cannot be a candidate for Hybrid Scan.
+  val INDEX_HYBRID_SCAN_DELETE_MAX_NUM_FILES =
+    "spark.hyperspace.index.hybridscan.delete.maxNumDeletedFiles"
+  val INDEX_HYBRID_SCAN_DELETE_MAX_NUM_FILES_DEFAULT = "10"
+
   // Identifier injected to HadoopFsRelation as an option if an index is applied.
   // Currently, the identifier is added to options field of HadoopFsRelation.
   // In Spark 3.0, we could utilize TreeNodeTag to mark the identifier for each plan.
   // See https://github.com/microsoft/hyperspace/issues/185
-  val INDEX_RELATION_IDENTIFIER: (String, String) = ("indexRelation" -> "true")
+  val INDEX_RELATION_IDENTIFIER: (String, String) = "indexRelation" -> "true"
 
   // Default number of buckets is set the default value of "spark.sql.shuffle.partitions".
   val INDEX_NUM_BUCKETS_DEFAULT: Int = SQLConf.SHUFFLE_PARTITIONS.defaultValue.get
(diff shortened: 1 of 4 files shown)


src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -42,11 +42,19 @@ object IndexConstants {
   val INDEX_HYBRID_SCAN_DELETE_ENABLED = "spark.hyperspace.index.hybridscan.delete.enabled"
```
imback82, line 42: Do we need to keep around this config?

sezruby: If we introduce other threshold configs for hybrid scan, this config will make things simple. I think we could keep this config until the feature is stable.

imback82: OK. If you are thinking through this, then we can leave it for now.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -105,12 +105,13 @@ object RuleUtils {
         }
       assert(filesByRelations.length == 1)
       indexes.filter(index =>
-        index.created && isHybridScanCandidate(index, filesByRelations.flatten))
+        index.created && index.deletedFiles.isEmpty && index.appendedFiles.isEmpty &&
```
imback82, line 108: why is isEmpty checks needed for hybrid scan?

sezruby: Because entry.allSourceFiles (source.Content) includes the appended files - so need to handle the appended files properly.

imback82: Is this an existing bug then? I wasn't sure why this is related to the config change.

imback82: also, shouldn't we support nonEmpty case since this is hybrid scan?

sezruby: Yes that's a bug. It's from non-empty appendedFiles or deletedFile, not the config.

imback82: Can you update the PR description please?

imback82: and can you add a test that fixes the bug?

sezruby: We can support but need some change - but there is no agreement on this. Do you think this also should be delivered in 0.4?

pirz:
LGTM, Thanks @sezruby

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: Yes, every bug fix should come with a test coverage. You can separate out the bug fix as a separate PR if you want (and just concentrate on the config changes in this PR).

imback82: Actually, can you just separate out the bug fix as a separate PR?

imback82: Discussed this offline, and we will add comment around this change.

imback82 (approved):
LGTM, thanks @sezruby!

PR #205, opened 2020-10-13
Use configs of previous LogEntry for refresh action

### LINKED ISSUE #196: Refresh index should pick bucket num and lineage column configs from previous entry instead of session config
### Describe the issue

https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala#L40-L42

https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala#L44-L48

This is a known issue that currently Refresh modes "full" and "incremental" pick values for `num-buckets` and `lineageColumnEnabled` from current spark session config. Creating this issue for followup.

sezruby:

 - **Tracking Issue**: #196 
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Fixes #196 

Instead of using the current spark.config, this PR gets numBuckets and lineage column config from previous index entry at refresh and use it for refresh.

Fixes a buggy behavior.

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -39,11 +39,11 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       .getOrElse(dataManager.getPath(0))
   }
 
-  protected def getNumBucketsConfig(spark: SparkSession): Int = {
+  protected def numBucketsForIndex(spark: SparkSession): Int = {
     HyperspaceConf.numBucketsForIndex(spark)
   }
 
-  protected def getLineageColumnConfig(spark: SparkSession): Boolean = {
+  protected def indexLineageEnabled(spark: SparkSession): Boolean = {
     HyperspaceConf.indexLineageEnabled(spark)
   }
 
@@ -53,7 +53,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       indexConfig: IndexConfig,
       path: Path): IndexLogEntry = {
     val absolutePath = PathUtils.makeAbsolute(path)
-    val numBuckets = getNumBucketsConfig(spark)
+    val numBuckets = numBucketsForIndex(spark)
 
     val signatureProvider = LogicalPlanSignatureProvider.create()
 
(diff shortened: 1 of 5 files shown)


sezruby:
@imback82 What if a user wants to change the configs - with full refresh mode? Should we consider that case also?

imback82:
> @imback82 What if a user wants to change the configs - with full refresh mode? Should we consider that case also?

Let's not support it for now (the user can create a new index with a different number of buckets if needed - since refreshing full is like creating a new one). We can introduce a "overwrite" config later if needed.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -39,17 +39,29 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
       .getOrElse(dataManager.getPath(0))
   }
 
+  protected def getNumBucketsConfig(spark: SparkSession): Int = {
+    spark.sessionState.conf
```
imback82, line 43: Can we move this to `HyperspaceConf`?

imback82: And we can call it `numBuckets` and `lineageEnabled` respectively. (or `numBucketsForIndex` and `indexLineageEnabled`)

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -49,6 +49,14 @@ private[actions] abstract class RefreshActionBase(
 
   protected lazy val previousIndexLogEntry = previousLogEntry.asInstanceOf[IndexLogEntry]
 
+  override protected final def getNumBucketsConfig(spark: SparkSession): Int = {
```
imback82, line 52: Let's add `// Refresh maintains the same number of buckets as the existing index because ...` and we can mention the about the full refresh here.

imback82 (commented):
Few minor comments, but the PR looks fine to me.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -39,17 +39,21 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
       .getOrElse(dataManager.getPath(0))
   }
 
+  protected def getNumBucketsConfig(spark: SparkSession): Int = {
```
imback82, line 42: Can we name this as `def numBucketsForIndex` and `def indexLineageEnabled` as well?

imback82 (commented):
One minor comment, but LGTM, thanks @sezruby!

imback82 (approved):

PR #214, opened 2020-10-15
Remove duplicates in appendedFiles and deletedFiles

### LINKED ISSUE #211: Dedupe appendedFiles / deletedFiles in RefreshActionBase.scala
I think we now need to dedupe now that we are considering size, etc. If we move to `FileInfo`, it will be most likely OK. Same applies `appendedFiles`

_Originally posted by @imback82 in https://github.com/microsoft/hyperspace/pull/207#discussion_r504830903_

sezruby:

 - **Tracking Issue**: #211  
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Fixes #211

Remove possible duplicate entries in appendedFiles and deletedFiles caused by the same file names but different properties - size or modified time. 
Duplicates in deletedFiles are not problem, but can be a problem for appendedFiles; duplicate rows can be included in the index data while incremental refresh.

Yes, fixes a possible bug when utilizing incremental refresh with modified files.

Unit test


src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -103,11 +103,14 @@ private[actions] abstract class RefreshActionBase(
...
     val originalFiles = relation.data.properties.content.fileInfos
-    val delFiles = (originalFiles -- currentFiles).map(_.name).toSeq
+    val delFiles = (originalFiles -- currentFiles).map(_.name)
+
+    // Remove duplicate deleted file names in the previous log entry.
+    val prevDeletedFiles = previousIndexLogEntry.deletedFiles.filterNot(delFiles.contains)
 
     // TODO: Add test for the scenario where existing deletedFiles and newly deleted
     //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
-    delFiles ++ previousIndexLogEntry.deletedFiles
```
imback82, line 110: Any reason not using `distinct` with the previous implementation? for example, `(delFiles ++ previousIndexLogEntry.deletedFiles).distinct`?

sezruby: Because delFiles is originally Set, so I wanted to utilize it.
It's okay to use `distinct` if you prefer it.

imback82 (approved):
LGTM, thanks @sezruby!

Issue #184, opened 2020-10-07
[FEATURE REQUEST]: Hybrid scan - performance validation & optimization for delete dataset


sezruby:
**Feature requested**

#171 enables to utilize Hybrid scan with delete dataset as well as append dataset.
However, the feature still requires some validation and optimization before GA.
So I created the temporary config below to make it switchable until then. 

```
  val INDEX_HYBRID_SCAN_DELETE_ENABLED = "spark.hyperspace.index.hybridscan.delete.enabled"
  val INDEX_HYBRID_SCAN_DELETE_ENABLED_DEFAULT = "false"
```

This issue is for tracking this config.

**Acceptance criteria** 

**Success criteria**

- [x] delete-only dataset TPCH
- [x] delete-only dataset TPCH - 100K files 
- [x] append & delete dataset TPCH

**Additional context**

PR #215, opened 2020-10-16
Optimize Hybrid Scan for deleted files

sezruby:

 - **Tracking Issue**: #184
 - **Parent Issue**: #150 
 - **Dependencies**: n/a

Apply catalyst's `OptimizeIn` rule to the injected filter-not-in plan for Hybrid Scan with delete files. (refer #171)

https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala#L237

Yes, performance optimization is done by using "InSet" instead of "In", in case deleted files are larger than `spark.sql.optimizer.inSetConversionThreshold` config, default value: 10.

Changed plan example (with the threshold value=1): 
```
  +- Project [clicks#684, Date#680]
      +- Filter ((isnotnull(clicks#684) && (clicks#684 <= 4000)) && (clicks#684 >= 2000))
         +- Project [Date#680, clicks#684]
            +- Filter NOT _data_file_name#712 INSET (file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/part-00003-6fb13b2d-abcc-46e2-81ba-42b3a933acec-c000.snappy.parquet,file:/C:/Users/eunsong/repo/hyperspace2/src/test/resources/data/sampleparquet4/part-00002-6fb13b2d-abcc-46e2-81ba-42b3a933acec-c000.snappy.parquet)
               +- Relation[Date#680,clicks#684,_data_file_name#712] parquet
```

Unit test

### DIFF
--- a/src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
@@ -19,7 +19,7 @@ package com.microsoft.hyperspace.index
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.{DataFrame, QueryTest}
-import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, In, Literal, Not}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, In, InSet, Literal, Not}
 import org.apache.spark.sql.catalyst.plans.logical.{Filter, Project, RepartitionByExpression, Union}
 import org.apache.spark.sql.execution.{FileSourceScanExec, InputAdapter, ProjectExec, UnionExec, WholeStageCodegenExec}
 import org.apache.spark.sql.execution.datasources._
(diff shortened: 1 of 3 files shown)


sezruby:
@rapoth @imback82 @pirz @apoorvedave1 Should we increase the max number of deleted files threshould ("spark.hyperspace.index.hybridscan.delete.maxNumDeletedFiles")
with this change? - I think this might requires some performance validation but we could increase it little more (like 30 or 50?) as the `spark.sql.optimizer.inSetConversionThreshold` is 10 by default.

And please review this change in case you'd like to deliver this PR in this release. Thanks!

pirz:
> @rapoth @imback82 @pirz @apoorvedave1 Should we increase the max number of deleted files threshould ("spark.hyperspace.index.hybridscan.delete.maxNumDeletedFiles")
> with this change? - I think this might requires some performance validation but we could increase it little more (like 30 or 50?) as the `spark.sql.optimizer.inSetConversionThreshold` is 10 by default.
> 
> And please review this change in case you'd like to deliver this PR in this release. Thanks!

That is a good point @sezruby - I would vote conservative for keeping this value small (i.e. 10) for now to avoid potential performance degrades. We can bump it up later once we see more evidence of it performing as expected with higher values for this threshold. However, @sezruby, as you are conducting the benchmarks for this part, you should really make the final call :).

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -565,17 +565,16 @@ class HybridScanTest extends QueryTest with HyperspaceSuite {
       var deletedFilesList: Seq[Seq[String]] = Nil
       val nodes = planWithHybridScan collect {
         case p @ Filter(
-              Not(In(attr, deletedFileNames)),
+              Not(expr),
               LogicalRelation(fsRelation: HadoopFsRelation, _, _, _)) =>
           // Check new filter condition on lineage column.
+          val attr = expr.children.head.asInstanceOf[Attribute]
+          val deletedFile = expr.children.last.asInstanceOf[Literal].toString
```
imback82, line 572: Can we match exact instead of `expr.children`?
```
        case p @ Filter(
              Not(EqualTo(left: Attribute, right: Literal)),
              LogicalRelation(fsRelation: HadoopFsRelation, _, _, _)) =>
          // Check new filter condition on lineage column.
          val colName = left.name
          val deletedFile = right.toString
```

pirz:
LGTM, Thanks @sezruby !

imback82 (approved):
LGTM, thanks @sezruby!

PR #216, opened 2020-10-16
Update documentation - Refresh index

pirz:
Update documentation for refresh index and refresh modes: full and incremental.

This PR adds details for refresh index in full and incremental modes to Hyperspace documentation in mutable datasets page.

Yes, user documentation is changed.

Proof reading.


docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
 
-Now, we offer several options to handle appended and deleted files more efficiently.
+Now, we offer several options to handle above scenario more efficiently.
 
 ## Options of using index when your dataset changes
 
 ### Refresh Index
-
-We provide several refresh modes to refresh index data.
+You can refresh an index according to its latest source data files by running the `refresh` command.
```
imback82, line 25:
```suggestion
You can refresh an index according to its latest source data files by running the `refreshIndex` command.
```

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+Hyperspace provide several modes to refresh an index. These modes differ in terms of the way they update the index and the amount of data scans and shuffle each does.
+You should pick a mode for refreshing an index according to its current size and total amount of data deleted from or appended to its source data files.
+You can specify the mode as an argument when calling the `refresh` command.
```
imback82, line 28:
```suggestion
You can specify the mode as an argument when calling the `refreshIndex` command.
```

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+Currently, there are two refresh modes available for an index: `"full"` and `"incremental"`.
 
 #### Full
```
imback82, line 31:
```suggestion
#### Refresh Index - Full Mode
```
Duplicate "Refresh Index" to have a better context?

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
-TODO
+After some changes in an index's original dataset files, using `refresh` with the `"full"` mode causes
```
imback82, line 32:
```suggestion
After some changes in an index's original dataset files, using `refreshIndex` with the `"full"` mode causes
```

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+the index refresh action perform a full rebuild of the index.
+This ends up creating a new version of the index and involves a full scan and shuffle of its latest source data.
+As a result, the amount of time it takes to refresh an index in this mode is similar to creating a new index, with the same index configuration, on the latest dataset content.
```
imback82, line 35: "latest dataset content." you mean "latest source data"?

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+```python
+from hyperspace import Hyperspace
+
+hs = Hyperspace(spark)
+hs.refreshIndex("empIndex", "full")
+```
 
 #### Incremental
-TODO
+After some changes in an index's original dataset files, using `refresh` with the `"incremental"` mode causes
```
imback82, line 58: "After some files are added or deleted to the original source files (from which an index was built),"?

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+the index refresh action fix any existing index file which has some deleted records and index newly added data files.
```
imback82, line 59: Does it "fix"?

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+An index needs needs to have lineage enabled to be eligible for refresh in the incremental mode.
```
imback82, line 60: needs needs

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+Check the [configuration](02-ug-configuration.md) page to see how lineage is enabled when creating an index.
+
+Once refresh is called for an index in the incremental mode, Hyperspace checks latest dataset files and identifies
```
imback82, line 63: Can we use the same phrase to refer to "source data files" (instead of dataset)?

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+deleted source data files and newly added ones. It recreates those portions of index which have records from the
+deleted files. Lineage is used to detect and fix these affected index files. Subsequently, Hyperspace creates new index files on
```
imback82, line 65: I think we can update "fix" since it sounds like we are "modifying" the existing index files.

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,81 @@ This guide helps you to utilize indexes on mutable dataset.
...
+
+```python
+from hyperspace import Hyperspace
+
+hs = Hyperspace(spark)
+hs.refreshIndex("empIndex", "incremental")
+```
 
-#### Quick
+##### Optimize Index
```
imback82, line 88: This should be "###"?

imback82 (approved):
LGTM, thanks @pirz!

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -11,28 +11,82 @@ This guide helps you to utilize indexes on mutable dataset.
...
+
+hs = Hyperspace(spark)
+hs.refreshIndex("empIndex", "full")
+```
+
+#### Refresh Index - Incremental Mode
+After some files are added to or deleted from the original source files, an index was built on,
+using `refreshIndex` with the `"incremental"` mode causes the index refresh action recreate any existing index file,
+which has some deleted records, to remove those records. Refresh also creates new index files by indexing newly added data files.
+An index needs to have lineage enabled to be eligible for refresh in the incremental mode.
```
sezruby, line 61: @pirz @imback82 @apoorvedave1 Append-only dataset doesn't need lineage column for "incremental" mode. 
As using linage column has some overhead, I think we need describe this in user document, in a clear and easy way..

imback82: @pirz Feel free to update this in #217. Thanks!

pirz: @sezruby @imback82 - Just to be clear, for supporting deletes, index needs lineage. Are you asking to modify above sentence as below?
"An index needs to have lineage enabled to support delete during refresh in the incremental mode."

pirz: @imback82 @sezruby - I modified and extended this part in #217 - Please take a look at it there and let me know about your comments (if any). Thanks!

sezruby: > Just to be clear, for supporting deletes, index needs lineage. Are you asking to modify above sentence as below?
"An index needs to have lineage enabled to support delete during refresh in the incremental mode."

Yes I'd like to point out that lineage is only required for deleted files, so something like:
In order to support deleted files in the incremental refresh mode, lineage should be enabled when it is built. In case it is append-only dataset, (only appended files are expected in your dataset), you do not need to set lineage column.

Issue #22, opened 2020-06-23
Add an API to determine if an index is invalid (i.e., if base table changed).


apoorvedave1:
**Describe the issue**

Add an API like `hs.isValid(indexName)` which returns true or false depending on whether the index is valid.
An Index can become invalid if the original data is changed. This would lead to signature mismatch and index would become unusable.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

rapoth:
We should subsume this as part of the new Why Not? API. @sezruby Can you please include this use case in your proposal?

PR #217, opened 2020-10-16
Update documentation - Optimize index

pirz:
Update documentation for optimize index and optimize modes: quick and full.

This PR adds details for optimize index in quick and full modes to Hyperspace documentation in mutable datasets page.

Yes, user documentation is changed.

Proof reading.


docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -87,7 +89,84 @@ hs.refreshIndex("empIndex", "incremental")
 ```
 
 ### Optimize Index
-TODO
+The fastest way to index newly added source data files and merge them into an existing index is by calling `refreshIndex`
+with the `"incremental"` mode. In this mode, each time refresh is called, it creates fresh index files for new data files
+and updates index metadata to include them in the index content.
+As these index files accumulate, they could affect query performance when the index is used.
```
sezruby, line 95: BTW, RefreshDeleteAction will merge them; so append-only incremental refresh ops will make that fragments. I think this could be described in design doc.

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -87,7 +89,84 @@ hs.refreshIndex("empIndex", "incremental")
...
+Once the index is leveraged for a query,
+the large number of these files could increase overall query time as more index files need to be accessed and potentially
+read to compute the query results.
+Hyperspace provides the `optimizeIndex` command to alleviate above issue by changing index files layout for an
```
sezruby, line 99: I think we need to show the status of index data files (e.g. number of files) and suggest to run optimizeIndex so that a user can easily determine and trigger it if needed. Could you create an issue for this?

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -87,7 +89,84 @@ hs.refreshIndex("empIndex", "incremental")
...
+index which has many index files, due to incremental refresh call(s). This is achieved by merging index files together,
+if possible, and replacing them with fewer larger files that capture exact same index records. This process is similar to
+compaction in append-only log structured merge index structures.
+
+You should note that the `optimizeIndex` command is a best effort to modify index files layout and its
+final outcome depends on how index records are stored in existing index files.
+Two or more index files can be merged with each other, if and only if they all have index records which belong to the
+same bucket (according to index configuration).
+When running `optimizeIndex` command, Hyperspace tries to find such groups of index files and merge them together.
+However, if most of the index files have records from different buckets, then final physical layout of index files
```
sezruby, line 109: > if most of the index files have records from different buckets, 

Could you rephrase this part?

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -87,7 +89,84 @@ hs.refreshIndex("empIndex", "incremental")
...
+
+hs = Hyperspace(spark)
+hs.optimizeIndex("empIndex", "quick")
+```
+
+#### Optimize Index - Full Mode
+In the `"full"` mode, `optimizeIndex` command considers all existing index files as candidates for merging. Therefore,
+Hyperspace does a full scan on current index content and identifies groups of index files which have index records belonging
+to the same bucket according to the index configuration. It then replaces each group with a single index file created through
+merging all the files in that group together. This mode tries  to achieve the best query performance improvement via a
```
sezruby, line 148:
```suggestion
merging all the files in that group together. This mode tries to achieve the best query performance improvement via a
```

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -58,8 +58,10 @@ hs.refreshIndex("empIndex", "full")
 After some files are added to or deleted from the original source files, an index was built on,
 using `refreshIndex` with the `"incremental"` mode causes the index refresh action recreate any existing index file,
 which has some deleted records, to remove those records. Refresh also creates new index files by indexing newly added data files.
-An index needs to have lineage enabled to be eligible for refresh in the incremental mode.
-Check the [configuration](02-ug-configuration.md) page to see how lineage is enabled when creating an index.
+An index needs to have lineage enabled to support deletes during refresh in the incremental mode. Lineage
```
sezruby, line 61: Can we create a sub section for lineage (with 6 of #?)? It's pre-requisite for delete refresh/hybrid scan, so I think it's better to make it noticeable.

pirz: Sure, but where do you exactly want to put that section? I am thinking it should be with "3 #" and comes before "Options of using index when your dataset changes", as it is used by refresh/hybridScan which come in that section. @sezruby What do you think?

docs/_docs/03-ug-mutable-dataset.md
pirz: Sure, are suggesting to add a new API to Hyperspace which can be called directly by user and shows some stats on the index, including how "fragmented" it is?

docs/_docs/03-ug-mutable-dataset.md
pirz: Yes, I made a bit of change in the text to emphasize more that index files for appended data cause the issue. I think we should not use `RefreshDeleteAction` term in the doc as that is an internal Action. From a user's perspective, it is only `refreshIndex` API. 
@sezruby Let me know if you still think more context is required here.

docs/_docs/03-ug-mutable-dataset.md
pirz: I made a bit of change here. Please let me know if that addressed your comment.

docs/_docs/03-ug-mutable-dataset.md
sezruby: How about "### Pre-requisite" before "### RefreshIndex"?

docs/_docs/03-ug-mutable-dataset.md
sezruby: Yes this might be enough for user document.

docs/_docs/03-ug-mutable-dataset.md
sezruby: Yes we could show 
- general index stats
  - name, index data bytes,
  - all versions for index data files of (latest version?)
     - can be multiple in case optimizeIndex or incremental refresh)
- source data context - (we could merge with isValid (#22))
  - number of current files / appended files / deleted files 
  - show 3~5 full paths of appended or deleted fils 
- optimize context
  - how many optimize-ndeeded files and their size
  - recommend optimizeIndex or not?
- root paths of index source files 

I think we could do this with 
hs.getIndexStatus(IndexName) or just hs.optimizeIndex(IndexName, mode, dryRun = true) only for "optimize context"?
WDYT? @pirz

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -106,9 +105,9 @@ final outcome depends on how index records are stored in existing index files.
 Two or more index files can be merged with each other, if and only if they all have index records which belong to the
 same bucket (according to index configuration).
 When running `optimizeIndex` command, Hyperspace tries to find such groups of index files and merge them together.
-However, if most of the index files have records from different buckets, then final physical layout of index files
-after running optimize would be similar to the original layout. An example of such an index is an index right after
-creation or full refresh, or an index which is refreshed incrementally with very few additional small source data files.    
+However, if for each given bucket, all index records are already stored in a single index file (i.e. we dont have several index
```
sezruby, line 108: ` (i.e. it does not have several` ?

sezruby: How about `already stored in a single index file` => `already stored in a single index file for each bucket`??

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -106,9 +105,9 @@ final outcome depends on how index records are stored in existing index files.
...
+files for a bucket), then final physical layout of index files after running optimize would be similar to the original
```
sezruby, line 109: I think it should be the same if all index records are already stored in a single index file for each bucket, we shouldn't change it. (Though current optimizeIndex rewrites the data and change it, but we don't need to describe it here.) and "final physical layout" term could be confusing. 

How about just removing "However ~~~ original layout" lines?

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -87,7 +89,83 @@ hs.refreshIndex("empIndex", "incremental")
...
+hs = Hyperspace(spark)
+hs.optimizeIndex("empIndex", "quick")
+```
+
+#### Optimize Index - Full Mode
+In the `"full"` mode, `optimizeIndex` command considers all existing index files as candidates for merging. Therefore,
+Hyperspace does a full scan on current index content and identifies groups of index files which have index records belonging
+to the same bucket according to the index configuration. It then replaces each group with a single index file created through
+merging all the files in that group together. This mode tries to achieve the best query performance improvement via a
+potentially slow optimize index process.
```
imback82, line 148: Can you mention the difference between "full" optimize vs. "full" refresh? It is subtle, so it will be worthwhile to call it out.

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -87,7 +89,83 @@ hs.refreshIndex("empIndex", "incremental")
 ```
 
 ### Optimize Index
-TODO
+The fastest way to index new data files and merge them into an existing index is by calling `refreshIndex`
```
imback82, line 92: Maybe change "The fastest way" to "One way" since we will have "quick" mode soon? 😄

imback82 (commented):
Generally looking fine to me (pending @sezruby's comments). @apoorvedave1, can you take a look as well?

docs/_docs/03-ug-mutable-dataset.md
pirz: @sezruby - Overall looks fine to me. I just think:
1) We should keep it as a separate command (like what you suggest: `getIndexStatus(ixName)` - or probably rename it with sth else) and we should not mix it with Optimize.
2) "recommend optimizeIndex or not?": I am not sure if it is a good idea to make a such a recommendation in the output of such an API. Normally, a "getStatus" type of API should be metadata read/summarization only. Making such a recommendation needs some (deeper) index content analysis. We need to check how much of overlap we have between new index file's buckets and the older ones. This does not necessarily need scanning index records, but can be done by extracting bucket-ids from file names and grouping them on that to see how many buckets with ">1 files" are there. But in general, "getStatus()" APIs and "recommend()" APIs are normally separated. 
Adding @rapoth - @apoorvedave1 - @imback82 to make suggestions

docs/_docs/03-ug-mutable-dataset.md
pirz: Added a paragraph on it.

docs/_docs/03-ug-mutable-dataset.md
pirz: I added it as "### Lineage"

docs/_docs/03-ug-mutable-dataset.md
pirz: I revised that part again. Plz take a look at it.

docs/_docs/03-ug-mutable-dataset.md
pirz: I revised that part again. Plz take a look at it.

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -21,6 +21,19 @@ Now, we offer several options to handle above scenario more efficiently.
...
+### Lineage
+Hyperspace uses lineage for tracing index entries back to the source data files from which they were generated.
+Lineage is required for removing deleted index entries during [index refresh in the incremental mode](#refresh-index---incremental-mode)
+or enforcing deletes at the query time when using [Hybrid Scan](#hybrid-scan).
+By default, lineage is disabled for indexes, and if required it should be enabled at the time of index creation.
+Check the [configuration](https://microsoft.github.io/hyperspace/docs/ug-configuration/) page to see how you can enable lineage for an index.
+
+Once lineage is enabled for an index, Hyperspace adds a new column to the index schema to save the source data file path for each
+index entry. This means enabling lineage increases total storage size used for index files, proportional to the
+number of distinct source data files the index is built on. If you know you wont delete any source data file after index
```
sezruby, line 33:
```suggestion
number of distinct source data files the index is built on. If you know you will not delete any source file after index
```

docs/_docs/03-ug-mutable-dataset.md
```diff
@@ -21,6 +21,19 @@ Now, we offer several options to handle above scenario more efficiently.
...
+creation or you are fine recreating the index using [index refresh in the full mode](#refresh-index---full-mode)
+after deleting some data files, then you dont need to enable lineage for the index.
```
sezruby, line 35:
```suggestion
after deleting some files, then you do not need to enable lineage for the index. 
```

docs/_docs/03-ug-mutable-dataset.md
sezruby: Thanks! and then now we can remove ... Lineage~ part here and, add link to the section. 

```suggestion
An index needs to have [lineage](#lineage) enabled to support deletes during refresh in the incremental mode. 
```

docs/_docs/05-ug-optimize.md
```diff
@@ -0,0 +1,91 @@
...
+hs.optimizeIndex("empIndex", "quick")
+```
+
+### Optimize Index - Full Mode
+In the `"full"` mode, `optimizeIndex` command considers all existing index files as candidates for merging. Therefore,
+Hyperspace does a full scan on current index content and identifies groups of index files which have index records belonging
+to the same bucket according to the index configuration. It then replaces each group with a single index file created through
+merging all the files in that group together. This mode tries to achieve the best query performance improvement via a
+potentially slow optimize index process.
+You should note that `optimizeIndex` differs from `refreshIndex` in the sense that optimize is an index-only operation.
```
sezruby, line 68:
```suggestion
You should note that `optimizeIndex` differs from `refreshIndex` in the sense that `optimizeIndex` is an index-only operation.
```
or optimization? or just use bold - **optimize**?

docs/_docs/05-ug-optimize.md
```diff
@@ -0,0 +1,91 @@
...
+Unlike refresh, when optimize is running on an index, it does not look at the current state of the source data files.
+If there were any source data file changes after index creation or last refresh, optimize wont apply those data changes
```
sezruby, line 70:
```suggestion
If there were any source data file changes after index creation or last refresh, optimize will not apply those data changes
```

docs/_docs/05-ug-optimize.md
```diff
@@ -0,0 +1,91 @@
...
+to the index files.
+
+Assume you have an index with the name `empIndex`. After adding some data files to the dataset `empIndex` is created on
```
sezruby, line 73:
```suggestion
Assume you have an index with the name `empIndex`. After adding some source files to the dataset `empIndex` is created on
```
Let's define two terms clearly and be consistent throughout user documents :)
- index source file: source files in source relations
- index data file: index data file that we built

And could you replace `source data file` with `source file` in other parts?

sezruby (changes requested):

docs/_docs/03-ug-mutable-dataset.md
pirz: "source data file" is a term that we are have been using consistently throughout coding, comments and documentation to refer to an index's original dataset files. I would rather keep "source data file" as is.

docs/_docs/05-ug-optimize.md
pirz: sure, I am using "source data files" and "index files" throughout coding, comments and documentation.

docs/_docs/03-ug-mutable-dataset.md
sezruby: I made this comment because of https://github.com/microsoft/hyperspace/pull/218#discussion_r509664352

I understand the point - "source data files" is also consistent in our code base and github issues, but
just wanted to be consistent in **user doc** and I think it will look better. WDYT? @pirz

sezruby: > Let's define two terms clearly and be consistent throughout user documents :)
> 
> - index source file: source files in source relations
> 
> - index data file: index data file that we built

@pirz @imback82 @apoorvedave1 
Ok let's decide "source data files" or "source files" in user doc. I prefer the latter one because it's shorter and clear. But "source data file" would be good as we're using it in the code base and github issue documentation.

sezruby (dismissed):
LGTM, thanks @priz! 
Let's just decide the term 😁

docs/_docs/05-ug-optimize.md
```diff
@@ -0,0 +1,91 @@
...
+### Optimize Index - Full Mode
+In the `"full"` mode, `optimizeIndex` command considers all existing index files as candidates for merging. Therefore,
+Hyperspace does a full scan on current index content and identifies groups of index files which have index records belonging
+to the same bucket according to the index configuration. It then replaces each group with a single index file created through
+merging all the files in that group together. This mode tries to achieve the best query performance improvement via a
+potentially slow optimize index process.
+You should note that `optimizeIndex` differs from `refreshIndex` in the sense that `optimizeIndex` is an index-only operation.
+Unlike `refreshIndex`, when `optimizeIndex` is running on an index, it does not look at the current state of the source data files.
+If there were any source data file changes after index creation or last refresh, `optimizeIndex` will not apply those data changes
+to the index files.
```
apoorvedave1, line 71: thanks @pirz , could you please move this to the beginning / introduction section of the page? Currently it shows up in "full mode" section. It might lead to minor confusion that it's something specific to full mode.

apoorvedave1 (dismissed):
minor comments but otherwise LGTM 👍 , thanks @pirz

docs/_docs/05-ug-optimize.md
pirz: Good point. Moved it. Thnx!

apoorvedave1 (commented):
LGTM, 👍 thanks @pirz

imback82 (approved):
LGTM, thanks @pirz!

PR #219, opened 2020-10-19
Check source data change before full refresh

### LINKED ISSUE #190: Full refresh should check -no data change- and abort instead of doing full refresh
I think it would be good to check the condition - no data change - for "full" refresh mode?
It doesn't need to be handled with this PR, but it seems necessary.

_Originally posted by @sezruby in https://github.com/microsoft/hyperspace/pull/189#issuecomment-705325520_

sezruby:

 - **Tracking Issue**: Fixes #190
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Check source data change before full refresh and abort if there is no change.

Yes, no version update (and unnecessary data shuffle) when full refresh without any source data change.

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
@@ -40,6 +40,20 @@ class RefreshAction(
     write(spark, df, indexConfig)
   }
 
+  /**
+   * Validate index is in active state for refreshing and there are some changes in
+   * source data file(s).
+   */
+  final override def validate(): Unit = {
+    super.validate()
+
+    if (currentFiles.equals(previousIndexLogEntry.allSourceFileInfos)) {
+      // scalastyle:off
+      println("here")
+      throw NoChangesException("Refresh full aborted as no source data changed.")
+    }
+  }
+
   final override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
     RefreshActionEvent(appInfo, logEntry.asInstanceOf[IndexLogEntry], message)
   }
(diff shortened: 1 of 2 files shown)


src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
```diff
@@ -40,6 +40,20 @@ class RefreshAction(
...
+  /**
+   * Validate index is in active state for refreshing and there are some changes in
+   * source data file(s).
+   */
+  final override def validate(): Unit = {
+    super.validate()
+
+    if (currentFiles.equals(previousIndexLogEntry.allSourceFileInfos)) {
+      // scalastyle:off
+      println("here")
```
pirz, line 52: I guess you had added below these two lines for quick debugging and you may wanna delete them now

// scalastyle:off
   println("here")

src/test/scala/com/microsoft/hyperspace/actions/RefreshActionTest.scala
```diff
@@ -83,16 +94,26 @@ class RefreshActionTest extends SparkFunSuite with SparkInvolvedSuite {
...
     val ex = intercept[HyperspaceException](action.validate())
     assert(ex.getMessage.contains("Refresh is only supported in ACTIVE state"))
   }
+
+  test("validate() fails if there is no source data change.") {
+    testLogEntry.state = ACTIVE
+    when(mockLogManager.getLog(anyInt)).thenReturn(Some(testLogEntry))
+    val action = new RefreshAction(spark, mockLogManager, mockDataManager)
+    val ex = intercept[NoChangesException](action.validate())
+    assert(ex.getMessage.contains("Refresh full aborted"))
```
pirz, line 117: What about checking the full exception msg as it is a fixed/non-parametrized one?

`assert(ex.getMessage.contains("Refresh full aborted as no source data changed"))`

pirz (commented):
LGTM, thanks @sezruby

imback82 (approved):
LGTM, thanks @sezruby!

PR #221, opened 2020-10-20
No operation for non-optimizable index data files in OptimizeIndex

### LINKED ISSUE #204: Support no-op in case of optimize if only 1 "optimizable" file per bucket
### Describe the issue

A small optimization in `optimize` api could be to ONLY run optimize for those buckets where number of small files per bucket is > 1

sezruby:

 - **Tracking Issue**: Fixes #204 
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

If there is only one index data file with the same bucket id, we don't need to optimize the index file in optimizeIndex API.
This PR excludes those unnecessary index data files for OptimizeIndex.

Yes, optimizeIndex API doesn't update the index data if there's no need to do it.

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
@@ -16,6 +16,8 @@
 
 package com.microsoft.hyperspace.actions
 
+import scala.collection.mutable
+
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.SparkSession
@@ -49,9 +51,6 @@ import com.microsoft.hyperspace.util.{HyperspaceConf, PathUtils}
  *
  * `Full` mode: This allows for slow but complete optimization. ALL index files are
  * picked for compaction.
- *
- * TODO: Optimize can be a no-op if there is at most one optimizable file per bucket.
- *  https://github.com/microsoft/hyperspace/issues/204.
  *
  * @param spark SparkSession.
  * @param logManager IndexLogManager for index being refreshed.
@@ -106,20 +105,39 @@ class OptimizeAction(
 
     if (filesToOptimize.isEmpty) {
       throw NoChangesException(
-        "Optimize aborted as no index files smaller than " +
+        "Optimize aborted as no applicable files smaller than " +
           s"${HyperspaceConf.optimizeFileSizeThreshold(spark)} found.")
     }
   }
 
   private lazy val (filesToOptimize, filesToIgnore): (Seq[FileInfo], Seq[FileInfo]) = {
-    if (mode.equalsIgnoreCase(IndexConstants.OPTIMIZE_MODE_QUICK)) {
-      val threshold = HyperspaceConf.optimizeFileSizeThreshold(spark)
-      previousIndexLogEntry.content.fileInfos.toSeq.partition(_.size < threshold)
-    } else {
-      // For 'full' mode, put all the existing index files into 'filesToOptimize' partition so
-      // that one file is created per bucket.
-      (previousIndexLogEntry.content.fileInfos.toSeq, Seq())
+    val (possibleCandidates, largeFilesToIgnore) =
+      if (mode.equalsIgnoreCase(IndexConstants.OPTIMIZE_MODE_QUICK)) {
+        val threshold = HyperspaceConf.optimizeFileSizeThreshold(spark)
+        previousIndexLogEntry.content.fileInfos.toSeq.partition(_.size < threshold)
+      } else {
+        // For 'full' mode, put all the existing index files into 'filesToOptimize' partition
+        // so that one file is created per bucket.
+        (previousIndexLogEntry.content.fileInfos.toSeq, Seq())
+      }
+
+    // Check if there is only one index file in a bucket and exclude the index file
+    // as it doesn't need to be optimized.
+    val filesPerBucket = mutable.Map[String, List[FileInfo]]()
+    possibleCandidates.foreach { f =>
+      // Extract key from the file path; "part-XXXXX" indicates each bucket id.
+      val key = new Path(f.name).getName.substring(0, 10)
+      if (!filesPerBucket.contains(key)) {
+        filesPerBucket.put(key, List(f))
+      } else {
+        filesPerBucket.update(key, filesPerBucket(key) :+ f)
+      }
     }
+
+    // Get candidate index files for optimization based on the length of file list per bucket.
+    val (filesToOptimize, singleFiles) = filesPerBucket.values.partition(_.length > 1)
+
+    (filesToOptimize.flatten.toSeq, singleFiles.flatten.toSeq ++ largeFilesToIgnore)
   }
 
   override def logEntry: LogEntry = {
--- a/src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
@@ -122,22 +122,13 @@ class OptimizeAction(
         (previousIndexLogEntry.content.fileInfos.toSeq, Seq())
       }
 
-    // Check if there is only one index file in a bucket and exclude the index file
-    // as it doesn't need to be optimized.
-    val filesPerBucket = mutable.Map[Int, List[FileInfo]]()
-    possibleCandidates.foreach { f =>
-      val key = BucketingUtils.getBucketId(new Path(f.name).getName).get
-      if (!filesPerBucket.contains(key)) {
-        filesPerBucket.put(key, List(f))
-      } else {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
```diff
@@ -106,20 +105,39 @@ class OptimizeAction(
 
     if (filesToOptimize.isEmpty) {
       throw NoChangesException(
-        "Optimize aborted as no index files smaller than " +
+        "Optimize aborted as no applicable files smaller than " +
```
pirz, line 108: Can you revise this message and make it more clear? (I am not sure "applicable" is an easy-to-know word in the context of optimize).

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
```diff
@@ -106,20 +105,39 @@ class OptimizeAction(
...
+        // so that one file is created per bucket.
+        (previousIndexLogEntry.content.fileInfos.toSeq, Seq())
+      }
+
+    // Check if there is only one index file in a bucket and exclude the index file
+    // as it doesn't need to be optimized.
+    val filesPerBucket = mutable.Map[String, List[FileInfo]]()
+    possibleCandidates.foreach { f =>
+      // Extract key from the file path; "part-XXXXX" indicates each bucket id.
+      val key = new Path(f.name).getName.substring(0, 10)
```
pirz, line 129: I think we should use Spark's `getBucketId()` function to extract bucket-id from a FileName. 
Take a look at below: 
https://github.com/apache/spark/blob/5264164a67df498b73facae207eda12ee133be7d/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/BucketingUtils.scala#L36

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -399,13 +401,46 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
               OptimizeActionEvent(_, _, msg)) =>
-            assert(msg.contains("Optimize aborted as no index files smaller than 1 found."))
+            assert(msg.contains("Optimize aborted as no applicable files smaller than 1 found."))
           case _ => fail()
         }
       }
     }
   }
 
+  test("Verify optimize is a no-op if it is not needed.") {
```
pirz, line 411: Can you make the test name a bit more clear? sth like "optimize is no-op if each bucket has a single index file.".

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
sezruby: How about  "no optimizable index files"?

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
sezruby: Thanks!

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
pirz: Yes, that is precise and much better. Thanks!

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
```diff
@@ -106,20 +106,38 @@ class OptimizeAction(
...
+    // as it doesn't need to be optimized.
+    val filesPerBucket = mutable.Map[Int, List[FileInfo]]()
+    possibleCandidates.foreach { f =>
+      val key = BucketingUtils.getBucketId(new Path(f.name).getName).get
+      if (!filesPerBucket.contains(key)) {
+        filesPerBucket.put(key, List(f))
+      } else {
+        filesPerBucket.update(key, filesPerBucket(key) :+ f)
+      }
     }
```
imback82, line 135: How about
```scala
val filesPerBucket = possibleCandidates.groupBy(f => BucketingUtils.getBucketId(new Path(f.name).getName))
```

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
```diff
@@ -106,20 +106,38 @@ class OptimizeAction(
...
+
+    // Get candidate index files for optimization based on the length of file list per bucket.
```
imback82, line 137: nit: How about explaining why:
```
// Files belonging to buckets that have only one file are ignored because they are already optimized.
```
or you can explain the opposite (e.g., multiple files within a bucket can be merged).

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
```diff
@@ -106,20 +106,38 @@ class OptimizeAction(
...
+    val (filesToOptimize, singleFiles) = filesPerBucket.values.partition(_.length > 1)
```
imback82, line 138: nit: `singleFilesToIgnore`?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -314,7 +314,9 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
       hyperspace.optimizeIndex(indexConfig.indexName)
 
       // Check if latest log file is updated with newly created index files.
-      validateMetadata("index", Set("v__=2"))
+      // Since we appended 3 records for incremental refresh (v_==1) and optimized as v_==2,
+      // some of index data files should be from v__=0.
+      validateMetadata("index", Set("v__=0", "v__=2"))
```
imback82, line 319: Is this an existing test bug?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -399,13 +401,49 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+      assert(latestId == logManager(systemPath, indexConfig.indexName).getLatestId().get)
+
+      // Check emitted events.
+      MockEventLogger.emittedEvents match {
+        case Seq(
+            OptimizeActionEvent(_, _, "Operation started."),
+            OptimizeActionEvent(_, _, msg)) =>
+          assert(
+            msg.contains(
+              "Optimize aborted as no optimizable index files smaller than 268435456 found."))
```
imback82, line 441: Can we check all the index files are less than this number to be complete?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
sezruby: No, the current version of optimizeIndex rewrites single files also. This PR fixes it.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
sezruby: Instead of this, added the check for the number of files per bucket.

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
sezruby: Super neat! Thanks 👍

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
imback82: Shouldn't we check the size as well? Otherwise, `possibleCandidates` can be empty?

sezruby: I thought the size check was for the result of optimizeIndex. I added size check before optimizeIndex. Thanks!

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -391,21 +393,90 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+        .write
+        .parquet(testPath)
+      val df = spark.read.parquet(testPath)
+      hyperspace.createIndex(df, indexConfig)
+
+      // Check if latest log file is updated with newly created index files.
+      val latestId = logManager(systemPath, indexConfig.indexName).getLatestId().get
+
+      // Ensure all index files are smaller than the size threshold so that they can be
+      // possible candidates for optimizeIndex.. In this test, as the test data is tiny,
```
imback82, line 437:
```suggestion
      // possible candidates for optimizeIndex. In this test, as the test data is tiny,
```

imback82 (approved):
LGTM, thanks @sezruby!

imback82 (approved):

Issue #222, opened 2020-10-21
Introduce auxiliary data structure for applying rules


sezruby:
**Feature requested**

In order to optimize rule application process, we need to introduce some auxiliary data structure to share some information, so that we could avoid unnecessary recalculations and share some context to other parts of the rule application.

We could think about 2 types of data structure (i.e. cache):
1) data structure cannot be shared for all rules; it's only available within a rule.
2) data structure can be shared for all rules.

For 1), 
- cons: the tags in IndexLogEntry cannot be shared between Rules, recalculation problem still remains - e.g. signature calculation for each rules - Filter Rule and Join Rule (+ join rule v2).
- pros: object management - candidate index entries are destructed after applying each rule, for each applicable plan. So no need to reset or manage complicated structure something like Map[ (index, relation), tags ].

For 2), 
- cons: hard to maintain the data structure; how/when/where to create/clear/access the data from each rule.
- pros: no recalculation

I tried 2) a little, but maintaining a cache (of IndexOptimizerContext) for all rules is not that simple as each rule is called by Spark RuleExecutor. It's possible to use a wrapper rule applying all other rules in it, though rule-specific statistics by Spark RuleExecutor won't be collected properly. ([ref](https://github.com/apache/spark/blob/eb9966b70055a67dd02451c78ec205d913a38a42/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/rules/RuleExecutor.scala#L221))

Other options:
- TreeNodeTag
  - only available from Spark3.0

Possible candidates for auxiliary data:
- General) signature value for each candidate logical relation.
- HybridScan) if the current index application requires Hybrid Scan or not.
- HybridScan) common file size of index source files and current source files in a given relation.
  - can be used for rank algorithm
- HybridScan) current files / appended files / deleted files
.. etc

**Acceptance criteria** 

**Success criteria**

**Additional context**

sezruby:
Closed by #223

PR #223, opened 2020-10-21
Introduce IndexLogEntryTag for auxiliary data while applying rules

sezruby:

 - **Tracking Issue**: #222 
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

This PR is one possible implementation of the first approach in #222.

In order to optimize rule application process, this PR introduces IndexLogEntryTag which can store auxiliary
data while applying rules and defines a tag for Hybrid Scan:

```
  // HYBRIDSCAN_REQUIRED indicates if Hybrid Scan is required for this index or not.
  // This is set in getCandidateIndexes and utilized in transformPlanToUseIndex.
```

Tags are stored in each `IndexLogEntry` instances which can be accessed by `indexManager.getIndexes`. 
Therefore the lifetime of each tag is the same of each index instance, which means if the indexManager returns cached index log entries, tags are available and reusable for other queries while the cached log entries are available.

ref) tag implementation from Spark 3.0 - https://github.com/apache/spark/blob/eb9966b70055a67dd02451c78ec205d913a38a42/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala#L93

No

Unit test


sezruby:
@imback82 @pirz @apoorvedave1 Please review this change when you have the time. Thanks!

imback82:
Sorry for the delay. I will try to get to this this weekend.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -172,11 +178,14 @@ object RuleUtils {
...
     assert(getLogicalRelation(plan).isDefined)
 
-    val transformed = if (HyperspaceConf.hybridScanEnabled(spark)) {
-      transformPlanToUseHybridScan(spark, index, plan, useBucketSpec)
-    } else {
-      transformPlanToUseIndexOnlyScan(spark, index, plan, useBucketSpec)
-    }
+    val transformed =
+      if (HyperspaceConf.hybridScanEnabled(spark) && index
+        .getTagValue(IndexConstants.INDEX_HYBRIDSCAN_REQUIRED_TAG)
```
imback82, line 183: Hmm. To me, this seems a bit fragile since this is calculated in `getCandidateIndexes` and used in `transformPlanToUseIndex`. How can we guarantee that the tag was generated with the correct `plan`?

sezruby: getCandidateIndex(plan) creates copy of indexLogEntry - so, for example, JoinIndexRule has `lIndex` and `rIndex` for left plan and right plan. Candidates indexes for both are selected and applied respectively. In this way, we could keep the tags for each plan.

sezruby: More example in rank:
https://github.com/microsoft/hyperspace/pull/164/files#diff-57915e1583c4c52ae6ccbf6324db8db200813f3c8bf3b6b4623b299ca80aaa64L44

imback82: Can't you just tag the logical plan used? I am just afraid that it may become harder to maintain in a long run.

sezruby: There's no tag in Spark 2.4... that's the problem. 😟

imback82: I mean tagging the plan to the log entry?

sezruby: Ok then I'll try sth like
```
private val tags: mutable.Map[IndexLogEntryTag[_], (plan, Any)] = mutable.Map.empty
```

imback82: Can't you have `IndexLogEntryTag[LogicalPlan]("plan")` as a separate key? And make sure you are operating on the same plan?

sezruby: How about this
```
@JsonIgnore
  private val tags: mutable.Map[(LogicalPlan, IndexLogEntryTag[_]), Any] = mutable.Map.empty

  def setTagValue[T](plan: LogicalPlan, tag: IndexLogEntryTag[T], value: T): Unit = {
    tags((plan, tag)) = value
  }

  def getTagValue[T](plan: LogicalPlan, tag: IndexLogEntryTag[T]): Option[T] = {
    tags.get((plan, tag)).map(_.asInstanceOf[T])
  }

  def unsetTagValue[T](plan: LogicalPlan, tag: IndexLogEntryTag[T]): Unit = {
    tags.remove((plan, tag))
  }
}

// A tag of a `IndexLogEntry`, which defines name and type.
case class IndexLogEntryTag[T](name: String)
```

imback82: Yea, this seems fine. And I guess we could use `null` if a tag is not associated with a plan?

```
def setTagValue[T](tag: IndexLogEntryTag[T], value: T): Unit = {
  tags((null, tag)) = value
}
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -82,6 +82,13 @@ object RuleUtils {
       val commonCnt = inputSourceFiles.count(entry.allSourceFileInfos.contains)
       val deletedCnt = entry.allSourceFileInfos.size - commonCnt
 
+      // If there is no change in source dataset, this index can be applied by
+      // transformPlanToUseIndexOnlyScan.
+      entry.setTagValue(
```
imback82, line 87: Do we need to tag even if `isHybridScanCandidate` returns false?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -172,11 +179,14 @@ object RuleUtils {
...
 
-    val transformed = if (HyperspaceConf.hybridScanEnabled(spark)) {
-      transformPlanToUseHybridScan(spark, index, plan, useBucketSpec)
-    } else {
-      transformPlanToUseIndexOnlyScan(spark, index, plan, useBucketSpec)
-    }
+    val transformed =
+      if (HyperspaceConf.hybridScanEnabled(spark) && index
+            .getTagValue(getLogicalRelation(plan).get, IndexConstants.INDEX_HYBRIDSCAN_REQUIRED_TAG)
+            .getOrElse(false)) {
```
imback82, line 185: this tag should be always present if hybrid scan is enabled, no?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -143,6 +144,10 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
             if (expectCandidateIndex) {
               assert(indexes.length == 1)
               assert(indexes.head.name == "index1")
+              assert(
+                indexes.head
+                  .getTagValue(plan, IndexConstants.INDEX_HYBRIDSCAN_REQUIRED_TAG)
+                  .getOrElse(false) == expectedHybridScanTag)
```
imback82, line 150: Can you test the return value of `getTagValue`? You can make `expectedHybridScanTag` as `Option[Boolean]` if needed.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -56,6 +56,12 @@ object IndexConstants {
   // See https://github.com/microsoft/hyperspace/issues/185
   val INDEX_RELATION_IDENTIFIER: (String, String) = "indexRelation" -> "true"
 
+  // Tags are used while applying rules.
+  // INDEX_HYBRIDSCAN_REQUIRED_TAG indicates if Hybrid Scan is required for this index or not.
+  // This is set in getCandidateIndexes and utilized in transformPlanToUseIndex.
+  val INDEX_HYBRIDSCAN_REQUIRED_TAG: IndexLogEntryTag[Boolean] =
+  IndexLogEntryTag[Boolean]("hybridScanRequired")
+
```
imback82, line 64: This file is getting out of control with random constants. Can you create `IndexLogEntryTags` and put tag related constants there? I should be able to use as `IndexLogEntryTags.HYBRID_SCAN_REQUIRED`, for example.

imback82 (commented):
@sezruby Could you rebase #164 against this PR? I want to check what the tag usage will be like before closing this one. Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
sezruby: @imback82 I updated #164 based on this PR. Thanks!

sezruby:
comment test

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
rapoth: Comment test 4

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
sezruby: @imback82 Could you check the change and merge this change? Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -486,8 +488,35 @@ case class IndexLogEntry(
...
     config.hashCode + signature.hashCode + numBuckets.hashCode + content.hashCode
   }
+
+  /**
+   * A mutable map for holding auxiliary information of this index log entry while applying rules.
+   */
+  @JsonIgnore
+  private val tags: mutable.Map[(LogicalPlan, IndexLogEntryTag[_]), Any] = mutable.Map.empty
+
+  def setTagValue[T](plan: LogicalPlan, tag: IndexLogEntryTag[T], value: T): Unit =
```
imback82, line 498: Can you add `{ }` around these `def`s?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.index
+
+object IndexLogEntryTags {
+  // INDEX_HYBRIDSCAN_REQUIRED_TAG indicates if Hybrid Scan is required for this index or not.
+  // This is set in getCandidateIndexes and utilized in transformPlanToUseIndex.
+  val INDEX_HYBRIDSCAN_REQUIRED_TAG: IndexLogEntryTag[Boolean] =
```
imback82, line 22: nit: just call this `HYBRIDSCAN_REQUIRED`? `IndexLogEntryTags.HYBRIDSCAN_REQUIRED` sounds fine?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -82,12 +82,23 @@ object RuleUtils {
...
-        // For append-only Hybrid Scan, deleted files are not allowed.
-        deletedCnt == 0 && commonCnt > 0
+
+      // For append-only Hybrid Scan, deleted files are not allowed.
+      lazy val isAppendOnlyCandidate = !hybridScanDeleteEnabled && deletedCnt == 0 &&
+        commonCnt > 0
+
+      val isCandidate = isDeleteCandidate || isAppendOnlyCandidate
+      if (isCandidate) {
+        // If there is no change in source dataset, the index can be applied by
```
imback82, line 94:
```suggestion
        // If there is no change in source dataset, the index will be applied by
```

imback82 (commented):
LGTM except for few nits, thanks @sezruby!

imback82:
@sezruby Can you update the description with the lifetime of tags with the current code? (e.g., per rule)

sezruby:
@imback82 Updated pr description. Thanks!

imback82:
> Tags are stored in each `IndexLogEntry` instances created by `indexManager.getIndexes`.
> Therefore the lifetime of each tag is the same of each index instance which means tags are only usable each during rule application of each candidate plan (Filter, Join).

The index manager could be `CachingIndexCollectionManager`. In this case, tags would be shared across rules, etc., right? Is that the intention? (Sorry, I didn't fully understand this part: "tags are only usable each during rule application of each candidate plan (Filter, Join).")

sezruby:
"each rule application" means if there are 3 filter nodes, there will be 3 tries of rule application. Tags are not shared between them because indexManager.getIndexes returns "cloned" index log entries. 
So even it's not shared within a rule application in case Join left & right.

imback82:
> Sorry I don't know why I thought it's "cloned"; this's the reason I didn't use "df" as a part of key of the map at first. 😖
> But anyway, if cached index entries are used, tags can be shared across rules and also queries! I think this is GREAT though it's not intended 😁

Great. As long as the cached tags do not affect other rules, it's good for now.

> Then, should we add "max number of caching entires" or is it enough to rely on INDEX_CACHE_EXPIRY_DURATION_SECONDS (default 300s)? Or even we could reuse the index log entry (with tags) if there's no update on the index.
> If needed, it would be better to work with another PR as this is a dependent pr for #164.

Yea, let's think about this separately.

 Merging to master!

imback82 (approved):

PR #225, opened 2020-10-23
Support Hybrid Scan for partitioned source data

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Support Hybrid Scan for partitioned source data. 

For partitioned data, we should set `basePath` option to read partitioned columns. Otherwise, partitioned columns are not included in the loaded data. In Hybrid Scan, newly appended source files in a partitioned dataset should be loaded with `basePath` so that all partitioned columns are handled properly.

Yes.
Without this change, Hybrid Scan with an index on partitioned source doesn't work - it throws an exception during query execution.

Unit test


src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -381,18 +381,42 @@ object RuleUtils {
...
+          if (location.partitionSchema.isEmpty) {
+            new InMemoryFileIndex(spark, filesAppended, Map(), None)
+          } else {
+            // Set "basePath" so that partitioned columns are also included in the output schema.
+            val basePath = location.partitionSpec.partitionColumns
+              .foldRight(location.partitionSpec.partitions.head.path) { (col, path) =>
+                if (path.getName.contains(col.name)) {
+                  path.getParent
+                } else {
+                  path
```
imback82, line 399: Is `else` path a possible scenario?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -381,18 +381,42 @@ object RuleUtils {
...
-            location = newLocation,
-            dataSchema = StructType(indexSchema.filter(baseRelation.schema.contains(_))),
-            options =
-              fsRelation.options + IndexConstants.INDEX_RELATION_IDENTIFIER)(spark)
+          if (location.partitionSchema.isEmpty) {
+            new InMemoryFileIndex(spark, filesAppended, Map(), None)
+          } else {
+            // Set "basePath" so that partitioned columns are also included in the output schema.
+            val basePath = location.partitionSpec.partitionColumns
+              .foldRight(location.partitionSpec.partitions.head.path) { (col, path) =>
```
imback82, line 395: Can you add a comment to the code why `foldRight` is required here? (e.g., traversing back from the child folder, etc.) Btw, is it guaranteed that `partitionColumns` will be in that order?

imback82: Btw, wouldn't rootPath be same as basePath if hive-partitioning is used (`location.partitionSchema.nonEmpty`)?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -381,18 +381,42 @@ object RuleUtils {
...
         val newLocation =
-          new InMemoryFileIndex(spark, filesAppended, Map(), None)
-        val newRelation =
-          fsRelation.copy(
-            location = newLocation,
-            dataSchema = StructType(indexSchema.filter(baseRelation.schema.contains(_))),
-            options =
-              fsRelation.options + IndexConstants.INDEX_RELATION_IDENTIFIER)(spark)
+          if (location.partitionSchema.isEmpty) {
+            new InMemoryFileIndex(spark, filesAppended, Map(), None)
```
imback82, line 391: How about:
```scala
val options = if (location.partitionSchema.isEmpty) {
  Map()
} else {if (location.partitionSchema.isEmpty) {
  // Map() with basePath
}
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: https://github.com/apache/spark/blob/d6a68e0b67ff7de58073c176dd097070e88ac831/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningUtils.scala#L220

partitionColumns seems be in order, though it cannot be guaranteed clearly.
But I think using "=" in the condition below is also risky..

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Seems it's not possible? null value is handled with default value - `A=a\B=__HIVE_DEFAULT_PARTITION__`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Ok let me just use the number of partition columns as a partition path always includes all partition column names and handle  possible exceptional cases(?) on demand..

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -35,26 +35,31 @@ import com.microsoft.hyperspace.util.FileUtils
...
+  protected val sampleDataFormatDelete = sampleDataLocationRoot + "sample2"
+  protected val sampleDataFormatDelete2 = sampleDataLocationRoot + "sample3"
+  protected val sampleDataFormatDelete3 = sampleDataLocationRoot + "sample4"
+  protected val sampleDataFormatDelete4 = sampleDataLocationRoot + "sample5"
+  protected val sampleDataFormatBoth = sampleDataLocationRoot + "sample6"
+  protected val sampleDataFormat2Append = sampleDataLocationRoot + "sample7"
+  protected val sampleDataFormat2Delete = sampleDataLocationRoot + "sample8"
+  var hyperspace: Hyperspace = _
+  var fileFormat = ""
+  var fileFormat2 = ""
```
imback82, line 51: Do you need to create `var` for this? You can just use `"parquet"` and `"json"` in the caller directly (much easier to follow).

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -35,26 +35,31 @@ import com.microsoft.hyperspace.util.FileUtils
...
+  var indexScanWithAppendFilesExpected = false
```
imback82, line 52: I didn't try hard to understand what this variable does between test suites, and what is this for?

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -35,26 +35,31 @@ import com.microsoft.hyperspace.util.FileUtils
...
 
   // Creates an index with given 'df' and 'indexConfig'. Then copies the first 'appendCnt'
```
imback82, line 54: Update

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -35,26 +35,31 @@ import com.microsoft.hyperspace.util.FileUtils
...
   // number of input files from 'df' and deletes the last 'deleteCnt' of the input files.
   def setupIndexAndChangeData(
-      df: DataFrame,
+      sourceFileFormat: String,
+      dataPath: String,
```
imback82, line 58: `sourceFileFormat` and `sourcePath`, or `dataFileFormat` and `dataPath`.

src/test/scala/com/microsoft/hyperspace/index/PartitionedDataHybridScanTest.scala
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+package com.microsoft.hyperspace.index
+
+import org.apache.hadoop.fs.Path
+
+import com.microsoft.hyperspace.Hyperspace
+import com.microsoft.hyperspace.util.FileUtils
+
+class PartitionedDataHybridScanTest extends HybridScanTest with HyperspaceSuite {
```
imback82, line 24: Can we add tests where users specified "basePath"?

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -252,37 +270,38 @@ class HybridScanTest extends QueryTest with HyperspaceSuite {
...
               assert(bucketSpec.numBuckets === 200)
               p
-            case p @ FileSourceScanExec(_, _, _, _, _, dataFilters, _) =>
+            case p @ FileSourceScanExec(_, _, _, partitionFilters, _, dataFilters, _) =>
               // Check filter pushed down properly.
               assert(
-                dataFilters.toString.contains(" >= 2000)") && dataFilters.toString.contains(
-                  " <= 4000)"))
+                (dataFilters.toString.contains(" >= 2000)") &&
+                  dataFilters.toString.contains(" <= 4000)")) ||
```
imback82, line 277: Once you start adding more conditions in `assert` - especially `||` testing different objects `dataFilters` and `partitionFilters`, it's hard to reason which condition is met. Can you separate these into different `assert`s?

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
imback82: And you can update the sample* variables as `sampleParquetDataAppend`, `sampleJsonDataAppend`, etc., unless your goal is to test different combinations like `csv`, etc. which I don't see now.

src/test/scala/com/microsoft/hyperspace/index/PartitionedDataHybridScanTest.scala
imback82: For example, we could have a test where an index was created with `spark.read.option("basePath", "/tmp").parquet("/tmp/a=1")`, and if a new partition, `a=2` is introduced, data under `a=2` shouldn't be read. (I believe that's the right behavior).

src/test/scala/com/microsoft/hyperspace/index/PartitionedDataHybridScanTest.scala
```diff
@@ -0,0 +1,103 @@
...
+        appendCnt = 1,
+        deleteCnt = 1)
+      setupIndexAndChangeData(
+        fileFormat2,
+        sampleDataFormat2Delete,
+        indexConfig1.copy(indexName = "index_JsonDelete"),
+        appendCnt = 0,
+        deleteCnt = 2)
+    }
+  }
```
imback82, line 102: Do we have a test that tests when a new partition is added? If it's combined with `HybridScanTest`, I suggest separating them out to better track.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -381,18 +383,36 @@ object RuleUtils {
...
+            _,
+            _) =>
+        val options = if (location.partitionSchema.isEmpty) {
+          Map[String, String]()
+        } else {
+          // Retrieve basePath from a partition path.
+          val basePath = location.partitionSpec.partitionColumns
+            .foldLeft(location.partitionSpec.partitions.head.path) { (path, _) =>
+              path.getParent
+            }
```
apoorvedave1, line 398: the foldLeft is confusing. could you change it as below:
```suggestion
          val basePath: Path = {
            var path = location.partitionSpec.partitions.head.path
            for (_ <- location.partitionSpec.partitionColumns.indices) {
              path = path.getParent
            }
            path
```

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
sezruby: Actually it was for `delta`. Can we keep this variable names for tests of Delta Lake Hybrid Scan?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -381,18 +383,36 @@ object RuleUtils {
...
+
+          // Set "basePath" so that partitioned columns are also included in the output schema.
+          Map("basePath" -> basePath.toString)
```
apoorvedave1, line 401: could you please point me to where this key-value is being used. I couldn't find it.

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
sezruby: This `var` is for the code reuse and skipping not related test. Partitioned source data doesn't support "Filter Rule + Parquet format" - scanning index data & source data using one scan node.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: I prefer FP way - not using `var` if possible.
How about this impl?
```
   val basePath =
            Iterator
              .iterate(location.partitionSpec.partitions.head.path)(_.getParent)
              .drop(location.partitionSpec.partitionColumns.length).next
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: It's used in https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningAwareFileIndex.scala#L221.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: +1 for `foldLeft`

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
imback82: OK, can you add some comments on how this `var` should be used? It was not intuitive why this was false for partitioned tests.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Ok then let me keep `foldLeft` impl. Thanks!

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
sezruby: I changed them as `val` and added some comment for this. Thanks!

src/test/scala/com/microsoft/hyperspace/index/PartitionedDataHybridScanTest.scala
sezruby: I added 2 tests using basePath but not sure if they cover what you meant.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -181,6 +166,21 @@ object RuleUtils {
     transformed
   }
 
+  /**
```
sezruby, line 169: rearranged by reformating

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -381,18 +381,36 @@ object RuleUtils {
...
+            baseOutput,
+            _,
+            _) =>
+        val options = if (location.partitionSchema.isEmpty) {
+          Map[String, String]()
+        } else {
+          // Retrieve basePath from a partition path.
+          val basePath = location.partitionSpec.partitionColumns
+            .foldLeft(location.partitionSpec.partitions.head.path) { (path, _) =>
+              path.getParent
```
imback82, line 395: Since we are more or less reverse engineering to compute base path, can we factor out to a function and add more descriptive comments? For example:
```scala
  private def extractBasePath(partitionSpec: PartitionSpec): Option[String] = {
    if (partitionSpec == PartitionSpec.emptySpec) {
      None
    } else {
      // For example, we could have the following in PartitionSpec:
      //   - partition columns = "col1", "col2"
      //   - partitions: "/path/col1=1/col2=1", "/path/col1=1/col2=2", etc.
      // , and going up the same number of directory levels as the number of partition columns
      // will compute the base path. Note that PartitionSpec.partitions will always contain
      // all the partitions in the path, so "partitions.head" is taken as an initial value.
      val basePath = partitionSpec.partitionColumns
        .foldLeft(partitionSpec.partitions.head.path)((path, _) => path.getParent)
      Some(basePath.toString)
    }
  }
```

imback82: Question: do we need to worry about the case where there are multiple base paths in `PartitioningAwareFileIndex.basePaths`?

imback82 (commented):
I have few minor comments, but generally looking good to me!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Seems we don't: https://github.com/apache/spark/blob/d6a68e0b67ff7de58073c176dd097070e88ac831/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningUtils.scala#L166

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -35,100 +35,123 @@ import com.microsoft.hyperspace.util.FileUtils
...
+  // Introduce fileFormat to support "delta" format soon.
+  val fileFormat = "parquet"
+  val fileFormat2 = "json"
+
+  // This flag is for testing plan transformation if appended files could be load with index data
+  // scan node.
+  // For now, it's applied for a very specific case: FilterIndexRule, Parquet source format,
+  // no partitioning, no deleted files. This is false by default and will be changed to true if
+  // the test is required.
+  var indexDataScanWithAppendedFilesExpected = false
```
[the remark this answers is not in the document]
sezruby: I created a new suite but not sure it's the best. Let me revisit the test structure later for Delta lake.

src/test/scala/com/microsoft/hyperspace/index/PartitionedDataHybridScanTest.scala
```diff
@@ -0,0 +1,259 @@
...
+          .toDF("Date", "RGUID", "Query", "imprs", "clicks")
+          .write
+          .partitionBy("Date")
+          .mode("append")
+          .format(fileFormat)
+          .save(testPath)
+
+        assert(FileUtils.getDirectorySize(new Path(testPath + "/Date=2019-10-03")) > 0)
+
+          {
```
imback82, line 233: indentation seems off.

src/test/scala/com/microsoft/hyperspace/index/PartitionedDataHybridScanTest.scala
```diff
@@ -0,0 +1,259 @@
...
+      val indexSourcePaths = Seq(testPath + "/Date=2017-09-03", testPath + "/Date=2018-09-03")
+      val df =
+        spark.read.format(fileFormat).option("basePath", testPath).load(indexSourcePaths: _*)
+      hyperspace.createIndex(df, IndexConfig("indexName", Seq("Date"), Seq("clicks")))
+      def filterQuery(df: DataFrame): DataFrame = {
+        df.filter(df("Date") > "2000-01-01").select(df("clicks"))
+      }
+
+      withIndex("indexName") {
+        // Check the index is applied regardless of Hybrid Scan config.
```
imback82, line 201: We could move this below `{` similar to line 234.

src/test/scala/com/microsoft/hyperspace/index/HybridScanTestSuite.scala
```diff
@@ -0,0 +1,636 @@
...
+import org.apache.spark.sql.execution.exchange.ShuffleExchangeExec
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.sources.DataSourceRegister
+
+import com.microsoft.hyperspace.{Hyperspace, Implicits, SampleData}
+import com.microsoft.hyperspace.index.execution.BucketUnionExec
+import com.microsoft.hyperspace.index.plans.logical.BucketUnion
+import com.microsoft.hyperspace.util.FileUtils
+
+trait HybridScanTestSuite extends QueryTest with HyperspaceSuite {
```
imback82, line 35: Could you not change the file name in this PR? All the diffs are gone. :(

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -381,24 +381,54 @@ object RuleUtils {
...
+            baseOutput,
+            _,
+            _) =>
+        val options = extractBasePath(location.partitionSpec) match {
+          case Some(basePath) =>
+            // Set "basePath" so that partitioned columns are also included in the output schema.
+            Map("basePath" -> basePath)
+          case _ =>
+            Map[String, String]()
+        }
```
imback82, line 395: ni: Can we utilize `map`, `getOrElse`, `orNull`, etc. of `Option` if applicable (as long as it doesn't hurt the readability)?
```scala
        val options = extractBasePath(location.partitionSpec)
          .map { basePath =>
            // Set "basePath" so that partitioned columns are also included in the output schema.
            Map("basePath" -> basePath)
          }
          .getOrElse(Map()) // I believe this may not need type info anymore (didn't check), although either one is fine.
```

src/test/scala/com/microsoft/hyperspace/index/HybridScanTest.scala
```diff
@@ -446,9 +326,11 @@ class HybridScanTest extends QueryTest with HyperspaceSuite {
     }
   }
 
-  test("Delete-only: join index, deleted files should be excluded from each index relation.") {
-    val df1 = spark.read.parquet(sampleParquetDataLocationDelete)
-    val df2 = spark.read.parquet(sampleParquetDataLocationDelete3)
+  test(
+    "Delete-only: join rule, deleted files should be excluded from each index data" +
+      " relation.") {
```
imback82, line 331: nit: Is this change required?

src/test/scala/com/microsoft/hyperspace/index/HybridScanGeneralTest.scala
```diff
@@ -0,0 +1,167 @@
...
+
+    dfFromSample.write.format(fileFormat).save(sampleDataFormatAppend)
+    dfFromSample.write.format(fileFormat).save(sampleDataFormatAppend2)
+    dfFromSample.write.format(fileFormat).save(sampleDataFormatDelete)
+    dfFromSample.write.format(fileFormat).save(sampleDataFormatDelete2)
+    dfFromSample.write.format(fileFormat).save(sampleDataFormatDelete3)
+    dfFromSample.write.format(fileFormat).save(sampleDataFormatDelete4)
+    dfFromSample.write.format(fileFormat).save(sampleDataFormatBoth)
+    dfFromSample.write.format(fileFormat2).save(sampleDataFormat2Append)
+    dfFromSample.write.format(fileFormat2).save(sampleDataFormat2Delete)
```
imback82, line 42: Can we remove unused sample data?

src/test/scala/com/microsoft/hyperspace/index/HybridScanGeneralTest.scala
```diff
@@ -0,0 +1,167 @@
...
+
+package com.microsoft.hyperspace.index
+
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.execution.datasources._
+
+import com.microsoft.hyperspace.Hyperspace
+
+class HybridScanGeneralTest extends HybridScanTestSuite {
```
imback82, line 25: Maybe `HybridScanForNonPartitionedDataSuite` and `HybridScanForPartitionedDataSuite`, if I understand the separation correctly? (A small description would also help)

imback82 (commented):
LGTM (pending few minor comments), thanks @sezruby!

src/test/scala/com/microsoft/hyperspace/index/HybridScanGeneralTest.scala
sezruby: They are all used.. do you think it's not necessary to run the testset for both partitioned data and non-partitioend data?

imback82: Ah sorry, I thought this suite was testing only the tests where partitioned data suite couldn't test.

imback82 (approved):

PR #227, opened 2020-10-25
Pluggable source provider

imback82:

Introduce pluggable source provider.

This PR proposes to introduce pluggable source provider APIs so that new source such as Delta Lake can be easily plugged in.

Yes. This PR exposes a new way to plug in a source from which an index can be created.

Existing tests


imback82:
@sezruby Did you get a chance to implement #197 using this PR?

src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.spark.sql.execution.datasources.LogicalRelation
+
+import com.microsoft.hyperspace.index.{FileIdTracker, Relation}
+
+/**
+ * ::Experimental::
+ * A trait that a data source should implement so that an index can be created/managed and
+ * utilized for the data source.
+ *
+ * @since 0.3.0
```
imback82, line 29: These need to be updated to `0.4.0`

src/main/scala/com/microsoft/hyperspace/util/CacheWithTransform.scala
```diff
@@ -0,0 +1,45 @@
...
+ * using the initial value.
+ *
+ * The cached initial/final values will be updated only if the initial value is modified.
+ *
+ * @param init Function to produce an initial value.
+ * @param transform Function to transform the initial value.
+ * @tparam A Type of the initial value.
+ * @tparam B Type of the final value.
+ */
+class CacheWithTransform[A, B](val init: () => A, val transform: A => B) {
```
imback82, line 31: I need to add a test for this class.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -0,0 +1,159 @@
...
+import com.microsoft.hyperspace.util.{CacheWithTransform, HashingUtils, HyperspaceConf}
+
+/**
+ * Default implementation for file-based Spark built-in sources such as parquet, csv, json, etc.
+ *
+ * This source can support relations that meet the following criteria:
+ *   - The relation is [[HadoopFsRelation]] with [[PartitioningAwareFileIndex]] as file index.
+ *   - Its file format implements [[DataSourceRegister]].
+ */
+class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedSourceProvider {
```
imback82, line 37: I need to create a test for this class.

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * [[FileBasedSourceProviderManager]] is responsible for loading source providers which implements
+ * [[FileBasedSourceProvider]] and running APIs for each provider loaded.
+ *
+ * Each API in [[FileBasedSourceProvider]] returns [[Option]] and this manager ensures that only
+ * one provider returns [[Some]] for each API.
+ *
+ * @param spark Spark session.
+ */
+class FileBasedSourceProviderManager(spark: SparkSession) {
```
imback82, line 38: I need to create a test for this.

src/main/scala/com/microsoft/hyperspace/util/HyperspaceConf.scala
```diff
@@ -71,6 +71,20 @@ object HyperspaceConf {
       .toBoolean
   }
 
+  def fileBasedSourceBuilders(spark: SparkSession): String = {
+    spark.sessionState.conf
+      .getConfString(
+        "spark.hyperspace.index.sources.fileBasedBuilders",
+        "com.microsoft.hyperspace.index.sources.default.DefaultFileBasedSourceBuilder")
```
sezruby, line 78: Why this config is necessary? Can we just create all builders?

imback82: To control what builders are being loaded similar to how plugins, listeners, etc. are loaded in `SparkContext`. What's the benefit of creating all builders available? The only benefit I can think of is not having to specify this config, but having a tight control seems better to me.

Issue #200, opened 2020-10-12
Numeric id-based lineage column


sezruby:
**Feature requested**

Currently, the lineage column in index data contains the full path of each source file, in String type.
This is an initial approach for the lineage column to support "delete" dataset.

There are some disadvantages for this approach:
- storage bloat
- high memory footprint when reading the index data
- high comparison cost - string comparison

though,
- easy to maintain
- can apply partition filter (not implemented yet)

We can optimize the lineage column by assigning a unique id to each source file.

The following metadata change is required and need to keep the last unique id for newly appended source files.
```
// cur
case class FileInfo(name: String, size: Long, modifiedTime: Long)

// suggested
case class FileInfo(name: String, size: Long, modifiedTime: Long, id: Long)
```

With this, the lineage column can be defined with the "id" value/Long type instead of the full path / String type.

**Acceptance criteria** 

Should work well with the current features using the lineage column:
- [ ] incremental refresh
- [ ] HybridScan

**Success criteria**

**Additional context**

rapoth:
@sezruby Thanks for writing this up! Can you also put up the success criteria? Also, how can we still support partitioning? (maybe use ids instead of actual partition keys for push-down?)

pirz:
Added by #234

Issue #198, opened 2020-10-12
Revisit index metadata for mutable dataset


sezruby:
### Describe the issue

This issue is for the discussion started in https://github.com/microsoft/hyperspace/pull/194#discussion_r502736181

We are currently proposing the features for mutable dataset, and introduced `appendedFiles` and `deletedFiles` for new refresh modes - `incremental`, `quick` (planned).
```
/**
   * Hdfs file properties.
   * @param content Content object representing Hdfs file based data source.
   * @param appendedFiles Appended files since the last time derived dataset was updated.
   * @param deletedFiles Deleted files since the last time derived dataset was updated.
   */
  case class Properties(
      content: Content,
      appendedFiles: Seq[String] = Nil,
      deletedFiles: Seq[String] = Nil)

relations.head.data.properties.deletedFiles
relations.head.data.properties.appendedFiles
```

The current master is based on "mixed" semantic approach for `source.Content` in IndexLogEntry - there could be  "unhandled" appendedFiles and deletedFiles during the new refresh mode. "unhandled" means their data is not included in the index data - so need to handle them appropriately for correctness. When updating unhandled appendedFiles and deletedFiles, "mixed" version updates the list of source files and its signature value though the index data doesn't include the data from "appendedFiles" and "deletedFiles". This "mixed" semantic might cause some complexity in code base.
@imback82 suggested a new semantic for this, always "since" semantic approach.

The following example might help understand what "mixed" and "since" (from @imback82 's https://github.com/microsoft/hyperspace/pull/194#discussion_r502857227)

There are source files: f1, f2, f3
* An index is created.
* A source file is added: f4
* A source file is deleted: f1
* Refresh "quick" (metadata only operation)

With the current semantics (mix of "latest" and "since"), we will have the following in the latest index log entry:
* source files / signature: f2, f3, f4 / sig(f2, f3, f4)
* appendedFiles: f4
* deletedFiles: f1

With the semantics of always using "since":
* source files / signature: f1, f2, f3 / sig(f1, f2, f3)
* appendedFiles: f4
* deletedFiles: f1
* latestSourceFilesSignature: sig(f2, f3, f4)

However, to support Hybrid scan effectively with the `quick` refresh, we need to keep the updated signature leveraging `appendedFiles` and `deletedFiles`.

imback82:
+1 for
```
With the semantics of always using "since":

source files / signature: f1, f2, f3 / sig(f1, f2, f3)
appendedFiles: f4
[...]

PR #228, opened 2020-10-25
Store appended files and deleted files as Content type in IndexLogEntry

### LINKED ISSUE #210: appendedFiles / deletedFiles in IndexLogEntry can be stored as Seq[FileInfo] instead of Seq[String]
> `RefreshDeleteAction` handles the modified files as deleted files, and also as appended files.

Does it make sense to update `appendedFiles` / `deletedFiles` to be `Seq[FileInfo]` in index log entry for completeness (esp. if someone is reading the index log entry)?

_Originally posted by @imback82 in https://github.com/microsoft/hyperspace/pull/207#issuecomment-707868254_

sezruby:

 - **Tracking Issue**: Fixes #210 
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Store the type of appendedFiles and deletedFiles as Content type instead of list of String.
This change is required for #200 and #198.

```
case class Properties(
      content: Content,
-      appendedFiles: Seq[String] = Nil,
-      deletedFiles: Seq[String] = Nil)
+      appendedFiles: Option[Content] = None,
+      deletedFiles: Option[Content] = None)
```

ref) `None` value is serialized as "null" in JSON
```
...
appendedFiles: null
...
```

Yes.
This is a breakage change.
Form of `appendedFiles` and `deletedFiles` in IndexLogEntry has changed - a user should rebuild indexes.

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -102,7 +102,10 @@ private[actions] abstract class RefreshActionBase(
   protected lazy val deletedFiles: Seq[FileInfo] = {
     val relation = previousIndexLogEntry.relations.head
     val originalFiles = relation.data.properties.content.fileInfos
-    (originalFiles -- currentFiles).toSeq
+
+    // TODO: Add test for the scenario where existing appendedFiles and newly appended
+    //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
+    ((originalFiles -- currentFiles) ++ previousIndexLogEntry.deletedFiles).toSeq
   }
 
   /**
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -109,7 +109,7 @@ private[actions] abstract class RefreshActionBase(
     val prevDeletedFiles =
       previousIndexLogEntry.deletedFiles.filterNot(f => delFileNames.contains(f.name))
 
-    // TODO: Add test for the scenario where existing appendedFiles and newly appended
+    // TODO: Add test for the scenario where existing deletedFiles and newly deleted
     //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
     delFiles.toSeq ++ prevDeletedFiles
   }
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -102,10 +102,16 @@ private[actions] abstract class RefreshActionBase(
   protected lazy val deletedFiles: Seq[FileInfo] = {
     val relation = previousIndexLogEntry.relations.head
     val originalFiles = relation.data.properties.content.fileInfos
+    val delFiles = originalFiles -- currentFiles
+    val delFileNames = delFiles.map(_.name)
+
+    // Remove duplicate deleted file names in the previous log entry.
+    val prevDeletedFiles =
+      previousIndexLogEntry.deletedFiles.filterNot(f => delFileNames.contains(f.name))
 
     // TODO: Add test for the scenario where existing appendedFiles and newly appended
     //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
-    ((originalFiles -- currentFiles) ++ previousIndexLogEntry.deletedFiles).toSeq
+    delFiles.toSeq ++ prevDeletedFiles
   }
 
   /**
@@ -138,9 +144,15 @@ private[actions] abstract class RefreshActionBase(
   protected lazy val appendedFiles: Seq[FileInfo] = {
     val relation = previousIndexLogEntry.relations.head
     val originalFiles = relation.data.properties.content.fileInfos
+    val newFiles = (currentFiles -- originalFiles)
+    val newFileNames = newFiles.map(_.name)
+
+    // Remove duplicate appended file names in the previous log entry.
+    val prevAppendedFiles =
+      previousIndexLogEntry.appendedFiles.filterNot(f => newFileNames.contains(f.name))
 
     // TODO: Add test for the scenario where existing appendedFiles and newly appended
     //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
-    (currentFiles -- (originalFiles -- previousIndexLogEntry.appendedFiles)).toSeq
+    newFiles.toSeq ++ prevAppendedFiles
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -99,17 +99,10 @@ private[actions] abstract class RefreshActionBase(
    * Finally, append the previously known deleted files to the result. These
    * are the files for which the index was never updated in the past.
    */
-  protected lazy val deletedFiles: Seq[String] = {
+  protected lazy val deletedFiles: Seq[FileInfo] = {
     val relation = previousIndexLogEntry.relations.head
     val originalFiles = relation.data.properties.content.fileInfos
-    val delFiles = (originalFiles -- currentFiles).map(_.name)
-
-    // Remove duplicate deleted file names in the previous log entry.
-    val prevDeletedFiles = previousIndexLogEntry.deletedFiles.filterNot(delFiles.contains)
-
-    // TODO: Add test for the scenario where existing deletedFiles and newly deleted
-    //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
-    delFiles.toSeq ++ prevDeletedFiles
+    (originalFiles -- currentFiles).toSeq
   }
 
   /**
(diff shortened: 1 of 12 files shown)


src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -126,8 +126,15 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
         |                  "properties" : { }
         |                }
         |              },
-        |              "deletedFiles" : ["file:/rootpath/f1"],
-        |              "appendedFiles" : ["file:/rootpath/f3"]
+        |              "deletedFiles" : [{
+        |                    "name" : "file:/rootpath/f1",
```
imback82, line 130: Now that I look that this, should we store these as `Directory` to be consistent? The thing that bugs me is that `FileInfo`'s behavior is inconsistent - one is storing full path, and another just name. Also, this can save the metadata file size. Thoughts?

(I know we are using it interchangeably in the code - e.g., `Content.fileInfos`, which can be addressed separately).

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -99,17 +99,10 @@ private[actions] abstract class RefreshActionBase(
...
+  protected lazy val deletedFiles: Seq[FileInfo] = {
     val relation = previousIndexLogEntry.relations.head
     val originalFiles = relation.data.properties.content.fileInfos
-    val delFiles = (originalFiles -- currentFiles).map(_.name)
-
-    // Remove duplicate deleted file names in the previous log entry.
-    val prevDeletedFiles = previousIndexLogEntry.deletedFiles.filterNot(delFiles.contains)
-
-    // TODO: Add test for the scenario where existing deletedFiles and newly deleted
-    //  files are updated. https://github.com/microsoft/hyperspace/issues/195.
```
imback82, line 111: Is this TODO handled in this PR?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -389,15 +389,16 @@ case class IndexLogEntry(
...
   }
 
-  def appendedFiles: Seq[String] = {
+  def appendedFiles: Seq[FileInfo] = {
     relations.head.data.properties.appendedFiles
   }
 
-  def withAppendedAndDeletedFiles(appended: Seq[String], deleted: Seq[String]): IndexLogEntry = {
+  def withAppendedAndDeletedFiles(appended: Seq[FileInfo], deleted: Seq[FileInfo]): IndexLogEntry
+  = {
```
imback82, line 401: is this by scalafmt? Looks a bit weird. Could you double check?

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -99,17 +99,10 @@ private[actions] abstract class RefreshActionBase(
    * Finally, append the previously known deleted files to the result. These
    * are the files for which the index was never updated in the past.
    */
-  protected lazy val deletedFiles: Seq[String] = {
+  protected lazy val deletedFiles: Seq[FileInfo] = {
     val relation = previousIndexLogEntry.relations.head
     val originalFiles = relation.data.properties.content.fileInfos
-    val delFiles = (originalFiles -- currentFiles).map(_.name)
```
pirz, line 105: Question: Why is the part for deduplication removed from deletedFiles and appendedFiles functions? (They have been added recently thru #214).

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
sezruby: I also thought about using `Content` as utility functions are coupled with it, not with `Directory`. (and inconsistency of FileInfo also bothers me). However, the format below didn't seem good to me, a little. Do you think this is more preferable?

```
case class Hdfs(properties: Hdfs.Properties) {
  val kind = "HDFS"
}
object Hdfs {
  /**
   * Hdfs file properties.
   * @param content Content object representing Hdfs file based data source.
   * @param appendedFiles Appended files since the last time derived dataset was updated.
   * @param deletedFiles Deleted files since the last time derived dataset was updated.
   */
  case class Properties(
      content: Content,
      appendedFiles: Content,
      deletedFiles: Content)
}
```

or with #198
```
case class Update {
  appendedFiles: Content,
  deletedFiles: Content,
  latestSignature: Signature
}
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -334,8 +336,8 @@ object Hdfs {
    */
   case class Properties(
       content: Content,
-      appendedFiles: Seq[String] = Nil,
-      deletedFiles: Seq[String] = Nil)
+      appendedFiles: Content = Content(Directory.emptyDirectory),
+      deletedFiles: Content = Content(Directory.emptyDirectory))
```
imback82, line 340: Thanks. I like this approach better. Could you check if we could make it as `Option[Content]`? (If it doesn't work out nicely, `Directory.empty` seems reasonable. (I got `empty` from `def empty[A] : Option[A] = None` for `Option`)

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
sezruby: Yes, seems we still need to remove duplicates. To my mind, it's a bit of mess to handle duplicates here but #198 will resolve the problem; with "since" semantic, we don't need to care about appendedFiles and deletedFiles of previousIndexLogEntry.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: `Option[Content]` works well. Thanks @imback82!

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -406,7 +426,8 @@ case class IndexLogEntry(
               relations.head.copy(
                 data = relations.head.data.copy(
                   properties = relations.head.data.properties.copy(
-                    appendedFiles = appended, deletedFiles = deleted))))))))
+                    appendedFiles = Content.fromLeafFiles(appendedFiles),
+                    deletedFiles = Content.fromLeafFiles(deletedFiles)))))))))
   }
```
imback82, line 432: How about:
```scala
def toFileStatus(f: FileInfo) = {
  new FileStatus(f.size, false, 0, 1, f.modifiedTime, new Path(f.name))
}
copy(
  source = source.copy(
    plan = source.plan.copy(
      properties = source.plan.properties.copy(
        relations = Seq(
          relations.head.copy(
            data = relations.head.data.copy(
              properties = relations.head.data.properties.copy(
                appendedFiles = Content.fromLeafFiles(appended.map(toFileStatus)),
                deletedFiles = Content.fromLeafFiles(deleted.map(toFileStatus))))))))))
```

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -286,11 +286,13 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
       "'appendedFiles' are not usable indexes if hybrid scan is disabled.") {
     withSQLConf(INDEX_HYBRID_SCAN_ENABLED -> "false") {
       val entry1 = createIndexLogEntry("t1iTest", Seq(t1c1), Seq(t1c3), t1ProjectNode)
-      val entry2 = entry1.withAppendedAndDeletedFiles(Seq(), Seq("f1"))
-      val entry3 = entry1.withAppendedAndDeletedFiles(Seq("f2"), Seq())
+      val entry2 = entry1.withAppendedAndDeletedFiles(Seq(), Seq(FileInfo("file:/dir/f1", 1, 1)))
+      val entry3 = entry1.withAppendedAndDeletedFiles(Seq(FileInfo("file:/dir/f2", 1, 1)), Seq())
+      entry2.state = "ACTIVE"
+      entry3.state = "ACTIVE"
```
imback82, line 292: Is this change required? (It helps reviewers if you make guide comments on changes that are not PR-related).

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -536,10 +537,11 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
     TestUtils.deleteFiles(dataPath, "*parquet", 1).head
   }
 
-  private def listFiles(path: String): Seq[String] = {
+  private def listFiles(path: String): Seq[FileInfo] = {
     val absolutePath = PathUtils.makeAbsolute(path)
     val fs = absolutePath.getFileSystem(new Configuration)
-    fs.listStatus(absolutePath).toSeq.map(_.getPath.toString)
+    fs.listStatus(absolutePath).toSeq.filter(f => DataPathFilter.accept(f.getPath)).map(f =>
```
imback82, line 543: Why is `DataPathFilter` change introduced?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -384,20 +390,34 @@ case class IndexLogEntry(
...
+    if (relations.head.data.properties.deletedFiles.isDefined) {
+      relations.head.data.properties.deletedFiles.get.fileInfos
+    } else {
+      Set()
+    }
   }
 
-  def appendedFiles: Seq[String] = {
-    relations.head.data.properties.appendedFiles
+  def appendedFiles: Set[FileInfo] = {
```
imback82, line 404: Can you add the following comments to `def`s that return `FileInfo` with full path so we can revisit this later?
```
// FileInfo's 'name' contains the full path to the file.
```

imback82 (commented):
Generally looking fine to me.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -91,7 +91,7 @@ object RuleUtils {
     }
 
     if (hybridScanEnabled) {
-      // TODO: Duplicate listing files for the given relation as in
+      // TODO: Duplicate listing files for the given 361relation as in
```
imback82, line 94: typo

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
imback82: Oh I see. This is very subtle and I will create an issue to see if we need to follow up (potentially any `copy()` on `IndexLogEntry` could be misleading).

Meanwhile could you add the following?
```
// IndexLogEntry.withAppendedAndDeletedFiles doesn't copy LogEntry's fields. Thus, set the 'state' to ACTIVE manually so that these entries are considered in RuleUtils.getCandidateIndexes.
```

imback82 (commented):
LGTM, thanks @sezruby!

imback82 (approved):

pirz (approved):
LGTM, thanks @sezruby

PR #232, opened 2020-10-28
Merge RefreshAppendAction and RefreshDeleteAction

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

This PR merges `RefreshAppendAction` and `RefreshDeleteAction` as `RefreshIncremental`, which is required for #198.

Without this, it is hard to calculate index signature value considering deletedFiles and appendedFiles - 
as spark.read API doesn't allow to create a DF with non-existing file paths.

No

Unit test


src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.spark.sql.{SaveMode, SparkSession}
+import org.apache.spark.sql.functions.col
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
+import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEvent, RefreshIncrementalActionEvent}
+
+/**
+ * Action to create indexes on newly arrived data. If the user appends new data to existing,
```
sezruby, line 28: To reviewers:
1) This file is copied from RefreshAppendAction.scala and I haven't revised overall comments yet. I'll update the comments later.
2) I haven't removed RefreshAppendAction.scala and RefreshDeleteAction.scala yet, for reference.
I'll remove them and events later.

rapoth:
> Without this, it is hard to calculate index signature value considering deletedFiles and appendedFiles -
> as spark.read API doesn't allow to create a DF with non-existing file paths.

Could you please add an example that clarifies why it is difficult?

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
imback82: This approach looks good. Let's move forward with this!

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,124 @@
...
+   * indexed in this operation). It also contains ALL index files (index files for previously
+   * indexed data as well as newly created files).
+   *
+   * @return Merged index log entry.
+   */
+  override def logEntry: LogEntry = {
+    // Log entry with complete data and newly index files.
+    val entry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
+
+    if (deletedFiles.isEmpty) {
```
pirz, line 111: It would be helpful if you add a comment, explaining in more details, why having deletedFiles affects the way new entry is created.

src/main/scala/com/microsoft/hyperspace/index/DataFrameWriterExtensions.scala
```diff
@@ -50,7 +50,8 @@ object DataFrameWriterExtensions {
         df: DataFrame,
         path: String,
         numBuckets: Int,
-        bucketByColNames: Seq[String]): Unit = {
+        bucketByColNames: Seq[String],
+        mode: SaveMode = SaveMode.Overwrite): Unit = {
```
pirz, line 54: If this function is being changed to support other modes, lets add a quick test case for `SaveMode.Append` under `DataFrameWriterExtensionsTests`.

pirz (commented):
Thanks @sezruby - The overall approach looks fine to me. I left a few comments, mostly for question/clarification.

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,124 @@
...
+    spark: SparkSession,
+    logManager: IndexLogManager,
+    dataManager: IndexDataManager)
+    extends RefreshActionBase(spark, logManager, dataManager) {
+  final override def op(): Unit = {
+    logInfo(
+      "Refresh index is updating index by removing index entries " +
+        s"corresponding to ${deletedFiles.length} deleted source data files.")
+
+    if (appendedFiles.nonEmpty) {
```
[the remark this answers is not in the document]
sezruby: Good point. Previously, with only filenames instead of fileInfos, RefreshDeleteAction is required to be done firstly. 
Because if we do RefreshAppendAction first, we cannot distinguish which rows should be removed while RefreshDeleteAction. 

Now we don't need to care about the order because we know the list of index data files not including appended data.
I changed the order just because AppendRefresh uses `write` function which uses `overwrite` mode to write the data.
(I tried to keep previous impl. mostly)

Either way, mode: SaveMode is required because there are 2 data writings for "delete" and "append" respectively.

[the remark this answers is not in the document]
sezruby: I added some code to use "overwrite" mode in case there is no appended files, to make sure everything is clear before the write.
Thanks :)

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -0,0 +1,140 @@
...
+
+    if (deletedFiles.nonEmpty) {
+      // For an index with lineage, find all the source data files which have been deleted,
+      // and use index records' lineage to mark and remove index entries which belong to
+      // deleted source data files as those entries are no longer valid.
+      val refreshDF =
+        spark.read
+          .parquet(previousIndexLogEntry.content.files.map(_.toString): _*)
+          .filter(
+            !col(s"${IndexConstants.DATA_FILE_NAME_COLUMN}").isin(deletedFiles.map(_.name): _*))
```
imback82, line 75:
```suggestion
            !col(IndexConstants.DATA_FILE_NAME_COLUMN).isin(deletedFiles.map(_.name): _*))
```

src/main/scala/com/microsoft/hyperspace/index/DataFrameWriterExtensions.scala
imback82, line 54: Let's not use default value for the mode and be explicit.

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -106,25 +106,14 @@ case class CancelActionEvent(appInfo: AppInfo, index: IndexLogEntry, message: St
...
- * @param index Related index.
- * @param message Message about event.
- */
-case class RefreshDeleteActionEvent(appInfo: AppInfo, index: IndexLogEntry, message: String)
-    extends HyperspaceIndexCRUDEvent
-
-/**
- * Index Refresh Event for appended source files. Emitted when refresh is called on an index
+ * Index Refresh Event for incremental mode. Emitted when refresh is called on an index
  * with config flag set to create index for appended source data files.
```
imback82, line 110: This needs to be updated.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -591,6 +591,7 @@ class E2EHyperspaceRulesTests extends QueryTest with HyperspaceSuite {
   private def queryPlanHasExpectedRootPaths(
       optimizedPlan: LogicalPlan,
       expectedPaths: Seq[Path]): Boolean = {
+    assert(getAllRootPaths(optimizedPlan) === expectedPaths)
```
imback82, line 594: Why do we need this check when the caller is doing assert? For printing diff using `===`?

imback82 (commented):
LGTM, thanks @sezruby!

imback82:
@sezruby I pushed https://github.com/microsoft/hyperspace/pull/232/commits/4f29cda9bc382462cf94cc27920f570496845fb4 to unblock #234. Feel free to create a separate PR if you don't agree with the changes I pushed. Thanks!

imback82 (approved):

PR #235, opened 2020-10-29
Add Update structure for appended files and deleted files

### LINKED ISSUE #169: Add index metadata update in RefreshIndex for quick mode

sezruby:

 - **Tracking Issue**: #198 
 - **Parent Issue**: n/a.
 - **Dependencies**: #232

This change is also required for #169, refresh quick mode.
Fixes #169

For #198, we need to keep the latest signature value leveraging appended and deleted files. 
This PR defines `Update` class for newly updated data.
```
case class Update(
  latestSignatures: Seq[Signature],
  appendedFiles: Option[Content] = None,
  deletedFiles: Option[Content] = None)
```

Yes, this is a breakage change, metadata change, so need to rebuild their index.


src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -39,14 +39,6 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
       .getOrElse(dataManager.getPath(0))
   }
 
-  protected def numBucketsForIndex(spark: SparkSession): Int = {
```
sezruby, line 42: rearranged by reformatting.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -40,28 +40,9 @@ private[actions] abstract class RefreshActionBase(
     dataManager: IndexDataManager)
     extends CreateActionBase(dataManager)
     with Action {
-  private lazy val previousLogEntry: LogEntry = {
```
sezruby, line 43: rearranged by reformatting

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -323,9 +323,22 @@ case class LogicalPlanFingerprint(properties: LogicalPlanFingerprint.Properties)
   val kind = "LogicalPlan"
 }
 object LogicalPlanFingerprint {
-  case class Properties(signatures: Seq[Signature])
+  case class Properties(signatures: Seq[Signature], planSignature: String)
```
sezruby, line 326: @imback82 @pirz @apoorvedave1 
I wonder this is good approach or no.  We need to keep `planSignature` separately for signature calculation at refreshing  - though only logical relation plan is allowed for now.

sezruby:
@imback82 @pirz @apoorvedave1 This PR is based on #232, so please check the last commit for less amount of diff.
https://github.com/microsoft/hyperspace/pull/235/commits/c7236605e7516e903de917abcdeb864792586470

This PR should be included in this release as this is the metadata change. But please note that there is no use case of `Update` structure for now - this is for refresh quick mode.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: Hmm, this seems a bit weird. :)  Can you point me where this is being used? I couldn't find.

One way to get around (depending on the use case), we can get rid of `IndexSignatureProvider` and just use two separate signatures: `FileBasedSignatureProvider` and `PlanSignatureProvider`. And we can have a utils class that combines the signature values on the rule path.

sezruby: @imback82  Outdated but for use case: https://github.com/sezruby/hyperspace/blob/metaupdate/src/main/scala/com/microsoft/hyperspace/actions/RefreshAppendAction.scala#L100
https://github.com/sezruby/hyperspace/blob/metaupdate/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala#L70

`signatures` defined as `Seq[Signature]` for multiple relations. So do you suggest `Seq[Seq[Signature]]`?

This is for quick refresh, so no use case for now. Can I create `RefreshQuickAction.scala` for this w/o exposing API?

imback82: > `signatures` defined as `Seq[Signature]` for multiple relations. So do you suggest `Seq[Seq[Signature]]`?

Does it make sense to move `fingerprint: LogicalPlanFingerprint` from `SparkPlan` to `Relation` so that the fingerprint always goes with the corresponding relation. (And we can break up `IndexSignatureProvider`). WDYT?

imback82: @apoorvedave1 / @pirz Any suggestions?

sezruby: As a source plan can contain multiple relations, it might be better to keep plan signature in SparkPlan.
But I think it's good to calculate "FileBasedSignatureProvider" for each relation respectively, in terms of design flexibility.

imback82: > As a source plan can contain multiple relations, it might be better to keep plan signature in SparkPlan.

Can't we have additional signature at `SparkPlan` in that case?

Can you explain the use case at a high level (instead of code), so we can better understand what we need to achieve?

sezruby: Refresh Quick mode
- collect appendedFiles, deletedFiles, latestSignature 
- latestSignature can be calculated by planSignature value + updated file based signature value
- current behavior at refresh (full and incremental):
  - use reconstructed `df` to get planSignature

Seems we could use the reconstructed `df` in the same way of the other refresh modes.

sezruby: I updated this PR w/o planSignature change and added simple RefreshQuickAction impl for use case.

sezruby: @imback82 @priz @apoorvedave1 Could you give any quick comment on this? 

BTW how about adding `stats: Map[String, String]` in CoveringIndex.properties that @imback82 mentioned in the previous discussion for `hasLineage` properties? it's just to avoid breaking changes in the next release expecting more statistics / usability changes.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -326,6 +326,19 @@ object LogicalPlanFingerprint {
   case class Properties(signatures: Seq[Signature])
 }
 
+/**
+ * Updated source data since lst time derived dataset was updated. Note that index data does not
```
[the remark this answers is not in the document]
imback82: Actually, `Update` is used under `Hdfs` which is a generic representation of file system. We can make a generic comment here and move this comment to `Relation` to explain `data: Hdfs`.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -335,13 +348,9 @@ object Hdfs {
   /**
    * Hdfs file properties.
    * @param content Content object representing Hdfs file based data source.
-   * @param appendedFiles Appended files since the last time derived dataset was updated.
-   * @param deletedFiles Deleted files since the last time derived dataset was updated.
+   * @param update Source data update since the last time derived dataset was updated.
```
imback82, line 351: Can we remove the mention of  "source' and "derived" data set?

```
@param update captures any updates since 'content' was created.
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -326,6 +326,19 @@ object LogicalPlanFingerprint {
...
+/**
+ * Updated source data since lst time derived dataset was updated. Note that index data does not
+ * cover appendedFiles and deleted Files.
+ *
+ * @param appendedFiles Appended files.
+ * @param deletedFiles Deleted files.
+ * @param latestSignatures Latest signatures including appendedFiles and deletedFiles.
+ */
+case class Update(
+  latestSignatures: Seq[Signature],
```
imback82, line 338: I thought about this field, and can we make this as `fingerprint: LogicalPlanFingerprint`? This goes against my previous comment about making this `Update` generic. Let me explain.

Whenever you see "kind" and "properties", that structure is supposed to be a "plug-in" point. For example, the "Source" has "SparkPlan", but it can be replaced with different sources "SqlDW", etc. It's easier to see if you look at the JSON representation. We never got a chance to work on the parser to support these scenarios and resorted to the `case class` to make reading/writing JSON easy for the time being.

Back to `Update`, I think the spec owner can decide what to do with the fingerprint for `Update`. In our case, it will be a `LogicalPlanFingerprint`. This also aligns with `Content` having a `fingerprint`. Does this make sense? We can update the comment as follows:
```scala
/**
 * Captures any HDFS updates.
 *
 * @note 'fingerprint' shouldn't be tied to [[LogicalPlanFingerprint]], but it's a free-form
 *       class, meaning it has "kind" and "properties" so that different classes can be plugged in.
 *       Thus, 'fingerprint' is still a generic field in terms of the metadata log specification.
 *
 * @param fingerprint Fingerprint of the update.
 * @param appendedFiles Appended files.
 * @param deletedFiles Deleted files.
 */
case class Update(
  fingerprint: LogicalPlanFingerprint,
  appendedFiles: Option[Content] = None,
  deletedFiles: Option[Content] = None)
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: Thanks @sezruby for `RefreshQuickAction`. The change looks good to me. I left a comment whether we can use `LogicalPlanFingerprint`. If it works out, could you separate out `RefreshQuickAction` to a separate PR to better track?

> BTW how about adding `stats: Map[String, String]` in CoveringIndex.properties that @imback82 mentioned in the previous discussion for `hasLineage` config? it's just to avoid breaking changes in the next release expecting more statistics / usability changes.

There may be an option to update `ObjectMapper` to allow adding new fields while maintaining compatibility? I am also fine with adding one now. Would this be a "stat" or more like "properties"?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -394,17 +403,22 @@ case class IndexLogEntry(
...
   // FileInfo's 'name' contains the full path to the file.
   def appendedFiles: Set[FileInfo] = {
-    relations.head.data.properties.appendedFiles.map(_.fileInfos).getOrElse(Set())
+    sourceUpdate.flatMap(_.appendedFiles).map(_.fileInfos).getOrElse(Set())
+  }
+
+  // FileInfo's 'name' contains the full path to the file.
+  def deletedFiles: Set[FileInfo] = {
+    sourceUpdate.flatMap(_.deletedFiles).map(_.fileInfos).getOrElse(Set())
   }
```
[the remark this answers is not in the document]
sezruby: It's required until
- RefreshQuickAction & refresh mode api
- getCandidateIndex fix
  - compare with the latest signature if exists
  - if so, tag hybrid scan required = true 
- transformPlanToUseIndex & hybrid scan fix
  - use appendedFiles and deletedFiles if exist

For this, we need #223 for tagging.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: That makes sense. Thanks for the suggestion :)

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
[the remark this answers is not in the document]
sezruby: Updated the comment. Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: > There may be an option to update ObjectMapper to allow adding new fields while maintaining compatibility? I am also fine with adding one now. 

Seems no problem to add new fields later. Let's handle it later and wrap up this release 👍

> Would this be a "stat" or more like "properties"?

I suggested `stats` as it will be added under `CoveringIndex.Properties`. How about `extra` or `aux` ..?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -335,16 +352,20 @@ object Hdfs {
...
-      deletedFiles: Option[Content] = None)
+  case class Properties(content: Content, update: Option[Update] = None)
 }
 
-// IndexLogEntry-specific Relation that represents the source relation.
+/**
+ * IndexLogEntry-specific Relation that represents the source relation.
+ *
+ * @param rootPaths List of root paths for the source relation.
+ * @param data Source data since last time derived dataset was updated.
```
imback82, line 364: This seems to be mentioning the `Update` field?
How about
```
@param data Source data for the relation.
            Hdfs.properties.content captures source data which derived dataset was created from.
            Hdfs.properties.update captures any updates since the derived dataset was created. 
            Note that Hdfs.properties.update.fingerprint is a fingerprint created from the latest source data.
```

imback82: @sezruby Writing this comment, `Update.fingerprint` still feels out of place. :(

What if we update `SparkPlan.Properties.fingerprint` to always point to the latest fingerprint (instead of having fingerprint inside `Update`)?

imback82 (approved):
LGTM, thanks @sezruby! Could you update the PR description?

PR #238, opened 2020-11-02
Support refresh quick mode by using Hybrid Scan

### LINKED ISSUE #169: Add index metadata update in RefreshIndex for quick mode

### LINKED ISSUE #174: Modify Hyperspace rules to leverage an index with deleted source data files after RefreshIndex in quick mode
### Describe the issue
Currently an index with non-empty `deletedFiles` or `appendedFiles` is ignored during optimization.
We need to extend optimizer rules to consider and leverage such indexes. For deletedFiles, the rule needs to inject a filter on top of index scan node to enforce delete by excluding index entries coming from deleted source data files.

**User Scenario.**
2. User deletes some files from the original data under "/path/to/dataset/".
3. User calls RefreshIndex with `mode = "quick"` to update index metadata. 
4. User runs a query on latest source data files (step 2). If index is applicable, Hyperspace should leverage the index and enforce delete by removing index entries for deleted source data files during query time.
 
- Pre-requisite: Issue #169 (Add index metadata update in RefreshIndex for quick mode).
- Parent: Issue #134 (Add support for deletes in RefreshIndex for quick mode).

### Expected behavior
A refreshed index with some deleted source data files should be leveraged as expected on user queries running on latest source data files.

### Environment 
All.

sezruby:

 - **Tracking Issue**: #174
 - **Parent Issue**: n/a
 - **Dependencies**: #235 

Fixes #174 
Fixes #169 

Support quick refresh of indexes which only updates the metadata - `Update` data - of the index.

```
hs.refreshIndex("indexName", "quick")
```

After the refresh, the latest signature will be used for index selection process (getCandidateIndexes) and the appended files and deleted files of `Update` will be handled by Hybrid Scan.

Yes

Unit test


src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -254,10 +246,11 @@ object RuleUtils {
             baseOutput,
             _,
             _) =>
-        val curFiles = location.allFiles
-          .map(f => FileInfo(f.getPath.toString, f.getLen, f.getModificationTime))
-
-        val (filesDeleted, filesAppended) =
+        val (filesDeleted, filesAppended) = if (index.hasSourceUpdate) {
```
imback82, line 249: Is it guaranteed that `location.allFiles` will always be captured in `index.deletedFiles` and `index.appendedFiles`? If so, can you add a comment why? (I didn't see a check on the signature if hybrid scan is enabled).

sezruby: Good point - we cannot guarantee that if hybrid scan is enabled. We need to check the hybrid scan config here:
```if (!hybridScanEnabled && index.hasSourceUpdate)```

File: src/main/scala/com/microsoft/hyperspace/actions/RefreshQuickAction.scala [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Validate index is in active state for refreshing and there are some changes
+   * in source data file(s).
+   */
+  final override def validate(): Unit = {
+    super.validate()
+
+    if (appendedFiles.isEmpty && deletedFiles.isEmpty) {
+      throw NoChangesException("Refresh incremental aborted as no source data change found.")
```
pirz, line 51: The message should be fixed. It is not refresh incremental.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -588,7 +693,7 @@ class E2EHyperspaceRulesTests extends QueryTest with HyperspaceSuite {
   private def verifyQueryPlanHasExpectedRootPaths(
       optimizedPlan: LogicalPlan,
       expectedPaths: Seq[Path]): Unit = {
-    assert(getAllRootPaths(optimizedPlan) === expectedPaths)
+    assert(getAllRootPaths(optimizedPlan).sortBy(_.getName) === expectedPaths.sortBy(_.getName))
```
pirz, line 696: can we use toSet() here?

sezruby: There can be multiple same rootPaths from join query tests.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -179,10 +170,15 @@ object RuleUtils {
...
     // This tag should always exist if Hybrid Scan is enabled.
-    lazy val hybridScanRequired =
+    val hybridScanRequired = HyperspaceConf.hybridScanEnabled(spark) &&
       index.getTagValue(getLogicalRelation(plan).get, IndexLogEntryTags.HYBRIDSCAN_REQUIRED).get
 
-    val transformed = if (HyperspaceConf.hybridScanEnabled(spark) && hybridScanRequired) {
+    // In case the index has appended files and/or deleted files which are updated by quick
+    // refresh mode, we should handle them in query time by using Hybrid Scan.
+    // Because quick refresh doesn't update index data, but just metadata in IndexLogEntry.
+    lazy val quickRefreshedIndex = index.hasSourceUpdate
```
pirz, line 179: If we are using it only once, do we need to create a separate (lazy) val for it?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -279,23 +275,30 @@ object RuleUtils {
...
-          .map(f => FileInfo(f, index.fileIdTracker.addFile(f)))
-
         val (filesDeleted, filesAppended) =
-          if (HyperspaceConf.hybridScanDeleteEnabled(spark) && index.hasLineageColumn) {
-            val (exist, nonExist) = curFiles.partition(index.sourceFileInfoSet.contains)
-            val filesAppended = nonExist.map(f => new Path(f.name))
-            if (exist.length < index.sourceFileInfoSet.size) {
-              (index.sourceFileInfoSet -- exist, filesAppended)
+          if (!HyperspaceConf.hybridScanEnabled(spark) && index.hasSourceUpdate) {
+            // This index is refreshed as quick mode and validated with the latest signature.
```
pirz, line 280: nit: "... was refreshed in quick mode..."

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexTests.scala
```diff
@@ -392,6 +392,67 @@ class RefreshIndexTests extends QueryTest with HyperspaceSuite {
     }
   }
 
+  test(
+    "Validate RefreshQuickAction updates appended and deleted files in metadata " +
+      "expected, when some file gets deleted and some appended to source data.") {
```
pirz, line 397: nit: "expected" -> "as expected"

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: It's mainly for readability and comments. + It's only required when HybridScan is enabled, so it's lazy val.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Thanks!

sezruby:
@imback82 @pirz @apoorvedave1 Could you review this change? Thanks!

pirz (dismissed):
LGTM, Thanks @sezruby

apoorvedave1:
LGTM, thanks @sezruby

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82, line 179: To me, RuleUtils doesn't need to know how the index metadata was generated (i.e., it just knows how to understand it). So I think we can just name it as `isSourceUpdated`? And you can update the comment that if `isSourceUpdated` is set to true, index data is not up to date with the source data.

imback82: I would add a comment if you added `lazy` to avoid creating sets for appended files, etc.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -279,23 +275,30 @@ object RuleUtils {
...
-          .map(f => FileInfo(f, index.fileIdTracker.addFile(f), asFullPath = true))
-
         val (filesDeleted, filesAppended) =
-          if (HyperspaceConf.hybridScanDeleteEnabled(spark) && index.hasLineageColumn) {
-            val (exist, nonExist) = curFiles.partition(index.sourceFileInfoSet.contains)
-            val filesAppended = nonExist.map(f => new Path(f.name))
-            if (exist.length < index.sourceFileInfoSet.size) {
-              (index.sourceFileInfoSet -- exist, filesAppended)
+          if (!HyperspaceConf.hybridScanEnabled(spark) && index.hasSourceUpdate) {
+            // This index was refreshed in quick mode and validated with the latest signature.
```
imback82, line 280: Same here. I don't think we need to mention "quick" mode. You can say that the index data is not up to date with source data if index.hasSourceUpdate is true.

imback82 (commented):
LGTM (few minor points), thanks @sezruby!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Good point, updated the comments :)

src/main/scala/com/microsoft/hyperspace/actions/RefreshQuickAction.scala
```diff
@@ -0,0 +1,80 @@
...
+ */
+class RefreshQuickAction(
+    spark: SparkSession,
+    logManager: IndexLogManager,
+    dataManager: IndexDataManager)
+    extends RefreshActionBase(spark, logManager, dataManager) {
+  final override def op(): Unit = {
+    logInfo(
+      s"Refresh index is updating metadata only with ${deletedFiles.size} of deleted" +
+        s" files and ${appendedFiles.size} of appended files.")
```
imback82, line 40:
```suggestion
      s"Refresh index is updating metadata only with ${deletedFiles.size} of deleted " +
        s"files and ${appendedFiles.size} of appended files.")
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -279,23 +275,30 @@ object RuleUtils {
...
-
         val (filesDeleted, filesAppended) =
-          if (HyperspaceConf.hybridScanDeleteEnabled(spark) && index.hasLineageColumn) {
-            val (exist, nonExist) = curFiles.partition(index.sourceFileInfoSet.contains)
-            val filesAppended = nonExist.map(f => new Path(f.name))
-            if (exist.length < index.sourceFileInfoSet.size) {
-              (index.sourceFileInfoSet -- exist, filesAppended)
+          if (!HyperspaceConf.hybridScanEnabled(spark) && index.hasSourceUpdate) {
+            // If the index contains the source update info, we need to handle
+            // appendedFiles and deletedFiles in IndexLogEntry.
```
imback82, line 281: Can you add back the comment that signature validation was done by this time?

imback82 (approved):

imback82 (approved):

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: I merged this PR. Can you address this either in #164 or #265?

PR #241, opened 2020-11-02
Update configurations for Python tests for better resource utilization

imback82:

This PR borrows configs from delta-io used for Python tests:
https://github.com/delta-io/delta/blob/819bcd5635dd47775d1c6adb51b3f40674d0ea81/python/delta/testing/utils.py. New configs allow better utilization of resources.

`"spark.hyperspace.index.num.buckets"` should be tweaked for Scala tests, but currently few tests depend on 200 buckets, so need to be revistied.

Update the configs so that resource is better utilized.

No

Existing tests


imback82 (commented):
Merging as this doesn't affect the critical path.

PR #242, opened 2020-11-02
Clean up Hyperspace-related configs

imback82:

Few clean ups related to configs:
  * Renamed `spark.hyperspace.index.num.buckets` to `spark.hyperspace.index.numBuckets`, but supports the legacy config.
  * Removed unused configs: `spark.hyperspace.index.creation.path` and `spark.hyperspace.index.search.paths`.

This PR proposes to rename `spark.hyperspace.index.num.buckets` to `spark.hyperspace.index.numBuckets` and clean up unused configs.

No, since legacy config is still honored.

Added new tests testing falling back to legacy configs.


src/main/scala/com/microsoft/hyperspace/index/PathResolver.scala
```diff
@@ -68,27 +68,6 @@ private[hyperspace] class PathResolver(conf: SQLConf) {
...
-  }
-
-  /**
-   * Get the Hyperspace index search paths.
-   *
-   * @return Hyperspace index search paths.
-   */
-  def indexSearchPaths: Option[Seq[String]] = {
-    getOption(IndexConstants.INDEX_SEARCH_PATHS)
-      .map(_.split(',').toSeq)
```
imback82, line 89: These are not currently not used. I will create a separate issue to implement this scenario.

imback82: Created #243.

sezruby:
LGTM thanks!

PR #244, opened 2020-11-05
Rename IndexLogEntry.extra to IndexLogEntry.properties

imback82:

`IndexLogEnry` has an `extra` field to capture any additional info for the "top-level" log entry. However, other entities have been using `properties` to capture these info. I think it's better to keep it consistent with `properties` instead of `extra`.

This PR proposes to use `properties` instead of `extra` as the name in one of the `IndexLogEntry`'s fields.

This is a breaking change for the `IndexLogEntry`.

Updated the existing test.


sezruby (approved):
LGTM thanks!

PR #224, opened 2020-10-21
Support Hybrid Scan with Delta Lake relation

sezruby:
### What is the context for this pull request?

 - **Tracking Issue**: n/a
 - **Parent Issue**: #148
 - **Dependencies**: #197, #227

**Please just check the last commit to check the diff as the base is not yet merged -  #197**

### What changes were proposed in this pull request?

This PR enables to do Hybrid Scan with Delta Lake relation. 
A new util function `def retrieveCurFileInfos(plan: LogicalPlan)` is defined and it returns the current file list in a given logical plan regardless of the type of the relation.

We can optimize `getCandidateIndex` for Hybrid Scan by using the version info and path of Delta Lake later.

### Does this PR introduce _any_ user-facing change?

Yes, a user can utilize Hybrid Scan with Delta Lake relation.

### How was this patch tested?

Unit test

PR #265, opened 2020-11-23
Support Delta Lake file-based source provider

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: #148 
 - **Dependencies**: n/a

Refactor #197 + #224 (except for hybrid scan test refactoring) based on #227.

This PR introduces DeltaLakeFileBasedSource to support indexes on Delta Lake Sources.

Yes,

Users can create/refresh indexes on Delta Lake table and also utilize Hybrid Scan.

Unit test


sezruby:
@imback82 Need to add some comments but could you review this PR?
I'll handle Hybrid Scan test refactoring in #224 with another PR.

imback82:
> @imback82 Need to add some comments but could you review this PR?

Cool! I will take a look. 

> I'll handle Hybrid Scan test refactoring in #224 with another PR.

Yes, I believe Hybrid Scan requires new APIs, so let's handle it separately. Thanks!

imback82:
`DeltaLakeFileBasedSource` looks good to me. If you are OK with overall approach with #227, I can merge it and resolve comments (unit tests) as a follow up PRs.

Btw, can you check the build failure?

sezruby:
> Yes, I believe Hybrid Scan requires new APIs, so let's handle it separately. Thanks!
@imback82 I added `allFiles` API for Hybrid Scan and seems that's all for Hybrid Scan.

BTW #227 also has the same test failure and it's not reproducible on my local. Let me check further.

imback82:
#227 has been merged. Thanks!

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -70,11 +70,19 @@ private[actions] abstract class RefreshActionBase(
     val latestRelation =
       Hyperspace.getContext(spark).sourceProviderManager.refreshRelation(relations.head)
     val dataSchema = DataType.fromJson(latestRelation.dataSchemaJson).asInstanceOf[StructType]
-    spark.read
-      .schema(dataSchema)
-      .format(latestRelation.fileFormat)
-      .options(latestRelation.options)
-      .load(latestRelation.rootPaths: _*)
+    if (latestRelation.rootPaths.size == 1) {
```
sezruby, line 73: Delta Lake only allows one path in load()

imback82: Yea, we can do something like:
```scala
val df = spark.read
  .schema(dataSchema)
  .format(latestRelation.fileFormat)
  .options(latestRelation.options)
// Due to the difference in how the "path" option is set: https://github.com/apache/spark/blob/ef1441b56c5cab02335d8d2e4ff95cf7e9c9b9ca/sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala#L197,
// load() with a single parameter needs to be handled differently.
if (latestRelation.rootPaths.size == 1) {
  df.load(latestRelation.rootPaths.head)
} else {
  df.load(latestRelation.rootPaths: _*)
}
```

imback82: Btw, what happens if the delta lake implementation adds the "path" option in `latestRelation.options`?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -174,9 +175,14 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
       //    + file:/C:/hyperspace/src/test/part-00003.snappy.parquet
       import spark.implicits._
       val dataPathColumn = "_data_path"
-      val lineageDF = fileIdTracker.getFileToIdMap.toSeq
+      val isDeltaLakeSource = DeltaLakeRuleUtils.isDeltaLakeSource(df.queryExecution.optimizedPlan)
```
imback82, line 178: Can we move this to source provider somehow? Basically, we don't want source specific implementation outside the provider.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -437,19 +443,24 @@ object RuleUtils {
...
+          // For example, we could have the following in PartitionSpec:
+          //   - partition columns = "col1", "col2"
+          //   - partitions: "/path/col1=1/col2=1", "/path/col1=1/col2=2", etc.
+          // , and going up the same number of directory levels as the number of partition columns
+          // will compute the base path. Note that PartitionSpec.partitions will always contain
+          // all the partitions in the path, so "partitions.head" is taken as an initial value.
+          val basePath = p.partitionSpec.partitionColumns
+            .foldLeft(p.partitionSpec.partitions.head.path)((path, _) => path.getParent)
+          Some(basePath.toString)
+        case d: TahoeLogFileIndex =>
```
imback82, line 461: same here. we need to push this to provider by introducing new API

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
imback82: question, why does delta not require the `replace`?

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
sezruby: https://github.com/microsoft/hyperspace/pull/265/files#diff-f7ecc68d1799e9c2c916973786081d5f35f312785537ccd2eed61bce41a10786R72
"path" key will be removed in `createRelation`.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
sezruby: input_file_neme() function of Delta returns "file:/" not "file:///"

imback82: I see. How about we normalize the output of `input_file_name` by removing "file:/" or "file:///"? (and same to the `getFileToIdMap` keys)

sezruby: replacing filename - id map is cheaper than normalizing all file paths of all rows. I'll add an API for this.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -24,10 +24,12 @@ import org.apache.spark.sql.catalyst.catalog.BucketSpec
 import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, In, Literal, Not}
 import org.apache.spark.sql.catalyst.optimizer.OptimizeIn
 import org.apache.spark.sql.catalyst.plans.logical._
+import org.apache.spark.sql.delta.files.TahoeLogFileIndex
```
sezruby, line 27: --

imback82 (commented):
Generally approach/integration looks good! I will do a detailed review this week.

@pirz @apoorvedave1 Could you take a look as well?

apoorvedave1:
LGTM, thanks @sezruby

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -0,0 +1,166 @@
...
+            .mode("append")
+            .save(dataPath)
+
+          // The index should be applied for the updated version.
+          assert(verifyIndexUsage(query().queryExecution.optimizedPlan, "deltaIndex", true))
+        }
+      }
+    }
+  }
+
```
sezruby, line 147: Note to reviewers: more hybrid scan + delta lake test cases - https://github.com/microsoft/hyperspace/pull/274

pirz (commented):
LGTM, Thanks @sezruby

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -0,0 +1,166 @@
...
+  def verifyIndexUsage(
```
imback82, line 148: nit: `isIndexApplied` or `isIndexUsed`?

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
imback82: Is the result comparison tested in #274?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -261,7 +262,8 @@ object RuleUtils {
       index: IndexLogEntry,
       plan: LogicalPlan,
       useBucketSpec: Boolean): LogicalPlan = {
-    val isParquetSourceFormat = index.relations.head.fileFormat.equals("parquet")
+    val fileFormat = index.relations.head.fileFormat
+    val isParquetSourceFormat = fileFormat.equals("parquet") || fileFormat.equals("delta")
```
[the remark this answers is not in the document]
imback82: You can do this as a separate PR if that is preferred. Please create an issue in that case.

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
sezruby: yes `checkAnswer` is used to compare the results in #274

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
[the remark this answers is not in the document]
sezruby: I think this case is too specific to create an API to source provider; and it refers fileformat string in index metadata, not relation.
So it's better to create the function in `IndexLogEntry` or some Utils class if needed. WDYT? @imback82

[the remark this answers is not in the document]
imback82: You should be able to plug in a source provider defined externally without changing Hyperspace codebase. For example, let's say I have a source format "blah" that uses parquet internally, and how can I plug in without modifying Hyperspace?



> I think this case is too specific to create an API to source provider; and it refers fileformat string in index metadata, not relation.

You can do this in the create path and record it in the metadata.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
```diff
@@ -0,0 +1,187 @@
...
+      logicalRelation: LogicalRelation,
+      fileIdTracker: FileIdTracker): Option[Relation] = {
+    logicalRelation.relation match {
+      case HadoopFsRelation(location: TahoeLogFileIndex, _, dataSchema, _, _, options) =>
+        val files = location
+          .getSnapshot(stalenessAcceptable = false)
+          .filesForScan(projection = Nil, location.partitionFilters, keepStats = false)
+          .files
+          .map { f =>
+            new FileStatus(
```
imback82, line 64: Can this also use `toFileStatus`?

imback82 (commented):
LGTM except for one pending comment about introducing `hasParquetAsSourceFormat`.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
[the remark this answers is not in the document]
sezruby: Got your point - it should be able to add a new source provider externally.
Let me handle this with a new pr & issue. Thanks!

imback82 (approved):

Issue #226, opened 2020-10-23
[FEATURE REQUEST]: Support mutable source data when it's loaded with globbing pattern


imback82:
**Feature requested**
There are many user scenarios where the data is stored as non-hive partitioning:
```
/source/1/1/data.json
/source/1/2/data.json
/source/2/1/data.json
```
, and the user can load the data as `spark.read.json("/source/*/*")`.

If the data is loaded with globbing pattern, you will see the `rootPaths` as follows:
```scala
val df = spark.read.json("/source/*/*")
val locs = df.queryExecution.analyzed.collect {
  case LogicalRelation(HadoopFsRelation(location: InMemoryFileIndex, _, _, _, _, _), _, _, _) => location }
locs.head.rootPaths

// Output:
// List(file:/source/1/1, file:/source/1/2, file:/source/2/1)
```
This could be problematic if new folders (e.g, `/source/2/2`) are added and if Hyperspace wants to keep track of newly appended files.

On a side note, if hive-partitioning is used, `rootPaths` will be as follows:
```scala
// Data:
// /source/month=1/date=1/data.json
// /source/month=1/date=2/data.json
// /source/month=2/date=1/data.json

val df = spark.read.json("/source")
val locs = df.queryExecution.analyzed.collect {
  case LogicalRelation(HadoopFsRelation(location: InMemoryFileIndex, _, _, _, _, _), _, _, _) => location }
locs.head.rootPaths

// Output:
// List(file:/source)
```

**Acceptance criteria** 
Hyperspace can handle mutable data that's loaded with globbing pattern.

**Success criteria**
One way to achieve this is to inspect `path` option to see if globbing pattern is used and handle it accordingly. (There could be an inconsistent behavior how the `path` option is set (e.g., `parquet(/tmp/*)` vs. `format("parquet").load("/tmp/*)` in Spark 2.4, so we may need to double-check on this.)

Or, we could introduce a config so that the user can do `spark.read.option("spark.hyperspace.source.globbingPattern", "/tmp/*").parquet("/tmp/*")` for the user to opt-in explicitly.

apoorvedave1:
> One way to achieve this is to inspect `path` option to see if globbing pattern is used and handle it accordingly. (There could be an inconsistent behavior how the `path` option is set (e.g., `parquet(/tmp/*)` vs. `format("parquet").load("/tmp/*)` in Spark 2.4, so we may need to double-check on this.)
> 
[...]

PR #269, opened 2020-11-23
Support globbing patterns in dataframes for creation/maintenance/usage of indexes.

apoorvedave1:

 - **Tracking Issue**: #226

Support for globbing patterns in data sources when creating/maintaining/using indexes.

Users will now be able to create and maintain indexes on dataframes created with globbing patterns e.g. 

```scala
  val path = "tmp/1/*"
  val path2 = "tmp/2/*"
  val df = spark.read
    .option("spark.hyperspace.source.globbingPattern", s"$path,$path2") // This is only required the first time index is being created.
    .parquet(path, path2)

  hs.createIndex(df, ...)
```
> Note: for multiple paths, as shown in the example above, please concatenate the paths with commas `,`.

unit tests

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -127,7 +127,8 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
         // "path" key in options can incur multiple data read unexpectedly.
         val opts = options - "path"
 
-        // If user provided a globbing pattern, set rootPaths to this pattern. Else, use file index.
+        // If user provided a globbing pattern, set rootPaths to globbing pattern.
+        // Else, use file index to identify rootPaths.
         val rootPaths = opts.get(GLOBBING_PATTERN_KEY) match {
           case Some(pattern) =>
             // Validate if globbing pattern matches actual source paths.
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -16,7 +16,9 @@
 
 package com.microsoft.hyperspace.actions
 
+import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
+import org.apache.spark.deploy.SparkHadoopUtil
 import org.apache.spark.sql.{DataFrame, SaveMode, SparkSession}
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
 import org.apache.spark.sql.functions.input_file_name
@@ -25,6 +27,7 @@ import org.apache.spark.sql.sources.DataSourceRegister
 import com.microsoft.hyperspace.HyperspaceException
 import com.microsoft.hyperspace.index._
 import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
+import com.microsoft.hyperspace.index.IndexConstants.GLOBBING_PATTERN_KEY
 import com.microsoft.hyperspace.util.{HyperspaceConf, PathUtils, ResolverUtils}
 
 /**
@@ -123,12 +126,33 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
         }
         // "path" key in options can incur multiple data read unexpectedly.
         val opts = options - "path"
-        Relation(
-          location.rootPaths.map(_.toString),
-          Hdfs(sourceDataProperties),
-          dataSchema.json,
-          fileFormatName,
-          opts)
+
+        // If user provided a globbing pattern, set rootPaths to this pattern. Else, use file index.
+        val rootPaths = opts.get(GLOBBING_PATTERN_KEY) match {
+          case Some(pattern) =>
+            // Validate if globbing pattern matches actual source paths.
+            // This logic is picked from the globbing logic at:
+            // https://github.com/apache/spark/blob/v2.4.4/sql/core/src/main/scala/org/apache/
+            // spark/sql/execution/datasources/DataSource.scala#L540
+            val globPaths = pattern.split(",").map(_.trim).map { path =>
+              val hdfsPath = new Path(path)
+              val fs = hdfsPath.getFileSystem(new Configuration)
+              val qualified = hdfsPath.makeQualified(fs.getUri, fs.getWorkingDirectory)
+              qualified.toString -> SparkHadoopUtil.get.globPathIfNecessary(fs, qualified)
+            }.toMap
+
+            val globPathValues = globPaths.values.flatten.toSet
+            if (!location.rootPaths.forall(globPathValues.contains)) {
+              throw HyperspaceException("Some glob patterns do not match with available root " +
+                s"paths of the source data. Please check if option $GLOBBING_PATTERN_KEY matches " +
+                "with the provided paths.")
+            }
+            globPaths.keySet.toSeq
+
+          case _ => location.rootPaths.map(_.toString)
+        }
+
+        Relation(rootPaths, Hdfs(sourceDataProperties), dataSchema.json, fileFormatName, opts)
     }
 
   protected def write(spark: SparkSession, df: DataFrame, indexConfig: IndexConfig): Unit = {
--- a/src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
@@ -59,17 +59,17 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
       fileIdTracker: FileIdTracker): Option[Relation] = {
     logicalRelation.relation match {
       case HadoopFsRelation(
-      location: PartitioningAwareFileIndex,
-      _,
-      dataSchema,
-      _,
-      fileFormat,
-      options) if isSupportedFileFormat(fileFormat) =>
+          location: PartitioningAwareFileIndex,
+          _,
+          dataSchema,
+          _,
+          fileFormat,
+          options) if isSupportedFileFormat(fileFormat) =>
         val files = location.allFiles
         // Note that source files are currently fingerprinted when the optimized plan is
         // fingerprinted by LogicalPlanFingerprint.
         val sourceDataProperties =
-        Hdfs.Properties(Content.fromLeafFiles(files, fileIdTracker).get)
+          Hdfs.Properties(Content.fromLeafFiles(files, fileIdTracker).get)
         val fileFormatName = fileFormat.asInstanceOf[DataSourceRegister].shortName
 
         // "path" key in options can incur multiple data read unexpectedly.
@@ -81,18 +81,23 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
             // This logic is picked from the globbing logic at:
             // https://github.com/apache/spark/blob/v2.4.4/sql/core/src/main/scala/org/apache/
             // spark/sql/execution/datasources/DataSource.scala#L540
-            val globPaths = pattern.split(",").map(_.trim).map { path =>
-              val hdfsPath = new Path(path)
-              val fs = hdfsPath.getFileSystem(new Configuration)
-              val qualified = hdfsPath.makeQualified(fs.getUri, fs.getWorkingDirectory)
-              qualified.toString -> SparkHadoopUtil.get.globPathIfNecessary(fs, qualified)
-            }.toMap
+            val fs = location.allFiles.head.getPath.getFileSystem(new Configuration)
+            val globPaths = pattern
+              .split(",")
+              .map(_.trim)
+              .map { path =>
+                val hdfsPath = new Path(path)
+                val qualified = hdfsPath.makeQualified(fs.getUri, fs.getWorkingDirectory)
+                qualified.toString -> SparkHadoopUtil.get.globPathIfNecessary(fs, qualified)
+              }
+              .toMap
 
             val globPathValues = globPaths.values.flatten.toSet
             if (!location.rootPaths.forall(globPathValues.contains)) {
-              throw HyperspaceException("Some glob patterns do not match with available root " +
-                s"paths of the source data. Please check if option $GLOBBING_PATTERN_KEY matches " +
-                "with the provided paths.")
+              throw HyperspaceException(
+                "Some glob patterns do not match with available root " +
+                  s"paths of the source data. Please check if $pattern matches all of " +
+                  s"${location.rootPaths.mkString(",")}.")
             }
             globPaths.keySet.toSeq
 
--- a/src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
@@ -25,7 +25,7 @@ import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, InMemoryFil
 import org.apache.spark.sql.execution.exchange.ShuffleExchangeExec
 
 import com.microsoft.hyperspace.{Hyperspace, Implicits, SampleData, TestUtils}
-import com.microsoft.hyperspace.index.IndexConstants.{REFRESH_MODE_INCREMENTAL, REFRESH_MODE_QUICK}
+import com.microsoft.hyperspace.index.IndexConstants.{GLOBBING_PATTERN_KEY, REFRESH_MODE_INCREMENTAL, REFRESH_MODE_QUICK}
 import com.microsoft.hyperspace.index.execution.BucketUnionStrategy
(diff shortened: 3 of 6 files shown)


src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -123,12 +126,33 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
         val opts = options - "path"
-        Relation(
-          location.rootPaths.map(_.toString),
-          Hdfs(sourceDataProperties),
-          dataSchema.json,
-          fileFormatName,
-          opts)
+
+        // If user provided a globbing pattern, set rootPaths to this pattern. Else, use file index.
+        val rootPaths = opts.get(GLOBBING_PATTERN_KEY) match {
```
sezruby, line 131: I think instead of keeping all paths in rootPaths at creation, 
- keep globPattern in option field in createIndex
- use the globPattern to get the current file list in refreshIndex

WDYT?

apoorvedave1: thanks @sezruby for the suggestion, I thought about that option too and it looked quite lucrative.

The only reason I chose this one was to simplify 2 things:
1. the rest of the index lifecycle / index maintenance gets handled automatically without any other change. This means that we only need to change one class.
2. simpler change in code.

Semantically also, it might be more meaningful to keep rootpaths containing "/temp/*/*" in source.relations. E.g. if the user looks at the metadata, they will immediately know what was the intention when creating index.

Please let me know WDYT. Adding @imback82 , @pirz , @rapoth for opinions/suggestions.

Actually now I also think that if we stick with the current implementation, we can completely remove the globbing config from relation.options. (i.e. one less thing to maintain)

sezruby: ok I realized that without this, Spark keeps all paths in `location.rootPaths`. 
Let's keep the glob paths in rootPaths to avoid duplicate paths in IndexLogEntry.
But with this approach, we still need to keep the globbing path config for refresh; rootPaths will be replaced after refresh, without the config.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -123,12 +126,33 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+            // https://github.com/apache/spark/blob/v2.4.4/sql/core/src/main/scala/org/apache/
+            // spark/sql/execution/datasources/DataSource.scala#L540
+            val globPaths = pattern.split(",").map(_.trim).map { path =>
+              val hdfsPath = new Path(path)
+              val fs = hdfsPath.getFileSystem(new Configuration)
+              val qualified = hdfsPath.makeQualified(fs.getUri, fs.getWorkingDirectory)
+              qualified.toString -> SparkHadoopUtil.get.globPathIfNecessary(fs, qualified)
+            }.toMap
+
+            val globPathValues = globPaths.values.flatten.toSet
```
sezruby, line 144: What if `spark.read.option(GLOBBING_PATTERN_KEY, "/*").parquet("/1/*")` ?
I think we could check it only at createIndex.. WDYT? @apoorvedave1

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -123,12 +126,33 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
         // "path" key in options can incur multiple data read unexpectedly.
         val opts = options - "path"
-        Relation(
-          location.rootPaths.map(_.toString),
-          Hdfs(sourceDataProperties),
-          dataSchema.json,
-          fileFormatName,
-          opts)
+
+        // If user provided a globbing pattern, set rootPaths to this pattern. Else, use file index.
```
sezruby, line 130: nit: Could you a bit revise this comment?

apoorvedave1, line 130:
```suggestion
        // If user provided a globbing pattern, set rootPaths to globbing pattern.
        // Else, use file index to identify rootPaths.
```

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
apoorvedave1: Could you please test this once? In my experience the original df creation fails for the above example

sezruby: Yep looks like the case is also checked. Thanks for confirming!

sezruby (commented):
Minor comments, but LGTM!

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -55,27 +59,48 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+              val fs = hdfsPath.getFileSystem(new Configuration)
+              val qualified = hdfsPath.makeQualified(fs.getUri, fs.getWorkingDirectory)
+              qualified.toString -> SparkHadoopUtil.get.globPathIfNecessary(fs, qualified)
+            }.toMap
+
+            val globPathValues = globPaths.values.flatten.toSet
+            if (!location.rootPaths.forall(globPathValues.contains)) {
+              throw HyperspaceException("Some glob patterns do not match with available root " +
+                s"paths of the source data. Please check if option $GLOBBING_PATTERN_KEY matches " +
+                "with the provided paths.")
```
imback82, line 95: Should we log globbing pattern and root paths for easier debugging?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -478,6 +478,128 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+      hyperspace.createIndex(df, indexConfig)
+
+      // Check if latest log file contains data source as globbing pattern.
+      var index = latestIndexLogEntry(systemPath, indexConfig.indexName)
+      assert(index.relations.head.rootPaths.equals(Seq(globPath1, globPath2)))
+      assert(index.relations.head.data.properties.content.files.forall(path =>
+        path.toString.startsWith(p1) || path.toString.startsWith(p2)))
+
+      // Append data to a new directory which matches the globbing pattern.
+      val p3 = absoluteTestPath + "/1/3"
```
imback82, line 582: Should we also add `/2/2` for completeness?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -478,6 +478,128 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+      assert(
+        index.relations.head.data.properties.content.files.forall(path =>
+          path.toString.startsWith(p1) || path.toString.startsWith(p2) || path.toString
+            .startsWith(p3)))
+      assert(index.relations.head.data.properties.content.files.exists(_.toString.startsWith(p1)))
+      assert(index.relations.head.data.properties.content.files.exists(_.toString.startsWith(p2)))
+      assert(index.relations.head.data.properties.content.files.exists(_.toString.startsWith(p3)))
+    }
+  }
+
```
imback82, line 602: Should we validate hybrid scan scenario or no?

imback82 (commented):
LGTM (except for pending comments), thanks @apoorvedave1!

Could you also do a follow up PR to update doc (config + behavior)?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
imback82: for example, globbing pattern was `1/*`. Initially, you had `1/1`, and `1/2` was added. And validate if data under `1/2` was considered.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -478,6 +478,128 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+      // Create index with globbing pattern.
+      val df =
+        spark.read.option("spark.hyperspace.source.globbingPattern", globPath).parquet(globPath)
+      val indexConfig = IndexConfig("index", Seq("RGUID"), Seq("imprs"))
+      hyperspace.createIndex(df, indexConfig)
+
+      // Check if latest log file contains data source as globbing pattern.
+      var index = latestIndexLogEntry(systemPath, indexConfig.indexName)
+      assert(index.relations.head.rootPaths.head.equals(globPath))
```
imback82, line 501: nit: shall we create `val relation = index.relations.head`?

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
sezruby: If the query plan contains `1/1` and `1/2` properly, hybrid scan will work as expected.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -478,6 +478,128 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
     }
   }
 
+  test("Verify index maintenance (create, refresh) works with globbing patterns.") {
```
sezruby, line 481: Can you add create & refresh index tests for multiple `*` patterns ? e.g. `table/*/m=1/*`

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 , @sezruby , from this discussion I will close this comment. Please reopen if additional test is required.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 ,added

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @imback82 ,added

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: thanks @sezruby , i added a test for multiple levels of `*`. 

For specialized case e.g. `/*/m=1/*`, i think it would be an overkill, because I mean if 
`spark.read.parquet("glob/*/m=1/*")` works then this will also work right, as long as we pass in the right value in `.parquet(<path>)`? WDYT?

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
apoorvedave1: thanks @imback82 ,added

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -55,27 +59,48 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+
+        val rootPaths = opts.get(GLOBBING_PATTERN_KEY) match {
+          case Some(pattern) =>
+            // Validate if globbing pattern matches actual source paths.
+            // This logic is picked from the globbing logic at:
+            // https://github.com/apache/spark/blob/v2.4.4/sql/core/src/main/scala/org/apache/
+            // spark/sql/execution/datasources/DataSource.scala#L540
+            val globPaths = pattern.split(",").map(_.trim).map { path =>
+              val hdfsPath = new Path(path)
+              val fs = hdfsPath.getFileSystem(new Configuration)
```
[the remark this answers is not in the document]
apoorvedave1: thanks @sezruby ,added

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
imback82: yes, let's add one.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
sezruby: You mean - is it required additional changes? or not necessary as it will work? 

And could you add some tests for the other glob patterns?  like 
```
* (match 0 or more character)
? (match single character)
[ab] (character class)
[^ab] (negated character class)
[a-b] (character range)
{a,b} (alternation)
\c (escape character)
```
I think we should check if the index with these patterns will be refreshed correctly.

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: added, thanks @imback82

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: > You mean - is it required additional changes? or not necessary as it will work?
> 
> And could you add some tests for the other glob patterns? like
> 
> ```
> * (match 0 or more character)
> ? (match single character)
> [ab] (character class)
> [^ab] (negated character class)
> [a-b] (character range)
> {a,b} (alternation)
> \c (escape character)
> ```
> 
> I think we should check if the index with these patterns will be refreshed correctly.

I think it's overkill to add test for every regex possible since we are not adding any regex/wildcard handling logic at all.
cc @imback82 , @pirz , @rapoth for opinions?

apoorvedave1: I mean if spark.read.parquet("any-wildcard-pattern") works and we make sure that we are passing the same argument always, then it will work always right?

sezruby: Yeah but we just have the tests using "*" - we can add a test using all of those patterns are used. 
I just worried about handling special characters in json, and not wanted to test regex behavior.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
```diff
@@ -684,6 +684,66 @@ class E2EHyperspaceRulesTests extends QueryTest with HyperspaceSuite {
...
+        // Check appended file is added to relation node or not.
+        val nodes = planWithHybridScan.collect {
+          case p @ LogicalRelation(fsRelation: HadoopFsRelation, _, _, _) =>
+            // Verify appended file is included or not.
+            assert(
+              fsRelation.location.inputFiles
+                .count(_.contains(absoluteTestPath.toUri.toString)) === 3)
+            // Verify number of index data files.
+            assert(fsRelation.location.inputFiles.count(_.contains("index")) === 4)
+            assert(fsRelation.location.inputFiles.length === 7)
```
imback82, line 737: How should I understand 3, 4, and 7? Don't we need to check if `p2` exists, etc?

apoorvedave1:
> LGTM (except for pending comments), thanks @apoorvedave1!
> 
> Could you also do a follow up PR to update doc (config + behavior)?

Thanks @imback82 , This issue for tracking documentation. I will start the pr soon
https://github.com/microsoft/hyperspace/issues/283

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
apoorvedave1: Thanks, I have added this https://github.com/microsoft/hyperspace/issues/284 issue for tracking json conversion tests

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTests.scala
apoorvedave1: thanks @imback82 , I cleaned up this test to check for p2, not-p1 and index files

imback82 (approved):

src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
imback82: Cool, I am merging this to master. Let's do a follow up PR for this discussion. Thanks!

imback82:
Can we also add the following case? Basically, we can inject this option after the table is created:
```
CREATE TABLE t(k int) USING json
ALTER TABLE t SET SERDEPROPERTIES ('spark.hyperspace.source.globbingPattern' = 'globbing pattern')
REFRESH TABLE t
```

imback82:
@apoorvedave1 Could you take a look at the follow ups asked?

Issue #248, opened 2020-11-11
Fix APIs for "REFRESHING" state of indexes though refresh was cancelled


sezruby:
### Describe the issue

If refresh index operation is canceled before it's finished, the index state remains "REFRESHING".
In that case, the index cannot be applied and also deleted or refreshed via Hyperspace APIs.

We could handle this case as follow:
- ignore incomplete state of the latest version
  - index candidate based on the latest "ACTIVE" version.
- delete/vacuum support so that a user can remove spoiled index. 
- refresh index based on the latest "ACTIVE" support? quick refresh can rewrite the invalid index log entry?

PR #272, opened 2020-11-25
Add Delta Lake version history to IndexLogEntry for efficient time travel query

### LINKED ISSUE #270: Improvement of Time Travel Query on Delta Lake
**Feature requested**

Currently, a time travel query on delta lake can be handled with Hybrid Scan.
However, in case the time travel version is too far from the latest version of index, there might be many appended and deleted files between the time travel version and the latest index. If so, we cannot utilize the index efficiently and Hybrid Scan overhead might outweigh the benefit of the index.

To mitigate this issue, we could pick an old version of index if it's available. 

For example, an index is built on a Delta Lake table, and it refreshed 5 times. 

Built: version 1
Refreshed: version  5, 7, 10, 15

And if a time travel query is requested with version 6, we can choose 5 or 7 depending on the similarity.

This can be done by following steps:

- Keep all the versions in `properties.properties` field / to avoid searching all index log entry files
  - key: DELTA_VERSION_HISTORY_PROPERTY 
  - value: comma separated "INDEX_VERSION_ID:DELTA_VERSION_ID"
    - e.g `1:1,3:5,5:7,7:10,9:15` for refreshed delta version `1, 5, 7, 10, 15`
- In getCandidateIndexes
  - pre-requisite: Hybird Scan enabled = true
  - getClosestIndexEntry
    - check the version list and return 2 available version (closest)
    - pick the one with more commonBytes
  - set IndexLogEntryTag
    - key: REPLACED_INDEX_LOG_ENTRY
    - value: older version of index log entry
- In transformPlanToUseIndex
  - use REPLACED_INDEX_LOG_ENTRY instead of the latest IndexLogEntry

**Acceptance criteria** 

**Success criteria**

**Additional context**

sezruby:

 - **Tracking Issue**: #270 
 - **Parent Issue**: #148 
 - **Dependencies**: #265  (merged)

Fixes #270 
`

This PR improves Hybrid Scan for a time travel query of Delta Lake table by using an old version of index which is closest to the given time travel delta version.

This PR includes:
- keep Delta Lake table versions at each create/refresh indexes in IndexLogEntry metadata.
  - in CoveringIndex.properties.properties, (to avoid searching all index log entry files)
    - **key: DELTA_VERSION_HISTORY_PROPERTY**
    - **value: comma separated "INDEX_VERSION_ID:DELTA_VERSION_ID"**
      - **e.g 1:1,3:5,5:7,7:10,9:15 for refreshed delta version 1, 5, 7, 10, 15**
- find and replace an old version of index in getCandidateIndexes
  - new `closestIndexVersion` API in source provider

Yes, an old version of index can be applied for a time travel query on Delta Lake.

Unit tests


src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -202,9 +193,14 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
       //    + file:/C:/hyperspace/src/test/part-00003.snappy.parquet
       import spark.implicits._
       val dataPathColumn = "_data_path"
-      val lineageDF = fileIdTracker.getFileToIdMap.toSeq
+      val isDeltaLakeSource = DeltaLakeRuleUtils.isDeltaLakeSource(df.queryExecution.optimizedPlan)
+        val lineageDF = fileIdTracker.getFileToIdMap.toSeq
         .map { kv =>
-          (kv._1._1.replace("file:/", "file:///"), kv._2)
+          if (isDeltaLakeSource) {
```
sezruby, line 199: Note: different input_file_name() value of Delta Lake df

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -35,7 +35,8 @@ class CreateAction(
     dataManager: IndexDataManager)
     extends CreateActionBase(dataManager)
     with Action {
-  final override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
+  final override def logEntry: LogEntry =
+    getIndexLogEntry(spark, df, indexConfig, indexDataPath, super[Action].endId)
```
sezruby, line 39: Note: need to get the index version id, but it's not available in CreateActionBase.getIndexLogEnrtry.
So I added `endId` function and `with Action`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -70,14 +69,27 @@ object RuleUtils {
...
+
+    def isHybridScanCandidate(index: IndexLogEntry, inputSourceFiles: Seq[FileInfo]): Boolean = {
       // TODO: Some threshold about the similarity of source data files - number of common files or
       //  total size of common files.
       //  See https://github.com/microsoft/hyperspace/issues/159
       // TODO: As in [[PlanSignatureProvider]], Source plan signature comparison is required to
       //  support arbitrary source plans at index creation.
       //  See https://github.com/microsoft/hyperspace/issues/158
 
+      val entry = if (isDeltaLakeSource) {
```
sezruby, line 85: @imback82 WDYT about this function? Should we also define new API inside source provider for this?
#270 ref for the design.

imback82: Yes, we should define new API inside source provider. I think the current approach always requires the Delta jar available.

sezruby:
@pirz  @apoorvedave1 Could you review the change and give any comment on the approach - keeping version history in properties -? Thanks!

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
```diff
@@ -213,6 +217,127 @@ class DeltaLakeFileBasedSource(private val spark: SparkSession) extends FileBase
...
+        if (equalOrLessLastIndex == versions.size - 1) {
+          // Use the latest version.
+          Some(index)
+        } else if (equalOrLessLastIndex == -1) {
+          Some(getIndexLogEntry(versions.head._1))
+        } else if (versions(equalOrLessLastIndex)._2 == location.tableVersion) {
+          Some(getIndexLogEntry(versions(equalOrLessLastIndex)._1))
+        } else {
+          val prevPair = versions(equalOrLessLastIndex)
+          val nextPair = versions(equalOrLessLastIndex + 1)
```
thugsatbay, line 280: nit: bestIndPair, nextBestIndPair

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -310,6 +321,264 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
...
+            .limit(3)
+            .write
+            .format("delta")
+            .mode("append")
+            .save(dataPath)
+
+          // Get the timestamp string for time travel query using "timestampAsOf" option.
+          val timestampStr = getSparkFormattedTimestamps(System.currentTimeMillis).head
+          // Sleep 1 second because the unit of commit timestamp is second.
+          Thread.sleep(1000)
```
thugsatbay, line 392: Avoid using sleep in ut as it will take too much time especially when this method is called multiple times.

To simulate time just take a global Time variable `currTime =  System.currentTimeMillis` and then keep on adding 1000 to it whenever you want to move time and then just use with `getSparkFormattedTimestamps` function

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
```diff
@@ -33,6 +33,7 @@ import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEvent, HyperspaceE
  */
 trait Action extends HyperspaceEventLogging with Logging with ActiveSparkSession {
   protected val baseId: Int = logManager.getLatestId().getOrElse(-1)
+  protected def endId: Int = baseId + 2
```
thugsatbay, line 36: Sorry, why +2, can you please explain ?

sezruby: it's because we write the log twice at begin() using `baseId + 1` and end()  using `baseId + 2`. 
As begin() has in-progress state (e.g. creating), Hyperspace always reads the log file written in "end()"

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
sezruby: I added this because delta lake internally records the time in **second** unit. So I forces to record different times for each version update, so that I could test the query with timestamp properly.

imback82:
@sezruby Could you fix the conflicts?

> Unit tests (test will be added)

This can be removed now?

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -130,6 +131,40 @@ class FileBasedSourceProviderManager(spark: SparkSession) {
...
+  def closestIndexVersion(
+      spark: SparkSession,
+      logicalRelation: LogicalRelation,
+      curFiles: Seq[FileInfo],
+      index: IndexLogEntry): IndexLogEntry = {
+    run(p => p.closestIndexVersion(spark, logicalRelation, curFiles, index))
+  }
+
+  /**
+   * Returns index version related properties.
```
jaceklaskowski, line 154: nit: That looks like something for `@return` below?

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -291,6 +291,53 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
     }
   }
 
+  /**
+   * Returns IndexLogEntry of the closest index version for the given relation.
```
jaceklaskowski, line 295: nit: s/Returns/@return

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
```diff
@@ -216,6 +220,127 @@ class DeltaLakeFileBasedSource(private val spark: SparkSession) extends FileBase
...
     }
   }
+
+  private def deltaLakeVersionHistory(index: IndexLogEntry): Seq[(Int, Int)] = {
+    val versions =
+      index.derivedDataset.properties.properties.getOrElse(DELTA_VERSION_HISTORY_PROPERTY, "")
+
+    versions.split(",").reverseIterator.foldLeft(Seq[(Int, Int)]()) { (list, versionStr) =>
+      val pair = versionStr.split(":")
+      val versions = (pair(0).toInt, pair(1).toInt) // (index log version, delta table version)
```
jaceklaskowski, line 230: This comment could be part of the code using pattern matching on assignment, e.g.

```
val Array(indexLogVersion, deltaTableVersion) = versionStr.split(":").map(_.toInt)
```

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
```diff
@@ -216,6 +220,127 @@ class DeltaLakeFileBasedSource(private val spark: SparkSession) extends FileBase
...
+      if (list.nonEmpty && list.head._2 == versions._2) {
+        // Omit if the delta lake version is the same. In this case, we only takes the higher
```
jaceklaskowski, line 232: nit: take

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
thugsatbay: I don't think so, I ran your test and used ` DeltaTable.forPath(spark, dataPath).history().show(10, false)` to view timestamp (2021-02-02 14:49:19.077). It used milliseconds and each op should have a readVersion ?

thugsatbay: You can use the timestamp in delta table to store read versions ?

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
thugsatbay: nit: Can we rename it to `firstLogEntryId` and `lastLogEntryId`. And add a comment for the same. 
Also should we have an assert that `firstLogEntryId` would always be -1 or an odd number ?

imback82:
@sezruby Could you resolve the conflicts? Let's try to get this in next (sorry for the delay).

@apoorvedave1 Can you also help with the review when you get a chance? Thanks.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -75,7 +80,12 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
             LogicalPlanFingerprint.Properties(Seq(Signature(signatureProvider.name, s)))))
 
         val coveringIndexProperties =
-          (hasLineageProperty(spark) ++ hasParquetAsSourceFormatProperty(spark, df)).toMap
+          (hasLineageProperty(spark) ++
+            hasParquetAsSourceFormatProperty(spark, df) ++
+            Hyperspace
+              .getContext(spark)
+              .sourceProviderManager
+              .versionHistoryProperties(versionId, relations.head, prevIndexProperties())).toMap
```
imback82, line 88: Instead of having this API, how about exposing something generic like `def enrichRelationProperties(relation: Relation, prop: Map[String, String]): Map[String, String]`? ' You will pack the `versionId` as an entry in the map; the default implementation will just return the given `prop`.

sezruby: in FileBasedRelation? but in `optimizeAction`, we don't construct source df.. 
In sourceProvider, yes I can make it more generic.

imback82: yes, this will be in source provider.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -73,7 +78,14 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
         val coveringIndexProperties =
-          (hasLineageProperty(spark) ++ hasParquetAsSourceFormatProperty(relation)).toMap
+          (hasLineageProperty(spark) ++
+            hasParquetAsSourceFormatProperty(relation)).toMap ++
+            Hyperspace
+              .getContext(spark)
+              .sourceProviderManager
+              .enrichIndexProperties(
+                sourcePlanProperties.relations.head,
+                prevIndexProperties() + (IndexConstants.INDEX_LOG_VERSION -> versionId.toString))
```
imback82, line 88: Can we pack lineage / parquet property here as well? The intention is to pack all the properties first then pass them to `enrichIndexProperties` to "enrich" them as needed by the provider.

src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
```diff
@@ -34,9 +34,10 @@ class RefreshAction(
     spark: SparkSession,
     logManager: IndexLogManager,
     dataManager: IndexDataManager)
-    extends RefreshActionBase(spark, logManager, dataManager) {
+    extends RefreshActionBase(spark, logManager, dataManager) with Action {
 
-  override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
+  override def logEntry: LogEntry =
+    getIndexLogEntry(spark, df, indexConfig, indexDataPath, super[Action].endId)
```
imback82, line 40: do you need `super[Action]`?

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
sezruby: baseId is not firstLogEntryId and it requires some change..
If an operation failed in op(), then firstLogEntryId could be an odd number. (though failures are not handled properly now. e.g. #248)

imback82: cc @apoorvedave1

sezruby: this is required because `LogEntry`.id is not available when creating - `IndexLogEntry`

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
sezruby: Seems Azure test framework needs sleep for some reason :D.. https://github.com/microsoft/hyperspace/pull/272/commits/f19614b5b32a284137d1823d3a73f06f024a9067
I might have added this after debugging / investigation.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,86 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+   * curFiles is used to calculate the similarity with each index version data.
+   * Delta Lake source provider utilizes the history information of DELTA_VERSION_HISTORY_PROPERTY
+   * to find the closet version of the index.
+   *
+   * @param curFiles List of FileInfo for the source files in the relation.
+   * @param index Candidate index to be applied.
+   * @return IndexLogEntry of the closest version among available index versions.
+   */
+  override def closestIndexVersion(
+      curFiles: Seq[FileInfo],
```
imback82, line 177: shouldn't this be available in the current relation?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,86 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
       /* modificationTime */ modificationTime,
       path)
   }
+
+  private def deltaLakeVersionHistory(index: IndexLogEntry): Seq[(Int, Int)] = {
+    val versions =
+      index.derivedDataset.properties.properties
+        .getOrElse(DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY, "")
+
+    versions.split(",").reverseIterator.foldLeft(Seq[(Int, Int)]()) { (list, versionStr) =>
```
imback82, line 152: Sample example in the comment would help.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,86 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+      case HadoopFsRelation(location: TahoeLogFileIndex, _, _, _, _, _) =>
+        val equalOrLessLastIndex = versions.lastIndexWhere(location.tableVersion >= _._2)
+        if (equalOrLessLastIndex == versions.size - 1) {
+          // Use the latest version.
+          index
+        } else if (equalOrLessLastIndex == -1) {
+          getIndexLogEntry(versions.head._1)
+        } else if (versions(equalOrLessLastIndex)._2 == location.tableVersion) {
+          getIndexLogEntry(versions(equalOrLessLastIndex)._1)
+        } else {
```
imback82, line 201: Could you add some comments for each block so that one can understand the logic without reading the code? 😄

src/main/scala/com/microsoft/hyperspace/actions/Action.scala
imback82: Can we add `final` modifier?

src/main/scala/com/microsoft/hyperspace/index/IndexManager.scala
```diff
@@ -104,4 +104,6 @@ trait IndexManager {
    * @return Index metadata and statistics as a [[DataFrame]].
    */
   def index(indexName: String): DataFrame
+
+  private[hyperspace] def getLogManager(indexName: String): Option[IndexLogManager]
```
imback82, line 108: I am not sure if this is a good idea to expose the `IndexLogManager` at this point (I understand this is package private, but since this can be used from source relation APIs, we should treat this as a public API).

One way is to make `IndexManager` to be more of a facade to the `IndexLogManager` and provide `def getIndexVersions(name: String): Option[Seq[Int]]` and `def getIndex(name: String, version: Int): Option[IndexLogEntry]`. `CachingIndexCollectionManager` can be used to cache `IndexLogEntry` this way if needed.

Any thoughts?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
sezruby: Yes we could use `allFiles`, but it will incur listing files for every api call. Can we change it as `lazy val`?

imback82: yes you can do that + you can also cache `Seq[FileInfo]` inside this object if needed since the same `relation` object will be used.

src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
```diff
@@ -61,9 +61,19 @@ trait FileBasedRelation extends SourceRelation {
...
-   * All the files that the current relation references to.
+   * FileStatus list for all source files that the current relation references to.
    */
-  def allFiles: Seq[FileStatus]
+  val allFiles: Seq[FileStatus]
+
+  /**
+   * FileInfo list for all source files that the current relation references to.
+   */
+  val allFileInfos: Seq[FileInfo]
```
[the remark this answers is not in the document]
sezruby: Thanks!

src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
```diff
@@ -34,9 +34,10 @@ class RefreshAction(
     spark: SparkSession,
     logManager: IndexLogManager,
     dataManager: IndexDataManager)
-    extends RefreshActionBase(spark, logManager, dataManager) {
+    extends RefreshActionBase(spark, logManager, dataManager) with Action {
```
imback82, line 37: nit: can you run scalafmt? in my case, `with Action` goes to a new line.

src/main/scala/com/microsoft/hyperspace/actions/RefreshAction.scala
```diff
@@ -34,9 +34,10 @@ class RefreshAction(
...
 
-  override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
+  override def logEntry: LogEntry =
+    getIndexLogEntry(spark, df, indexConfig, indexDataPath, endId)
```
imback82, line 40: super nit: line break is not needed? (diff would be better)

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -48,7 +48,7 @@ class RefreshIncrementalAction(
     spark: SparkSession,
     logManager: IndexLogManager,
     dataManager: IndexDataManager)
-    extends RefreshActionBase(spark, logManager, dataManager) {
+    extends RefreshActionBase(spark, logManager, dataManager) with Action {
```
imback82, line 51: ditto

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -102,6 +102,8 @@ object IndexConstants {
   private[hyperspace] val LINEAGE_PROPERTY = "lineage"
   // Indicate whether the source file format is parquet.
   private[hyperspace] val HAS_PARQUET_AS_SOURCE_FORMAT_PROPERTY = "hasParquetAsSourceFormat"
+  // Indicate index log version.
+  private[hyperspace] val INDEX_LOG_VERSION = "indexLogVer"
```
imback82, line 106: nit: let's just spell it out: `indexLogVersion`

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -76,56 +73,60 @@ object RuleUtils {
...
-          // If there is no change in source dataset, the index will be applied by
-          // transformPlanToUseIndexOnlyScan.
-          entry.setTagValue(
-            relation.plan,
-            IndexLogEntryTags.HYBRIDSCAN_REQUIRED,
-            !(commonCnt == entry.sourceFileInfoSet.size && commonCnt == inputSourceFiles.size))
-        }
-        isCandidate
+      if (entry.withCachedTag(relation.plan, IndexLogEntryTags.IS_HYBRIDSCAN_CANDIDATE) {
+            // Find the number of common files between the source relation and index source files.
```
imback82, line 87: indentation looks off?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -76,56 +73,60 @@ object RuleUtils {
...
+                commonBytes)
+
+              // If there is no change in source dataset, the index will be applied by
+              // transformPlanToUseIndexOnlyScan.
+              entry.setTagValue(
+                relation.plan,
+                IndexLogEntryTags.HYBRIDSCAN_REQUIRED,
+                !(commonCnt == entry.sourceFileInfoSet.size
+                  && commonCnt == relation.allFileInfos.size))
+            }
```
imback82, line 124: no code change in this block? Guided comment would have been helpful...

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -152,20 +153,15 @@ object RuleUtils {
...
-          f.getLen,
-          f.getModificationTime,
-          IndexConstants.UNKNOWN_FILE_ID)
-      }
       prepareHybridScanCandidateSelection(spark, relation.plan, indexes)
-      val totalSizeInBytes = inputSourceFiles.map(_.size).sum
-      indexes.filter(index =>
-        index.created && isHybridScanCandidate(index, inputSourceFiles, totalSizeInBytes))
+
+      indexes.flatMap { index =>
```
imback82, line 158:
```scala
indexes.filter(_.created).flatMap(getHybridScanCandidate)
```
? iterating twice seems fine here, but no strong opinion.

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -90,6 +90,19 @@ class FileBasedSourceProviderManager(spark: SparkSession) {
     run(p => p.getRelation(plan))
   }
 
+  /**
+   * Returns enriched index properties.
+   *
+   * @param relation Relation to retrieve necessary information.
+   * @param previousProperties Index properties of previous index version.
```
imback82, line 97: `previousProperties` could be misleading if you are creating an index? just `properties` seems fine to me.

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -90,6 +90,19 @@ class FileBasedSourceProviderManager(spark: SparkSession) {
...
+   * @return New property entries for index creation or refreshment.
```
imback82, line 98: nit: refreshment => refresh

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedRelation.scala
```diff
@@ -235,4 +235,5 @@ class DefaultFileBasedRelation(spark: SparkSession, override val plan: LogicalRe
       (kv._1._1.replace("file:/", "file:///"), kv._2)
     }
   }
+
```
imback82, line 238: nit: unnecessary change?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
```diff
@@ -21,17 +21,23 @@ import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
...
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation}
 
-import com.microsoft.hyperspace.index.Relation
+import com.microsoft.hyperspace.index.{IndexConstants, Relation}
 import com.microsoft.hyperspace.index.sources.{FileBasedRelation, FileBasedSourceProvider, SourceProvider, SourceProviderBuilder}
 
+object DeltaLakeConstants {
+  val DELTA_FORMAT_STR = "delta"
+
+  // JSON property name used in index metadata to store Delta Lake version history of refreshment.
```
imback82, line 30: nit: refreshment => refreshes?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
```diff
@@ -21,17 +21,23 @@ import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
...
+  val DELTA_VERSION_HISTORY_PROPERTY = "deltaVers"
```
imback82, line 31: `deltaVersions`?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
imback82: Can you go over other occurrences of refreshment?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
```diff
@@ -86,6 +92,34 @@ class DeltaLakeFileBasedSource(private val spark: SparkSession) extends FileBase
...
+    if (!relation.fileFormat.equals(DeltaLakeConstants.DELTA_FORMAT_STR)) {
+      None
+    } else {
+      val indexVersion = previousProperties(IndexConstants.INDEX_LOG_VERSION)
+      val deltaVerHistory = relation.options.get("versionAsOf").map { deltaVersion =>
+        DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY ->
+          previousProperties.get(DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY).map { str =>
+            s"$str,$indexVersion:$deltaVersion"
+          }.getOrElse(s"$indexVersion:$deltaVersion")
+      }
```
imback82, line 119:
```suggestion
      val deltaVerHistory = relation.options.get("versionAsOf").map { deltaVersion =>
        val newVersionMapping = s"$indexVersion:$deltaVersion"
        DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY ->
          previousProperties.get(DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY).map { prop =>
            s"$prop,$newVersionMapping"
          }.getOrElse(newVersionMapping)
      }
```

src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
```diff
@@ -61,9 +61,29 @@ trait FileBasedRelation extends SourceRelation {
...
+        f.getModificationTime,
+        IndexConstants.UNKNOWN_FILE_ID)
+    }
+  }
+
+  /**
+   * Summation of all source file size.
+   */
+  final lazy val allFileSizeInBytes: Long = {
+    allFileInfos.map(_.size).sum
```
imback82, line 85: nit: should this populate `allFileInfos` or should this call `allFiles`?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,95 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+
+  private def deltaLakeVersionHistory(index: IndexLogEntry): Seq[(Int, Int)] = {
+    val versions =
+      index.derivedDataset.properties.properties
+        .getOrElse(DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY, "")
+
+    // The value is comma separated versions - <index log version>:<delta table version>.
+    // e.g. "1:2,3:5,5:9"
+    versions.split(",").reverseIterator.foldLeft(Seq[(Int, Int)]()) { (list, versionStr) =>
+      val pair = versionStr.split(":")
```
imback82, line 155: remove?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,95 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+      val Array(indexLogVersion, deltaTableVersion) = versionStr.split(":").map(_.toInt)
+      if (list.nonEmpty && list.head._2 == deltaTableVersion) {
+        // Omit if the delta lake version is the same. In this case, we only take the higher
+        // index log version as this is usually from index optimizations.
```
imback82, line 159: "usually" or "always"? If "usually", what are the other scenarios?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,95 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+    val versions =
+      index.derivedDataset.properties.properties
+        .getOrElse(DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY, "")
+
+    // The value is comma separated versions - <index log version>:<delta table version>.
+    // e.g. "1:2,3:5,5:9"
+    versions.split(",").reverseIterator.foldLeft(Seq[(Int, Int)]()) { (list, versionStr) =>
+      val pair = versionStr.split(":")
+      val Array(indexLogVersion, deltaTableVersion) = versionStr.split(":").map(_.toInt)
+      if (list.nonEmpty && list.head._2 == deltaTableVersion) {
```
imback82, line 157: Can't you do this with `filter` instead of `foldLeft` unless I misunderstood something (in this case, can you add some comments?)

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -163,6 +162,14 @@ class IndexCollectionManager(
     }
   }
 
+  private[hyperspace] override def getIndexLogEntry(
```
imback82, line 165: If you check https://github.com/microsoft/hyperspace/pull/272#discussion_r579565340, I think it's fine to expose this as public (this is accessed from source relation, meaning it may be helpful to expose publicly), as long as we also expose APIs to get the log versions.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,95 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+          // Now, there are 2 candidate versions for the given table version:
+          // prevPair(indexVer1, tablePrevVersion), nextPair(indexVer2, tableNextVersion)
+          // which is (tablePrevVersion < TimeTravel table version < tableNextVersion).
+          val prevPair = versions(equalOrLessLastIndex)
+          val nextPair = versions(equalOrLessLastIndex + 1)
+
+          val prevLog = getIndexLogEntry(prevPair._1)
+          val nextLog = getIndexLogEntry(nextPair._1)
+
+          val curFilesSize = allFileInfos.map(_.size).sum
```
imback82, line 216: is this `allFileSizeInBytes`?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,95 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+
+          val prevCommonBytes =
+            allFileInfos.filter(prevLog.sourceFileInfoSet.contains).map(_.size).sum
+          val prevAppendedBytes = curFilesSize - prevCommonBytes
+          val prevDeletedBytes = prevLog.sourceFilesSizeInBytes - prevCommonBytes
+
+          val nextCommonBytes =
+            allFileInfos.filter(nextLog.sourceFileInfoSet.contains).map(_.size).sum
+          val nextAppendedBytes = curFilesSize - nextCommonBytes
+          val nextDeletedBytes = nextLog.sourceFilesSizeInBytes - nextCommonBytes
```
imback82, line 226: I think you should just define a `def` to handle this.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,95 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+    val versions = deltaLakeVersionHistory(index)
+
+    lazy val indexManager = Hyperspace.getContext(spark).indexCollectionManager
+    def getIndexLogEntry(logVersion: Int): IndexLogEntry = {
+      indexManager
+        .getIndexLogEntry(index.name, logVersion)
+        .get
+        .asInstanceOf[IndexLogEntry]
+    }
+    // TODO: Currently assume all versions of index data exist.
```
imback82, line 188: Possible to create an issue and link here?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -142,4 +143,95 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+    //  Need to check and remove candidate indexes.
+
+    plan.relation match {
+      case HadoopFsRelation(location: TahoeLogFileIndex, _, _, _, _, _) =>
+        val equalOrLessLastIndex = versions.lastIndexWhere(location.tableVersion >= _._2)
```
imback82, line 193: equalOrLessThanLastIndex? Bt the way, could you add a comment what this variable actually means (e.g, what is this index supposed to point to)?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
sezruby: Used `foldLeft` to skip the same log version described below.. we could do `filter` with `var` of previous entry.
Which one is more preferable?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
sezruby: It's the only case; removed "usually".

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
sezruby: Added #387

src/main/scala/com/microsoft/hyperspace/index/IndexManager.scala
```diff
@@ -104,4 +104,15 @@ trait IndexManager {
...
   def index(indexName: String): DataFrame
+
+  /**
+   * Get IndexLogEntry of the given index log version.
+   *
+   * @param indexName Name of the index to get.
+   * @param logVersion Index log version to retrieve.
+   * @return IndexLogEntry if the index of the given log version exists, otherwise None.
+   */
+  def getIndexLogEntry(
```
imback82, line 115: We have `def getIndexes(states: Seq[String] = Seq()): Seq[IndexLogEntry]`. Should we make the naming consistent? Also, we should expose an API to return log versions: https://github.com/microsoft/hyperspace/pull/272#discussion_r598210846.

imback82: If you want, you can create a separate PR that just updates `IndexManager` APIs, but I am also fine if you do it in this PR.

src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
```diff
@@ -113,6 +133,18 @@ trait FileBasedRelation extends SourceRelation {
    * Returns list of pairs of (file path, file id) to build lineage column.
    */
   def lineagePairs(fileIdTracker: FileIdTracker): Seq[(String, Long)]
+
+  /**
+   * Returns IndexLogEntry of the closest index version for the given relation.
```
imback82, line 138: This needs update? No relation is given in this `def`?

src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
```diff
@@ -113,6 +133,18 @@ trait FileBasedRelation extends SourceRelation {
...
+   *
+   * curFiles is used to calculate the similarity with each index version data.
```
imback82, line 140: this needs update?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -76,29 +73,28 @@ object RuleUtils {
...
+    val hybridScanEnabled = HyperspaceConf.hybridScanEnabled(spark)
+    val hybridScanDeleteEnabled = HyperspaceConf.hybridScanDeleteEnabled(spark)
+
+    def getHybridScanCandidate(index: IndexLogEntry): Option[IndexLogEntry] = {
       // TODO: As in [[PlanSignatureProvider]], Source plan signature comparison is required to
       //  support arbitrary source plans at index creation.
       //  See https://github.com/microsoft/hyperspace/issues/158
 
-      entry.withCachedTag(relation.plan, IndexLogEntryTags.IS_HYBRIDSCAN_CANDIDATE) {
+      val entry = relation.closestIndexVersion(index)
```
imback82, line 84: `cloestIndexVersion` sounds a bit weird since it returns an entry. Maybe just `closestIndex`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -123,9 +119,14 @@ object RuleUtils {
           entry.setTagValue(
             relation.plan,
             IndexLogEntryTags.HYBRIDSCAN_REQUIRED,
-            !(commonCnt == entry.sourceFileInfoSet.size && commonCnt == inputSourceFiles.size))
+            !(commonCnt == entry.sourceFileInfoSet.size
+              && commonCnt == relation.allFileInfos.size))
         }
         isCandidate
+      }) {
```
imback82, line 126: `)` here is closing `if`, and for the following it's easy to consider `{` as the starting block of `if`.
```
if (entry.withCachedTag(relation.plan, IndexLogEntryTags.IS_HYBRIDSCAN_CANDIDATE) {`
```
Is there a better way to restructure?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
imback82: Ah OK, this is fine then. Could you add a block of comment of some examples for `deltaLakeVersionHistory`?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -144,4 +145,94 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
   }
+
+  private def deltaLakeVersionHistory(index: IndexLogEntry): Seq[(Int, Int)] = {
+    val versions =
+      index.derivedDataset.properties.properties
+        .getOrElse(DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY, "")
+
+    // The value is comma separated versions - <index log version>:<delta table version>.
+    // e.g. "1:2,3:5,5:9"
+    versions.split(",").reverseIterator.foldLeft(Seq[(Int, Int)]()) { (list, versionStr) =>
```
imback82, line 156: `list` => `acc`?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
imback82: thanks.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -144,4 +145,94 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+          def getAppendedAndDeletedBytes(logEntry: IndexLogEntry): (Long, Long) = {
+            val commonBytes =
+              allFileInfos.filter(logEntry.sourceFileInfoSet.contains).map(_.size).sum
+            (allFileSizeInBytes - commonBytes, logEntry.sourceFilesSizeInBytes - commonBytes)
+          }
+
+          val (prevAppendedBytes, prevDeletedBytes) = getAppendedAndDeletedBytes(prevLog)
+          val (nextAppendedBytes, nextDeletedBytes) = getAppendedAndDeletedBytes(nextLog)
+
+          // Pick one with less diff bytes, so as to reduce the regression from Hybrid Scan.
```
imback82, line 229: `reduce` => you mean `prevent` or `reduce the chance of regression`?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -144,4 +145,94 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+          val prevPair = versions(equalOrLessThanLastIndex)
+          val nextPair = versions(equalOrLessThanLastIndex + 1)
+
+          val prevLog = getIndexLogEntry(prevPair._1)
+          val nextLog = getIndexLogEntry(nextPair._1)
+
+          def getAppendedAndDeletedBytes(logEntry: IndexLogEntry): (Long, Long) = {
+            val commonBytes =
+              allFileInfos.filter(logEntry.sourceFileInfoSet.contains).map(_.size).sum
+            (allFileSizeInBytes - commonBytes, logEntry.sourceFilesSizeInBytes - commonBytes)
```
imback82, line 223: looking at the usage, you can just return the sum of extra bytes? Do you need to separate these?

src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
imback82: ok

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -310,6 +321,262 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
...
+      val dfFromSample = sampleData.toDF("Date", "RGUID", "Query", "imprs", "clicks")
+      dfFromSample.write.format("delta").save(dataPath)
+
+      val timestampMap = mutable.Map[Long, String]()
+      timestampMap.put(0, getSparkFormattedTimestamps(System.currentTimeMillis).head)
+
+      val indexName = "deltaIndex2"
+      val deltaDf = spark.read.format("delta").load(dataPath)
+
+      def appendData(doRefresh: Boolean): Unit = {
```
imback82, line 442: Is it possible to remove duplicate code? I see very similar code in previous test.

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -310,6 +321,262 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
...
+          checkExpectedIndexUsed(Some(0), 1)
+          appendData(doRefresh = false) // delta version 5, index log version 3
+          appendData(doRefresh = false) // delta version 6, index log version 3
+          appendData(doRefresh = true) // delta version 7, index log version 5
+          hyperspace.optimizeIndex(indexName) // delta version 8, index log version 7
+          appendData(doRefresh = true) // delta version 9, index log version 9
+          hyperspace.optimizeIndex(indexName) // delta version 10, index log version 11
+          appendData(doRefresh = false) // delta version 11, index long version 11
+
+          checkExpectedIndexUsed(None, 11)
```
imback82, line 530: Is it possible to add comment for each why the given parameter was used?

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -310,6 +321,262 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
...
+          } else {
+            spark.read.format("delta").option("versionAsOf", version.get).load(dataPath)
+          }
+        } else {
+          spark.read.format("delta").load(dataPath)
+        }
+        deltaDf.filter(deltaDf("clicks") <= 2000).select(deltaDf("query"))
+      }
+
+      def checkExpectedIndexUsed(
```
imback82, line 480: Is it possible to merge with previous `checkExpectedIndexUsed` and test both path and version? (then you can just have one function for validation?)

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
imback82: > I added this because delta lake internally records the time in **second** unit. So I forces to record different times for each version update, so that I could test the query with timestamp properly.

Maybe add this comment to the code? Reading the existing comment, I didn't know why we had to sleep.

imback82 (commented):
Almost there

src/main/scala/com/microsoft/hyperspace/index/IndexManager.scala
sezruby: will do this with a follow up PR

imback82: sounds good!

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -144,4 +145,97 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+   * @param index Index to retrieve the version info.
+   * @return List of (index log version, delta table version) in ascending order.
+   */
+  private def deltaLakeVersionHistory(index: IndexLogEntry): Seq[(Int, Int)] = {
+    val versions =
+      index.derivedDataset.properties.properties
+        .getOrElse(DeltaLakeConstants.DELTA_VERSION_HISTORY_PROPERTY, "")
+
+    // The value is comma separated versions - <index log version>:<delta table version>.
+    // e.g. "1:2,3:5,5:9"
```
imback82, line 161:
```scala
// Versions are processed in a reverse order to keep the higher index log version in case different index
// log versions refer to the same delta lake version.
// For example, "1:1, 2:2, 3:2" will become Seq((1, 1), (3, 2)).
```
, if I understand this correctly?

imback82 (commented):

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -144,4 +145,97 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
...
+          val prevPair = versions(equalOrLessThanLastIndex)
+          val nextPair = versions(equalOrLessThanLastIndex + 1)
+
+          val prevLog = getIndexLogEntry(prevPair._1)
+          val nextLog = getIndexLogEntry(nextPair._1)
+
+          def getDiffBytes(logEntry: IndexLogEntry): Long = {
+            val commonBytes =
+              allFileInfos.filter(logEntry.sourceFileInfoSet.contains).map(_.size).sum
+            allFileSizeInBytes - commonBytes + logEntry.sourceFilesSizeInBytes - commonBytes
```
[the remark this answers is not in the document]
imback82: I think the comment `// appended bytes + deleted bytes` would help?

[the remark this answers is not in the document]
imback82: > I think the comment `// appended bytes + deleted bytes` would help (thus the parenthesis)?

imback82 (approved):

Issue #280, opened 2020-12-03
Hive-partition columns are not correctly stored in index content when refresh-incremental is called


apoorvedave1:
### Describe the issue

When incremental refresh is called on hive-partitioned data, if partitoin columns are part of index columns (indexed/included), they are not picked by the refresh call and are filled with nulls.

### To Reproduce
1. create hive partitioned data. e.g. `df.write.partitionBy("c1").parquet...`
2. create index where c1 is used in either indexed or included columns
3. append new data to source
4. `hs.refresh("index", "incremental")`
5. check index data. 'c1' column will be filled with 'nulls' for all appended data.

### Expected behavior
'c1' column should contain proper partition values instead of nulls

PR #281, opened 2020-12-03
Support incremental refresh index with hive-partition columns

apoorvedave1:

 - **Tracking Issue**: #280

Set "basePath" option during creation and refresh of indexes on a hive-partitioned data source. This will ensure that all calls to `refreshIndex` will pick this basePath while identifying newly appended data files for incremental refresh.

When 'basePath' is known, spark will identify the partition columns automatically and properly pick them to store in the index.

> NOTE: This PR doesn't include changes which may be required for Deltalake data source. Please follow up with https://github.com/microsoft/hyperspace/issues/295 for delta lake

no

unit tests


src/test/scala/com/microsoft/hyperspace/index/IndexManagerTests.scala
```diff
@@ -478,6 +478,41 @@ class IndexManagerTests extends HyperspaceSuite with SQLHelper {
...
+        .toDF("Date", "RGUID", "Query", "imprs", "clicks")
+        .write
+        .partitionBy("Date")
+        .mode("append")
+        .parquet(testPath)
+      hyperspace.refreshIndex("indexName", "incremental")
+
+      // Check if partition columns are correctly stored in index contents.
+      val indexDf2 = spark.read.parquet(s"$systemPath/indexName").where("Date = '2019-10-03'")
+      assert(appendData.size == indexDf2.count())
```
apoorvedave1, line 511: Note to reviewers: this section of the test will fail without the code changes. for the appended data (date = 2019-10-03), the indexDf2.count() will return 0 because "Date = null" for all appended rows).

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -67,8 +67,19 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
         val sourceDataProperties =
           Hdfs.Properties(Content.fromLeafFiles(files, fileIdTracker).get)
         val fileFormatName = fileFormat.asInstanceOf[DataSourceRegister].shortName
+
+        // Store basePath of hive-partitioned data sources, if applicable.
+        val basePath = PathUtils.extractBasePath(location.partitionSpec())
```
sezruby, line 72: Could you use basePath if it's given? as it's already validated.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -67,8 +67,19 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+
         // "path" key in options can incur multiple data read unexpectedly.
-        val opts = options - "path"
+        // Since "options" is case-insensitive map, it will change any previous entries of
+        // "basePath" to lowercase "basepath", making it unusable.
+        // Remove lowercase "basepath" and add proper cased "basePath".
+        val opts = basePath match {
+          case Some(path) => Map("basePath" -> path) ++ options - "path" - "basepath"
+          case _ => options - "path" - "basepath"
+        }
```
sezruby, line 81:
```suggestion
        // Since "options" is case-insensitive map, it will change any previous entries of
        // "basePath" to lowercase "basepath", making it unusable.
        // Remove lowercase "basepath" and add proper cased "basePath".
        val opts =  options - "path" - "basepath" + (basePath match {
          case Some(path) => "basePath" -> path
          case _ => Nil
        }
```
Could you check this? suggestion is not tested.

apoorvedave1: no this won't work because `options` is case insensitive. opts will change "basePath" to "basepath" making it unusable

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
apoorvedave1: thanks @sezruby , added

imback82:
@apoorvedave1 Can you fix the conflicts please?

apoorvedave1:
> @apoorvedave1 Can you fix the conflicts please?

@imback82 , thanks , done

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -440,22 +440,6 @@ object RuleUtils {
...
-      //   - partitions: "/path/col1=1/col2=1", "/path/col1=1/col2=2", etc.
-      // , and going up the same number of directory levels as the number of partition columns
-      // will compute the base path. Note that PartitionSpec.partitions will always contain
-      // all the partitions in the path, so "partitions.head" is taken as an initial value.
-      val basePath = partitionSpec.partitionColumns
-        .foldLeft(partitionSpec.partitions.head.path)((path, _) => path.getParent)
-      Some(basePath.toString)
-    }
-  }
-
```
apoorvedave1, line 458: Note: moved to PathUtils

sezruby: @apoorvedave1 Could you rebase this change as #265 is merged?
We could use `Hyperspace.getContext(spark).sourceProviderManager.partitionBasePath(location)`.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -72,8 +72,20 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
```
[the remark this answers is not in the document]
apoorvedave1: thanks, opts will still be case insensitive. This means what we store in json will be case insensitive. We want case sensitive "basePath" option to store on disk.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
apoorvedave1: thanks @sezruby , fixed

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -203,7 +210,7 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
    */
   override def partitionBasePath(location: FileIndex): Option[String] = {
     location match {
-      case p: PartitioningAwareFileIndex =>
+      case p: PartitioningAwareFileIndex if p.partitionSpec.partitions.nonEmpty =>
```
sezruby, line 213: return `None` in source provider API means the location is not applicable for this source provider.
So if the check is required in this function, we need to change the return type as `Option[Option[String]]`.
Could you fix the return type? Thanks!

apoorvedave1: do you mean changing the definition to 
```scala
def partitionBasePath(location: FileIndex): Option[Option[String]]
```
if that's the case, isn't it easier to change the description of this function to just be: if none, then partitionBasePath is not defined, for any reason whatsoever. Reason could be location is not compatible or location is not hive-partitioned?

Alternately, if we define Option[Option[String]], could you suggest what these scenarios will mean:
Some(Some(path)) => location is supported and basepath is valid
Some(None) => location is supported but basePath is not valid
None => location is not supported

My only concern is this will create difficulty in understanding of the api. I would suggest it's better to keep it simple meaning:
partitioBasePath = 
Some(path) => for valid scenarios
None => for invalid scenarios for any reason. (I am just clubbing the Some(None) and None above.
Please let me know what you think

sezruby: >Alternately, if we define Option[Option[String]], could you suggest what these scenarios will mean:
Some(Some(path)) => location is supported and basepath is valid
Some(None) => location is supported but basePath is not valid
None => location is not supported

I meant this. It's because all of the source provider return None for a API call, the source provider manager throws `No source provider returned valid results.` exception. 
Otherwise, we should check partitonSpec is empty or not before calling API. In this way, if we call the API with non-partitioned location, it will throw the null exception which is easier to know the reason.

imback82: > My only concern is this will create difficulty in understanding of the api.

The source provider APIs return `Option[T]` so that only provider is picked for a relation: https://github.com/microsoft/hyperspace/blob/8fdf28b0019fc6b443081202e0c440fd8b908b6b/src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala#L123-L133

Thus, if `T` needs to be `Option[String]`, the return value for the API should be `Option[Option[String]]` as @sezruby pointed out. Does it make sense?

apoorvedave1: cool. I have pushed another iteration with this suggestion. Please have a look

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeFileBasedSource.scala
```diff
@@ -140,10 +140,10 @@ class DeltaLakeFileBasedSource(private val spark: SparkSession) extends FileBase
    * @param location Partitioned data location.
    * @return basePath to read the given partitioned location.
    */
-  override def partitionBasePath(location: FileIndex): Option[String] = {
+  override def partitionBasePath(location: FileIndex): Option[Option[String]] = {
     location match {
       case d: TahoeLogFileIndex =>
```
sezruby, line 145: Could you add `if (location.partitionSchema.nonEmpty)` here? and return `Some(None)` if `isEmpty` (same for DefaultFileBasedSource)

And could you remove `extractBasePath` function in RuleUtils.scala accordingly? Thanks!

apoorvedave1: thanks @sezruby , added

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -72,8 +72,20 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+        // Remove lowercase "basepath" and add proper cased "basePath".
+        val opts = options.get("basePath").orElse {
+          partitionBasePath(location) match {
+            case Some(Some(path)) => Some(path)
+            case _ => None
+          }
+        } match {
+          case Some(path) => Map("basePath" -> path) ++ options - "path" - "basepath"
+          case _ => options - "path" - "basepath"
+        }
```
sezruby, line 88: For readability, 

```suggestion
        // Store basePath of hive-partitioned data sources, if applicable.
        val basePathOpt = options
          .get("basePath")
          .orElse(partitionBasePath(location).map(Some(_)))
          .map(path => "basePath" -> path)

        // "path" key in options can incur multiple data read unexpectedly.
        // Since "options" is case-insensitive map, it will change any previous entries of
        // "basePath" to lowercase "basepath", making it unusable.
        // Remove lowercase "basepath" and add proper cased "basePath".
        val opts = options - "path" - "basepath" ++ basePathOpt
```

sezruby: And I think we also need to add "basePath" option for DeltaLakeFileBasedSource.
Could you handle this for delta lake and add a test in DeltaLakeIntegrationTest likewise? Thanks!

apoorvedave1: could you please try it once and see if the options is correctly populated with case? As I mentioned previously, this will not work because options is case insensitive.

For deltalake, I cannot take this request at this point because it's been too long for this PR and it's out of scope of this PR. I will update its description and create another issue/PR to tackle this

apoorvedave1: https://github.com/microsoft/hyperspace/issues/295 for testing delta lake for incremental refresh

apoorvedave1: updated the description of this PR that delta is not handled in this case.

sezruby:
```suggestion
       val originalOptions =
          options.asInstanceOf[CaseInsensitiveMap[String]].originalMap

        // Extract "basePath" for partitioned source data even if it's already defined in options,
        // to get the full path of basePath.
        val basePathOpt = partitionBasePath(location).map("basePath" -> _.get)

        // "path" key in options can incur multiple data read unexpectedly.
        val opts = originalOptions - "path" ++ basePathOpt
```

Ok, I tested some and tested this code. 
1) existing basePath can be a relative path; so we need to get the full path anyway
2)  If case insensitive map is the problem, any other options including upper case should have the same problem when refreshing the index.  mergeSchema or ... etc I confirmed that `spark.read.option("basepath", "asdf").parquet("asdf/id=3")` doesn't work.

Could you apply this and revise the comment accordingly?

apoorvedave1: thanks. I made some changes to it because of some unhandled cases.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -72,8 +73,20 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
         val fileFormatName = fileFormat.asInstanceOf[DataSourceRegister].shortName
 
+        // Use case-sensitive map if the provided options are case insensitive.
+        val caseSensitiveOptions = options match {
+          case map: CaseInsensitiveMap[String] => map.originalMap
+          case map => map
+        }
+
+        // Get basePath of hive-partitioned data sources, if applicable.
+        val basePathOpt = partitionBasePath(location) match {
```
sezruby, line 83: `val basePathOpt = partitionBasePath(location).flatten.map("basePath" -> _)`
Could you use this? instead of creating a Map for this.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -72,8 +73,20 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
           Hdfs.Properties(Content.fromLeafFiles(files, fileIdTracker).get)
         val fileFormatName = fileFormat.asInstanceOf[DataSourceRegister].shortName
 
+        // Use case-sensitive map if the provided options are case insensitive.
+        val caseSensitiveOptions = options match {
```
sezruby, line 77: `val caseSensitiveOptions: Map[String, String] = `
Could you add type here instead of below `opt`?  And it seems spark always uses the case insensitive map but this would be ok just in case.

nit: `case _ => _ `

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
apoorvedave1: thanks for the suggestion. this can be done only if we make partitionBasePath a `final` and guarantee that there will never be a None returned from it. This will just introduce bugs in the future if we ever extend DefaultFileBasedSource.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
apoorvedave1: thanks @sezruby , removed the type. 

> nit: `case _ => _`

could you please explain what does this mean?

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
sezruby: @apoorvedave1 There is no hole in this. 

flatten makes 'Option[Option[String]] to Option[String]'

And map only performed the Option is Some and it makes the result Option[(String,String)]

Then Map() ++ Option[(String, String)] added the element only if it's Some.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -72,8 +73,20 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+          case map: CaseInsensitiveMap[String] => map.originalMap
+          case map => map
```
sezruby, line 79: nit: case _ => _ 

I meant this.

apoorvedave1: i mean what's the expectation here? this doesn't set caseSensitiveOptions to options

apoorvedave1:
@imback82 could you please take a look when you get a chance?

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
apoorvedave1: got it. to me the current one is clear enough (adding map to a map vs option to a map). I will let others take a call on this one. Thanks @sezruby

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
sezruby: Sorry, I didn't know that correctly. Thanks for the explanation!

apoorvedave1: 👍

sezruby (commented):
Except for one https://github.com/microsoft/hyperspace/pull/281#discussion_r542981683, LGTM Thanks @apoorvedave1!

And please handle Delta Lake #295 asap as `options` in lowercase can cause a bug.

apoorvedave1:
> Except for one [#281 (comment)](https://github.com/microsoft/hyperspace/pull/281#discussion_r542981683), LGTM Thanks @apoorvedave1!
> 
> And please handle Delta Lake #295 asap as `options` in lowercase can cause a bug.

thank you. I will see if I get the bw. But please feel free to start a PR if you get a chance.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
sezruby: Note that the overhead is not the same..

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -72,8 +73,20 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
           Hdfs.Properties(Content.fromLeafFiles(files, fileIdTracker).get)
         val fileFormatName = fileFormat.asInstanceOf[DataSourceRegister].shortName
 
+        // Use case-sensitive map if the provided options are case insensitive.
```
imback82, line 76: Can you add a comment "why" this is necessary?

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -201,9 +214,9 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
    * @param location Partitioned data location.
    * @return basePath to read the given partitioned location.
```
imback82, line 215: Can you update the comment what `Some(None)` will mean? (also in the `interfaces.scala`)

imback82:
> Store "basePath" in metadata while creating the index on a hive-partitioned data source.

Can you update this? (I think we are setting `basePath` option when creating a relation). I misunderstood it as updating the index metadata to store this info.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
imback82: or you can just say `optional basePath`.

apoorvedave1: thanks @imback82 , done

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
apoorvedave1: added, thanks @imback82

apoorvedave1:
> > Store "basePath" in metadata while creating the index on a hive-partitioned data source.
> 
> Can you update this? (I think we are setting `basePath` option when creating a relation). I misunderstood it as updating the index metadata to store this info.

thanks @imback82 , yeah that was the original implementation. It changed during the review. I have updated the description

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
andrei-ionescu: I'm not fond of `Option[Option[String]]`. It is very hard to understand and very rarely used -- it looks like a coding mistake. Other option would be to create a custom created class/type/object (i.e: `LocationWrapper` or something) to model the 3 states that are needed or maybe use the `Either` class.

If it is still decided to go with `Option[Option[String]]` I would suggest to add to the scala doc the meaning defined above: 

```
Some(Some(path)) => location is supported and basepath is valid
Some(None) => location is supported but basePath is not valid
None => location is not supported
```

imback82 (approved):
LGTM, thanks @apoorvedave1!

@sezruby Can you also finish reviewing this PR?

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
imback82: @andrei-ionescu Fair enough. Do you want to update this as a follow up PR once this PR is checked in? Otherwise, @apoorvedave1 can do a follow up. Please let us know.

sezruby:
@imback82 one comment: https://github.com/microsoft/hyperspace/pull/281#discussion_r542981683
but it seems minor. Let's merge this PR :) Thanks @apoorvedave1!

imback82:
Ah OK. I missed that comment since it was resolved. @apoorvedave1 Can you update it with `flatten` approach? That seems cleaner. Thanks!

apoorvedave1:
> Ah OK. I missed that comment since it was resolved. @apoorvedave1 Can you update it with `flatten` approach? That seems cleaner. Thanks!

done, thanks!

imback82 (approved):

Issue #275, opened 2020-12-02
[PROPOSAL]: Obtain Enhanced Index Statistics


pirz:
## Goal
This design proposal is for implementing enhancement #146

We want to make it possible for a user to get some detailed statistics on existing Hyperspace indexes. These statistics summarize some important logical/physical properties of index and they can be useful for different scenarios on the user side s.a. deciding about triggering index optimization.

## Current Situation
Currently, the only way for a user to get some information and statistics about existing indexes is by checking the output of `Hyperspace.indexes` API. This output lists below information for each index:
1. Index name
2. Indexed columns
3. Included columns
4. Number of buckets
5. Index schema
6. Index state

## Proposed Changes
We propose to make two changes:
1. Introduce `getIndexStat()` API: This is a new API in Hyperspace that is invoked by user to obtain information and statistics on an index.
2. Extend the set of information and statistics that are already shown for indexes by adding more/richer details.

**Note.** This change does not add/modify existing index metadata. Those new statistics that are not explicitly captured in index metadata get computed and added to the output of `getIndexStat()`, as the API is invoked.

### `getIndexStat` API
This new API is defined as:

```
def getIndexStat(indexName: String): DataFrame
```

Once user invokes `getIndexStat` API for a given index, Hyperspace uses the "latest stable"  index metadata to collect and compute required stats. 

Note that no concurrency control happens as this command is running (i.e. It does not lock the index in a "transitioning" state according to Hyperspace concurrency model). As a result, if index gets changed during this time, those changes may or may not be included in the command output, depending on when they are added to the latest stable index metadata log.

### Additional Index Information & Statistics
In addition to info/stats that are already available to user (listed above), we include below items to the output of `getIndexStat`:

1. Index kind
2. Number/Size of index files (latest version)
3. Number/Size of source data files (latest version)
4. Lineage enabled
5. Number of index versions
6. Number/Size of appended files
7. Number/Size of deleted files
8. Index root paths for active versions

### Comments
**c1.** One stat that could be added is "index build time". A couple of points about it:
1. For a fresh index, which is an index that is created and has not been refreshed, "build time" is well-defined. It can be estimated using the difference between timestamps in index metadata logs of "creating" and "active" states. However, once the index goes thru one or more refresh cycles, it is not clear how the build time should be defined, as the index content is gradually/incrementally changed.
2. While reporting "build time" seems interesting, a major problem with it is its accuracy:
[...]

Issue #146, opened 2020-09-03
[FEATURE REQUEST]: Enhancement of Index statistics


sezruby:
### Describe the issue

It would be good to have more detailed statistics about available indexes. This can be done by
- create new APIs such as 
  - `hs.getIndexStat()` for all indexes and `hs.getIndexStat(index)` for an index
- extend IndexSummary class (hs.indexes)
  - IndexSummary (current)
    - Index name
    - indexed columns
    - included columns
    - number of buckets
    - index schema json
    - index location
    - index state
  - candidate properties
    - Total index data size including all versions (disk)
    - latest version info
      - index build time (creation or refresh)
      - index data size
      - index bucket number
    - number of versions
    - number of index data files (increased by incremental refresh) for optimizeIndex

We can provide additional information about source relations of an index to give users some insights so that they could refresh the invalidated index or turn on hybrid scan config. To detect delta of source relations, an input DataFrame is required. So API might look like `hs.getAvailableIndexes(df)` which shows both valid and invalidated indexes for the given DataFrame (or plan). For invalidated indexes, additional properties might be:
- Number of appended files
- Total size of appended data
- Number of deleted file
- Total size of deleted data

rapoth:
@imback82 Could you open a related feature to list the data source location when users utilize `indexes.show` API?

PR #286, opened 2020-12-08
Obtain Enhanced Index Statistics

pirz:
This PR adds a new API for getting detailed index statistics for a given index.
 For more details refer to below design proposal.

- **Tracking Issue**: #146 
- **Design proposal**: #275 

Extending Hyperspace APIs by adding a new API `index(indexName)` to get index metadata and detailed index statistics for a given index.

Here is a sample output DataFrame for an index with lineage enabled which has gone thru 2 rounds of incremental refresh:
```
+------+--------------+---------------+----------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+-------------+----------+-------------+--------------+--------------+---------------+----------------+-----------------+---------------+----------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|name  |indexedColumns|includedColumns|numBuckets|schema                                                                                                                                                                                                                         |state |kind         |hasLineage|numIndexFiles|sizeIndexFiles|numSourceFiles|sizeSourceFiles|numAppendedFiles|sizeAppendedFiles|numDeletedFiles|sizeDeletedFiles|indexRootPaths                                                                                                                                                                                                                         |
+------+--------------+---------------+----------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+-------------+----------+-------------+--------------+--------------+---------------+----------------+-----------------+---------------+----------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|index1|[RGUID]       |[Date]         |200       |{"type":"struct","fields":[{"name":"RGUID","type":"string","nullable":true,"metadata":{}},{"name":"Date","type":"string","nullable":true,"metadata":{}},{"name":"_data_file_id","type":"long","nullable":false,"metadata":{}}]}|ACTIVE|CoveringIndex|true      |16           |1184          |10            |9244           |0               |0                |0              |0               |[file:/C://hyperspace/src/test/resources/indexStatsTest/index1/v__=0, file:/C://hyperspace/src/test/resources/indexStatsTest/index1/v__=1, file:/C://hyperspace/src/test/resources/indexStatsTest/index1/v__=2]|
+------+--------------+---------------+----------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+-------------+----------+-------------+--------------+--------------+---------------+----------------+-----------------+---------------+----------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

```
 
Yes, it adds a new API that can be invoked by user to get statistics for a given index.

Unit tests added.


src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -206,3 +220,67 @@ private[hyperspace] object IndexSummary {
     indexDirPath.toString
   }
 }
+
+/**
```
pirz, line 224: @Reviewers:
Using a case class is a clean/easy way to create a DataFrame with proper schema.
`IndexStatistics` includes the members of `IndexSummary`. Ideally it should have extended `IndexSummary`, however AFAIK Case classes in Scala do not support inheritance. I am open to hear your suggestions to avoid duplicating `IndexSummary` members here.

sezruby:
Could you add an example result of `getIndexStats` API in PR description? Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
sezruby: I'm okay with the approach as it looks more concise.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -110,6 +111,19 @@ class IndexCollectionManager(
       .map(toIndexLogEntry)
   }
 
+  override def getIndexStats(indexName: String): DataFrame = {
+    getLogManager(indexName).fold(
+      throw HyperspaceException(s"Index with name $indexName could not be found")) {
```
sezruby, line 116: nit
```suggestion
      throw HyperspaceException(s"Index with name $indexName could not be found.")) {
```

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -206,3 +220,86 @@ private[hyperspace] object IndexSummary {
...
+ * Case class representing index metadata and index statistics from latest index version.
+ *
+ * @param name index name.
+ * @param indexedColumns indexed columns.
+ * @param includedColumns included columns.
+ * @param numBuckets number of buckets.
+ * @param schema index schema json.
+ * @param indexLocation index location.
+ * @param state index state.
+ * @param kind index kind.
```
sezruby, line 234: how about `type` instead of `kind`?

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -206,3 +220,86 @@ private[hyperspace] object IndexSummary {
     indexDirPath.toString
   }
 }
+
+/**
+ * Case class representing index metadata and index statistics from latest index version.
+ *
+ * @param name index name.
```
sezruby, line 227: nit: 
```suggestion
 * @param name Index name.
```

src/main/scala/com/microsoft/hyperspace/index/IndexManager.scala
```diff
@@ -96,4 +96,12 @@ trait IndexManager {
...
    */
   def getIndexes(states: Seq[String] = Seq()): Seq[IndexLogEntry]
+
+  /**
+   * Get index metadata and detailed index statistics for a given index.
+   *
+   * @param indexName Name of the index to get stats for.
+   * @return Index metadata and statistics as a [[DataFrame]].
+   */
+  def getIndexStats(indexName: String): DataFrame
```
sezruby, line 106: How about adding a utility function - `printIndexStats` which prints the result dataframe in a transposed way?  e.g.
```
name | index1
indexedColumns | [RGUID]
includedColumns | [Date]
numBuckets | 200
schema | "type":"struct","fields":[{"name":"RGUID","type":"string","nullable":true,"metadata":{}}, {"name":"Date","type":"string","nullable":true,"metadata":{}},{"name":"_data_file_id","type":"long","nullable":false,"metadata":{}}]}|
...
```
It's because - the result dataframe has only 1 rows and it can be a little bit inconvenient to check stats in a line.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
pirz: It is called `kind` in IndexLogEntry and the same term is used here.

src/main/scala/com/microsoft/hyperspace/index/IndexManager.scala
pirz: I think such beautification is really up to the consumer. I am not sure adding more and more APIs and user facing functions into the codebase is a good idea. DataFrame is a standard output and easy to use. User can easily interpret/convert it into any desired format/representation.

sezruby: But if they are using Notebook or spark-shell, they need to write such function every time. I think 1-2 utility functions would be good to have. In any case it's just for usability and let's add the print function later if needed.

sezruby (commented):
LGTM Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -206,3 +220,85 @@ private[hyperspace] object IndexSummary {
...
+  }
+
+  /**
+   * Extract top-most index directories which contain index files. Each index directory
+   * contains index files created for an active version of index.
+   *
+   * @param entry Index log entry.
+   * @return List of paths to index directories for all active versions of index.
+   */
+  private def getIndexRootPaths(entry: IndexLogEntry): Seq[String] = {
```
imback82, line 293: "Root" seems a bit misleading. (I was a bit confused until I checked the tests).
I thought if we had `idx1/v=0/` and `idx1/v=1`, it would return `idx` since that's the root. Maybe you meant `getIndexDirectoryForAllVersions` or something similar?

src/test/scala/com/microsoft/hyperspace/index/IndexStatisticsTest.scala
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  test("getIndexStats() on a fresh index returns correct result.") {
+    Seq(true, false).foreach { enableLineage =>
+      withSQLConf(IndexConstants.INDEX_LINEAGE_ENABLED -> enableLineage.toString) {
+        withIndex(indexConfig.indexName) {
+          hyperspace.createIndex(dataDF, indexConfig)
+          validateIndexStats(
+            indexConfig.indexName,
+            Seq(IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX + "=0"))
```
imback82, line 61:
```suggestion
            Seq(s"${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=0"))
```

src/test/scala/com/microsoft/hyperspace/index/IndexStatisticsTest.scala
```diff
@@ -0,0 +1,154 @@
...
+              .write
+              .mode("append")
+              .parquet(testPath)
+
+            hyperspace.refreshIndex(indexConfig.indexName, mode)
+
+            val version = if (mode.equals("quick")) 0 else 1
+            validateIndexStats(
+              indexConfig.indexName,
+              Seq(IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX + s"=$version"))
```
imback82, line 92: ditto on the string interpolation

src/test/scala/com/microsoft/hyperspace/index/IndexStatisticsTest.scala
```diff
@@ -0,0 +1,154 @@
...
+            SampleData.testData
+              .take(3)
+              .toDF(dataColumns: _*)
+              .write
+              .mode("append")
+              .parquet(testPath)
+
+            hyperspace.refreshIndex(indexConfig.indexName, mode)
+
+            val version = if (mode.equals("quick")) 0 else 1
```
imback82, line 89: general comment. If you add a short description on why 0 and 1, it helps the readers. For example,
```
// "quick" mode doesn't create a new index, no index is created (v=0). Whereas in "incremental", blah. 
```

src/test/scala/com/microsoft/hyperspace/index/IndexStatisticsTest.scala
```diff
@@ -0,0 +1,154 @@
...
+  test("getIndexStats() on an index with multiple active versions returns correct result.") {
+    withTempPathAsString { testPath =>
+      withSQLConf(IndexConstants.INDEX_LINEAGE_ENABLED -> "true") {
+        withIndex(indexConfig.indexName) {
+          SampleData.save(spark, testPath, Seq("Date", "RGUID", "Query", "imprs", "clicks"))
+          val df = spark.read.parquet(testPath)
+          hyperspace.createIndex(df, indexConfig)
+
+          import spark.implicits._
+          var expectedVersions = Seq(IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX + "=0")
```
imback82, line 108: ditto on the string interpolation

src/test/scala/com/microsoft/hyperspace/index/IndexStatisticsTest.scala
```diff
@@ -0,0 +1,154 @@
...
+          for (i <- 1 to 2) {
+            SampleData.testData
+              .take(3)
+              .toDF(dataColumns: _*)
+              .write
+              .mode("append")
+              .parquet(testPath)
+
+            hyperspace.refreshIndex(indexConfig.indexName, "incremental")
+            expectedVersions :+= IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX + s"=$i"
```
imback82, line 118: ditto

src/test/scala/com/microsoft/hyperspace/index/IndexStatisticsTest.scala
```diff
@@ -0,0 +1,154 @@
...
+      .getFileSystem(new Configuration())
+      .listStatus(indexPath)
+      .filter(f => f.isDirectory && expectedIndexPaths.contains(f.getPath.getName))
+      .map(_.getPath.toString)
+      .toSet
+
+    val actualPaths = indexStatsDF
+      .select("indexRootPaths")
+      .map(r => r(0).asInstanceOf[Seq[String]])
+      .collect()(0)
```
imback82, line 149: just use `indexStats.indexRootPaths`?

src/test/scala/com/microsoft/hyperspace/index/IndexStatisticsTest.scala
```diff
@@ -0,0 +1,154 @@
...
+            hyperspace.refreshIndex(indexConfig.indexName, "incremental")
+            expectedVersions :+= IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX + s"=$i"
+          }
+          validateIndexStats(indexConfig.indexName, expectedVersions)
+        }
+      }
+    }
+  }
+
+  private def validateIndexStats(indexName: String, expectedIndexPaths: Seq[String]): Unit = {
```
imback82, line 126: It should be `expectedIndexVersions`? In this case, can we just pass `Seq[Int]` to simplify?

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
imback82: Have we thought about using "extended" similar to `DESCRIBE TABLE EXTENDED` or `SHOW TABLE EXTENDED`?

For example, we have the following in `Hyperspace`:
```
def indexes: DataFrame
```

We can maybe update it as the following:
```
def indexes(extended: Boolean = false): DataFrame

def index(extended: Boolean = false): DataFrame
```

In that way, we can have one structure and pick short version if `extended` is false.

imback82 (commented):
Since we are updating the public API, we need to update Python API as well (can be done as a follow up PR)

imback82:
```
> +------+--------------+---------------+----------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+-------------+----------+-------------+--------------+--------------+---------------+----------------+-----------------+---------------+----------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
> |name  |indexedColumns|includedColumns|numBuckets|schema                                                                                                                                                                                                                         |state |kind         |hasLineage|numIndexFiles|sizeIndexFiles|numSourceFiles|sizeSourceFiles|numAppendedFiles|sizeAppendedFiles|numDeletedFiles|sizeDeletedFiles|indexRootPaths                                                                                                                                                                                                                         |
> +------+--------------+---------------+----------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+-------------+----------+-------------+--------------+--------------+---------------+----------------+-----------------+---------------+----------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
> |index1|[RGUID]       |[Date]         |200       |{"type":"struct","fields":[{"name":"RGUID","type":"string","nullable":true,"metadata":{}},{"name":"Date","type":"string","nullable":true,"metadata":{}},{"name":"_data_file_id","type":"long","nullable":false,"metadata":{}}]}|ACTIVE|CoveringIndex|true      |16           |1184          |10            |9244           |0               |0                |0              |0               |[file:/C://hyperspace/src/test/resources/indexStatsTest/index1/v__=0, file:/C://hyperspace/src/test/resources/indexStatsTest/index1/v__=1, file:/C://hyperspace/src/test/resources/indexStatsTest/index1/v__=2]|
> +------+--------------+---------------+----------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+-------------+----------+-------------+--------------+--------------+---------------+----------------+-----------------+---------------+----------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
This is a good start, but maybe we can transpose it so that it's easier to read (as a followup PR)? For example:
```
scala> sql("describe extended t").show(false)
+----------------------------+----------------------------------------------------------------------------+-------+
|col_name                    |data_type                                                                   |comment|
+----------------------------+----------------------------------------------------------------------------+-------+
|val                         |string                                                                      |null   |
|id                          |bigint                                                                      |null   |
|# Partition Information     |                                                                            |       |
|# col_name                  |data_type                                                                   |comment|
|id                          |bigint                                                                      |null   |
|                            |                                                                            |       |
|# Detailed Table Information|                                                                            |       |
|Database                    |test                                                                        |       |
|Table                       |t                                                                           |       |
|Owner                       |terryk                                                                      |       |
|Created Time                |Mon Dec 07 22:28:57 PST 2020                                                |       |
|Last Access                 |UNKNOWN                                                                     |       |
|Created By                  |Spark 3.0.1                                                                 |       |
|Type                        |MANAGED                                                                     |       |
|Provider                    |csv                                                                         |       |
|Table Properties            |[key1=val1]                                                                 |       |
|Location                    |file:/Users/terryk/spark/spark-3.0.1-bin-hadoop2.7/spark-warehouse/test.db/t|       |
|Serde Library               |org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe                          |       |
|InputFormat                 |org.apache.hadoop.mapred.SequenceFileInputFormat                            |       |
|OutputFormat                |org.apache.hadoop.hive.ql.io.HiveSequenceFileOutputFormat                   |       |
+----------------------------+----------------------------------------------------------------------------+-------+
```

I believe @Niharikadutta has the transpose code and we can ping her offline.

imback82:
I created #290 if we want to follow up.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
imback82: I thought about this a bit more and this is what I am proposing:
* The APIs can be similar to `SHOW TABLES` and `SHOW TABLE EXTENDED` as follows:
```scala
def indexes: DataFrame // same as existing (behavior is similar to "SHOW TABLES")

def index(name: String): DataFrame // just def name change from "getIndexStats" (behavior is similar to "SHOW TABLE EXTENDED)
```

* Have one structure (`IndexSummary`) to represent the index info:
For `indexes`, we can select fewer fields when we create the dataframe. For `index`, we will select all the fields by default ("extended" behavior). If we need to groups "stats", we can create `IndexStats` and make it as a member of `IndexSummary`; it will create nested fields, but it's fine. But, let's avoid duplicating fields across different case classes.

WDYT, @pirz @sezruby ?

sezruby: How about using only `IndexStats` and filter additional columns for `indexes`? 
I think it's better than using nested fields. e.g.
for indexes: `Seq(IndexStatistics(spark, toIndexLogEntry(l), extended=false)).toDF().select(name, indexedColumns, includedColumns ...`
for index: `Seq(IndexStatistics(spark, toIndexLogEntry(l), extended=true)).toDF()`

For indexes & index naming, it seems better for consistency :)

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -206,3 +220,85 @@ private[hyperspace] object IndexSummary {
...
+      entry.appendedFiles.size,
+      entry.appendedFiles.map(_.size).sum,
+      entry.deletedFiles.size,
+      entry.deletedFiles.map(_.size).sum,
+      getIndexRootPaths(entry))
+  }
+
+  /**
+   * Extract top-most index directories which contain index files. Each index directory
+   * contains index files created for an active version of index.
```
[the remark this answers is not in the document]
sezruby: it's for the "latest" version of index. I suggested to add this field and thought this would be helpful to show which (version of) index data is referred for the latest index. (so that users can delete outdated versions or perform optimizeIndex.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
sezruby: `getIndexDataLocations` or `getIndexDirectoriesForIncludingVersions` ...? 😅

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
imback82: Yes, either way is fine as long as we don't duplicate fields (and nested fields/flatten only if we want to group fields which may not be necessary at this point)

pirz: @imback82 I had thought about using the same structure for both commands and selecting fewer fields when creating DF for `indexes`. My concern was performance:
1. Fields like `sizeIndexFiles` and `sizeSourceFiles` can take some time to calculate if the index has thousands of files or sources. But with `indexes` we calculate them and then exclude them from the output (i.e. waste of effort).
2. `indexes` fetches information for "all" indexes, so calculating expensive fields for all indexes and truncating them is just not elegant.

Regarding your suggestion to consolidate structures, are you suggesting switching from a "case class" to a normal class and then adding functions to convert an instance of class to a DF with proper schema?

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -110,6 +111,19 @@ class IndexCollectionManager(
       .map(toIndexLogEntry)
   }
 
+  override def getIndexStats(indexName: String): DataFrame = {
```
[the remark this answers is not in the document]
pirz: I dont think so. This API does not use `getIndexes` and directly goes after details of only one index. Our policy for cache usage is the function implementation should use `getIndexes` (which fetches info on all indexes and populates cache if there is cache miss). Hence I think it is Ok to keep this function only in `IndexCollectionManager` (the parent) and `CachingIndexCollectionManager` (the child) simply uses it from the super class.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
sezruby: You can avoid the overhead with a new arg(extended) in apply function, for 'indexes'.

Case class should work

pirz: This is done - `IndexSummary` is removed and a single case class `IndexStatistics` is used for both APIs: `indexes(..)` and `index(..)`.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
pirz: renamed to `getIndexContentDirectoryPaths`.

sezruby (commented):
LGTM Thanks @pirz

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -153,37 +170,124 @@ object IndexCollectionManager {
 }
 
 /**
- * Case class representing index summary
- *
- * TODO: Finalize about adding these: data location, signatures, file lists etc.
+ * Case class representing index metadata and index statistics from latest index version.
```
imback82, line 173: Can you move `IndexStatistics` related code to `IndexStatistics.scala` (not to pollute `IndexCollectionManager.scala`)? You can also move `getIndexContentDirectoryPaths` and `indexDirPath` to `object IndexStatistics`.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -97,8 +98,11 @@ class IndexCollectionManager(
     import spark.implicits._
     getIndexes()
       .filter(!_.state.equals(Constants.States.DOESNOTEXIST))
```
imback82, line 100: `DOESNOTEXIST` now that you are importing it?

imback82 (commented):
LGTM (except for pending comments), thanks @pirz!

pirz:
@imback82 This change has been waiting to be merged for some time. Can you please help with that?

imback82 (commented):
LGTM (two minor comments), thanks @pirz!

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -110,6 +114,19 @@ class IndexCollectionManager(
       .map(toIndexLogEntry)
   }
 
+  override def index(indexName: String): DataFrame = {
+    getLogManager(indexName).fold(
```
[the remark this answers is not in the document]
pirz: @imback82 I originally did not use `withLogManager` as it expects `f: IndexLogManager => Unit`, while here we need to return a `DataFrame`. I made a change to use `withLogManager` and capture generated `df` in a var to return it.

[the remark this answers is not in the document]
imback82: Why not update `withLogManager` as `private def withLogManager[T](indexName: String)(f: IndexLogManager => T): T`?

src/main/scala/com/microsoft/hyperspace/index/IndexStatistics.scala
```diff
@@ -0,0 +1,212 @@
...
+        entry.content.fileInfos.map(_.size).sum,
+        entry.sourceFileInfoSet.size,
+        entry.sourceFileInfoSet.map(_.size).sum,
+        entry.appendedFiles.size,
+        entry.appendedFiles.map(_.size).sum,
+        entry.deletedFiles.size,
+        entry.deletedFiles.map(_.size).sum,
+        getIndexContentDirectoryPaths(entry))
+    } else {
+      IndexStatistics(
```
imback82, line 108: Can't you utilize the new `apply`? (That was the purpose)

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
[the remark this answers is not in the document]
pirz: Yes, that is much better. Thanks. I made the change.

pirz:
Thanks @imback82 - All suggestions are addressed.

imback82 (approved):
LGTM, thanks @pirz!

imback82:
@pirz Can you also update https://microsoft.github.io/hyperspace/docs/ug-quick-start-guide/ since this PR exposes a public API at `Hyperspace` level?

imback82:
> @pirz Can you also update https://microsoft.github.io/hyperspace/docs/ug-quick-start-guide/ since this PR exposes a public API at `Hyperspace` level?

Gentle ping

PR #291, opened 2020-12-10
Add hasParquetAsSourceFormat API in source provider

### LINKED ISSUE #288: Support pluggable internal parquet format via source provider API
Can we introduce `hasParquetAsSourceFormat` to the provider API and record this info in the metadata?

_Originally posted by @imback82 in https://github.com/microsoft/hyperspace/pull/265#discussion_r537989294_

sezruby:

 - **Tracking Issue**: #288 
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Fixes #288

Added a new API - hasParquetAsSourceFormat to determine each source relation uses parquet file format internally in a pluggable way.
For example, Delta Lake stores the file format as "delta" but uses parquet as source file format internally.

Hyperspace applies an additional optimization for parquet source format when using Hybrid Scan & filter index rule.
Therefore to support this optimization for a new source provider, we need this API.

Yes, added new property in index metadata.
(added only if `hasParquetAsSourceFormat` is true.

This is refactoring change so will be tested by existing testcases.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -99,7 +99,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       df: DataFrame): Option[(String, String)] = {
     val relation = df.queryExecution.optimizedPlan.asInstanceOf[LogicalRelation]
     if (Hyperspace.getContext(spark).sourceProviderManager.hasParquetAsSourceFormat(relation)) {
-      Some(IndexConstants.IS_PARQUET_SOURCE_PROPERTY -> "true")
+      Some(IndexConstants.HAS_PARQUET_AS_SOURCE_FORMAT -> "true")
     } else {
       None
     }
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -99,7 +99,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       df: DataFrame): Option[(String, String)] = {
     val relation = df.queryExecution.optimizedPlan.asInstanceOf[LogicalRelation]
     if (Hyperspace.getContext(spark).sourceProviderManager.hasParquetAsSourceFormat(relation)) {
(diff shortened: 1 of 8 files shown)


src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -99,8 +99,11 @@ object IndexConstants {
   // Default id used for a file which does not have an id or its id is not known.
   private[hyperspace] val UNKNOWN_FILE_ID: Long = -1L
 
-  // JSON property name used in index metadata to store whether lineage is enabled for an index.
+  // JSON property names used in index metadata.
+  // Indicate whether lineage is enabled for an index.
   private[hyperspace] val LINEAGE_PROPERTY = "lineage"
+  // Indicate whether the source file format is parquet.
+  private[hyperspace] val IS_PARQUET_SOURCE_PROPERTY = "isParquetSource"
```
imback82, line 106: Can we just name this as `val HAS_PARQUET_AS_SOURCE_FORMAT = "hasParquetAsSourceFormat"`?

imback82 (commented):

apoorvedave1:
LGTM, Thanks @sezruby

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -97,6 +94,25 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
     }
   }
 
+  private def hasParquetAsSourceFormatProperty(
+      spark: SparkSession,
+      df: DataFrame): Option[(String, String)] = {
+    val relation = df.queryExecution.optimizedPlan.asInstanceOf[LogicalRelation]
+    if (Hyperspace.getContext(spark).sourceProviderManager.hasParquetAsSourceFormat(relation)) {
+      Some(IndexConstants.IS_PARQUET_SOURCE_PROPERTY -> "true")
```
imback82, line 102: Doesn't this need to be updated?

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
imback82: Oops, sorry, I meant to keep the `_PROPERTY` at the end to be consistent with `LINEAGE_PROPERTY`: `HAS_PARQUET_AS_SOURCE_FORMAT_PROPERTY`.

PR #293, opened 2020-12-11
Add new IndexLogEntryTag to avoid duplicate calculation in getCandidateIndexes

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Introduce new tags to avoid duplicate operation while applying rules.

- `SIGNATURE_MATCHED` tag keeps the result of `def signatureValid` for each relation & indexes, and reuse if exists. Since there's no config to affect the result, it can be reused without any check.
- `IS_HYBRIDSCAN_CANDIDATE' tag keeps the result of `def isHybridScanCandidate` for each relation & indexes. If some hybrid scan related configs are changed, the result can be changed. So before using this tag, reset the tag values if related configs are changed.
 - related configs
    - "spark.hyperspace.index.hybridscan.maxDeletedRatio"
    - "spark.hyperspace.index.hybridscan.maxAppendedRatio"
- `HYBRIDSCAN_RELATED_CONFIGS` keeps the related config values.

##### Test results
Test data
- 100k chunk lineitem table
- 1 non-candidate index (but with applicable column schema; so that we could measure the getCandidateIndex performance)
- unit: ms

Test scripts
```scala
val linetable = spark.read.parquet(tableName)
val filter = linetable.filter(linetable("l_orderkey") isin (1234,12341234, 123456)).select("l_orderkey")
measure(filter.queryExecution.optimizedPlan)
val filter = linetable.filter(linetable("l_orderkey") isin (1234,12341234, 123456)).select("l_orderkey")
measure(filter.queryExecution.optimizedPlan)
val filter = linetable.filter(linetable("l_orderkey") isin (1234,12341234, 123456)).select("l_orderkey")
measure(filter.queryExecution.optimizedPlan)
```

Result w/o this change:
```
// Non Hybrid Scan
ilter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 1533
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 1359
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 1347

// Hybrid Scan
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 3505
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 1314
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 1392
```

Result with this change
```
// Non Hybrid Scan
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 1690
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 11
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 10

// Hybrid Scan
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 1190
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 313 // cost from constructing the list of FileInfo for the given plan
filter: org.apache.spark.sql.DataFrame = [l_orderkey: bigint]
duration: 306
```

No, it's an internal optimization.

existing tests will cover these changes.


src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,9 +16,50 @@
...
+  val IS_HYBRIDSCAN_CANDIDATE: IndexLogEntryTag[Boolean] =
+    IndexLogEntryTag[Boolean]("hybridScanCandidate")
+
+  // IS_HYBRIDSCAN_CANDIDATE indicates if the index can be applied to the plan using Hybrid Scan.
+  val HYBRIDSCAN_CONFIG_CAPTURE: IndexLogEntryTag[Seq[String]] =
+    IndexLogEntryTag[Seq[String]]("hybridScanConfigCapture")
+
+  private def getHybridScanConfigs(spark: SparkSession): Seq[String] = {
+    Seq(
+      HyperspaceConf.hybridScanEnabled(spark).toString,
```
sezruby, line 44: Note to reviewers: In fact this is not necessary as the hybrid scan related tags are referred only if hybrid scan is enabled.
I added this for the context of the function name.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,9 +16,50 @@
...
+
+  // IS_SIGNATURE_MATCH indicates if the plan has the same signature value with the index or not.
+  val IS_SIGNATURE_MATCH: IndexLogEntryTag[Boolean] =
+    IndexLogEntryTag[Boolean]("isSignatureMatch")
+
+  // IS_HYBRIDSCAN_CANDIDATE indicates if the index can be applied to the plan using Hybrid Scan.
+  val IS_HYBRIDSCAN_CANDIDATE: IndexLogEntryTag[Boolean] =
+    IndexLogEntryTag[Boolean]("hybridScanCandidate")
+
+  // IS_HYBRIDSCAN_CANDIDATE indicates if the index can be applied to the plan using Hybrid Scan.
```
pirz, line 38: Comment seems related to the above API and needs a change.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,9 +16,50 @@
...
+
+  private[hyperspace] def resetHybridScanTagsIfNeeded(
+      spark: SparkSession,
+      plan: LogicalPlan,
+      indexes: Seq[IndexLogEntry]): Unit = {
+    val curConfigs = getHybridScanConfigs(spark)
+
+    indexes.foreach { index =>
+      val taggedConfigs = index.getTagValue(plan, HYBRIDSCAN_CONFIG_CAPTURE)
+      if (taggedConfigs.isEmpty || !taggedConfigs.get.equals(curConfigs)) {
```
pirz, line 57: Question: It seems to be a force reset, so do we really need to check the previous value here? Why not skipping prev value check and simply doing an upsert and in case the prev value is the same as new value it is a simple overwrite of same value (which on avg should be O(1) - the same cost as prev value lookup).

sezruby: prev value check is required for `unsetTagValue(index.unsetTagValue(plan, IS_HYBRIDSCAN_CANDIDATE)) in any case. 
So I updated after the check.

imback82:
> Result (only this PR - including InMemoryFileIndex cost)

So, what's the summary? I wasn't sure which number to look at to see the gain. Also, is the duration in milliseconds?

sezruby:
@imback82  I updated the test results with 2 different binaries. 
The previous result was incorrect because of the cost for cachedIndexCollectionManager.

imback82:
Cool, the result makes more sense now. I will get to the review this weekend. @apoorvedave1 / @pirz can you review this PR meanwhile? Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,12 +16,52 @@
...
```
imback82, line 60: nit: prefer using `map` or `foreach` for this pattern `taggedConfigs.isEmpty || taggedConfigs.get`:
```
      index.getTagValue(plan, HYBRIDSCAN_CONFIG_CAPTURE).foreach { taggedConfigs =>
        if (taggedConfigs.equals(curConfigs)) {
          // Need to reset cached tags as these config changes can change the result.
          index.unsetTagValue(plan, IS_HYBRIDSCAN_CANDIDATE)
          index.setTagValue(plan, HYBRIDSCAN_CONFIG_CAPTURE, curConfigs)
        }
      }
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,12 +16,52 @@
...
+  // HYBRIDSCAN_REQUIRED indicates if Hybrid Scan is required for the index or not.
   val HYBRIDSCAN_REQUIRED: IndexLogEntryTag[Boolean] =
     IndexLogEntryTag[Boolean]("hybridScanRequired")
 
   // COMMON_SOURCE_SIZE_IN_BYTES stores overlapping bytes of index source files and given relation.
   val COMMON_SOURCE_SIZE_IN_BYTES: IndexLogEntryTag[Long] =
     IndexLogEntryTag[Long]("commonSourceSizeInBytes")
+
+  // IS_SIGNATURE_MATCH indicates if the plan has the same signature value with the index or not.
+  val IS_SIGNATURE_MATCH: IndexLogEntryTag[Boolean] =
```
imback82, line 34: nit: `DOES_SIGNATURE_MATCH` or `SIGNATURE_MATCHED`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -58,6 +58,11 @@ object RuleUtils {
     val hybridScanDeleteEnabled = HyperspaceConf.hybridScanDeleteEnabled(spark)
 
     def signatureValid(entry: IndexLogEntry): Boolean = {
+      val signatureMatchTag = entry.getTagValue(plan, IndexLogEntryTags.IS_SIGNATURE_MATCH)
+      if (signatureMatchTag.isDefined) {
+        return signatureMatchTag.get
```
imback82, line 63: nit:
```scala
entry.getTagValue(plan, IndexLogEntryTags.IS_SIGNATURE_MATCH).foreach { sigMatched =>
  return sigMatched
}
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -66,8 +71,13 @@ object RuleUtils {
...
           .signature(plan)) match {
-        case Some(s) => s.equals(sourcePlanSignature.value)
-        case None => false
+        case Some(s) =>
+          val result = s.equals(sourcePlanSignature.value)
+          entry.setTagValue(plan, IndexLogEntryTags.IS_SIGNATURE_MATCH, result)
+          result
+        case None =>
+          entry.setTagValue(plan, IndexLogEntryTags.IS_SIGNATURE_MATCH, false)
+          false
```
imback82, line 80: You can leave this unchanged, but change the following:
```scala
val sigMatched = signatureMap.getOrElseUpdate(
...
}
entry.setTagValue(plan, IndexLogEntryTags.IS_SIGNATURE_MATCH, sigMatched)
sigMatched
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,12 +16,52 @@
...
+    IndexLogEntryTag[Seq[String]]("hybridScanConfigCapture")
+
+  private def getHybridScanConfigs(spark: SparkSession): Seq[String] = {
+    Seq(
+      HyperspaceConf.hybridScanEnabled(spark).toString,
+      HyperspaceConf.hybridScanDeletedRatioThreshold(spark).toString,
+      HyperspaceConf.hybridScanAppendedRatioThreshold(spark).toString)
+  }
+
+  private[hyperspace] def resetHybridScanTagsIfNeeded(
```
imback82, line 52: Should we move this logic out of `IndexLogEntryTags` which should just define tags or tag related functionality without any business logics? We can move this to where it's used.

imback82: This seems to be also unsetting `IS_HYBRIDSCAN_CANDIDATE`. Maybe we can rename this to something like `prepareHybridScanCandidateSelection`, etc.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -143,6 +159,7 @@ object RuleUtils {
             }
       }
       assert(filesByRelations.length == 1)
+      IndexLogEntryTags.resetHybridScanTagsIfNeeded(spark, plan, indexes)
```
imback82, line 162: Do we have a test where configs change? If so, can you update the PR description?

imback82 (commented):
General approach seems fine to me.

imback82:
> * "spark.hyperspace.index.hybridscan.delete.enabled"
> * "spark.hyperspace.index.hybridscan.delete.maxNumDeletedFiles"

Can you update these, which are not in the code base anymore?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,12 +16,37 @@
...
   val HYBRIDSCAN_REQUIRED: IndexLogEntryTag[Boolean] =
     IndexLogEntryTag[Boolean]("hybridScanRequired")
 
   // COMMON_SOURCE_SIZE_IN_BYTES stores overlapping bytes of index source files and given relation.
   val COMMON_SOURCE_SIZE_IN_BYTES: IndexLogEntryTag[Long] =
     IndexLogEntryTag[Long]("commonSourceSizeInBytes")
+
+  // SIGNATURE_MATCHED indicates if the plan has the same signature value with the index or not.
+  val SIGNATURE_MATCHED: IndexLogEntryTag[Boolean] =
+    IndexLogEntryTag[Boolean]("isSignatureMatch")
```
imback82, line 35: update "isSignatureMatch" as well?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,12 +16,37 @@
...
+
+  // IS_HYBRIDSCAN_CANDIDATE indicates if the index can be applied to the plan using Hybrid Scan.
+  // This tag is reset when HYBRIDSCAN_RELATED_CONFIGS was changed.
+  val IS_HYBRIDSCAN_CANDIDATE: IndexLogEntryTag[Boolean] =
+    IndexLogEntryTag[Boolean]("hybridScanCandidate")
```
imback82, line 40: ditto: `isHybridScanCandidate`

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -16,12 +16,37 @@
...
+
+  // HYBRIDSCAN_RELATED_CONFIGS contains Seq of value strings of Hybrid Scan related configs.
+  val HYBRIDSCAN_RELATED_CONFIGS: IndexLogEntryTag[Seq[String]] =
+    IndexLogEntryTag[Seq[String]]("hybridScanConfigCapture")
```
imback82, line 44: ditto

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
imback82: Should `getHybridScanRelatedConfigs` belong to `IndexLogEntryTags`? Why not define this where it's used? We can make a generic version if required later.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -118,9 +130,26 @@ object RuleUtils {
           IndexLogEntryTags.HYBRIDSCAN_REQUIRED,
           !(commonCnt == entry.sourceFileInfoSet.size && commonCnt == inputSourceFiles.size))
       }
+      entry.setTagValue(plan, IndexLogEntryTags.IS_HYBRIDSCAN_CANDIDATE, isCandidate)
       isCandidate
     }
 
+    def prepareHybridScanCandidateSelection(
```
imback82, line 137: Not related to this PR, but we need to think about refactoring RuleUtils.scala. This is already two levels deep (`getCandidateIndexes` -> `isHybridScanCandidate` -> `prepareHybridScanCandidateSelection`).

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -58,17 +59,24 @@ object RuleUtils {
...
+      val sigMatched = signatureMap.getOrElseUpdate(
         sourcePlanSignature.provider,
         LogicalPlanSignatureProvider
           .create(sourcePlanSignature.provider)
           .signature(plan)) match {
         case Some(s) => s.equals(sourcePlanSignature.value)
         case None => false
       }
+      entry.setTagValue(plan, IndexLogEntryTags.SIGNATURE_MATCHED, sigMatched)
+      sigMatched
```
imback82, line 79: Now that I see this pattern is repeating, does it make sense to introduce something like the following:
```scala
def withCachedTag[A](entry: IndexLogEntry, plan: LogicalPlan, tag: IndexLogEntryTag[A])(
    f: () => A): A = {
  entry.getTagValue(plan, tag).foreach { return _ }
  val ret = f()
  entry.setTagValue(plan, tag, ret)
  ret
}
```
, so you can call it like 
```scala
withCachedTag(index, plan, IndexLogEntryTags.IS_HYBRIDSCAN_CANDIDATE) {
  () => isHybridScanCandidate(index, inputSourceFiles, totalSizeInBytes)
}
```
or even push this helper to `IndexLogEntry`:
```scala
index.withCachedTag(plan, IndexLogEntryTags.IS_HYBRIDSCAN_CANDIDATE) {
  () => isHybridScanCandidate(index, inputSourceFiles, totalSizeInBytes)
}
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -101,10 +115,8 @@ object RuleUtils {
         appendedBytesRatio < HyperspaceConf.hybridScanAppendedRatioThreshold(spark) &&
         deletedBytesRatio < HyperspaceConf.hybridScanDeletedRatioThreshold(spark)
 
-
       // For append-only Hybrid Scan, deleted files are not allowed.
-      lazy val isAppendOnlyCandidate = !hybridScanDeleteEnabled && deletedCnt == 0 &&
-        commonCnt > 0 &&
+      lazy val isAppendOnlyCandidate = deletedCnt == 0 && commonCnt > 0 &&
```
imback82, line 119: Isn't this change addressed by a different PR?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -121,19 +121,24 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
       val dataPath = tempPath.getAbsolutePath
       df.write.parquet(dataPath)
 
-      withIndex("index1") {
+      val indexNameWithLineage = "index1"
+      val indexNameWithoutLineage = "index2"
```
imback82, line 125: ditto with this change.

imback82:
> > * "spark.hyperspace.index.hybridscan.delete.enabled"
> > * "spark.hyperspace.index.hybridscan.delete.maxNumDeletedFiles"
> 
> Can you update these, which are not in the code base anymore?

Btw, this comment was regarding the PR description.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -572,6 +572,14 @@ case class IndexLogEntry(
     tags.remove((plan, tag))
   }
 
+  def withCachedTag[T](plan: LogicalPlan, tag: IndexLogEntryTag[T])(
+    f: () => T): T = {
```
imback82, line 576: Let's actually use the following so that the caller doesn't have to specify `() =>` (sorry for the wrong suggestion):
```suggestion
  def withCachedTag[T](plan: LogicalPlan, tag: IndexLogEntryTag[T])(f: => T): T = {
```

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -82,41 +86,62 @@ object RuleUtils {
...
-        entry.setTagValue(
-          plan,
-          IndexLogEntryTags.HYBRIDSCAN_REQUIRED,
-          !(commonCnt == entry.sourceFileInfoSet.size && commonCnt == inputSourceFiles.size))
+    def prepareHybridScanCandidateSelection(
+        spark: SparkSession,
+        plan: LogicalPlan,
+        indexes: Seq[IndexLogEntry]): Unit = {
+      val curConfigs = Seq(
+        HyperspaceConf.hybridScanEnabled(spark).toString,
```
imback82, line 133: Should we remove this? We can add `assert(HyperspaceConf.hybridScanEnabled(spark))` if needed.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -261,6 +261,55 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
...
+        withSQLConf(IndexConstants.INDEX_LINEAGE_ENABLED -> "true") {
+          indexManager.create(readDf, IndexConfig("index1", Seq("id")))
+        }
+
+        val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
+        df.limit(5).write.mode("append").parquet(dataPath)
+        val optimizedPlan = spark.read.parquet(dataPath).queryExecution.optimizedPlan
+
+        withSQLConf(IndexConstants.INDEX_HYBRID_SCAN_ENABLED -> "true") {
+
```
imback82, line 283:
```suggestion
```

imback82 (commented):

imback82 (approved):

imback82:
@sezruby Can you update the PR description? It is now out of sync with code. Thanks!

Issue #276, opened 2020-12-02
[PROPOSAL]: Support for Scan Pattern for Scalable Refresh Index for Large Data Sources


apoorvedave1:
**Describe the problem**
Goals
1. Support incremental refresh call on scan pattern
2. Prevent full file listing of df creation when incremental refresh is called with scan pattern

Independent Scenarios to support
1. Globbing data sources
	E.g. original source data path = data/*/*/*/*
	- create index supports this <PR # handles this>
	- Incremental refresh on scanPattern = data/*/a/*
	- Incremental refresh on scanPattern = data/*/a
hs.refreshIndex("indexname", scanPattern = "data/*/m=11/*")

2. Hive-partitioned data sources
	E.g. original source data path = data
	- create index supports this today
	- Incremental refresh on scanPattern = data/y=2020/m=11/
	- Incremental refresh on scanPattern = data/y=2020/*
	- Incremental refresh on scanPattern = data/*/m=11
hs.refreshIndex("indexname", scanPattern = "data/*/b/*")

**Describe your proposed solution**
General Ideas:
1. Setting "basePath" option in a dataframe is sufficient to inform spark of hive-partition schema if it exists.
2. For hive-partitioned data source: setting rootpath + scanPattern should suffice for listing the right set of files for refresh.
	- e.g. if actual files look like data/y=2020/m=11/d=1/f1.parquet and scanPattern is data/y=2020/*,
	dfUsedForRefresh = spark.read.option("basePath", "data").load("data/y=2020") can be used to refresh index. This will ensure the partition columns are picked and only the files from "data/y=2020" are picked.

2. For globbing data sources: Setting "path" option to a globbing pattern (starting from data root, until the leaf directory) will tell spark to list which files to use in refresh.
	- e.g. if actual files look like data/a/b/c/d/f1.parquet, and therefore, globbing pattern = data/*/*/*/*, and scanPattern is data/*/b/*,
	dfUsedForRefresh = spark.read.load("data/*/b/*/*") can be used to refresh index. This will ensure only the files from "data/y=2020" are picked.

Algo:
1. Globbing Data Sources
	CreateIndex:
		- Set rootPath (in metadata) = "data/*/*/*/*" (same as globbing pattern)
	Refresh Incremental:
		- scanPattern = data/*/b/*
		- Note: "basePath" is not required for globbing patterns but it does no harm even if present
		- mergedPath = merge(globPath, scanPattern) =>
[...]

Issue #233, opened 2020-10-28
[FEATURE REQUEST]: Scalable Refresh Index for Large Data Sources


rapoth:
**Feature requested**

Imagine a scenario where users have a folder `/data/` where they ingest data into partitions (hive-style or not) and each partition contains ~10 files. Then we'd have something like the following:

  - `/data/year=2020/month=10/day=20/hour=00/<30 files>`
  - `/data/year=2020/month=10/day=20/hour=01/<30 files>`
  - `/data/year=2020/month=10/day=20/hour=02/<30 files>`
  - ...

That is, we would have `30 files/hour * 24 hours = 720 files/day`, `21,600 files/month`, `259,200 files/year`. There are users who want to create a Hyperspace index on top of this data. Since data comes in at hourly granularity, users would want to call `refreshIndex` (in incremental mode) to continue maintaining the index. 

Currently, Hyperspace relies on a full enumeration on the underlying file system to determine what data has been appended/deleted by comparing it against what is stored in the `_hyperspace_log`. However, this process could be prohibitively expensive in scenarios like the one above, especially if the user wants to call `refreshIndex` more frequently. 

There is a potential scope for optimization if the user provides us with information that indicates which portion of the data has changed. This can be done potentially by allowing users to specify the **mutable portion** of the underlying source data. For instance: 

`hyperspace.refreshIndex("indexName", scanPattern="<scan-pattern>")`

Therefore, the user would use this API in one of the following ways:

  - Data change in a particular partition: `refreshIndex("idx1", scanPattern="/data/year=2020/month=10/day=29/hour=03/*")`
  - Data change in a particular hierarchy: `refreshIndex("idx1", scanPattern="/data/year=2020/month=07/*")`

In both cases, Hyperspace can avoid doing a naive enumeration on the underlying data source and instead look for changes only in the given `scanPattern`.

**Acceptance criteria** 

- [ ] New `refreshIndex` API that can take in the `scanPattern` parameter

**Success criteria**

- [ ] Measure the time difference for invoking the current implementation of `refreshIndex` which does a full enumeration on a data source containing ~300k files with a single partition change and compare against the new implementation which scans for changes only for a given partition(s)

PR #298, opened 2020-12-15
Incremental Refresh support with scan patterns

apoorvedave1:
Note to reviewers: Please follow the **tracking issue** for general algorithm used.

 - **Tracking Issue**: https://github.com/microsoft/hyperspace/issues/276
 - **Parent Issue**: https://github.com/microsoft/hyperspace/issues/233
 - **Dependencies**: #281 

This PR supports **incremental index refresh when scan patterns are provided**

`hyperspace.refreshIndex("indexName", scanPattern="<scan-pattern>")`

Therefore, the user would use this API in one of the following ways:

Data change in a particular partition: 
`refreshIndex("idx1", "incremental", scanPattern=Some("data/year=2020/month=10/day=29/hour=03/*"))`
Data change in a particular hierarchy: 
`refreshIndex("idx1", "incremental", scanPattern=Some("data/year=2020/month=07/*"))`

Supported scenarios:
```scala
// Case 1: Hive Partitioned Data
val df = spark.read.format(...).load("/path/to/data/")
hs.createIndex(..., IndexConfig("index", ...)
hs.refreshIndex("index", "incremental", Some("data/*"))  // everything
hs.refreshIndex("index", "incremental", Some("data/y=2020/*")) // everything from y=2020
hs.refreshIndex("index", "incremental", Some("data/y=*/m=10")) // everything where y = anything, m = 10
hs.refreshIndex("index", "incremental", Some("data/y=20*/m=10/")) // everything where y starts with 20, m = 10

// Case 2: Globbing Pattern Data
val df = spark.read.format(...).option("spark.hyperspace.source.globbingPattern", "/path/to/data/*/*/*").load("/path/to/data/*/*/*") // Globbing pattern option as per #269 
hs.createIndex(..., IndexConfig("index", ...)
hs.refreshIndex("index", "incremental", Some("data/*"))  // everything
hs.refreshIndex("index", "incremental", Some("data/a/*")) // everything from data/a
hs.refreshIndex("index", "incremental", Some("data/a/b/*")) // everything from data/a/b
hs.refreshIndex("index", "incremental", Some("data/*/b/")) // every folder at depth 2 which ends with b
```

Yes we introduce new refresh apis which can take scan pattern.

**Constraints:**
**The scan pattern MUST start with the data folder which was picked for indexing. In the above examples, the scan pattern starts with `data` which is the base path for indexing.**

unit tests added

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshScan.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshScan.scala
@@ -1,0 +1,158 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.actions
+
+import scala.util.{Failure, Success, Try}
+
+import org.apache.commons.io.FilenameUtils
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
+import org.apache.spark.sql.types.{DataType, StructType}
+
+import com.microsoft.hyperspace.index._
+
+class RefreshScan(
+    spark: SparkSession,
+    logManager: IndexLogManager,
+    dataManager: IndexDataManager,
+    scanPattern: String)
+    extends RefreshIncrementalAction(spark, logManager, dataManager) {
+
+  /** df representing the complete set of data files which will be indexed once this refresh action
+   * finishes. */
+  override protected lazy val df = {
+    val relation = previousIndexLogEntry.relations.head
+    val previouslyIndexedData = relation.data.properties.content
+    val newlyIndexedData = previouslyIndexedData.fileInfos -- deletedFiles ++ appendedFiles
+    val newlyIndexedDataFiles: Seq[String] = newlyIndexedData.map(_.name).toSeq
+
+    val dataSchema = DataType.fromJson(relation.dataSchemaJson).asInstanceOf[StructType]
+    spark.read
+      .schema(dataSchema)
+      .format(relation.fileFormat)
+      .options(relation.options)
+      .load(newlyIndexedDataFiles: _*)
+  }
+
+  def isMatch(path: String, scanPattern: String): Boolean = {
+    val scanSplits = scanPattern.split(Path.SEPARATOR)
+    scanSplits.nonEmpty && path.split(Path.SEPARATOR).contains(scanSplits.head)
+  }
+
+  def resolve(path: String, scanPattern: String): String = {
+    val scanSplits: Array[String] = scanPattern.split(Path.SEPARATOR)
+    val pathSplits: Array[String] = path.split(Path.SEPARATOR)
+    val splitPoint: Int = pathSplits.lastIndexOf(scanSplits.head)
+    var (prefix, suffix) = pathSplits.splitAt(splitPoint)
+
+    for (j <- 0 until math.max(scanSplits.length, suffix.length)) {
+      val resolvedPart = (Try(suffix(j)), Try(scanSplits(j))) match {
+        case (Success(path), Success(scan)) if FilenameUtils.wildcardMatch(path, scan) => path
+        case (Success(path), Success(scan)) if FilenameUtils.wildcardMatch(scan, path) => scan
+        case (Success(_), Success(_)) => throw new Exception("Incompatible scan pattern")
+        case (Success(path), Failure(_)) => path
+        case (Failure(_), Success(scan)) => scan
+        case _ => throw new Exception("Unexpected Exception")
+      }
+
+      prefix :+= resolvedPart
+    }
+    prefix.mkString(Path.SEPARATOR)
+  }
+
+  /** paths resolved with scan pattern.
+   * Paths merged With scan pattern to choose more selective option
+   * e.g. if rootPath
+   */
+  private lazy val resolvedPaths = {
+    val relation = previousIndexLogEntry.relations.head
+    relation.rootPaths.collect {
+      case path if isMatch(path, scanPattern) => resolve(path, scanPattern)
+    }
+
+    // Remove this after testing
+    // Seq("glob2/y=2023")
+  }
+
+  override def logEntry: LogEntry = {
+    // TODO: Deduplicate from super.logEntry()
+    val entry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
+
+    // If there is no deleted files, there are index data files only for appended data in this
+    // version and we need to add the index data files of previous index version.
+    // Otherwise, as previous index data is rewritten in this version while excluding
+    // indexed rows from deleted files, all necessary index data files exist in this version.
+    val updatedEntry = if (deletedFiles.isEmpty) {
+      // Merge new index files with old index files.
+      val mergedContent = Content(previousIndexLogEntry.content.root.merge(entry.content.root))
+      entry.copy(content = mergedContent)
+    } else {
+      // New entry.
+      entry
+    }
+
+    val relation = entry.source.plan.properties.relations.head
+    val updatedRelation =
+      relation.copy(rootPaths = previousIndexLogEntry.relations.head.rootPaths)
+    updatedEntry.copy(
+      source = updatedEntry.source.copy(plan = updatedEntry.source.plan.copy(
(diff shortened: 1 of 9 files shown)


apoorvedave1:
Note to reviewers: Tests for delete scenarios have not been added yet. Please have a look for general idea and comments. Nit comments are ok but not necessary at this time. Thank you.

src/main/scala/com/microsoft/hyperspace/actions/RefreshScan.scala
```diff
@@ -0,0 +1,158 @@
...
+class RefreshScan(
+    spark: SparkSession,
+    logManager: IndexLogManager,
+    dataManager: IndexDataManager,
+    scanPattern: String)
+    extends RefreshIncrementalAction(spark, logManager, dataManager) {
+
+  /** df representing the complete set of data files which will be indexed once this refresh action
+   * finishes. */
+  override protected lazy val df = {
```
apoorvedave1, line 38: Note: This `df` represents all the data this will be indexed after this Action finishes.
E.g. if previously indexed files are f1, f2, f3
Newly added files are f4, f5, f6
but scan pattern matches only f4

Then this `df` will represent f1, f2, f3, f4

`df` is used in creation of metadata to represent the current snapshot of indexed data files (as before)

src/main/scala/com/microsoft/hyperspace/actions/RefreshScan.scala
```diff
@@ -0,0 +1,158 @@
...
+      .schema(dataSchema)
+      .format(relation.fileFormat)
+      .options(relation.options)
+      .load(newlyIndexedDataFiles: _*)
+  }
+
+  def isMatch(path: String, scanPattern: String): Boolean = {
+    val scanSplits = scanPattern.split(Path.SEPARATOR)
+    scanSplits.nonEmpty && path.split(Path.SEPARATOR).contains(scanSplits.head)
+  }
```
apoorvedave1, line 55: true if a path matches scan pattern

src/main/scala/com/microsoft/hyperspace/actions/RefreshScan.scala
```diff
@@ -0,0 +1,158 @@
...
+        case (Success(_), Success(_)) => throw new Exception("Incompatible scan pattern")
+        case (Success(path), Failure(_)) => path
+        case (Failure(_), Success(scan)) => scan
+        case _ => throw new Exception("Unexpected Exception")
+      }
+
+      prefix :+= resolvedPart
+    }
+    prefix.mkString(Path.SEPARATOR)
+  }
```
apoorvedave1, line 76: resolves a path with scan pattern. E.g.
```
path = "c:/data/"
scanPattern = "data/y=20*"
resolvedPath = "c:/data/y=20*"

e.g.
path = "c:/root/data/*/*/*"
scanPattern = "data/a*"
resolvedPath = "c:/root/data/a*/*/*"   // for every level, we pick the more restrictive part. e.g. from * and a* we choose a* because it's more restrictive.
```

src/main/scala/com/microsoft/hyperspace/actions/RefreshScan.scala
apoorvedave1: this is also required for correct signature calculation of indexed data files

src/main/scala/com/microsoft/hyperspace/actions/RefreshScan.scala
```diff
@@ -0,0 +1,158 @@
...
+              // For each file, if it already has a file id, add that id to its corresponding
+              // FileInfo. Note that if content of an existing file is changed, it is treated
+              // as a new file (i.e. its current file id is no longer valid).
+              val id = fileIdTracker.addFile(f)
+              FileInfo(f, id, asFullPath = true)
+            }
+      }
+      .flatten
+      .toSet
+  }
```
apoorvedave1, line 157: `current files` represents **ALL files which match the scan pattern**. Some of these files may already be indexed"

src/main/scala/com/microsoft/hyperspace/actions/RefreshScan.scala
```diff
@@ -0,0 +1,158 @@
...
+    // Helper function to check if a file belongs to one of the resolved paths.
+    def fromResolvedPaths(file: FileInfo): Boolean = {
+      resolvedPaths.exists(p => FilenameUtils.wildcardMatch(file.name, p))
+    }
+
+    val originalFiles = previousIndexLogEntry.relations.head.data.properties.content.fileInfos
+      .filter(fromResolvedPaths)
+
+    (originalFiles -- currentFiles).toSeq
+  }
```
apoorvedave1, line 128: deleted files which match the scan pattern. We do not handle deleted files outside of scan pattern to avoid full file listing.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -72,8 +73,20 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+
         // "path" key in options can incur multiple data read unexpectedly.
-        val opts = options - "path"
+        // Since "options" is case-insensitive map, it will change any previous entries of
+        // "basePath" to lowercase "basepath", making it unusable.
+        // Remove lowercase "basepath" and add proper cased "basePath".
+        val opts = basePath match {
+          case Some(path) => Map("basePath" -> path) ++ options - "path" - "basepath"
+          case _ => options - "path" - "basepath"
+        }
```
apoorvedave1, line 89: this is from another dependency PR #281

src/main/scala/com/microsoft/hyperspace/util/PathUtils.scala
```diff
@@ -27,6 +28,27 @@ object PathUtils {
...
+      //   - partitions: "/path/col1=1/col2=1", "/path/col1=1/col2=2", etc.
+      // , and going up the same number of directory levels as the number of partition columns
+      // will compute the base path. Note that PartitionSpec.partitions will always contain
+      // all the partitions in the path, so "partitions.head" is taken as an initial value.
+      val basePath = partitionSpec.partitionColumns
+        .foldLeft(partitionSpec.partitions.head.path)((path, _) => path.getParent)
+      Some(basePath.toString)
+    }
+  }
+
```
apoorvedave1, line 51: from dependency PR #281

sezruby (commented):
Could you simply measure and share the performance described in "Success criteria" in https://github.com/microsoft/hyperspace/issues/233 before moving forward?

apoorvedave1:
> Could you simply measure and share the performance described in "Success criteria" in #233 before moving forward?
> 
> I think the only gain is not performing `listFiles` for all source files ( list files => toSet => calc appended / deleted files), but a user should specify the exact scan patterns for all update. (if not, the index won't be applied)
> If a user wants to run a query only for that portion of the data, then that makes sense though it makes hard to maintain the index.

Yes that's the only goal. If user appended a small bunch of files, we could avoid listing all files. I am working on the perf part but other than that, could you please review the code for functionality? (thanks for the other comments as well).

(We are not aiming for this specialized scenario where the user has partial indexes and also wants to query the partial data set on which the index is created)

imback82:
> Question to @rapoth : Does the customer see listing file during index creation as bottleneck, or if CX really wanted to refresh only couple of partitions where the data still comes in for other partitions?
> As @sezruby pointed above, if the customer only refreshes index on partial data (which satisfies scan pattern), the index will not be applied (without enabling hybrid scan)

We should be able to figure out whether the partition in question is up to data (instead of enabling hybrid scan), no?

rapoth:
> > Could you simply measure and share the performance described in "Success criteria" in #233 before moving forward?
> > I think the only gain is not performing `listFiles` for all source files ( list files => toSet => calc appended / deleted files), but a user should specify the exact scan patterns for all update. (if not, the index won't be applied)
> > If a user wants to run a query only for that portion of the data, then that makes sense though it makes hard to maintain the index.
> 
> Yes that's the only goal. If user appended a small bunch of files, we could avoid listing all files. I am working on the perf part but other than that, could you please review the code for functionality? (thanks for the other comments as well).
> 
> (We are not aiming for this specialized scenario where the user has partial indexes and also wants to query the partial data set on which the index is created)
> 
> cc @rapoth, @imback82
> Question to @rapoth : Does the customer see listing file during index creation as bottleneck, or if CX really wanted to refresh only couple of partitions where the data still comes in for other partitions?
> As @sezruby pointed above, if the customer only refreshes index on partial data (which satisfies scan pattern), the index will not be applied (without enabling hybrid scan)

The specific ask has been only for the case where refresh is needed on a couple of partitions. How can we solve this: `listing file during index creation`? Isn't that mandatory? Maybe I am missing something subtle here.

sezruby:
Actually I doubt the usability of this feature.. What if a user tires to refresh index with 
hs.refreshIndex("index", "incremental", Some("data/y=2020/*")) after hs.refreshIndex("index", "incremental", Some("data/y=2021/*")) ? 

To apply the index, a user should set the exact file lists when they are writing the query, without Hybrid Scan. In other words, the relation in their query should contain only original source file list + refreshed source file list (with scan pattern), historically.
In addition, push down partition filters might not work with partially refreshed indexes. 

@rapoth @imback82 WDYT?

imback82:
> To apply the index, a user should set the exact file lists when they are writing the query, without Hybrid Scan. In other words, the relation in their query should contain only original source file list + refreshed source file list (with scan pattern), historically.

I assume if the user is using this feature, they would select the right partition in their query.

I think we need to look at partition columns in the user query and only consider source files under that partition when applying index & Hybrid scan. (This should improve Hybrid scan scenario quite a bit)

PR #300, opened 2020-12-22
Add similarity thresholds for Hybrid Scan

### LINKED ISSUE #159: [FEATURE REQUEST]: Introduce a similarity threshold for Hybrid Scan

sezruby:

 - **Tracking Issue**: Fixes #159 
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Currently, there's only a temporary config to determine whether a candidate index can be applied to the given relation or not using Hybrid Scan. (`spark.hyperspace.index.hybridscan.delete.maxNumDeletedFiles`, introduced by #202)

This config was temporarily introduced to avoid some regression from Hybrid Scan in case there are a number of deleted files.
However, it cannot handle:
- different size of source files
- total amount of source files
- appended files

This can cause some regression when most of index data is outdated; for example, except for 1 source file, the other files are deleted in the source dataset and/or there are many appended files. 

Therefore, I introduce the following thresholds for Hybrid Scan. 
```
spark.hyperspace.index.hybridscan.maxDeletedRatio, default: 0.2
spark.hyperspace.index.hybridscan.maxAppendedRatio, default: 0.3
```
- `commonBytes` = total file size of common source files between a candidate index and a given (latest) relation in the query
- ratio of deleted files: 1 - (`commonBytes` / `total size of source files of the index`)
- ratio of appended files: 1 - (`commonBytes` / `total size of source files in the given relation`)

With this, we could apply Hybrid Scan only if both appended and deleted data are less than the thresholds.
The default values are determined based on tpch benchmark results, but it might vary depending on the dataset and query type.

The previous config (`spark.hyperspace.index.hybridscan.delete.maxNumDeletedFiles`) is removed.

Yes, added new threshold configs for Hybrid Scan. 
If the ratio of mutated data is larger than the thresholds, the index won't be applied.

Unit test.


sezruby:
@imback82 @pirz @apoorvedave1 Could you review this change? This PR is required to enable hybrid scan by default.
Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -451,6 +451,11 @@ case class IndexLogEntry(
     relations.head.data.properties.content.fileInfos
   }
 
+  @JsonIgnore
+  lazy val sourceFilesBytes: Long = {
```
imback82, line 455: `sourceFilesTotalSizeInBytes` or `sourceFilesSizeInBytes`. If you just end with `Bytes`, it sounds like you are referring to the actual bytes. Applies to other places if applicable.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -71,7 +71,10 @@ object RuleUtils {
       }
     }
 
-    def isHybridScanCandidate(entry: IndexLogEntry, inputSourceFiles: Seq[FileInfo]): Boolean = {
+    def isHybridScanCandidate(
+        entry: IndexLogEntry,
+        inputSourceFiles: Seq[FileInfo],
+        totalInputBytes: Long): Boolean = {
```
imback82, line 77: To imply that `inputSourceFiles` and `totalInputBytes` are related, we can either make a pair, or have `inputSourceFiles` and `inputSourceFilesSizeInBytes`.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -451,6 +451,11 @@ case class IndexLogEntry(
...
+    sourceFileInfoSet.foldLeft(0L)(_ + _.size)
```
imback82, line 456: nit: you can also do `sourceFileInfoSet.map(_.size).sum` similar to `inputSourceFiles.map(_.size).sum` below. But no strong opinion here (you don't need to change if you don't want).

imback82 (commented):
Generally seems fine to me.

@sezruby, so the "number" of files is not the deciding factor for the regression? We could have lots of small files deleted, but the ratio is small, for example.

sezruby:
> so the "number" of files is not the deciding factor for the regression? We could have lots of small files deleted, but the ratio is small, for example.

Yes, since we use `InSet`, it doesn't affect much. (although there's some overhead compared to small number of deleted files) For deleted files, ratio is more significant factor. In general, if there are many deleted files, rest of "undeleted" source files also might be numerous, which means index can be very effective for the dataset. There could be corner cases, but I think ratio might be enough.

imback82:
Ok, thanks

imback82:
@apoorvedave1 @pirz Can you review this PR? Thanks.

apoorvedave1 (dismissed):
LGTM, thanks @sezruby

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -46,13 +46,17 @@ object IndexConstants {
   val INDEX_HYBRID_SCAN_DELETE_ENABLED = "spark.hyperspace.index.hybridscan.delete.enabled"
```
imback82, line 46: Should we remove this config now? #184 seems to be closed.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -88,13 +91,21 @@ object RuleUtils {
...
       }
+
+      val appendedBytesRatio = 1 - commonBytes / inputSourceFilesSizeInBytes.toFloat
+      val deletedBytesRatio = 1 - commonBytes / entry.sourceFilesSizeInBytes.toFloat
+
       val deletedCnt = entry.sourceFileInfoSet.size - commonCnt
       lazy val isDeleteCandidate = hybridScanDeleteEnabled && entry.hasLineageColumn &&
-        commonCnt > 0 && deletedCnt <= HyperspaceConf.hybridScanDeleteMaxNumFiles(spark)
+        commonCnt > 0 &&
+        appendedBytesRatio < HyperspaceConf.hybridScanAppendedRatioThreshold(spark) &&
```
imback82, line 101: `isDeleteCandiate` needs to be updated now we also check append ratio?

src/main/scala/com/microsoft/hyperspace/util/HyperspaceConf.scala
```diff
@@ -48,12 +48,20 @@ object HyperspaceConf {
       .toLong
   }
 
-  def hybridScanDeleteMaxNumFiles(spark: SparkSession): Int = {
+  def hybridScanDeletedRatioThreshold(spark: SparkSession): Float = {
```
imback82, line 51: Shall we change this to `Double`? I am planning to move our conf system to something similar to Spark, and it uses `Double` instead of `Float` for decimal configs.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
imback82: And `HyperspaceConf.hybridScanDeleteEnabled` can be implemented based on `INDEX_HYBRID_SCAN_DELETED_RATIO_THRESHOLD` (disabled if it is set to 0, for example)

imback82 (commented):
LGTM (few minor points), thanks @sezruby!

sezruby:
Thanks for the reviews! I'll update the user document with a follow-up PR. 
BTW shouldn't we enable lineage column by default as we'd like to enable Hybrid Scan by default? WDYT?

imback82:
Can you fix the build failure please?

src/test/scala/com/microsoft/hyperspace/index/HybridScanForNonPartitionedDataTest.scala
```diff
@@ -110,7 +110,9 @@ class HybridScanForNonPartitionedDataTest extends HybridScanSuite {
       assert(basePlan.equals(filter.queryExecution.optimizedPlan))
     }
 
-    withSQLConf(IndexConstants.INDEX_HYBRID_SCAN_ENABLED -> "true") {
+    withSQLConf(
+      IndexConstants.INDEX_HYBRID_SCAN_ENABLED -> "true",
+      IndexConstants.INDEX_HYBRID_SCAN_DELETED_RATIO_THRESHOLD -> "0") {
```
imback82, line 115: Can this use `HybridScanEnabledAppendOnly`?

src/test/scala/com/microsoft/hyperspace/TestUtils.scala
```diff
@@ -107,3 +107,15 @@ object MockEventLogger {
     emittedEvents = Seq()
   }
 }
+
+object TestConfig {
+  val HybridScanEnabled = Seq(
+    IndexConstants.INDEX_HYBRID_SCAN_ENABLED -> "true",
+    IndexConstants.INDEX_HYBRID_SCAN_APPENDED_RATIO_THRESHOLD -> "0.99",
+    IndexConstants.INDEX_HYBRID_SCAN_DELETED_RATIO_THRESHOLD -> "0.99")
```
imback82, line 115: Can we use "1.0" instead? (not sure what it means by "0.99")

imback82: Are the tests failing if we use the default ratio? I am a bit confused with when to use this config, looking at other tests below.

sezruby: Yes because the sample data is too small, so almost all tests failed with the default config. 
"1.0" means Hybrid scan is performed even if there's no common source data, so I set it "0.9" or "0.99"

imback82 (approved):

PR #301, opened 2020-12-22
Support incremental refresh for Delta Lake

### LINKED ISSUE #295: Investigate if refresh incremental works fine with Delta lake
Investigate whether refresh-incremental works well with delta lake data source. Does it need "basePath" variable set as mentioned in the below comment.

And I think we also need to add "basePath" option for DeltaLakeFileBasedSource.
Could you handle this for delta lake and add a test in DeltaLakeIntegrationTest likewise? Thanks!

_Originally posted by @sezruby in https://github.com/microsoft/hyperspace/pull/281#discussion_r542115718_

### Context
https://github.com/microsoft/hyperspace/pull/281 Support incremental refresh index with hive-partition columns

sezruby:

 - **Tracking Issue**: #295 
 - **Parent Issue**: #148 
 - **Dependencies**: n/a
  
Fixes #295

This PR includes following bug fixes for Delta Lake source.
- support incremental refresh - added a new API to source provider `partialReadFileFormat` to read appended files separately
- fix incremental refresh for partitioned data (#281) + test
- fix options as the original map instead of caseInsensitiveMap 
- quick refresh test

Yes, this PR fixes several issues described above.

Unit test


src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
```diff
@@ -89,6 +89,14 @@ trait FileBasedSourceProvider extends SourceProvider {
    */
   def refreshRelation(relation: Relation): Option[Relation]
 
+  /**
+   * Returns a file format name to read partial data for a given [[Relation]].
+   *
+   * @param relation [[Relation]] object to read partial data files.
+   * @return File format to read partial data files.
+   */
+  def partialReadFileFormat(relation: Relation): Option[String]
```
imback82, line 98: I think this should be `def internalFileFormatName` or something to denote the actual file format used for storing the files?

imback82:
@apoorvedave1 @pirz Can you review this PR? Thanks.

apoorvedave1 (dismissed):
LGTM, thanks @sezruby

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -253,7 +268,7 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
   override def hasParquetAsSourceFormat(logicalRelation: LogicalRelation): Option[Boolean] = {
     logicalRelation.relation match {
       case HadoopFsRelation(_: PartitioningAwareFileIndex, _, _, _, format, _)
-        if isSupportedFileFormat(format) =>
+          if isSupportedFileFormat(format) =>
```
pirz, line 271: nit: Is indentation here correct?

pirz (dismissed):
LGTM, Thanks.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
sezruby: yes it's fixed by scalafmt; it's because this `if` is for the `case`.

imback82 (approved):
LGTM, thanks @sezruby!

Issue #302, opened 2020-12-23
Running Hyperspace with Databricks Runtime 5.5 LTS / 6.4 results in ClassCastException


andrei-ionescu:
### Describe the issue

Running Hyperspace with Databricks Runtime 5.5 LTS, 6.4 and older is resulting in the following error:

```
java.lang.ClassCastException: org.apache.spark.sql.execution.datasources.SerializableFileStatus cannot be cast to org.apache.hadoop.fs.FileStatus at 
  com.microsoft.hyperspace.index.sources.default.DefaultFileBasedSource$$anonfun$3.apply(DefaultFileBasedSource.scala:134) at 
  scala.math.Ordering$$anon$5.compare(Ordering.scala:122) at 
  java.util.TimSort.countRunAndMakeAscending(TimSort.java:355) at java.util.TimSort.sort(TimSort.java:220) at 
  java.util.Arrays.sort(Arrays.java:1438) at scala.collection.SeqLike$class.sorted(SeqLike.scala:648) at 
  scala.collection.AbstractSeq.sorted(Seq.scala:41) at scala.collection.SeqLike$class.sortBy(SeqLike.scala:623) at 
  scala.collection.AbstractSeq.sortBy(Seq.scala:41) at 
  com.microsoft.hyperspace.index.sources.default.DefaultFileBasedSource.signature(DefaultFileBasedSource.scala:134) at 
  com.microsoft.hyperspace.index.sources.FileBasedSourceProviderManager$$anonfun$signature$1.apply(FileBasedSourceProviderManager.scala:80) at 
  com.microsoft.hyperspace.index.sources.FileBasedSourceProviderManager$$anonfun$signature$1.apply(FileBasedSourceProviderManager.scala:80) at 
  com.microsoft.hyperspace.index.sources.FileBasedSourceProviderManager$$anonfun$run$1.apply(FileBasedSourceProviderManager.scala:97) at 
  com.microsoft.hyperspace.index.sources.FileBasedSourceProviderManager$$anonfun$run$1.apply(FileBasedSourceProviderManager.scala:96) at scala.collection.IndexedSeqOptimized$class.foldl(IndexedSeqOptimized.scala:57) at 
  scala.collection.IndexedSeqOptimized$class.foldLeft(IndexedSeqOptimized.scala:66) at
  scala.collection.mutable.ArraySeq.foldLeft(ArraySeq.scala:46) at 
[...]

PR #303, opened 2020-12-23
Fix Hyperspace with Databricks Runtime 5.5 LTS & 6.4 ClassCastException

andrei-ionescu:

 - **Tracking Issue**: https://github.com/microsoft/hyperspace/issues/302
 - **Discussion**: https://github.com/microsoft/hyperspace/discussions/285

This PR fixes the `ClassCastException` encountered when running Hyperspace with Databricks 5.5 LTS Runtime.

Technically, in the case of `org.apache.spark.sql.execution.datasources.SerializableFileStatus cannot be cast to org.apache.hadoop.fs.FileStatus` exception we fall back dynamically getting the needed properties from the `SerializableFileStatus` instance.

This exception is happening because the Spark running in the Databricks 5.5 Runtime has the `PartitioningAwareFileIndex` modified and instead of returning `FileStatus` it returns the `SerializableFileStatus` class. `SerializableFileStatus` is part of Delta open-source project, and normally it should be present at `org.apache.spark.sql.delta.util` but Databricks decided to incorporate into their own version of Spark in another location as seen in the exception.

I tried the following 3 solutions:
1. Use the `inputFiles` of the `PartitioningAwareFileIndex`
2. Create the `SerializableFileStatus` class under `org.apache.spark.sql.execution.datasources` inside the Hyperspace project
3. Dynamically retrieve the properties from the `org.apache.spark.sql.execution.datasources.SerializableFileStatus` class instance.

I decided to go with the **3rd solution** because:
- solution 1 gives all the input files not the ones resulting after the scan
- solution 2 adds code that is part of another open-source project

The code in this PR represents the 3rd solution.

No.

This was tested in a Databricks 5.5 LTS & 6.4 Runtime by executing the example code.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
@@ -24,6 +24,7 @@ import org.apache.spark.deploy.SparkHadoopUtil
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.apache.spark.sql.execution.datasources.{FileFormat, FileIndex, HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
 import org.apache.spark.sql.sources.DataSourceRegister
+import org.apache.spark.util.hyperspace.Utils
 
 import com.microsoft.hyperspace.HyperspaceException
 import com.microsoft.hyperspace.index.{Content, FileIdTracker, Hdfs, Relation}
@@ -65,7 +66,7 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
           _,
           fileFormat,
           options) if isSupportedFileFormat(fileFormat) =>
-        val files = location.allFiles
+        val files = filesFromIndex(location)
         // Note that source files are currently fingerprinted when the optimized plan is
         // fingerprinted by LogicalPlanFingerprint.
         val sourceDataProperties =
@@ -81,7 +82,7 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
             // This logic is picked from the globbing logic at:
             // https://github.com/apache/spark/blob/v2.4.4/sql/core/src/main/scala/org/apache/
             // spark/sql/execution/datasources/DataSource.scala#L540
-            val fs = location.allFiles.head.getPath.getFileSystem(new Configuration)
+            val fs = filesFromIndex(location).head.getPath.getFileSystem(new Configuration)
             val globPaths = pattern
               .split(",")
               .map(_.trim)
@@ -161,7 +162,7 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
     logicalRelation.relation match {
       case HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, format, _)
           if isSupportedFileFormat(format) =>
-        val result = location.allFiles.sortBy(_.getPath.toString).foldLeft("") {
+        val result = filesFromIndex(location).sortBy(_.getPath.toString).foldLeft("") {
           (acc: String, f: FileStatus) =>
             HashingUtils.md5Hex(acc + fingerprint(f))
         }
@@ -190,7 +191,7 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
   override def allFiles(logicalRelation: LogicalRelation): Option[Seq[FileStatus]] = {
     logicalRelation.relation match {
       case HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _) =>
-        Some(location.allFiles)
+        Some(filesFromIndex(location))
       case _ => None
     }
   }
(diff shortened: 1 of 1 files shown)


sezruby:
Thanks for providing this change, @andrei-ionescu! Could you confirm this also works with Databricks runtime 6.4?

@imback82 Could you check this PR and the approach? Thanks!

andrei-ionescu:
@sezruby Successfully tested with Databricks 6.4 Runtime.

andrei-ionescu:
@rapoth, @imback82, @sezruby: Can you check my PR? Thank you very much!

rapoth:
Thank you @andrei-ionescu! Sorry for the delay. Most of us were on vacation. We will review this PR this week. Thank you!

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -259,6 +259,31 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+      // standard API's `FileStatus`
+      index.allFiles.map(_.asInstanceOf[FileStatus])
+    } catch {
+      case e: ClassCastException if e.getMessage.contains("SerializableFileStatus") =>
+        val dbClassName = "org.apache.spark.sql.execution.datasources.SerializableFileStatus"
+        // scalastyle:off classforname
+        val clazz = Class.forName(dbClassName)
+        // scalastyle:on classforname
+        index.allFiles.asInstanceOf[Seq[_]].map { f =>
+          new LocatedFileStatus(
```
imback82, line 276: Any reason we are using `LocatedFileStatus` instead of `FileStatus`?

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -259,6 +259,31 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+            new FileStatus(
+              clazz.getMethod("length").invoke(f).asInstanceOf[Long],
```
imback82, line 278: Can we cache `getMethod`s?
```scala
val lengthMethod = clazz.getMethod("length")
...
new FileStatus(lengthMethod.invoke(f).asInstanceOf[Long],
...

```

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -259,6 +259,31 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
...
+    try {
+      // keep the `asInstanceOf` to force casting or fallback because Databricks `InMemoryFileIndex`
+      // implementation does return a `SerializableFileStatus` instead of the
+      // standard API's `FileStatus`
+      index.allFiles.map(_.asInstanceOf[FileStatus])
+    } catch {
+      case e: ClassCastException if e.getMessage.contains("SerializableFileStatus") =>
+        val dbClassName = "org.apache.spark.sql.execution.datasources.SerializableFileStatus"
+        // scalastyle:off classforname
+        val clazz = Class.forName(dbClassName)
```
imback82, line 273: Can we use `org.apache.spark.util.hyperspace.Utils.classForName` instead?

imback82 (commented):
Thanks @andrei-ionescu for fixing this! Generally looking good to me except for few minor comments.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
andrei-ionescu: I've been inspired by the current implementation in Delta - https://github.com/delta-io/delta/blob/master/src/main/scala/org/apache/spark/sql/delta/util/SerializableFileStatus.scala#L37-L41. 

I'll try without `LocatedFileStatus` and hope it still works. It should.

andrei-ionescu: 👍

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
andrei-ionescu: 👍

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
andrei-ionescu: 👍

andrei-ionescu:
Thanks @imback82 for your feedback! I integrated your suggestions and did tests it in DB Runtime.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -259,6 +260,31 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
         None
     }
   }
+
+  private def filesFromIndex(index: PartitioningAwareFileIndex): Seq[FileStatus] =
+    try {
+      // keep the `asInstanceOf` to force casting or fallback because Databricks `InMemoryFileIndex`
+      // implementation does return a `SerializableFileStatus` instead of the
+      // standard API's `FileStatus`
```
imback82, line 268:
```suggestion
      // Keep the `asInstanceOf` to force casting or fallback because Databrick's
      // `InMemoryFileIndex` implementation returns `SerializableFileStatus` instead of the
      // standard API's `FileStatus`.
```

imback82 (approved):
LGTM, thanks @andrei-ionescu!

imback82 (approved):

Issue #263, opened 2020-11-20
[PROPOSAL]: Support variable schema for included columns


pirz:
## Problem Description
This design proposal is for adding feature request #229.
 
Currently, Hyperspace supports creating indexes only on data with fixed schema. This means:
 
- All columns from "indexed" and "included" column should always be present in all data records.
- When refreshing, no column can be added/removed from "indexed" or "included" columns.

This makes it impossible to support creating index on data with evolving schema.

It is inevitable to enforce "no change" restriction on "indexed" columns as index records are bucketized and sorted on them; However, "included" columns are really payload data and do not affect physical layout of index files, therefore "no change" restriction on them could be lifted.

There are two cases to consider for schema changes:
- (During refresh) (some) new data records do not contain some included columns.
- (During refresh) (some) new data records contain some new included columns.

### Goal
The user should be able to create an index on data with evolving schema with ease.

### Assumptions/Requirements
1. Indexed columns do not change and they are always present in all source data records during index creation and any subsequent index refresh.
2. Data records can not have conflicting columns with same name but different data types added as included columns.
(Parquet's "mergeSchema" does not support it either. As an example, if we have a column whose value is `String` in a set of records and `Int` in other records, when trying to load all these records into a DataFrame using the "mergeSchema" option, ParquetReader fails with error: `Failed to merge incompatible data types string and int`.)
3. Included columns are nullable as they may not exist in all index records. We do not support defining a customized default value for an included column. We follow/use what Parquet does to missing columns when reading data with pre-defined schema.  

### Solution Overview
**Index Creation and Refresh API changes**

We make changes to the `createIndex` and `refreshIndex` APIs so the user provides information about included columns. Following cases should be supported at the API level:
- Specify the included columns, but new data may not contain (some of) those columns.
- Do not specify included columns. Hyperspace takes all the columns in the data schema (minus indexed columns) as included columns.
- Do not specify included columns, but specify columns to exclude from data schema for the included columns.

We add two arguments to the `createIndex` and `refreshIndex` APIs for included columns: 
1. `columns: Seq[String]`: A set of columns, picked from "data schema minus indexed column(s)".
2. `mode: String`: Specifies whether the list of columns should be excluded or included as included columns:
   - `mode = "include"`: Pick the list of columns as included columns.
   - `mode = "exclude"`: Remove the listed columns from data schema and pick the rest of columns as included columns.
 
As an example, assume DataFrame `df` has a schema with 5 columns:

[...]

Issue #229, opened 2020-10-26
[FEATURE REQUEST]: Support variable schema for included columns


imback82:
**Feature requested**

Hyperspace currently supports creating an index on data with fixed schema. However, there have been customer asks where an index needs to be created on data with evolving schema.

Hyperspace should be able to support this scenario - data with evolving schema - as long as the schema for indexed columns do not change; only the schema for included columns change.

**Success criteria**

##### API changes
The user should be able to create an index on data with evolving schema with ease.

There are few scenarios to consider:
* The user specifies the included columns but new data may not contain those columns.
* The user does not specify included columns - take all the columns (minus indexed columns) as included columns.
* The user does not specify included columns, but specify columns to exclude for the included columns.

To support the above scenarios, the current `Hyperspace.createIndex` API needs to be updated to take additional configurations. This is up for the discussion, but we can add `configs: Map[String, String]` to either `createIndex` or `IndexConfig`.

##### Benchmark
If schema needs to be merged, it is known that there will be some performance hits: https://spark.apache.org/docs/latest/sql-data-sources-parquet.html#schema-merging -> `Since schema merging is a relatively expensive operation, and is not a necessity in most cases, we turned it off by default starting from 1.5.0.`

When we load the source data, we should always use the "mergeSchema" option. On the other hand, when loading index data, we should be able to provided merged schema from metadata: see "additional context" below.

Benchmark should be performed to get some ideas on working with merged schema.

##### Supported sources
All the supported source types (parquet, json, etc.) should be clearly documented.

**Additional context**
If you create two parquet files with the following:
```scala
Seq(1).toDF("i").write.mode("append").parquet("/tmp/schema")
Seq((1, "hello")).toDF("i", "str").write.mode("append").parquet("/tmp/schema")
```
, you can read them as:
```scala
spark.read.option("mergeSchema", "true").parquet("/tmp/schema").show
```
, or since we record the schema in the metadata, we can do the merge ourselves and provide it when we read it:
```scala
val schema = new StructType().add("i", IntegerType, true).add("str", StringType, true)
spark.read.schema(schema).parquet("/tmp/schema").show
```
We change the schema in the optimizer, thus the latter approach should work out nicely (but needs further testing for non-existent columns).

PR #309, opened 2021-01-06
Add include/exclude columns to IndexConfig and modify included columns resolution

pirz:
This PR is part of changes to add support for indexing datasets with evolving schema. Check proposal #263 for more details.
This PR extends `IndexConfig` so that user can define included columns for an index via specifying `include` or `exclude` columns.

**Tracking Issue**: #229

This PR adds changes to:
- `IndexConfig` to support defining an index's included columns via specifying `include' or `exclude` columns.
- Column resolution in create index path is modified to extract list of index's included columns from `IndexConfig.include` and `IndexConfig.exclude` .

**Example:**
Assume user wants to create an index on DataFrame `df` whose schema has 5 columns:

`{"C0", "C1", "C2", "C3", "C4"}`

and `"C0"` is picked as the indexed column. 

Below is how `IndexConfig.includedColumns` can be used to define different sets of included columns:
- Pick `{"C1", "C2"}` as included columns:
```
val cols = IncludedColumns(include = Seq("C1", "C2"))
```  
- Pick no column as included columns (i.e. Index only has indexed column):
```
val cols = IncludedColumns(include = Seq())
```
- Exclude `{"C1", "C2"}` from data schema and pick `{"C3", "C4"}` as included columns:
```
val cols = IncludedColumns(exclude = Seq("C1", "C2"))
```
- Exclude no column from schema and pick all columns as included columns. So index will have `{"C1", "C2", "C3", "C4"}` as its included columns:
```
val cols = IncludedColumns(exclude = Seq())
```
We use above `IncludedColumns` instance (i.e. `cols`) for index creation as:
```
val ixConfig = IndexConfig(indexName = "ix", indexedColumns = Seq("C0"), includedColumns = cols)
createIndex(df, ixConfig)
```
Yes, it extends `IndexConfig` definition which is used by user when creating an index.

Unit testing.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -145,14 +145,34 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
     val spark = df.sparkSession
     val dfColumnNames = df.schema.fieldNames
     val indexedColumns = indexConfig.indexedColumns
-    val includedColumns = indexConfig.includedColumns
+    val includeColumns = indexConfig.includedColumns.include
+    val excludeColumns = indexConfig.includedColumns.exclude
     val resolvedIndexedColumns = ResolverUtils.resolve(spark, indexedColumns, dfColumnNames)
-    val resolvedIncludedColumns = ResolverUtils.resolve(spark, includedColumns, dfColumnNames)
+    val resolvedIncludeColumns = ResolverUtils.resolve(spark, includeColumns, dfColumnNames)
+    val resolvedExcludeColumns = ResolverUtils.resolve(spark, excludeColumns, dfColumnNames)
 
-    (resolvedIndexedColumns, resolvedIncludedColumns) match {
-      case (Some(indexed), Some(included)) => (indexed, included)
+    (resolvedIndexedColumns, resolvedIncludeColumns, resolvedExcludeColumns) match {
+      case (Some(indexed), Some(include), Some(exclude)) =>
+        if (include.intersect(exclude).nonEmpty) {
+          throw HyperspaceException(
+            "IncludedColumns include and exclude columns can not overlap.")
+        }
+
+        // If IndexConfig has both 'include' and 'exclude' columns, we pick the columns
+        // listed in indexConfig.include as index's included columns and ignore `exclude`.
+        if (includeColumns.nonEmpty) {
+          return (indexed, include)
+        } else if (excludeColumns.nonEmpty) {
+          val columns = dfColumnNames.filterNot(c => indexed.contains(c) || exclude.contains(c))
+          return (indexed, columns)
+        }
+
+        // There is no 'include' or 'exclude' column lists in indexConfig. So index will be created
+        // with only indexed columns and without any included column.
+        (indexed, Seq[String]())
+
       case _ =>
-        val unresolvedColumns = (indexedColumns ++ includedColumns)
+        val unresolvedColumns = (indexedColumns ++ includeColumns ++ excludeColumns)
           .map(c => (c, ResolverUtils.resolve(spark, c, dfColumnNames)))
           .collect { case c if c._2.isEmpty => c._1 }
         throw HyperspaceException(
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
@@ -28,25 +28,28 @@ import java.util.Locale
 case class IndexConfig(
     indexName: String,
     indexedColumns: Seq[String],
-    includedColumns: Seq[String] = Seq()) {
+    includedColumns: IncludedColumns = IncludedColumns()) {
   if (indexName.isEmpty || indexedColumns.isEmpty) {
     throw new IllegalArgumentException("Empty index name or indexed columns are not allowed.")
   }
 
   val lowerCaseIndexedColumns = toLowerCase(indexedColumns)
-  val lowerCaseIncludedColumns = toLowerCase(includedColumns)
-  val lowerCaseIncludedColumnsSet = lowerCaseIncludedColumns.toSet
+  val lowerCaseIncludedColumnsIncludeSet = includedColumns.lowerCaseIncludeColumnsSet
+  val lowerCaseIncludedColumnsExcludeSet = includedColumns.lowerCaseExcludeColumnsSet
 
   if (lowerCaseIndexedColumns.toSet.size < lowerCaseIndexedColumns.size) {
     throw new IllegalArgumentException("Duplicate indexed column names are not allowed.")
   }
 
-  if (lowerCaseIncludedColumnsSet.size < lowerCaseIncludedColumns.size) {
-    throw new IllegalArgumentException("Duplicate included column names are not allowed.")
+  if (lowerCaseIncludedColumnsIncludeSet.size < includedColumns.lowerCaseIncludeColumns.size ||
+      lowerCaseIncludedColumnsExcludeSet.size < includedColumns.lowerCaseExcludeColumns.size) {
+    throw new IllegalArgumentException(
+      "Duplicate include or exclude column names in included columns are not allowed.")
   }
 
   for (indexedColumn <- lowerCaseIndexedColumns) {
-    if (lowerCaseIncludedColumns.contains(indexedColumn)) {
+    if (lowerCaseIncludedColumnsIncludeSet.contains(indexedColumn) ||
+        lowerCaseIncludedColumnsExcludeSet.contains(indexedColumn)) {
       throw new IllegalArgumentException(
         "Duplicate column names in indexed/included columns are not allowed.")
     }
@@ -57,20 +60,20 @@ case class IndexConfig(
       case IndexConfig(thatIndexName, thatIndexedColumns, thatIncludedColumns) =>
         indexName.equalsIgnoreCase(thatIndexName) &&
           lowerCaseIndexedColumns.equals(toLowerCase(thatIndexedColumns)) &&
-          lowerCaseIncludedColumnsSet.equals(toLowerCase(thatIncludedColumns).toSet)
+          includedColumns.equals(thatIncludedColumns)
       case _ => false
     }
   }
 
   override def hashCode(): Int = {
-    lowerCaseIndexedColumns.hashCode + lowerCaseIncludedColumnsSet.hashCode
+    lowerCaseIndexedColumns.hashCode + includedColumns.hashCode()
   }
 
   override def toString: String = {
     val indexedColumnNames = lowerCaseIndexedColumns.mkString(", ")
-    val includedColumnNames = lowerCaseIncludedColumns.mkString(", ")
+    val includedColumnsString = includedColumns.toString
     s"[indexName: $indexName; indexedColumns: $indexedColumnNames; " +
-      s"includedColumns: $includedColumnNames]"
+      s"includedColumns: $includedColumnsString]"
   }
 
   private def toLowerCase(seq: Seq[String]): Seq[String] = seq.map(_.toLowerCase(Locale.ROOT))
@@ -82,13 +85,25 @@ case class IndexConfig(
  */
 object IndexConfig {
 
+  def apply(indexName: String, indexedColumns: Seq[String]): IndexConfig = {
+    IndexConfig(indexName, indexedColumns, IncludedColumns())
+  }
+
+  def apply(
+      indexName: String,
+      indexedColumns: Seq[String],
+      columnsToInclude: Seq[String]): IndexConfig = {
+    IndexConfig(indexName, indexedColumns, IncludedColumns(columnsToInclude))
+  }
+
   /**
    * Builder for [[IndexConfig]].
    */
   class Builder {
 
     private[this] var indexedColumns: Seq[String] = Seq()
-    private[this] var includedColumns: Seq[String] = Seq()
+    private[this] var includedColumnsInclude: Seq[String] = Seq()
+    private[this] var includedColumnsExclude: Seq[String] = Seq()
     private[this] var indexName: String = ""
 
     /**
@@ -129,20 +144,38 @@ object IndexConfig {
     }
 
     /**
-     * Updates included columns for [[IndexConfig]].
+     * Updates included columns for [[IndexConfig.includedColumns]].
      *
      * Note: API signature supports passing one or more argument.
      *
-     * @param includedColumn included column for [[IndexConfig]].
-     * @param includedColumns included columns for [[IndexConfig]].
+     * @param includedColumn included column for [[IndexConfig.includedColumns]].
+     * @param includedColumns included columns for [[IndexConfig.includedColumns]].
      * @return an [[IndexConfig.Builder]] object with updated included columns.
      */
     def include(includedColumn: String, includedColumns: String*): Builder = {
-      if (this.includedColumns.nonEmpty) {
+      if (this.includedColumnsInclude.nonEmpty) {
         throw new UnsupportedOperationException("Included columns are already set.")
       }
 
-      this.includedColumns = includedColumn +: includedColumns
+      this.includedColumnsInclude = includedColumn +: includedColumns
+      this
+    }
+
+    /**
+     * Updates excluded columns for [[IndexConfig.includedColumns]].
+     *
+     * Note: API signature supports passing one or more argument.
+     *
+     * @param excludedColumn excluded column for [[IndexConfig.includedColumns]].
+     * @param excludedColumns excluded columns for [[IndexConfig.includedColumns]].
+     * @return an [[IndexConfig.Builder]] object with updated excluded columns.
+     */
+    def exclude(excludedColumn: String, excludedColumns: String*): Builder = {
+      if (this.includedColumnsExclude.nonEmpty) {
+        throw new UnsupportedOperationException("Excluded columns are already set.")
+      }
+
+      this.includedColumnsExclude = excludedColumn +: excludedColumns
       this
     }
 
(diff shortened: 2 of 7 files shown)


python/hyperspace/includedcolumns.py
```diff
@@ -0,0 +1,12 @@
+class IncludedColumns:
```
sezruby, line 1: Do we have a better name for this? How about `IndexColumnConfig` or `IndexConfig.Columns` ?
It looks bit confusing because of "exclude"

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -145,14 +145,34 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+
+    (resolvedIndexedColumns, resolvedIncludeColumns, resolvedExcludeColumns) match {
+      case (Some(indexed), Some(include), Some(exclude)) =>
+        if (include.intersect(exclude).nonEmpty) {
+          throw HyperspaceException(
+            "IncludedColumns include and exclude columns can not overlap.")
+        }
+
+        // If IndexConfig has both 'include' and 'exclude' columns, we pick the columns
+        // listed in indexConfig.include as index's included columns and ignore `exclude`.
```
sezruby, line 162: I think it would be better to throw an exception if a user set both configs. It might cause some confusion to users.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -145,14 +145,34 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+        if (includeColumns.nonEmpty) {
+          return (indexed, include)
+        } else if (excludeColumns.nonEmpty) {
+          val columns = dfColumnNames.filterNot(c => indexed.contains(c) || exclude.contains(c))
+          return (indexed, columns)
+        }
+
+        // There is no 'include' or 'exclude' column lists in indexConfig. So index will be created
+        // with only indexed columns and without any included column.
+        (indexed, Seq[String]())
```
sezruby, line 172: In this case, can we include all columns except indexed columns? I think that is a more expected behavior.

#229 

> The user does not specify included columns - take all the columns (minus indexed columns) as included columns.

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
```diff
@@ -153,14 +186,51 @@ object IndexConfig {
...
   def builder(): Builder = new Builder
 }
+
+/**
+ * Specifies list of included columns in terms of columns to include and/or exclude.
+ *
+ * @param include List of column names to include as include columns.
+ * @param exclude List of column names to exclude to form list of included columns.
+ */
+case class IncludedColumns(include: Seq[String] = Nil, exclude: Seq[String] = Nil) {
```
sezruby, line 210: Same here. How about including indexed columns? 
e.g.
```
IndexConfig.ColumnInfo.indexed
IndexConfig.ColumnInfo.included
IndexConfig.ColumnInfo.excluded
```
`ColumnInfo` or `Columns` or `ColumnConfig` or `ColumnSchema`

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
```diff
@@ -129,20 +144,38 @@ object IndexConfig {
...
      *
      * Note: API signature supports passing one or more argument.
      *
-     * @param includedColumn included column for [[IndexConfig]].
-     * @param includedColumns included columns for [[IndexConfig]].
+     * @param includedColumn included column for [[IndexConfig.includedColumns]].
+     * @param includedColumns included columns for [[IndexConfig.includedColumns]].
      * @return an [[IndexConfig.Builder]] object with updated included columns.
      */
     def include(includedColumn: String, includedColumns: String*): Builder = {
```
sezruby, line 155: Just out of curiosity, why do we have these two args, not just `include(includedColumns: String*)`?
If we'd like to enforce non-emptiness, we could add assert here.

python/hyperspace/includedcolumns.py
pirz: IncludedColumns -> IncludedColumnsConfig

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
pirz: IncludedColumns -> IncludedColumnsConfig

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: This is added to be as close as possible to today's behavior: when a user wants to create an index with only indexed columns, they simply set empty included columns seq. If we make the change you suggest, then in-order to create such an index, the user has to explicitly add each and every columns to the `exclude` list.

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
pirz: This is a natural way to capture API's requirement in API's signature: "The API should be called with at least one argument" and it is used with builder pattern. As an example, you can check Spark's `Dataset` APIs that follow the builder pattern s.a. [`groupBy`](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#groupBy(col1:String,cols:String*):org.apache.spark.sql.RelationalGroupedDataset) and [`agg`](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#agg(expr:org.apache.spark.sql.Column,exprs:org.apache.spark.sql.Column*):org.apache.spark.sql.DataFrame)

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
```diff
@@ -153,14 +186,51 @@ object IndexConfig {
...
   def builder(): Builder = new Builder
 }
+
+/**
+ * Specifies list of included columns in terms of columns to include and/or exclude.
+ *
+ * @param include List of column names to include as include columns.
+ * @param exclude List of column names to exclude to form list of included columns.
+ */
+case class IncludedColumnsConfig(include: Seq[String] = Nil, exclude: Seq[String] = Nil) {
```
sezruby, line 210: IMO, having `IncludedColumnsConfig` inside `IndexConfig` is a bit confusing. Any reason to introduce this class? @pirz 

Since this is a part of user APIs, it should look clear and concise.
This is the current API using IncludedColumnsConfig:

```
val indexConfig = IndexConfig("index1", Seq("RGUID"), IncludedColumnsConfig(Seq(), Seq("Date")))
```

I think it's better to accept w/o the additional class, e.g. `IndexConfig(name, indexedColumns, includedColumns, excludedColumns)` and we could create it internally if needed.

@imback82 Could you give any comment on this?
I suggested including all 3 types of columns (indexed/included/excluded) https://github.com/microsoft/hyperspace/pull/309#discussion_r553808258, but it might be better not to use another class in user APIs.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
sezruby: Ok then.. any idea for including all columns without specifying them?
I think it would be good to have that option and it's one requirement of that issue.

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
pirz: @sezruby Lets be clear here: `IncludedColumnsConfig` is not *inside* `IndexConfig`, it is a completely separate case class, both in definition and usage. I added its definition in the same file just to keep it close to `IndexConfig` as they are used together. The definition can be moved to its own separate file.
 
I am not sure what exactly is wrong with adding a new class, `include` and `exclude` lists are both related to the definition of included columns and packing them together makes sense. The API call is clean and clear from a user perspective. I am not sure how unwrapping the config and bringing out two lists and add them separately removes any confusion. Can you be more precise about the *confusion* you are concerned about?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: An index is typically created on a subset of columns and not the whole data. Given that picking each an every column in source data should be a rare case, one can do that by listing all columns (minus indexed ones) in the include list of IncludedColumnsConfig.

sezruby: In the reporting issue(#229) and even in the pr description, there's an example picking all columns.

>Exclude no column from schema and pick all columns as included columns. So index will have {"C1", "C2", "C3", "C4"} as its included columns:
val cols = IncludedColumns(exclude = Seq())

Hyperspace indexes can be used to remove shuffle or improve filter query performance, so I think it's a bit different from  traditional indexes.

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
sezruby: In terms of :
* as we always take either of "included" or "excluded", it's confusing to handle them using one class
* they can use both included and excluded but excluded won't be used 
* we don't need to keep "excluded" columns after creation

minor points:
* need to import `IncludedColumnConfig` other than `IndexConfig` in their applications 
* public `lazy val`s will be exposed unnecessarily

We could keep the class internally if needed but I don't think it's good to expose the class in API.

imback82: Yea, I think we can update `IndexConfig` to have `name`, `indexedColumns`, `includedColumns` and `excludedColumns` similar to the `Builder`? (In this case, it's easier to add a verification check - inside `IndexConfig` - whether any of excluded columns are present in indexed/included columns, as well.)

Later when we need to add more info (if any) to the included columns,  we can come back to refactor? What do you guys think?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -145,14 +145,37 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    val excludeColumns = indexConfig.includedColumns.exclude
     val resolvedIndexedColumns = ResolverUtils.resolve(spark, indexedColumns, dfColumnNames)
-    val resolvedIncludedColumns = ResolverUtils.resolve(spark, includedColumns, dfColumnNames)
+    val resolvedIncludeColumns = ResolverUtils.resolve(spark, includeColumns, dfColumnNames)
+    val resolvedExcludeColumns = ResolverUtils.resolve(spark, excludeColumns, dfColumnNames)
+
+    (resolvedIndexedColumns, resolvedIncludeColumns, resolvedExcludeColumns) match {
+      case (Some(indexed), Some(include), Some(exclude)) =>
+        if (include.nonEmpty && exclude.nonEmpty) {
+          throw HyperspaceException(
```
imback82, line 157: This check can go inside `IndexConfig` as `require`.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
pirz: My prev comment does not say that an index can't/shouldn't be created on all columns, and obviously yes I know one can use an index to remove shuffles. 
Here we really talk about two cases:
1. Creating an index with only indexed columns.
2. Creating an index on all columns of source data.

We also have two syntax:
1. Listing no include/exclude column. (Simpler syntax)
2. Listing each and every column as include OR exclude.

We are trying to assign these two syntax to above cases. (Do you agree? :))
I am saying that, first case happens more frequent than second and by using an `IncludedColumns` with no include/exclude columns for the first case, we are:
1. Using the simpler syntax for a more frequent case, and
2. Keep this syntax change (switching from `Seq[String]()` to `IncludedColumns`) closer to what we have today. (When user wants to create an index with only indexed columns, he passes an empty sequence).

If we go with your suggestion, then the simpler syntax will be used for the second case, which is a more rare case to happen and that is why I am not in favor of it. @sezruby Let me know how you think.

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
pirz: The reason I have boxed `include` and `exclude` together is because they are related to a single concept in our covering index definition: "Included Columns". 
Removing `IncludedColumnsConfig` class and adding two separate lists `includedColumns` and `excludedColumns` does not add any more clarity that both `include` and `exclude` can not be used together. Whatever check/requirement we had inside `IncludedColumnsConfig` now comes one level up, so I am not sure why you say that clarifies/removes such requirements.

I am not religious about keeping `IncludedColumnsConfig` class, however one side reason that I considered when adding it was avoiding any confusion in naming arguments: In our Covering Index definition, "Included Columns" is a well-defined concept: "The payload columns that piggyback on index records which are physically organized according to indexed columns". Now if we change the API as:
`IndexConfig(name, indexedColumns, includedColumns, excludedColumns)`

Then `includedColumns` and `excludedColumns` arg names seems confusing as the first arg has the same name as the "included columns" concept. However, boxing them together and under the name `IncludedColumnsConfig` clarifies that these two args are used to define "included columns" list for an index. @imback82 If you are absolutely in favor of dropping `IncludedColumnsConfig`, then what do you suggest for arg names?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
sezruby: Yea I agree with your point (empty included & excluded columns => only indexed columns) but how could we support both (included all or nothing) cases? Because specifying all columns is also not a good API in case there are many columns in source data.

pirz: I agree with you, I am in favor of avoiding necessity of listing all columns.
As both `include` and `exclude` should not be set together, one option could be using:
1. One Seq[String] () to list column(s) names, and
2.  Using a `mode` to specify `include` or `exclude`.

(This is somehow related to the [other comment's discussion](https://github.com/microsoft/hyperspace/pull/309#discussion_r557067683))

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
```diff
@@ -28,25 +28,38 @@ import java.util.Locale
...
+    throw new IllegalArgumentException(
+      "Duplicate include or exclude column names in included columns are not allowed.")
+  }
+
+  if (lowerCaseIncludedColumnsIncludeSet.nonEmpty && lowerCaseIncludedColumnsExcludeSet.nonEmpty) {
+    throw new IllegalArgumentException(
+      "IncludedColumnsConfig include and exclude columns can not both be non-empty.")
+  }
+
+  if (lowerCaseIncludedColumnsIncludeSet.intersect(lowerCaseIncludedColumnsExcludeSet).nonEmpty) {
```
pirz, line 55: **To reviewers:** This check is actually redundant and should be dropped, as both lists can not be non-empty and that is verified in the previous check. I will drop this one.

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
imback82: I don't have a strong opinion here. But we need to document either way since we should not expect the users have a prior knowledge on the" included columns" concept, not to mention the covering index.

Also, even if we introduce `IncludedColumnsConfg`, I suggest exposing `apply` that takes these four arguments for the ease of use, as @sezruby pointed out above.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
sezruby: Yep this looks better in terms of 1) receiving 1 list 2) we can support both including all or nothing

Could you update PR? Thanks!

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
pirz: @imback82 WDYT about the alternative approach from last comment? (I copied it below).

One more idea (coming from [discussion](https://github.com/microsoft/hyperspace/pull/309#discussion_r560600669)) could be:
As both `include` and `exclude` should not be set together, one option could be using:
1. One Seq[String] () to list column(s) names, and
2. Using a mode to specify include or exclude.

Then we can list the columns that we want to include or exclude (depending on whichever is shorter to add) w.r.t to the DF schema passed to `createIndex` and create our desired "included columns" list. Moreover, with `("exclude", Seq())` we can include all columns and with `("include", Seq())` we can exclude all columns.

sezruby:
@pirz 
Could you create a new issue for determining createIndex / refershIndex APIs for other index types & exclude/include columns? 
We discussed some for that, but there's no issue for the API changes.

I think we could adapt builder APIs to support various types/columns config flexibly.

src/main/scala/com/microsoft/hyperspace/index/IndexConfig.scala
clee704: @pirz +1 for keeping included/excluded columns in a single class, because what really matters at the end of the day is which columns are included for the covering index. Included/excluded column spec from user is merely a mean to determine those final included columns.

Concerning the possibility of invalid spec (both included/excluded columns are set), how about making separate classes for each case?

```scala
sealed trait IncludedColumnsConfig
case class IncludeColumns(columns: Seq[String]) extends IncludedColumnsConfig
case class ExcludeColumns(columns: Seq[String]) extends IncludedColumnsConfig
```

Note that the imperative form is used in the names because they are sort of telling Hyperspace that it should "include" the specified columns only, or "exclude" the specified columns and must include all other columns.

PR #310, opened 2021-01-06
Fix for setting COMMON_SOURCE_SIZE_IN_BYTES tag 

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

This PR is a quick fix for #164.

Bug fix - COMMON_SOURCE_SIZE_IN_BYTES should be set with the given plan.

Yes, the bug above is fixed.

Unit test

### DIFF
--- a/src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
@@ -121,11 +121,10 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
       val df = spark.range(1, 5).toDF("id")
       val dataPath = tempPath.getAbsolutePath
       df.write.parquet(dataPath)
-      var expectedCommonSourceBytes = 0L
 
       withIndex("index1") {
         val readDf = spark.read.parquet(dataPath)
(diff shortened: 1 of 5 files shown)


src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -147,6 +149,10 @@ class RuleUtilsTest extends HyperspaceRuleTestSuite with SQLHelper {
               assert(
                 indexes.head.getTagValue(plan, IndexLogEntryTags.HYBRIDSCAN_REQUIRED)
                   === expectedHybridScanTag)
+              assert(
+                !hybridScanEnabled || indexes.head
+                  .getTagValue(plan, IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES) === Some(
+                  expectedCommonSourceBytes))
```
imback82, line 155: Instead of using `var` and capturing the variable as a closure, can we just pass it to `verify` and create a val for each scenario (if necessary)?

imback82 (commented):
LGTM (one comment), thanks @sezruby!

imback82:
Can you rename the title with a more descriptive one?

imback82:
Bt the way, can you also rename `setCommonBytesTags` to `setCommonSourceSizeInBytesTag` in this PR?

imback82 (approved):

Issue #67, opened 2020-06-24
Use a clock trait in index cache


pirz:
### Describe the issue
Add clock trait to the code and modify [`CreationTimeBasedIndexCache`](https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/index/CachingIndexCollectionManager.scala) to use the clock with System time by default. Tests cases under [`IndexCacheTest`](https://github.com/microsoft/hyperspace/blob/master/src/test/scala/com/microsoft/hyperspace/index/IndexCacheTest.scala) should be modified to not use `Thread.sleep`.

### To Reproduce
NA

### Expected behavior
`CreationTimeBasedIndexCache` should be modified to use a clock.

`Thread.sleep` usage should be removed from index cache test cases.

clee704:
Is it still open?

sezruby:
closed by #313

PR #313, opened 2021-01-06
Use Hadoop yarn clock utils for all time based calculations

thugsatbay:
Use hadoop yarn utils clock for all time based calculations, also makes it easy during testing.

 - **Tracking Issue**:[Issue67](https://github.com/microsoft/hyperspace/issues/67)
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

Just using inbuilt utils of hadoop lib for getting time.

None

Unit tests were modified to test changes.


src/main/scala/com/microsoft/hyperspace/index/CachingIndexCollectionManager.scala
```diff
@@ -16,6 +16,8 @@
 
 package com.microsoft.hyperspace.index
 
+import org.apache.hadoop.yarn.util.Clock
+import org.apache.hadoop.yarn.util.SystemClock
```
imback82, line 20: unused import

src/main/scala/com/microsoft/hyperspace/index/CachingIndexCollectionManager.scala
```diff
@@ -40,11 +42,11 @@ class CachingIndexCollectionManager(
...
-    extends IndexCollectionManager(
-      spark,
-      indexLogManagerFactory,
-      indexDataManagerFactory,
-      fileSystemFactory) {
+  extends IndexCollectionManager(
+    spark,
+    indexLogManagerFactory,
+    indexDataManagerFactory,
+    fileSystemFactory) {
```
imback82, line 49: Can you use the scalafmt to format: https://github.com/microsoft/hyperspace/blob/master/docs/coding-guidelines/scala-coding-style.md? Applies to other files.

src/main/scala/com/microsoft/hyperspace/index/CachingIndexCollectionManager.scala
```diff
@@ -119,7 +121,9 @@ object CachingIndexCollectionManager {
  * Cache entry is stale if it has been in the cache for some (configurable) time.
  * @param spark Spark session
  */
-class CreationTimeBasedIndexCache(spark: SparkSession) extends Cache[Seq[IndexLogEntry]] {
+class CreationTimeBasedIndexCache(spark: SparkSession, systemClock: Clock)
```
imback82, line 124: we can just name this as `clock` instead of `systemClock`.

src/main/scala/com/microsoft/hyperspace/index/IndexCacheFactory.scala
```diff
@@ -25,14 +26,20 @@ object IndexCacheType {
 }
 
 trait IndexCacheFactory {
-  def create(spark: SparkSession, cacheType: String): Cache[Seq[IndexLogEntry]]
+  def create(spark: SparkSession, cacheType: String, clock: Clock = new SystemClock)
```
imback82, line 29: Shall we remove default param `= new SystemClock` from this?

src/main/scala/com/microsoft/hyperspace/index/IndexCacheFactory.scala
```diff
@@ -25,14 +26,20 @@ object IndexCacheType {
...
+    : Cache[Seq[IndexLogEntry]]
 }
 
 object IndexCacheFactoryImpl extends IndexCacheFactory {
-  override def create(spark: SparkSession, cacheType: String): Cache[Seq[IndexLogEntry]] = {
-    cacheType match {
-      case IndexCacheType.CREATION_TIME_BASED => new CreationTimeBasedIndexCache(spark)
-      case _ => throw HyperspaceException(s"Unknown cache type: $cacheType.")
-    }
+  override def create(spark: SparkSession, cacheType: String, clock: Clock = new SystemClock)
```
imback82, line 34: Can we also remove default from here and pass `SystemClock` from `CachingIndexCollectionManager` so that it's easier to track which one is providing the specific impl.

src/test/scala/com/microsoft/hyperspace/index/IndexCacheTest.scala
```diff
@@ -166,18 +165,30 @@ class IndexCacheTest extends SparkFunSuite with SparkInvolvedSuite {
   }
 }
 
+class MockClock(startTime: Long = 1L) extends Clock {
```
imback82, line 168: Can you add a small description of this class similar to `MockIndexCacheFactoryImpl` below?

src/test/scala/com/microsoft/hyperspace/index/IndexCacheTest.scala
```diff
@@ -166,18 +165,30 @@ class IndexCacheTest extends SparkFunSuite with SparkInvolvedSuite {
...
+  private var time: Long = startTime
+
+  override def getTime: Long = time
+
+  def moveTime(value: Long): Unit = {
```
imback82, line 173: nit: `def advance`?

imback82: Also, since we don't use `startTime`, we can just simplify as:
```scala
class MockClock(private var curTime: Long = 0L) extends Clock {
  override def getTime: Long = curTime

  def advance(value: Long): Unit = {
    curTime = curTime + value
  }
}
```

src/test/scala/com/microsoft/hyperspace/index/IndexCacheTest.scala
```diff
@@ -166,18 +165,30 @@ class IndexCacheTest extends SparkFunSuite with SparkInvolvedSuite {
...
  */
 class MockIndexCacheFactoryImpl(spark: SparkSession) extends IndexCacheFactory {
-  val indexCache = new CreationTimeBasedIndexCache(spark)
-
-  override def create(spark: SparkSession, cacheType: String): Cache[Seq[IndexLogEntry]] = {
-    cacheType match {
-      case IndexCacheType.CREATION_TIME_BASED => indexCache
-      case _ => throw HyperspaceException(s"Unknown cache type: $cacheType.")
+  private val clock: Clock = new MockClock
+  val indexCache = new CreationTimeBasedIndexCache(spark, clock)
```
imback82, line 185: nit:
```suggestion
  val indexCache = new CreationTimeBasedIndexCache(spark, new MockClock)
```

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -36,7 +36,7 @@ class Hyperspace(spark: SparkSession) {
   /**
    * Create index.
    *
-   * @param df the DataFrame object to build index on.
+   * @param df          the DataFrame object to build index on.
```
imback82, line 39: revert?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -84,7 +84,7 @@ class Hyperspace(spark: SparkSession) {
    * refresh modes as listed below.
    *
    * @param indexName Name of the index to refresh.
-   * @param mode Refresh mode. Currently supported modes are `incremental` and `full`.
+   * @param mode      Refresh mode. Currently supported modes are `incremental` and `full`.
```
imback82, line 87: ditto and other places below.

imback82 (commented):
@thugsatbay Can you first go over the changes and make sure unnecessary formatting changes are not introduced? I see too many to review. Thanks.

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
thugsatbay: Hm, interesting, this change was introduced by scalafmt, I didn't even touch this. Let me see why it is doing that. Sorry about the formatting trouble.

thugsatbay: I have manually resolved the formatting, however it seems scalafmt is consistently assigning tab spaces for parameter definitions to align content in a tabular format.

thugsatbay: At this point I don't know what specific standard we are following or am I using scalafmt in an incorrect manner, any help in this would be appreciated. Thank you.

thugsatbay: @apoorvedave1 helped me out, I will update the doc later it wasn't straightforward for me. Thanks !

src/main/scala/com/microsoft/hyperspace/index/IndexCacheFactory.scala
```diff
@@ -16,6 +16,7 @@
 
 package com.microsoft.hyperspace.index
 
+import org.apache.hadoop.yarn.util.SystemClock
```
imback82, line 19: Is this OK to use? I see the following from https://hadoop.apache.org/docs/stable2/api/org/apache/hadoop/yarn/util/SystemClock.html#getTime(): 
`NOTE: Do not use this to calculate a duration of expire or interval to sleep`

imback82 (commented):
LGTM (one question), thanks @thugsatbay!

@sezruby Can you also take a look?

src/main/scala/com/microsoft/hyperspace/index/IndexCacheFactory.scala
thugsatbay: SystemClock in hadoop internally implementation uses `System.currentTimeMillis()` which is what was used before.

But also what it says is correct, an admin can come and change time of machine which could give wrong computation for our caching. We should be running internal clocks for caching, and can actually do that as a separate change ?

imback82: > SystemClock in hadoop internally implementation uses `System.currentTimeMillis()` which is what was used before.

Ah OK. In that case we can just leave it as it is.

thugsatbay: Can use something like this https://guava.dev/releases/21.0/api/docs/com/google/common/base/Ticker.html

imback82: We can just use `MonotonicClock` as the above doc suggests: https://hadoop.apache.org/docs/stable2/api/org/apache/hadoop/yarn/util/MonotonicClock.html.

thugsatbay: Ah, MonotonicClock only comes into hadoop utils from 2.8 and we are using 2.65 package, can I update it to 2.8, I don't know how much backward compatibility hyperspace can handle with hadoop git update ?

imback82: For now, let's just leave it as `SystemClock`.

src/test/scala/com/microsoft/hyperspace/index/IndexCacheTest.scala
```diff
@@ -157,13 +156,26 @@ class IndexCacheTest extends SparkFunSuite with SparkInvolvedSuite {
   }
 }
 
+/**
+ * Mock for testing purposes so we can validate and invalidate entries based on time.
+ * @param time long value that tells current time
```
sezruby, line 161: nit:

```suggestion

 * @param time Current time.
```

sezruby (commented):
LGTM Thanks @thugsatbay!

src/test/scala/com/microsoft/hyperspace/index/IndexCacheTest.scala
imback82: Applied the suggestion.

imback82 (approved):

PR #314, opened 2021-01-07
Fix the usage of getting the commons source size tag in JoinIndexRanker 

imback82:

Fixing the issue where incorrect logical plan is passed when retrieving the tag for common source size.

Fixing a bug

No

Existing test, but prob. need to add a new test.


src/main/scala/com/microsoft/hyperspace/index/rankers/JoinIndexRanker.scala
```diff
@@ -56,7 +56,7 @@ object JoinIndexRanker {
       indexPairs: Seq[(IndexLogEntry, IndexLogEntry)]): Seq[(IndexLogEntry, IndexLogEntry)] = {
     val hybridScanEnabled = HyperspaceConf.hybridScanEnabled(spark)
     def getCommonBytes(logicalPlan: LogicalPlan, index: IndexLogEntry): Long = {
-      index.getTagValue(leftChild, IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES).getOrElse(0L)
+      index.getTagValue(logicalPlan, IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES).getOrElse(0L)
```
imback82, line 59: @sezruby is this a bug?

sezruby: Yes, thanks for the fix!

PR #315, opened 2021-01-07
Clean up unused import/code by enabling -Ywarn-unused in scalacOptions

imback82:

After finding a bug by an unused variable, I enabled `-Ywarn-unused` in`scalacOptions` temporarily and cleaned unused import/default param/variable/etc. Note that I didn't remove unused imports if they are used in comments as `[[ClassName]]`, but this needs to be discussed separately since I feel that turning on `-Ywarn-unused` outweighs supporting click-thru for `[[ClassName]]`.

Btw, we cannot make warning as error since there are deprecation usage warning, which is from dependent packages.

Remove unused import/code.

No.

Existing tests.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -47,7 +47,8 @@ assemblyMergeStrategy in assembly := {
 }
 
 scalacOptions ++= Seq(
-  "-target:jvm-1.8"
+  "-target:jvm-1.8",
+  "-Ywarn-unused"
 )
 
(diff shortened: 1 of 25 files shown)


sezruby (approved):
LGTM thanks!

imback82 (commented):
Thanks @sezruby for the review!

PR #325, opened 2021-01-19
Remove logical plan serde related code

imback82:

Logical plans were serialized and stored in index metdata, but we have moved away from supporting this due to the compatibility issues between Spark versions. Either we can come up with a generic representation of all the logical plans or take the shim approach similar to spark-rapids.

This PR proposes to remove dead code.

No

Existing tests.


sezruby:
LGTM Thanks!

imback82 (commented):
Thanks @sezruby!

PR #321, opened 2021-01-11
Support DataSourceV2 sources

andrei-ionescu:
### What is the context for this pull request?

 - **Tracking Issue**: #306
 - **Proposal**: #318
 - **Dependencies**: N/A
 - Fixes: #306
 - Fixes: #318  

### What changes were proposed in this pull request?

This PR adds support for DataSourceV2.

The following changes are in this PR and each of them are separate commits:

- Use `LogicaPlan` instead of `LogicalRelation`. This gives us the possibility to add support for other kinds of relations like Spark datas source version 2.
- Add support for `DataSourceV2`.

### Does this PR introduce _any_ user-facing change?

The source interfaces has changed. Now instead of using `LogicalRelation` as parameter type it now uses `LogicalPlan`. There has been added support for `DataSourceV2` sources.

Detailed information can be found in the #318 proposal.

### How was this patch tested?

1. All already present tests are passing
2. Locally & Databricks Runtime tests based on the [Hyperspace usage API in Apache Spark](https://microsoft.github.io/hyperspace/#hyperspace-usage-api-in-apache-spark).

PR #320, opened 2021-01-09
Support Iceberg table format

andrei-ionescu:
### What is the context for this pull request?

 - **Tracking Issue**: #306
 - **Proposal**: #318
 - **Dependencies**: #321
 - Fixes: #306
 - Fixes: #318

### What changes were proposed in this pull request?

This PR adds support for Iceberg.

The following changes are in this PR and each of them are separate commits:

- Add Iceberg source.
- Add support for incremental refresh. This is based on #301 PR from @sezruby.
- Add integration test

### Does this PR introduce _any_ user-facing change?

No. The main changes to user-facing APIs are in the #321 PR. Detailed information can be found in the #318 proposal.

### How was this patch tested?

1. Integration test added for the new functionality
2. Locally & Databricks Runtime tests

- Local build

```
sbt publishLocal
```

- Run Spark shell with Hyperspace and Iceberg libraries loaded

```
$ spark-shell \
--driver-memory 4g \
--packages "com.microsoft.hyperspace:hyperspace-core_2.11:0.4.0-SNAPSHOT,org.apache.iceberg:iceberg-spark-runtime:0.10.0" \
--driver-java-options "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5006 -XX:+UseG1GC -Dlog4j.debug=true"
```

- Paste the following code

```
import org.apache.spark.sql._
import com.microsoft.hyperspace._
import com.microsoft.hyperspace.index._
import scala.collection.JavaConverters._
import org.apache.iceberg.PartitionSpec
import org.apache.iceberg.TableProperties
import org.apache.iceberg.spark._
import org.apache.iceberg.hadoop._

val hs = new Hyperspace(spark)

// create Iceberg table
val props = Map(TableProperties.WRITE_NEW_DATA_LOCATION -> "table3").asJava
val sourceDf = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
val schema = SparkSchemaUtil.convert(sourceDf.schema)
val part = PartitionSpec.builderFor(schema).build()
val icebergTable = new HadoopTables().create(schema, part, props, "table3")
sourceDf.write.mode("overwrite").format("iceberg").save("./table3")

// read created table
val iceDf = spark.read.format("iceberg").load("./table3")

// create indexes
hs.createIndex(iceDf, IndexConfig("index_ice0", indexedColumns = Seq("id"), includedColumns = Seq("name")))
[...]

PR #329, opened 2021-01-20
Add config to use bucketed scan for filter indexes

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Add a config for filter index rule to apply bucketing information when reading the index data.
```scala
  // Config used to set bucketSpec for Filter Index. If bucketSpec is used, Spark can prune
  // not applicable buckets, so we could read less files in case of a high selectivity query.
```

* if the value is true
  * Give bucketSpec to Spark so when reading the index data, Spark prunes if each bucket is not applicable. This can reduce the number of files to read for the query.
* else
  * spark will generate splits based on https://github.com/apache/spark/blob/2145f07e22a53e2da2f6a424c6fd169c2bb63a48/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FilePartition.scala#L86  (usually less number of buckets) but each bucket is not pruned.

Yes, plan is changed a bit for filter indexes if the config is true.

Unit test


src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -47,6 +47,11 @@ object IndexConstants {
     "spark.hyperspace.index.hybridscan.maxAppendedRatio"
   val INDEX_HYBRID_SCAN_APPENDED_RATIO_THRESHOLD_DEFAULT = "0.3"
 
+  // Config used to set bucketSpec for Filter Index. If bucketSpec is used, Spark can prune
```
imback82, line 50: nit: Filter rule?

imback82 (commented):
I will check back when the test is added.

imback82 (commented):
What's the plan for this PR vs. #332? This PR wants to enable bucketing so that bucket pruning can kick in; but it may regress if bucket pruning does not kick in. Is it possible to do some check if bucket pruning would kick in or not? If this is not feasible, maybe introduce a strategy to look at the physical plan?

sezruby:
The config is disabled by default. 2 main points:
- Bucket union handling
- Make it switchable option so if a user would like to force it, then they can use this config
  - other than bucket pruning, bucketing might be beneficial if there's a sort merge join which are not covered by indexes.

For #332, I'd like filter the file paths only using the bucket pruning logic, and after that setting bucket spec could be a different option depending on their workloads.

We might be able to detect such SMJ and setting bucket spec or not - as filter indexes are the last rule.

imback82:
> * Bucket union handling

Seems like a byproduct of introducing the new config so that we can always disable bucket union for filter rule (I see `useBucketUnionForAppended = false` for filter rule)? What happens if we enable bucket union for filter rule if bucket spec is enabled?

> * other than bucket pruning, bucketing might be beneficial if there's a sort merge join which are not covered by indexes.

Hmm, if this is the case (benefits SMJ), shouldn't join rule have already kicked in?

> For #332, I'd like to exclude index data files only, using the bucket pruning logic. After that, setting bucket spec could be a different option depending on their workloads.

Yea, excluding index data files makes sense. But we are going introduce "additional" option for setting the bucket spec? It's usually better for Spark to decide how to distribute the tasks (instead of limiting it to the number of buckets). @apoorvedave1 has done some benchmarks on this.

imback82:
Btw, I am +1 for having this config, but just have few questions.

sezruby:
> Seems like a byproduct of introducing the new config so that we can always disable bucket union for filter rule (I see useBucketUnionForAppended = false for filter rule)? What happens if we enable bucket union for filter rule if bucket spec is enabled?

Oh I forgot to disable bucket union for Filter 🤔.. 
Yeah if we could extract the join and apply the index properly with join rule(+v2) then that would be the best.

> But we are going introduce "additional" option for setting the bucket spec? It's usually better for Spark to decide how to distribute the tasks (instead of limiting it to the number of buckets). 
If we don't set the bucket spec, spark will create splits with the input files - 
https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/DataSourceScanExec.scala#L594

One benefit using bucketSpec is that we could enforce each file will be handled separately by an executor.
This might be good or bad depending on the environment/workloads, but for cache thing, this would help. 
(though I didn't measure the performance after filtering the file path + no bucket spec)

We could use this config internally, and then decide to expose this or not later? WDYT?

imback82:
> If we don't set the bucket spec, spark will create splits with the input files by bin packing -
> https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/DataSourceScanExec.scala#L594

Yes, but from my experience, this works out better by utilizing the full cluster than limiting the number of tasks to the number of buckets.

> We could use this config internally, and then decide to expose this or not later? WDYT?

Yes, I think we can have this config and document the behavior clearly.

andrei-ionescu:
I do like this new feature even if added as an undocumented flag. I did some tests over big datasets (billions of records) and having the index on a high cardinality field without bucketing and bucket pruning is not adding any performance improvement. Bucketing (and maybe partitioning by the indexed column) will add the benefit of pruning.

I'm looking forward for such a feature.

rapoth:
> I do like this new feature even if added as an undocumented flag. I did some tests over big datasets (billions of records) and having the index on a high cardinality field without bucketing and bucket pruning is not adding any performance improvement. Bucketing (and maybe partitioning by the indexed column) will add the benefit of pruning.
> 
> I'm looking forward for such a feature.

@andrei-ionescu This is very interesting. Do you have any more information (e.g., workload, numbers) you could share? Also, was it a single query or were you running a concurrent workload?

I do agree with supporting partitioning though (I'm assuming you are referring to low cardinality columns). We haven't been able to get to working on partitioning yet (but it's definitely I've heard from multiple customers). It also has several other benefits like being able to support advanced scenarios like index retention (e.g., if partitioned by timestamp, we could do clean-up very easily - in the current implementation, it'd have to rewrite the index to some large degree) and I guess we can also leverage dynamic partition pruning. :)

andrei-ionescu:
@rapoth Here are some details:

## Query

```
val sql = s"""
  SELECT   ts.timestamp
  FROM     ts 
  WHERE    ts.timestamp >= to_timestamp('2020-03-17')
  AND      ts.timestamp < to_timestamp('2020-03-18')
  LIMIT    1000
"""
```

Executed with: 

```
spark.sql(sql).collect
```

## Dataset

- schema size is about 20 top fields and about 17 of these are heavily nested
- about `14 Billion` rows
- the `timestamp` field is of `timestamp` type and is up to seconds
- the format is Iceberg

## Index

```
hs.createIndex(
  ts, 
  IndexConfig(
    "idx_ts3", 
    indexedColumns = Seq("timestamp"), 
    includedColumns = Seq("ns", "id")))
```

The index has:
- `434GB` file size
- `200` files

## Explained query

```
Plan with indexes:
CollectLimit 1000
+- *(1) Project [timestamp#207]
   +- *(1) Filter ((isnotnull(timestamp#207) && (timestamp#207 >= 1584403200000000)) && (timestamp#207 < 1584489600000000))
      <----+- *(1) FileScan Hyperspace(Type: CI, Name: idx_ts3, LogVersion: 1) [timestamp#207] Batched: true, DataFilters: [isnotnull(timestamp#207), (timestamp#207 >= 1584403200000000), (timestamp#207 < 1584489600000000)], Format: Parquet, Location: InMemoryFileIndex[dbfs:/u.../spark-warehouse/indexes/idx_ts3/v__=0/part-00000-tid-451174797136..., PartitionFilters: [], PushedFilters: [IsNotNull(timestamp), GreaterThanOrEqual(timestamp,2020-03-17 00:00:00.0), LessThan(timestamp,20..., ReadSchema: struct<timestamp:timestamp>---->

Plan without indexes:
CollectLimit 1000
+- *(1) Project [timestamp#207]
   +- *(1) Filter ((isnotnull(timestamp#207) && (timestamp#207 >= 1584403200000000)) && (timestamp#207 < 1584489600000000))
      <----+- *(1) ScanV2 iceberg[timestamp#207] (Filters: [isnotnull(timestamp#207), (timestamp#207 >= 1584403200000000), (timestamp#207 < 1584489600000000)], Options: [...)---->

Indexes used:
idx_ts3:/.../spark-warehouse/indexes/idx_ts3/v__=0

Physical operator stats:
+--------------------------------------------------------+-------------------+------------------+----------+
|                                       Physical Operator|Hyperspace Disabled|Hyperspace Enabled|Difference|
+--------------------------------------------------------+-------------------+------------------+----------+
|                                       *DataSourceV2Scan|                  1|                 0|        -1|
|*Scan Hyperspace(Type: CI, Name: idx_ts3, LogVersion: 1)|                  0|                 1|         1|
|                                            CollectLimit|                  1|                 1|         0|
|                                                  Filter|                  1|                 1|         0|
|                                                 Project|                  1|                 1|         0|
|                                       WholeStageCodegen|                  1|                 1|         0|
+--------------------------------------------------------+-------------------+------------------+----------+
```

## The cluster

I did run the experiment on Databricks cluster with the following details:
- driver: `64` cores `432GB` memory
- `6` workers: `32` cores `256GB` memory
- Spark version `2.4.5`

## Results

Time to get the `1000` rows:
- with Hyperspace is `17.24s`
- without Hyperspace is `16.86s`

Hope all these helps.

imback82:
Thanks @andrei-ionescu for the info.

I guess one solution to this is to allow transformation of the indexed keys (bucketing to minutes/hours instead of seconds, for example).

andrei-ionescu:
Even if we don't modify the timestamp, I'm having in average about 2K of the same timestamp value in the index. Bucketing or partitioning just by the values will reduce the time for query tremendously.

For example, instead of having the index dataset layer out like this: 

```
/.../spark-warehouse/indexes/idx_ts3/v__=0/.....parquet
```

I would suggest the following:

```
/.../spark-warehouse/indexes/idx_ts3/v__=0/timestamp=2020-03-17T00:03:07
```

rapoth:
@andrei-ionescu Would you mind opening a feature request with your original benchmark? You are bringing up a very interesting topic of indexing for timeseries data (I did not know you were looking at timestamp predicates :)). Alternately, I can copy some of this discussion and create one. Please let me know which one you prefer. I have some follow-up questions but I will ask them in the new issue.

We are also heavily interested in building specialized indexes for timeseries data so it is definitely awesome you have the same scenarios!

imback82:
> I would suggest the following:
> 
> ```
> /.../spark-warehouse/indexes/idx_ts3/v__=0/timestamp=2020-03-17T00:03:07/.....parquet
> ```

Hive-partitioning was explored before but abandoned due to the fact that we need to create bucket files for each partition  and wasn't scalable in our scenario.

But now that we have a specific use case, we can explore this again (prob. in the form of specialized index).

andrei-ionescu:
@imback82 I'm proposing to add `.partitionBy(resolvedIndexedColumns: _*).` in between `write` and `parquet` similar to this:

```
.repartition(resolvedIndexedColumns.map(df(_)): _*).write.partitionBy(resolvedIndexedColumns: _*).parquet(...)
```

somewhere around this place https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala#L129-L139.

This can be just a flag as in cases of high cardinality it may throw out a lot of folders/partitions. 

We can go even a step further and detect it and choose the best approach.

andrei-ionescu:
@rapoth I'll open up a feature request as you suggest but currently I'm focused on the following things: **DataSourceV2 support** (#321), **Iceberg support** (#320) and the **Nested fields support** (#347).

rapoth:
@andrei-ionescu  I apologize. I only meant to open up a new thread so we can continue the conversation there (I did not mean to say we'd work on it immediately).

andrei-ionescu:
@rapoth One more thing, I don't think it really matter if is a time series dataset. In any dataset that has lots of data and the resulted index is also massive adding this option will bring a lot of benefits.

rapoth:
@andrei-ionescu Fair point. I recently talked with a lot of customers who had timeseries data specifically so seeing your comment on timestamps made me forget every other use case 🦖

andrei-ionescu:
@rapoth Here is the feature request: #351.

imback82:
> @imback82 I'm proposing to add `.partitionBy(resolvedIndexedColumns: _*).` in between `write` and `parquet` similar to this:

Yes, that's the hive partitioning I was referring to.

@apoorvedave1 has done some prototyping on this and write wasn't really scaling. So you may want to try it first on your dataset as well.

src/test/scala/com/microsoft/hyperspace/index/HybridScanSuite.scala
```diff
@@ -460,12 +461,77 @@ trait HybridScanSuite extends QueryTest with HyperspaceSuite {
...
+
+        checkAnswer(baseQuery, filter)
+      }
+    }
+  }
+
+  test(
+    "Append-only: filter rule and non-parquet format," +
+      "appended data should be shuffled and merged by Union even with bucketSpec.") {
+    // Note: for delta lake, this test is also eligible as the dataset is partitioned.
```
imback82, line 484: How should we handle this "Note"? This sounds like a TODO?

imback82 (commented):
LGTM (few nits/questions), thanks @sezruby!

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -62,7 +62,8 @@ object FilterIndexRule
                   spark,
                   index,
                   originalPlan,
-                  useBucketSpec = false)
+                  useBucketSpec = HyperspaceConf.useBucketSpecForFilterRule(spark),
+                  useBucketUnionForAppended = false)
```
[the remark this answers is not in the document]
sezruby: It might be beneficial for later ops which requires bucketing, but just for filter index, we don't need it. 
I think it's better to write a new rule for the cases if needed.

src/test/scala/com/microsoft/hyperspace/index/HybridScanSuite.scala
sezruby: It's note because this test is already included in delta lake hybrid scan tests.

imback82: Shall we just remove the comment then? (not sure about the importance)

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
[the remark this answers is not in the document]
imback82: Shall we add the comment to the code (since it's not straightforward to understand)?

imback82 (approved):

PR #330, opened 2021-01-21
Remove deleted threshold condition for append-only Hybrid Scan

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

For append-only hybrid scan, the ratio threshold of deleted files shouldn't matter. 
Previously, we have different config for Hybrid Scan (append only & append + delete) but now we only have 1 config for Hybrid Scan, we don't need to check the delete threshold.
The condition blocks to apply the index if hybrid scan is enabled & maxDeletedRatio config is non-zero.

Yes, fixes the bug described above.

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -101,10 +101,8 @@ object RuleUtils {
         appendedBytesRatio < HyperspaceConf.hybridScanAppendedRatioThreshold(spark) &&
         deletedBytesRatio < HyperspaceConf.hybridScanDeletedRatioThreshold(spark)
 
-
       // For append-only Hybrid Scan, deleted files are not allowed.
-      lazy val isAppendOnlyCandidate = !hybridScanDeleteEnabled && deletedCnt == 0 &&
-        commonCnt > 0 &&
+      lazy val isAppendOnlyCandidate = deletedCnt == 0 && commonCnt > 0 &&
         appendedBytesRatio < HyperspaceConf.hybridScanAppendedRatioThreshold(spark)
 
       val isCandidate = isAppendAndDeleteCandidate || isAppendOnlyCandidate
--- a/src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
@@ -136,7 +136,7 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
             plan: LogicalPlan,
             hybridScanEnabled: Boolean,
             hybridScanDeleteEnabled: Boolean,
-            numExpectCandidateIndex: Int,
+            numExpectedCandidateIndex: Int,
             expectedHybridScanTag: Option[Boolean],
             expectedCommonSourceBytes: Option[Long]): Unit = {
           withSQLConf(
@@ -145,16 +145,19 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
             IndexConstants.INDEX_HYBRID_SCAN_DELETED_RATIO_THRESHOLD ->
               (if (hybridScanDeleteEnabled) "0.99" else "0")) {
             val indexes = RuleUtils
-              .getCandidateIndexes(spark, allIndexes, plan).sortBy(_.name)
-            if (numExpectCandidateIndex > 0) {
-              assert(indexes.length === numExpectCandidateIndex)
+              .getCandidateIndexes(spark, allIndexes, plan)
+              .sortBy(_.name)
+            if (numExpectedCandidateIndex > 0) {
+              assert(indexes.length === numExpectedCandidateIndex)
               assert(indexes.head.name === "index1")
-              assert(
-                indexes.head.getTagValue(plan, IndexLogEntryTags.HYBRIDSCAN_REQUIRED)
-                  === expectedHybridScanTag)
-              assert(
-                indexes.head.getTagValue(plan, IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES)
-                  === expectedCommonSourceBytes)
+              indexes.foreach { index =>
+                assert(
+                  index.getTagValue(plan, IndexLogEntryTags.HYBRIDSCAN_REQUIRED)
+                    === expectedHybridScanTag)
+                assert(
+                  index.getTagValue(plan, IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES)
+                    === expectedCommonSourceBytes)
+              }
             } else {
               assert(indexes.isEmpty)
             }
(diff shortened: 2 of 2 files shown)


src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -101,10 +101,8 @@ object RuleUtils {
         appendedBytesRatio < HyperspaceConf.hybridScanAppendedRatioThreshold(spark) &&
         deletedBytesRatio < HyperspaceConf.hybridScanDeletedRatioThreshold(spark)
 
-
       // For append-only Hybrid Scan, deleted files are not allowed.
-      lazy val isAppendOnlyCandidate = !hybridScanDeleteEnabled && deletedCnt == 0 &&
-        commonCnt > 0 &&
+      lazy val isAppendOnlyCandidate = deletedCnt == 0 && commonCnt > 0 &&
```
imback82, line 105: The fix doesn't seem to align with the PR description: "For append-only hybrid scan, the threshold of deleted files doesn't matter."?

Could you also update the comment (line 104) to reflect the change (regarding the `hybridScanDeleteEnabled`)?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -211,21 +214,21 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
...
-            expectCandidateIndex = false,
+            numExpectCandidateIndex = 0,
             expectedHybridScanTag = None,
             expectedCommonSourceBytes = None)
           verify(
             optimizedPlan,
             hybridScanEnabled = true,
             hybridScanDeleteEnabled = true,
-            expectCandidateIndex = true,
+            numExpectCandidateIndex = 1,
```
imback82, line 231: Question: Is the fix affecting this scenario only or other scenarios as well?

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -121,19 +121,22 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
...
+          indexManager.create(readDf, IndexConfig("index2", Seq("id")))
+        }
         val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
 
         def verify(
             plan: LogicalPlan,
             hybridScanEnabled: Boolean,
             hybridScanDeleteEnabled: Boolean,
-            expectCandidateIndex: Boolean,
+            numExpectCandidateIndex: Int,
```
imback82, line 139: nit: `numExpectedCandidateIndexes`

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -142,9 +145,9 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
             IndexConstants.INDEX_HYBRID_SCAN_DELETED_RATIO_THRESHOLD ->
               (if (hybridScanDeleteEnabled) "0.99" else "0")) {
             val indexes = RuleUtils
-              .getCandidateIndexes(spark, allIndexes, plan)
-            if (expectCandidateIndex) {
-              assert(indexes.length === 1)
+              .getCandidateIndexes(spark, allIndexes, plan).sortBy(_.name)
+            if (numExpectCandidateIndex > 0) {
+              assert(indexes.length === numExpectCandidateIndex)
               assert(indexes.head.name === "index1")
```
imback82, line 151: no need to check `index2`?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: I think the comment (line 104) is still valid - `deletedCnt == 0 ` 

Yes it doesn't matter but shouldn't be considered.. I updated the PR.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
sezruby: Because I always enabled the lineage for all hybrid scan tests, it doesn't affect other test scenarios.
(If there's a such test, it would have detected this bug...)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: I am still confused with the PR description: "we don't need to check the delete threshold." Where do we check the delete threshold for this change?

sezruby: It's because `hybridScanDeleteEnabled` is 
```scala
def hybridScanDeleteEnabled(spark: SparkSession): Boolean = {
    hybridScanDeletedRatioThreshold(spark) > 0.0
  }
```

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -143,15 +146,18 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
               (if (hybridScanDeleteEnabled) "0.99" else "0")) {
             val indexes = RuleUtils
               .getCandidateIndexes(spark, allIndexes, plan)
-            if (expectCandidateIndex) {
-              assert(indexes.length === 1)
+              .sortBy(_.name)
+            if (numExpectedCandidateIndex > 0) {
+              assert(indexes.length === numExpectedCandidateIndex)
               assert(indexes.head.name === "index1")
```
imback82, line 152: Should we just check all the names? `assert(indexes.map(_.name) === (1 to numExpectedCandidateIndexes).map("index" + _))` (didn't test compiling)

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: Ah sorry. My local repo had older version :)

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
imback82: nit: missed `es`?

imback82 (dismissed):
LGTM, thanks @sezruby!

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
imback82: or it may just be clear to pass the name of expected indexes instead of numExpectedCandidateIndexes: `expectedCandidateIndexNames: Seq[String]`

imback82 (approved):

PR #332, opened 2021-01-25
Support adaptive bucketed scan for FilterIndexRule

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: #329 

Bucketed scan can prune unused buckets for the given query.
A switchable config is proposed in #329, however, it might be difficult to change the config from user-side, for each filter query.

This PR provides adaptive bucketSpec application for FilterIndexRule as following:
- generate logical plan with bucketSpec
- check the number of selected buckets => using `FileSourceStrategy`
- if the number is smaller than `total bucket number * 0.8`, apply the bucketSpec.
- otherwise, create a new plan w/o bucketSpec and return it
```scala
               val fileSourceScanNodes = if (transformedPlanWithBucketSpec.isInstanceOf[Union]) {
                  FileSourceStrategy(transformedPlanWithBucketSpec.children.head)
                } else {
                  FileSourceStrategy(transformedPlanWithBucketSpec)
                }
                val foundPrunedBuckets = fileSourceScanNodes.head.collect {
                  case _ @FileSourceScanExec(_, _, _, _, optionalBucketSet, _, _)
                    if optionalBucketSet.isDefined && (optionalBucketSet.get
                      .cardinality() * 1.0 < index.bucketSpec.numBuckets * 0.8) =>
                    optionalBucketSet.get.cardinality()
                }
```

Currently, there's a bug in bucket pruning for `Filter(InSet, _))`.
I'll add a new strategy to fix the issue.

Yes, filter index can be applied with or without bucket spec according to the selectivity for each query.

Unit test


src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -56,13 +58,46 @@ object FilterIndexRule
...
+                    useBucketUnionForAppended = false)
+                val fileSourceScanNodes = if (transformedPlanWithBucketSpec.isInstanceOf[Union]) {
+                  FileSourceStrategy(transformedPlanWithBucketSpec.children.head)
+                } else {
+                  FileSourceStrategy(transformedPlanWithBucketSpec)
+                }
+                val foundPrunedBuckets = fileSourceScanNodes.head.collect {
+                  case _ @FileSourceScanExec(_, _, _, _, optionalBucketSet, _, _)
+                    if optionalBucketSet.isDefined && (optionalBucketSet.get
+                      .cardinality() * 1.0 < index.bucketSpec.numBuckets * 0.8) =>
```
imback82, line 78: Can you explain how you came up with 0.8?

imback82: Is it possible to get the list of files after bucket pruning here and use it with `useBucektSpec = false`? For example, let's say only one bucket is selected. In this case, only one task will be launched to process the file (and the file can be huge). It may be better to split the files to multiple tasks.

sezruby: I think it would be possible if we extract genBucketSet 
https://github.com/apache/spark/blob/d1177b52304217f4cb86506fd1887ec98879ed16/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileSourceStrategy.scala#L116

and use it to filter the file paths before transforming the plan. 
In this way, we don't need to generate the plan twice so I think it would be good.

imback82: sgtm

sezruby: @imback82 I tried several options.
1) create InMemoryFileIndex with filtered file list - this isn't compatible with INMEMORYFILEINDEX_INDEX_ONLY tag and it incurs additional overhead from the newly created InMemoryFileIndex. I'd like to keep the current tag behavior.
2) IndexHadoopFsRelation returns filtered file list - we need to rewrite `InMemoryFileIndex` as most of spark code refers `relation.location.xxx` to get the file list.

A possible better(?) option is introducing a custom FileSourceScanExec or InMemoryFileIndex.. but it will cause some complexity in codebase.

(btw I found the related issue: https://issues.apache.org/jira/browse/SPARK-32986)

So.. how about setting bucketSpec for FilterIndexRule if bucket pruning is applicable for now?

imback82: > So.. how about setting bucketSpec for FilterIndexRule if bucket pruning is applicable for now?

Sure we can do this. How will this work with #329?

sezruby: This change is based on #329. 

I added this config - https://github.com/microsoft/hyperspace/pull/332/files#diff-90c330f74c7d2fca88819041feeb77a90a663100358955b09f7518506dd8f051R60
And if it's true, bucketSpec is applied if bucket pruning is possible.
Otherwise, it refers the config(spark.hyperspace.index.filterRule.useBucketSpec) added in #329.

imback82: sorry, maybe I wasn't clear. Is there a scenario where bucket read performs better when there are no buckets pruned?

sezruby: @imback82  if the config "bucketCheck.enabled" is enabled, bucket scan won't be performed if there's no pruned bucket.

#329 will be used as a static config, for experimental purposes?

bucket read might be faster when job scheduling does matter(e.g. cache) or as non-bucketed scan will depend on the default spark configs - https://github.com/apache/spark/blob/2145f07e22a53e2da2f6a424c6fd169c2bb63a48/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FilePartition.scala#L86
it might not work well with every dataset/landscape/env setup.

imback82: > bucket read might be faster when job scheduling does matter(e.g. cache)

Hmm. usually scheduling is very fast that it shouldn't be a bottleneck; how much (percentage) was the scheduling overhead in the cache scenario?

OK, let me finish #329 and let me think if we can remove the inter-dependent configs.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -56,6 +57,11 @@ object FilterIndexRule
             findCoveringIndexes(filter, outputColumns, filterColumns)
           FilterIndexRanker.rank(spark, filter, candidateIndexes) match {
             case Some(index) =>
+              val useBucketSpec = if (HyperspaceConf.filterRuleBucketCheckEnabled(spark)) {
```
imback82, line 60: Should we do a ratio? If we selected 180 / 200 buckets, it may not be good to fall back to use bucket spec since we are limiting parallelism further without pruning many buckets.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -52,6 +52,14 @@ object IndexConstants {
   val INDEX_FILTER_RULE_USE_BUCKET_SPEC = "spark.hyperspace.index.filterRule.useBucketSpec"
   val INDEX_FILTER_RULE_USE_BUCKET_SPEC_DEFAULT = "false"
 
+  // If this config is true, Hyperspace generates a plan with bucketSpec first and check
+  // the selectivity of the filter query by creating the physical plan in advance.
+  // If less than half number of buckets are selected, Filter Rule uses the plan with bucketSpec.
+  // Otherwise, newly generated bucketSpec is not used for Filter Rule.
```
imback82, line 58: This seems to be out of date?

src/main/scala/com/microsoft/hyperspace/util/LogicalPlanUtils.scala
```diff
@@ -35,4 +45,123 @@ object LogicalPlanUtils {
       case _ => false
     }
   }
+
+  /**
+   * BucketSelector returns the selected buckets if bucket pruning is applicable for the given
+   * query plan. The logic is extracted from [[FileSourceScanStrategy]] in Spark.
+   */
+  object BucketSelector {
```
imback82, line 53: Can we move this out to a separate file?

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -56,6 +57,11 @@ object FilterIndexRule
...
+                BucketSelector(plan, index.bucketSpec).isDefined
+              } else {
+                HyperspaceConf.useBucketSpecForFilterRule(spark)
```
imback82, line 63: This can be a confusing experience since `useBucketSpec` even when `spark.hyperspace.index.filterRule.useBucketSpec` is set to true.

How about:
```scala
if (HyperspaceConf.useBucketSpecForFilterRule(spark)) {
   // always use bucket spec
} else {
  val selectedBuckets = BucketSelector(plan, index.bucketSpec)
  selectedBuckets.isDefined && (getSelectedRatio(selectedBuckets) < HyperspaceConf. prunedBucketRatioToAutoEnableBucketRead)
}
```

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
imback82: > 1. create InMemoryFileIndex with filtered file list - this isn't compatible with INMEMORYFILEINDEX_INDEX_ONLY tag and it incurs additional overhead from the newly created InMemoryFileIndex. I'd like to keep the current tag behavior.

BTW, this seems like a better option as Spark 3.2 will also decouple bucket scan vs. bucket filter: https://github.com/apache/spark/commit/76baaf746502b50f6f466d53bdce2b5a9a35b2df

This will require some changes in `RuleUtils` to support `INMEMORYFILEINDEX_INDEX_ONLY` but seems doable?

sezruby: Yes plan + INMEMORYFILEINDEX_INDEX_ONLY will cache it properly, but different file lists might incur unnecessary overhead. Filtering files in DataSourceScanExec is clear and doable like the link.

PR #355, opened 2021-02-11
Introduce SourceRelation/FileBasedRelation traits to remove direct dependency on LogicalRelation from actions/rules

imback82:

#321 and #320 require to support `DatasourceV2Relation`, but the current provider was designed to support only the `LogicalRelation`, so adding a new provider that supports a different relation type requires lots of code changes across actions/rules.

This proposes to introduce one more abstraction to `FileBasedSourceProvider` such that each provider now needs to implement `FileBasedRelation`. For example, `DeltaLakeFileBasedSource` needs to implement `DeltaRelation` which extends `FileBasedRelation` to handle a Delta Lake specific relation.

By decoupling this, actions/rules do not depend on `LogicalRelation` directly. 

No

API refactoring. Existing tests should be enough.


andrei-ionescu:
@imback82 Just some early feedback... 

I've seen you added the `FileBasedRelation` trait. The `DataSourceV2Relation` is not necessarily a file base relation. I would suggest to have this trait defined as a more generic relation, not only file based ones. As you suggest on my #321 PR `SourceRelation` fits better.

imback82:
We need to have files-related APIs such as `allFiles` at `FileBasedRelation` (I thought about having `SupportsFiles` trait, but didn't want to complicate APIs at this stage). Until we have a good use case/example of supporting non-filed relations, making it explicit seems reasonable (similar to we store `FileBasedSourceProviderManager` in the hyperpsace context). 

Similar to `SourceProvider`, I can have a top level trait `SourceRelation`, but will still use `FileBasedRelation` in `FileBasedSourceProviderManager.getRelation`.

andrei-ionescu:
First, regarding Iceberg and DataSourceV2...

The Iceberg source does not use and index (like the `InMemoryFileIndex`) and the files attached to an Iceberg table are not attached to a relation. To retrieve the files of an Iceberg table you need to use the table file scan api.

See https://github.com/microsoft/hyperspace/pull/320/files#diff-01b72d2de6f62e2696e47b850eb7e40039fa85120d40ef392df1cf7b44aff9f8R246-R259

Second, it does NOT look right to have the `DataSourceV2Relation` hard linked to the `FileBasedRelation` and I don't know if this refactor make the code clearer in the case of DataSourceV2 sources like Iceberg.

In the end, if you still consider that this will bring clarity, please complete this PR ASAP and merge it, so that I could modify my PRs to fit these changes.

Thank you.

imback82:
> Second, it does NOT look right to have the `DataSourceV2Relation` hard linked to the `FileBasedRelation` and I don't know if this refactor make the code clearer in the case of DataSourceV2 sources like Iceberg.

Why not? I see that you are implementing `allFiles`? Your provider for iceberg can just check if the given datasourcev2relation is applicable.

I will get this ready for review by eod today as promised.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -113,12 +109,6 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
     }
   }
 
-  protected def sourceRelations(spark: SparkSession, df: DataFrame): Seq[Relation] =
```
imback82, line 116: This is no longer needed as creating `Relation` has moved to `FileBasedRelation`.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -206,15 +208,4 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
 
     (indexDF, resolvedIndexedColumns, resolvedIncludedColumns)
   }
-
-  private def getPartitionColumns(df: DataFrame): Seq[String] = {
```
imback82, line 210: This is no longer needed as  `FileBasedRelation` supports getting the partition schema.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -121,15 +120,9 @@ private[actions] abstract class RefreshActionBase(
    * Build Set[FileInfo] to compare the source file list with the previous index version.
    */
   protected lazy val currentFiles: Set[FileInfo] = {
-    val curFiles = df.queryExecution.optimizedPlan.collect {
-      case relation: LogicalRelation =>
-        Hyperspace
-          .getContext(spark)
-          .sourceProviderManager
-          .allFiles(relation)
```
imback82, line 129: `allFiles` has moved to `FileBasedRelation`.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -106,10 +106,18 @@ object JoinIndexRule
...
+    }
+
+    lazy val optLeftRel = RuleUtils.getRelation(spark, l)
+    lazy val optRightRel = RuleUtils.getRelation(spark, r)
+
     isJoinConditionSupported(condition) &&
-    RuleUtils.getLogicalRelation(l).isDefined && RuleUtils.getLogicalRelation(r).isDefined &&
-    isPlanLinear(l) && isPlanLinear(r) && !isPlanModified(l) && !isPlanModified(r) &&
-    ensureAttributeRequirements(l, r, condition)
+    isPlanLinear(l) && isPlanLinear(r) &&
```
imback82, line 118: @apoorvedave1 we don't need `isPlanLinear` if `getRelation` implementation is as follows, right (collecting leaves and making sure it's only one)?:
```scala
  def getRelation(spark: SparkSession, plan: LogicalPlan): Option[FileBasedRelation] = {
    val provider = Hyperspace.getContext(spark).sourceProviderManager
    val leaves = plan.collectLeaves()
    if (leaves.size == 1 && provider.isSupportedRelation(leaves.head)) {
      Some(provider.getRelation(leaves.head))
    } else {
      None
    }
  }
```

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -157,21 +165,6 @@ object JoinIndexRule
...
     plan.children.length <= 1 && plan.children.forall(isPlanLinear)
 
-  /**
-   * Check if the candidate plan is already modified by Hyperspace or not.
-   * This can be determined by an identifier in options field of HadoopFsRelation.
-   *
-   * @param plan Logical plan.
-   * @return true if the relation in the plan is modified by Hyperspace.
-   */
-  private def isPlanModified(plan: LogicalPlan): Boolean = {
```
imback82, line 167: removed because this is directly testes in `isApplicable` above.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -148,26 +152,17 @@ object RuleUtils {
       // TODO: Duplicate listing files for the given relation as in
       //  [[transformPlanToUseHybridScan]]
       //  See https://github.com/microsoft/hyperspace/issues/160
-      val filesByRelations = plan.collect {
-        case rel: LogicalRelation =>
-          Hyperspace
-            .getContext(spark)
-            .sourceProviderManager
-            .allFiles(rel)
```
imback82, line 156: again, `allFiles` has moved to `FileBasedRelation` (see below).

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -470,17 +467,12 @@ object RuleUtils {
...
-            baseOutput,
-            _,
-            _) =>
-        val options = Hyperspace
-          .getContext(spark)
-          .sourceProviderManager
-          .partitionBasePath(location)
+      case l: LeafNode if provider.isSupportedRelation(l) =>
+        val relation = provider.getRelation(l)
+        val options = relation.partitionBasePath
```
imback82, line 475: @sezruby Question. Below line 482, do we need to create a tag on the `originalPlan` or can it be just the logical plan for the relation?:
```scala
val newLocation = index.withCachedTag(
          originalPlan,
          IndexLogEntryTags.INMEMORYFILEINDEX_HYBRID_SCAN_APPENDED) {
          new InMemoryFileIndex(spark, filesAppended, options, None)
        }
```
`getCandidateIndexes` uses the relation plan, so I was curious.

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -45,28 +44,15 @@ class FileBasedSourceProviderManager(spark: SparkSession) {
...
    * @param relation [[Relation]] to refresh.
    * @return Refreshed [[Relation]].
-   * @throws HyperspaceException if multiple providers returns [[Some]] or
-   *                             if no providers return [[Some]].
+   * @throws [[HyperspaceException]] if multiple providers returns [[Some]] oe
+   *         if no providers return [[Some]].
    */
-  def refreshRelation(relation: Relation): Relation = {
-    run(p => p.refreshRelation(relation))
+  def refreshRelationMetadata(relation: Relation): Relation = {
```
imback82, line 54: We have so many "Relation"s, so I am explicitly calling it out as relation metadata here now.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedRelation.scala
```diff
@@ -0,0 +1,238 @@
...
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index.{Content, FileIdTracker, Hdfs, Relation}
+import com.microsoft.hyperspace.index.IndexConstants.GLOBBING_PATTERN_KEY
+import com.microsoft.hyperspace.index.sources.FileBasedRelation
+import com.microsoft.hyperspace.util.HashingUtils
+
+/**
+ * Implementation for file-based relation used by [[DefaultFileBasedSource]]
+ */
+class DefaultFileBasedRelation(spark: SparkSession, override val plan: LogicalRelation)
```
imback82, line 38: Most of the implementation is from `DefaultFileBasedSource` with minor changes to change the return type: we don't need to wrap with `Option` any longer.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedSource.scala
```diff
@@ -47,82 +42,6 @@ class DefaultFileBasedSource(private val spark: SparkSession) extends FileBasedS
       formats.toLowerCase(Locale.ROOT).split(",").map(_.trim).toSet
     })
 
-  /**
```
imback82, line 50: Removed code has moved to `DefaultFileBasedRelation`.

imback82:
@andrei-ionescu @sezruby @apoorvedave1 This is ready for review now (I will update the PR description soon, but I believe you already know the context. 😄 ). Basically, I removed the direct dependency on `LogicalRelation` from actions/rules code.

@andrei-ionescu Now, you can just implement `IcebergProvider` and `IcebergRelation` without modifying anywhere else hopefully, 🤞🏼

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
sezruby: Yea getCandidateIndexes using a relation plan, but for here I used originalPlan.
It might be better to use the relation plan here as it's more reusable.

imback82:
@andrei-ionescu I also add you as a co-author in this commit https://github.com/microsoft/hyperspace/pull/355/commits/50ed5cd30bc4727b73dc9c0e46345d8c16aa5864, so when this PR is merged, you will get the co-authorship. Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
imback82: thanks, we can do that as a follow up (wanted to minimize the behavior change in this PR)

andrei-ionescu:
@imback82 Thanks for co-authoring!

andrei-ionescu (approved):

sezruby:
The approach looks great. Thanks :)

sezruby (approved):

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
apoorvedave1: Great Thanks, i will update the join rule v2 to incorporate this as well

apoorvedave1 (approved):
LGTM, 👍 thanks @imback82

imback82 (commented):
Thanks @andrei-ionescu @sezruby @apoorvedave1 for the review!

I will create a follow up PR to address a couple of comments.

Issue #318, opened 2021-01-07
[PROPOSAL]: Support Iceberg table format


andrei-ionescu:
## Problem description

This design proposal is for adding feature request #306.

Currently, Hyperspace supports creating indexes only simple formats (like parquet, etc.) and Delta, but not Iceberg. 

As [Iceberg table format](https://iceberg.apache.org) is close to Delta it would be very useful to have Iceberg support in Hyperspace.

## Goal

The user should be able to do use the current API to create an index over an Iceberg dataset and use it in queries.

## Proposed solution

Using the same Hyperspace APIs indexes over Iceberg datasets should work in the same way as for the other implementations.

Given the `icebergDataset` Iceberg dataset with schema

```
root
 |-- id: long (nullable = false)
 |-- str: string (nullable = true)
 |-- date: timestamp (nullable = true)
```

This is the way and index on `str` field with other 2 projection fields can be created

```
hs.createIndex(
  icebergDataset, 
  IndexConfig(
    "index_ice_str",
    indexedColumns = Seq("str"),
    includedColumns = Seq("date", "id")
  )
)
```

Adding the support for Iceberg into Hyperspace has many aspects and I'll try to explain each of these aspects.

### DataSourceV2 support

Iceberg is built on top of the Spark DataSourceV2 API thus Hyperspace should be able to work with these kind of sources and alter the plans containing the `DataSourceV2Relation` types. `DataSourceV2Relation` gives us access to the `IcebergSource` and although the access to its properties is limited we can use it and together with Iceberg table API we can get the latest version of the table and the linked files.

On the same note the Hyperspace rules will need also to be able to work with `DataSourceV2Relation` and alter the plan as needed.

For example, given a query like this

```
val query = icebergDataset.filter(icebergDataset("str") === "f81qj").select("date", "str", "id")
hs.explain(query, true)
```

The rules should be able to transform the plan like bellow:

```
=============================================================
Plan with indexes:
=============================================================
Project [date#3419, str#3418, id#3417L]
+- Filter (isnotnull(str#3418) && (str#3418 = f81qj))
[...]

Issue #306, opened 2021-01-04
[FEATURE REQUEST]: Add support for Iceberg table format


andrei-ionescu:
**Feature requested**

As a **data engineer**, **data scientist** or **developer**, I want **support for Iceberg table format**, in order to **speed-up  data processing**.

**Acceptance criteria** 

- [ ] Create index against an Iceberg dataset
- [ ] Run queries provided in the example
- [ ] Manage Iceberg related indexes

**Success criteria**

N/A

**Additional context**

N/A

andrei-ionescu:
Can I take this one? I'm already doing some progress on it.

rapoth:
Sounds good to me. Please feel free to ping us when you have a draft PR. Could you also kindly add an initial Proposal (so it serves as a design reference)? You can see some sample proposals here: #263 #275 #276 Thank you!

andrei-ionescu:
I created the proposal and added it to the description. It's https://github.com/microsoft/hyperspace/issues/318.

PR #358, opened 2021-02-15
Support Iceberg table format

andrei-ionescu:

 - **Tracking Issue**: #306
 - **Proposal**: #318
 - **Dependencies**: #355 
 - Fixes: #306
 - Fixes: #318
 - Closes: #321 
 - Closes: #320 

This PR adds support for Iceberg.

The following changes are in this PR and each of them are separate commits:

- Add Iceberg source.
- Add support for incremental refresh. This is based on #301 PR from @sezruby.
- Add integration test

No. The main changes to user-facing APIs are in the #321 PR. Detailed information can be found in the #318 proposal.

1. Integration test added for the new functionality
2. Locally & Databricks Runtime tests

- Local build

```
sbt publishLocal
```

- Run Spark shell with Hyperspace and Iceberg libraries loaded

```
$ spark-shell \
--driver-memory 4g \
--packages "com.microsoft.hyperspace:hyperspace-core_2.11:0.4.0-SNAPSHOT,org.apache.iceberg:iceberg-spark-runtime:0.10.0" \
--driver-java-options "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5006 -XX:+UseG1GC -Dlog4j.debug=true"
```

- Paste the following code

```
import org.apache.spark.sql._
import com.microsoft.hyperspace._
import scala.collection.JavaConverters._
import org.apache.iceberg.PartitionSpec
import org.apache.iceberg.TableProperties
import org.apache.iceberg.spark._
import org.apache.iceberg.hadoop._

// create Iceberg table
val props = Map(TableProperties.WRITE_NEW_DATA_LOCATION -> "table3").asJava
val sourceDf = Seq((1, "name1"), (2, "name2")).toDF("id", "name")
val schema = SparkSchemaUtil.convert(sourceDf.schema)
val part = PartitionSpec.builderFor(schema).build()
val icebergTable = new HadoopTables().create(schema, part, props, "table3")
sourceDf.write.mode("overwrite").format("iceberg").save("./table3")

// read created table
val iceDf = spark.read.format("iceberg").load("./table3")

// create indexes
hs.createIndex(iceDf, IndexConfig("index_ice0", indexedColumns = Seq("id"), includedColumns = Seq("name")))
hs.createIndex(iceDf, IndexConfig("index_ice1", indexedColumns = Seq("name")))

// verify plans
val query = iceDf.filter(iceDf("id") === 1).select("name")
hs.explain(query, verbose = true)
```

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergFileBasedSource.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergFileBasedSource.scala
@@ -29,9 +29,9 @@ import com.microsoft.hyperspace.index.sources.{FileBasedRelation, FileBasedSourc
  *
  * This source can support relations that meet the following criteria:
  *   - The relation is with [[DataSourceV2Relation]]
+ *   - The source of [[DataSourceV2Relation]] is an [[IcebergSource]]
  */
-class IcebergFileBasedSource(private val spark: SparkSession)
-    extends FileBasedSourceProvider {
+class IcebergFileBasedSource(private val spark: SparkSession) extends FileBasedSourceProvider {
 
   private val ICEBERG_FORMAT_STR = "iceberg"
 
@@ -64,15 +64,14 @@ class IcebergFileBasedSource(private val spark: SparkSession)
     }
   }
 
-
   /**
-   * Returns true if the given logical plan is a relation for Delta Lake.
+   * Returns true if the given logical plan is a relation for Iceberg.
    *
    * @param plan Logical plan to check if it's supported.
    * @return Some(true) if the given plan is a supported relation, otherwise None.
    */
-  def isSupportedRelation(plan: LogicalPlan): Option[Boolean] = plan match {
-    case _ @ DataSourceV2Relation(_: IcebergSource, _, _, _, _) =>
+  override def isSupportedRelation(plan: LogicalPlan): Option[Boolean] = plan match {
+    case _ @DataSourceV2Relation(_: IcebergSource, _, _, _, _) =>
       Some(true)
     case _ => None
   }
@@ -84,7 +83,7 @@ class IcebergFileBasedSource(private val spark: SparkSession)
    * @param plan Logical plan to wrap to [[FileBasedRelation]]
    * @return [[FileBasedRelation]] that wraps the given logical plan.
    */
-  def getRelation(plan: LogicalPlan): Option[FileBasedRelation] = plan match {
+  override def getRelation(plan: LogicalPlan): Option[FileBasedRelation] = plan match {
     case l @ DataSourceV2Relation(_: IcebergSource, _, _, _, _) =>
       Some(new IcebergRelation(spark, l))
     case _ => None
--- a/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergFileBasedSource.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergFileBasedSource.scala
@@ -1,0 +1,99 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.index.sources.iceberg
+
+import org.apache.iceberg.spark.source.IcebergSource
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
+
+import com.microsoft.hyperspace.index.Relation
+import com.microsoft.hyperspace.index.sources.{FileBasedRelation, FileBasedSourceProvider, SourceProvider, SourceProviderBuilder}
+
+/**
+ * Iceberg file-based source provider.
+ *
+ * This source can support relations that meet the following criteria:
+ *   - The relation is with [[DataSourceV2Relation]]
+ */
+class IcebergFileBasedSource(private val spark: SparkSession)
+    extends FileBasedSourceProvider {
+
+  private val ICEBERG_FORMAT_STR = "iceberg"
+
+  /**
+   * Given a [[Relation]], returns a new [[Relation]] that will have the latest source.
+   *
+   * @param relation [[Relation]] object to reconstruct [[DataFrame]] with.
+   * @return [[Relation]] object if the given 'relation' can be processed by this provider.
+   *         Otherwise, None.
+   */
+  override def refreshRelationMetadata(relation: Relation): Option[Relation] = {
+    if (relation.fileFormat.equals(ICEBERG_FORMAT_STR)) {
+      Some(relation.copy(options = relation.options - "snapshot-id" - "as-of-timestamp"))
+    } else {
+      None
+    }
+  }
+
+  /**
+   * Returns a file format name to read internal data files for a given [[Relation]].
+   *
+   * @param relation [[Relation]] object to read internal data files.
+   * @return File format to read internal data files.
+   */
+  override def internalFileFormatName(relation: Relation): Option[String] = {
+    if (relation.fileFormat.equals(ICEBERG_FORMAT_STR)) {
+      Some("parquet")
+    } else {
+      None
+    }
+  }
+
+
+  /**
+   * Returns true if the given logical plan is a relation for Delta Lake.
+   *
+   * @param plan Logical plan to check if it's supported.
+   * @return Some(true) if the given plan is a supported relation, otherwise None.
+   */
+  def isSupportedRelation(plan: LogicalPlan): Option[Boolean] = plan match {
+    case _ @ DataSourceV2Relation(_: IcebergSource, _, _, _, _) =>
+      Some(true)
+    case _ => None
+  }
+
+  /**
+   * Returns the [[FileBasedRelation]] that wraps the given logical plan if the given
+   * logical plan is a supported relation.
+   *
+   * @param plan Logical plan to wrap to [[FileBasedRelation]]
+   * @return [[FileBasedRelation]] that wraps the given logical plan.
+   */
+  def getRelation(plan: LogicalPlan): Option[FileBasedRelation] = plan match {
+    case l @ DataSourceV2Relation(_: IcebergSource, _, _, _, _) =>
+      Some(new IcebergRelation(spark, l))
+    case _ => None
+  }
+}
+
+/**
+ * Builder for building [[IcebergFileBasedSource]].
+ */
+class IcebergFileBasedSourceBuilder extends SourceProviderBuilder {
+  override def build(spark: SparkSession): SourceProvider = new IcebergFileBasedSource(spark)
+}
--- a/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
@@ -1,0 +1,300 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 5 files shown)


andrei-ionescu:
@imback82 I closed the #321, #320 PRs which were obsolete after your refactoring work on #355 - thanks for it. Please review this PR that adds support for Iceberg table format.

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -0,0 +1,300 @@
...
+    else {
+      val hiveCatalog = HiveCatalogs.loadCatalog(conf)
+      val tableIdentifier = TableIdentifier.parse(path.get)
+      hiveCatalog.loadTable(tableIdentifier)
+    }
+  }
+
+  // TODO: need Iceberg's `tasks()` method to be public to get rid of this method
+  private def tasks(table: Table, options: DataSourceOptions, pushedFilters: Seq[Filter],
+      requestedSchema: StructType): Seq[CombinedScanTask] = {
```
imback82, line 218: Could you run `scalafmt` in this file?

imback82 (commented):
Few minor/nit comments, but generally looking good to me.

I don't know every detail of Iceberg, but since the changes are self-contained, I think this is good to go. @sezruby could you also take a look? Thanks.

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
imback82: and `IcebergIntegrationTest.scala` as well.

sezruby (commented):
Generally looks good to me! Could you also consider adding `HybridScanForIcebergTest` like `HybridScanForDeltaLakeTest`?

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -0,0 +1,300 @@
...
+    try {
+      scan.planTasks().asScala.toSeq
+    } catch {
+      case e: IOException =>
+        throw new RuntimeException(s"Failed to close table scan: $scan", e)
+    }
+  }
+
+  private def toFileStatus(fileScanTask: FileScanTask): FileStatus = {
+    val jPath = new Path(fileScanTask.file().path().toString)
```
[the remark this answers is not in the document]
andrei-ionescu: Renamed to `path`.

andrei-ionescu:
@sezruby The test for hybrid scan is in the `IcebergIntegrationTest` lines [299-338](https://github.com/microsoft/hyperspace/pull/358/files#diff-ce1f32f296e1683385beb0fe1954b154710c0ba0120f028167afbe5953347dd3R299-R338).

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -0,0 +1,286 @@
...
+   */
+  override def createLogicalRelation(
+      hadoopFsRelation: HadoopFsRelation,
+      newOutput: Seq[AttributeReference]): LogicalRelation = {
+    val updatedOutput =
+      newOutput.filter(attr => hadoopFsRelation.schema.fieldNames.contains(attr.name))
+    new LogicalRelation(hadoopFsRelation, updatedOutput, None, false)
+  }
+
+  private def getIcebergTable: Table = {
```
imback82, line 195: nit: can this be just `private lazy val icebergTable: Table` so that we don't have to do `val t = getIcebergTable` on the caller side?

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -0,0 +1,286 @@
...
+      SparkSchemaUtil.prune(
+        table.schema,
+        requestedSchema,
+        pushedExpressions.reduceOption(Expressions.and).getOrElse(Expressions.alwaysTrue()),
+        caseSensitive)
+    } else {
+      table.schema
+    }
+
+    var scan: TableScan = table.newScan.caseSensitive(caseSensitive).project(lazySchema)
```
imback82, line 237: nit: we don't specify types (unless required) for local variables.
```suggestion
    var scan = table.newScan.caseSensitive(caseSensitive).project(lazySchema)
```

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergFileBasedSource.scala
```diff
@@ -0,0 +1,98 @@
...
+  override def internalFileFormatName(relation: Relation): Option[String] = {
+    if (relation.fileFormat.equals(ICEBERG_FORMAT_STR)) {
+      Some("parquet")
+    } else {
+      None
+    }
+  }
+
+  /**
+   * Returns true if the given logical plan is a relation for Delta Lake.
```
imback82, line 68: let's update delta

imback82 (commented):
LGTM (pending minor/nit comments + @sezruby's comments), thanks @andrei-ionescu!

sezruby:
> @sezruby The test for hybrid scan is in the IcebergIntegrationTest lines 299-338.

Yes, but we need to verify the exact plan transformation and the result of query.

andrei-ionescu:
@sezruby I did take the Delta Lake hybrid test and modified it for Iceberg. That is what is in the integration test from line 299 to 338. Is there other hybrid test for Delta Lake that I missed?

sezruby:
@andrei-ionescu There's another test set:
https://github.com/microsoft/hyperspace/blob/master/src/test/scala/com/microsoft/hyperspace/index/HybridScanForDeltaLakeTest.scala 
based on https://github.com/microsoft/hyperspace/blob/master/src/test/scala/com/microsoft/hyperspace/index/HybridScanSuite.scala

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
andrei-ionescu: There are some Iceberg options that may change from other writers and in this case it is desired to get the latest table instance.

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -0,0 +1,286 @@
...
+   * Options of the current relation.
+   */
+  override def options: Map[String, String] = plan.options
+
+  /**
+   * The partition schema of the current relation.
+   */
+  override def partitionSchema: StructType = plan.source match {
+    case _: IcebergSource =>
+      val dsOpts = new DataSourceOptions(plan.options.asJava)
```
imback82, line 158: not used?

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
imback82: ah ok, make sense. Shall we rename this to `loadIcebergTable` to be explicit about it? "load" sounds more like it's getting the latest version than "get" does.

andrei-ionescu:
@sezruby I added [HybridScanForIcebergTest.scala](https://github.com/microsoft/hyperspace/pull/358/files#diff-4905e3cb43d3852943ec05af7f34eb6e2d48dc0b707e700a9516cdda897f8729). Please have another look.

src/test/scala/com/microsoft/hyperspace/index/HybridScanForIcebergTest.scala
```diff
@@ -0,0 +1,171 @@
...
+          case p @ LogicalRelation(fsRelation: HadoopFsRelation, _, _, _) =>
+            // Verify appended file is included or not.
+            val files = fsRelation.location.inputFiles
+            assert(files.count(_.contains(sourcePath)) === 2)
+            // Verify number of index data files.
+            assert(files.count(_.contains("index_Append")) === 4)
+            assert(files.length === 6)
+            p
+        }
+        // Filter Index and Parquet format source file can be handled with 1 LogicalRelation
```
sezruby, line 149: super nit:
```suggestion
        // Filter Index and Parquet format source file can be handled with 1 LogicalRelation.
```

src/test/scala/com/microsoft/hyperspace/index/HybridScanForIcebergTest.scala
```diff
@@ -0,0 +1,171 @@
...
+      .asScala
+      .toSeq
+      .map(f => s"file:/${f.file().path().toString.replaceAll("^/", "")}")
+    (inputFilesB diff inputFilesA, inputFilesA diff inputFilesB)
+  }
+
+  test(
+    "Append-only: filter rule & parquet format, " +
+      "index relation should include appended file paths.") {
+    // This flag is for testing plan transformation if appended files could be load with index
```
sezruby, line 99: Sorry I didn't revised this comment after refactoring 😅.

```suggestion
    // This test is for verifying plan transformation if appended files could be load with index
```

sezruby (commented):
@andrei-ionescu Thanks! Seems it works as expected! 👍👍  
For start & end snapshot id test, I'd just like to check the use case & how those configs work well (or not well) with Hyperspace.

andrei-ionescu:
@sezruby 

> For start & end snapshot id test, I'd just like to check the use case & how those configs work well (or not well) with Hyperspace.

Is there such check for Delta so that I could inspire from?

sezruby:
@andrei-ionescu No, not sure Delta has the feature. 
How about this?

- build test data
  - snapshot 1) add a row with value 1 (append 1 file)
  - snapshot 2) add a row with value 2 (append 1 file)
  - snapshot 3) add a row with value 3 (append 1 file)
  - snapshot 4) add a row with value 4 (append 1 file)
  - snapshot 5) add a row with value 5 (append 1 file)

- build index with df with start-snapshot-id 2, end-snapshot-id 4 
  - query with df with start-snapshot-id 2, end-snapshot-id 4 w/o hybrid scan
  - query with df with start-snapshot-id 3, and check w/ hybrid scan
  - query with df with end-snapshot-id 3, and check w/ hybrid scan

andrei-ionescu:
@sezruby I'll try to see it is possible but from my knowledge on Iceberg I don't think this kind of test is possible. The snapshot ids are used for time travel and to isolate date at read time. A snapshot is a sort of full image of the dataset -- is not a delta from the previous. This is a bit different than Delta.

Delta has the same parameter under `versionAsOf`. If there are tests to validate the functionality of Delta with time travel I'll do my best add it to Iceberg too although the implementation is a bit different.

Otherwise I would suggest to have a separate PR with time travel for Iceberg.

sezruby:
@andrei-ionescu I guess `versionAsOf` of Delta Lake is the same as Iceberg only with `end-snapshot-id`.  
But both `start-snapshot-id` or `end-snapshot-id` handles "delta" dataset between two snapshots, right?

Hybrid scan will work w/o any link of version info - it utilizes the list of source files from DataFrame.

For time travel query optimization, to pick a proper version of index (after some refresh),
I added version history info to link delta version and index version in #272.

andrei-ionescu:
@sezruby: I refactored it a bit and it is no longer needed to have knowledge of `start-snapshot-id`, `end-snapshot-id` and other internal Iceberg properties.

sezruby:
LGTM thanks @andrei-ionescu!

andrei-ionescu:
@rapoth, @imback82, @sezruby: The PR has 2 `LGTM`s already - thanks @imback82 & @sezruby. What's the process? What next steps are ahead to have it merged and complete the feature?

imback82:
Let me take a final look since there were some changes since my last LGTM.

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -0,0 +1,227 @@
...
+   * @return Always true since Iceberg table files are stored as Parquet.
+   */
+  override def hasParquetAsSourceFormat: Boolean = true
+
+  /**
+   * Returns list of pairs of (file path, file id) to build lineage column.
+   *
+   * File paths should be the same format as "input_file_name()" of the given relation type.
+   * For [[IcebergRelation]], each file path should be in this format:
+   *   `file:/path/to/file`
```
imback82, line 126: Can you update this based on the implementation below?

imback82 (approved):

imback82:
@andrei-ionescu Can you do a follow up PR to address https://github.com/microsoft/hyperspace/pull/358#discussion_r580425973? Thanks!

andrei-ionescu:
@imback82: I create #362 to address the doc comment. 

@imback82, @sezruby, @rapoth: Thanks for help with this PR & feature.

rapoth:
It has been a fantastic collaboration so far and I have personally learned a lot in the process! Thank you @andrei-ionescu! 🙂

sezruby:
Thanks for the work @andrei-ionescu!

Could you also simply update the user document if possible? 
ref: https://microsoft.github.io/hyperspace/docs/ug-supported-data-formats/ 
https://github.com/microsoft/hyperspace/blob/master/docs/_docs/08-ug-supported-data-formats.md

Issue #345, opened 2021-01-29
[FEATURE REQUEST]: Include the Hyperspace version in the index metadata


imback82:
**Feature requested**

It would be helpful if we include the Hyperspace version in the index metadata to better support backward compatibility. Currently, it's very hard to see which version of Hyperspace was used when indexes were created.

thugsatbay:
I can take a look at this.

imback82:
One thing to remember is that the version should be read from [version.sbt](https://github.com/microsoft/hyperspace/blob/master/version.sbt).

Maybe https://github.com/sbt/sbt-buildinfo can help? (I haven't tried it though)

imback82:
@sezruby let's include the PR # (or link it) when you close an issue.

clee704:
I think it's #364 and #367.

PR #364, opened 2021-02-23
Add a sbt-buildinfo plugin to generate BuildInfo object

imback82:
This PR adds `sbt-buildinfo` to generate `BuildInfo` so that the package version, etc. can be accessed programatically.

After this PR, `target/scala-2.*/src_managed/main/sbt-buildinfo/BuildInfo.scala` is auto-generated with a following content:
```

import scala.Predef._

/** This object was generated by sbt-buildinfo. */
case object BuildInfo {
  /** The value is "hyperspace-core". */
  val name: String = "hyperspace-core"
  /** The value is "0.5.0-SNAPSHOT". */
  val version: String = "0.5.0-SNAPSHOT"
  /** The value is "2.12.8". */
  val scalaVersion: String = "2.12.8"
  /** The value is "0.13.18". */
  val sbtVersion: String = "0.13.18"
  override val toString: String = {
    "name: %s, version: %s, scalaVersion: %s, sbtVersion: %s" format (
      name, version, scalaVersion, sbtVersion
    )
  }
}
```

Now, it can be simply accessed as `BuildInfo.version` in Hyperspace code base.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,34 +27,37 @@ scalaVersion := scala212
 crossScalaVersions := supportedScalaVersions
 
 libraryDependencies ++= Seq(
-  "org.apache.spark" %% "spark-sql" % sparkVersion.value % "provided" withSources(),
-  "org.apache.spark" %% "spark-core" % sparkVersion.value % "provided" withSources(),
-  "org.apache.spark" %% "spark-catalyst" % sparkVersion.value % "provided" withSources(),
-  "io.delta" %% "delta-core" % "0.6.1" % "provided" withSources(),
-  "org.apache.iceberg" % "iceberg-spark-runtime" % "0.11.0" % "provided" withSources(),
-
+  "org.apache.spark" %% "spark-sql" % sparkVersion.value % "provided" withSources (),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -27,34 +27,37 @@ scalaVersion := scala212
...
 
-/*****************************
- * ScalaStyle configurations *
- *****************************/
+// The following creates target/scala-2.*/src_managed/main/sbt-buildinfo/BuildInfo.scala.
+lazy val root = (project in file("."))
+  .enablePlugins(BuildInfoPlugin)
+  .settings(
+    buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion),
+    buildInfoPackage := "com.microsoft.hyperspace")
```
imback82, line 56: To reviewers, these are the only relevant changes for this PR. Other changes are to conform to the scala style (I ran scalafmt + fixed the comment style)

imback82:
@thugsatbay, this should unblock you on #345 . Please let me know if you encounter other issues.

imback82:
Merging to unblock #345.

PR #367, opened 2021-02-25
Add hyperspace version to index log entry metadata

thugsatbay:

Add hyperspace project version to index log entry metadata.

 - **Tracking Issue**: #345 
 - **Parent Issue**: NA
 - **Dependencies**: NA

Adding a new version field in `IndexLogEntry` scala calass

No

Current Unit tests already present should suffice.


src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -438,6 +438,13 @@ case class IndexLogEntry(
     properties: Map[String, String])
     extends LogEntry(IndexLogEntry.VERSION) {
 
+  /*
+   * Include the project version in the index log entry file.
+   * This may help prevent in mismatching of project version and index
+   * log entry files.
+   */
+  val projectVersion: String = BuildInfo.version
```
sezruby, line 446: Could you add this into `properties`? to avoid breaking changes.
Thanks!

thugsatbay: I put it in the properties of IndexLogEntry.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -72,6 +72,13 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
           LogicalPlanFingerprint(
             LogicalPlanFingerprint.Properties(Seq(Signature(signatureProvider.name, s)))))
 
+        /*
+         * Include the project version in the index log entry file.
+         * This may help prevent in mismatching of project version and index
+         * log entry files.
+         */
+        val projectVersion: Option[(String, String)] = Some("projectVersion", BuildInfo.version)
```
sezruby, line 80: I think "hyperspaceVersion" would be clearer? what do you think?

imback82, line 80: How about the following?
In `case class IndexLogEntry`, add the following `require`:
```scala
case class IndexLogEntry(
    name: String,
    derivedDataset: CoveringIndex,
    content: Content,
    source: Source,
    properties: Map[String, String])
    extends LogEntry(IndexLogEntry.VERSION) {
  require(properties(IndexLogEntry.HYPERSPACE_VERSION_PROP_KEY) == BuildInfo.version)
  ...
```

In `object IndexLogEntry`, add the following:
```scala
  private val HYPERSPACE_VERSION_PROP_KEY = "hyperspaceVersion"

  def create(
    name: String,
    derivedDataset: CoveringIndex,
    content: Content,
    source: Source,
    properties: Map[String, String]): IndexLogEntry = {
    IndexLogEntry(
      name,
      derivedDataset,
      content,
      source,
      properties + (HYPERSPACE_VERSION_PROP_KEY -> BuildInfo.version))
```

And here, you can just call `IndexLogEntry.create`. This should ensure to always have the property and may help when we support multiple index types.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -86,7 +93,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
               coveringIndexProperties)),
           Content.fromDirectory(absolutePath, fileIdTracker),
           Source(SparkPlan(sourcePlanProperties)),
-          Map())
+          projectVersion.toMap)
 
       case None => throw HyperspaceException("Invalid plan for creating an index.")
```
imback82, line 98: Can you also update `IndexLogEntryTest`?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
imback82: `require(properties(IndexLogEntry.HYPERSPACE_VERSION_PROP_KEY) == BuildInfo.version)` could be too restrictive now since we need to handle the older version without this property.

imback82: We can introduce this (`require(properties.contains(IndexLogEntry.HYPERSPACE_VERSION_PROP_KEY)`) when we introduce a breaking change in the log entry.

thugsatbay: I believe what you are proposing is that you want the projectVersion property to be defined inside IndexLogEntry rather than from multiple places outside. Where best scenario is that it is a required property and should be checked inside while creating the case class instance.

However i don't like your approach as we now have two ways to create IndexLogEntry one from create static method and the other using the case class itself. We also can't make case class private as it is needed throughout code.

I propose we make the properties map mutable and add the property projectVersion if it is not present. IndexLogEntry is a dev class so having a mutable property should be fine? Also changing the equals method to compare properties. 

That way we don't introduce the create method and the prroperty becomes default and required in the construction of the instance of IndexLogEntry without breaking backward compatibility.

Let me know your thoughts, if I missed something. If so I can revert to the create approach.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
thugsatbay: Done, added some tests

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
imback82: seems like having mutable as a public field is worse since it opens up the possibility of modifying the object more?

having `require` (after a breaking change) seems like a reasonable tradeoff to me. (btw, you can't just naively set one now on the deserialize path since it wouldn't be semantically correct)

thugsatbay: Fair point, even though it is dev code someone may end up breaking it. Changed back to mutable.

I read online just now we can overwrite the case class constructor apply method, so instead of create method i introduced apply and there I added the property.

Also I kept the require statement, which only checks `projectVersion` with respect to project build property only if incoming `properties` contain `projectVersion` in the first place.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +620,26 @@ case class IndexLogEntryTag[T](name: String)
...
   def schemaString(schema: StructType): String = schema.json
+
+  /*
+   * Default IndexLogEntry Properties
+   */
+  def apply(name: String,
+            derivedDataset: CoveringIndex,
+            content: Content,
+            source: Source
+           ): IndexLogEntry = {
```
sezruby, line 634:
```suggestion
            source: Source): IndexLogEntry = {
```
Could you check scalafmt setting?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +620,26 @@ case class IndexLogEntryTag[T](name: String)
 object IndexLogEntry {
   val VERSION: String = "0.1"
 
+  val HYPERSPACE_PROJECT_VERSION: String = "hyperspaceVersion"
+
   def schemaString(schema: StructType): String = schema.json
+
+  /*
+   * Default IndexLogEntry Properties
```
sezruby, line 628: Could you revise the comment?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -438,6 +438,17 @@ case class IndexLogEntry(
...
 
+  /*
+   * Include the project version in the index log entry file.
+   * This may help prevent in mismatching of project version and index
+   * log entry files.
+   * TODO: breaking change remove if condition. Since currently there are index
+   *  files that don't have project version info and we need to support them as well.
+   */
+  if (properties.contains(IndexLogEntry.HYPERSPACE_PROJECT_VERSION)) {
+    require(properties(IndexLogEntry.HYPERSPACE_PROJECT_VERSION) == BuildInfo.version)
```
imback82, line 449: hmm isn't this too restrictive as I said in the other comment? If there is no breaking change from version 0.5 to 0.6, this check will fail unnecessarily when 0.6 is reading 0.5 log entry.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +620,26 @@ case class IndexLogEntryTag[T](name: String)
...
+
   def schemaString(schema: StructType): String = schema.json
+
+  /*
+   * Default IndexLogEntry Properties
+   */
+  def apply(name: String,
+            derivedDataset: CoveringIndex,
+            content: Content,
+            source: Source
```
imback82, line 633: seems like this doesn't help either since I can create IndexLogEntry directly? In that case, `create` is just fine supporting all fields?

thugsatbay: Won't you be able to create `IndexLogEntry` directly with create method as well ? 

This apply signature signifies that it is a default constructor for `IndexLogEntry` when you specify no properties and include default properties. Whereas if you specify properties then specify all properties, the create method seems more hacky to me ?

imback82: But it doesn't add any value of enforcing the requirement.

what we want is not supported by the language (unless we restructure this class with `sealed abstract` - there are few examples on the web), so anything we are doing is not the ideal approach (if you mean hacky by that). Also, I am not inventing create method approach: https://github.com/apache/spark/blob/56e664c7179eadeb5134b4418f3aaa6a9d742ef6/sql/catalyst/src/main/scala/org/apache/spark/sql/execution/datasources/v2/DataSourceV2Relation.scala#L169

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
thugsatbay: fair enough, i will change it to immutable.

We might need to maintain a set of combinations of index log file project Version supported to buildVersion, i guess you are correct, will remove require for now.

imback82: yes. we need to store a min compatible version. Could you help driving this effort as well separately? cc @rapoth

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: Another downside of this approach is that if you now want to add a prop on the caller side, you have to always manually add the version prop as well.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
thugsatbay: Okay, sure!

rapoth: Can you please open an issue and link here when you get a chance? Thank you!

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
thugsatbay: Okay, switched to create format that is commonly used in the project.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
thugsatbay: Done.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +609,30 @@ case class IndexLogEntryTag[T](name: String)
...
   val VERSION: String = "0.1"
 
+  val HYPERSPACE_PROJECT_VERSION: String = "hyperspaceVersion"
+
   def schemaString(schema: StructType): String = schema.json
+
+  /*
+   * Use this method to create a new IndexLogEntry. Which automatically includes
+   * all common default hyperspace project properties.
+   * TODO: force dev to use this method as this takes into account all
```
sezruby, line 619: Could you create an issue for this and note the link here?

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -72,14 +128,7 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
...
 
-  test("IndexLogEntry spec example") {
-    val schemaString =
-      """{\"type\":\"struct\",
-        |\"fields\":[
-        |{\"name\":\"RGUID\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}},
-        |{\"name\":\"Date\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}}]}
-        |""".stripMargin.replaceAll("\r", "").replaceAll("\n", "")
-
+  test("IndexLogEntry spec example, missing project Version") {
```
sezruby, line 131: super nit
```suggestion
  test("IndexLogEntry spec example, missing project Version.") {
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +609,30 @@ case class IndexLogEntryTag[T](name: String)
 object IndexLogEntry {
   val VERSION: String = "0.1"
 
+  val HYPERSPACE_PROJECT_VERSION: String = "hyperspaceVersion"
```
sezruby, line 612: Could you move this to IndexConstants? since other key strings are also in there.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +609,31 @@ case class IndexLogEntryTag[T](name: String)
...
+
   def schemaString(schema: StructType): String = schema.json
+
+  /*
+   * Use this method to create a new IndexLogEntry. Which automatically includes
+   * all common default hyperspace project properties.
```
imback82, line 618:
```suggestion
   * Use this method to create a new IndexLogEntry, which automatically includes
   * all common default hyperspace project properties.
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +609,31 @@ case class IndexLogEntryTag[T](name: String)
...
+   * TODO: force dev to use this method as this takes into account all
+   *  system properties that needed to be added by default. Currently, dev can also
+   *  create IndexLogEntry from case class.
+   *  https://github.com/microsoft/hyperspace/issues/370
```
imback82, line 622: Can you add a similar comment to the `IndexLogEntry` constructor and we need to add `require` when there is a breaking change?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +609,31 @@ case class IndexLogEntryTag[T](name: String)
...
 
+  val HYPERSPACE_PROJECT_VERSION: String = "hyperspaceVersion"
+
   def schemaString(schema: StructType): String = schema.json
+
+  /*
+   * Use this method to create a new IndexLogEntry. Which automatically includes
+   * all common default hyperspace project properties.
+   * TODO: force dev to use this method as this takes into account all
+   *  system properties that needed to be added by default. Currently, dev can also
```
imback82, line 620: I see `hyperspace project properties` and `system properties`. Can we make this consistent?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -608,7 +609,31 @@ case class IndexLogEntryTag[T](name: String)
...
+   *  create IndexLogEntry from case class.
+   *  https://github.com/microsoft/hyperspace/issues/370
+   */
+  def create(name: String,
+            derivedDataset: CoveringIndex,
+            content: Content,
+            source: Source,
+            properties: Map[String, String] = Map()): IndexLogEntry = {
+    new IndexLogEntry(
```
imback82, line 629: We can remove `new` here.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: And it should be `HYPERSPACE_VERSION_PROPERTY` if we move this to `IndexConstants`.

imback82 (commented):
Few minor comments, but generally looks fine to me

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -430,7 +430,12 @@ object SparkPlan {
 // IndexLogEntry-specific Source that uses SparkPlan as a plan.
 case class Source(plan: SparkPlan)
 
-// IndexLogEntry that captures index-related information.
+/*
```
imback82, line 433:
```suggestion
/**
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -610,6 +616,29 @@ object IndexLogEntry {
...
+   *  project properties that needed to be added by default. Currently, dev can also
+   *  create IndexLogEntry from case class.
+   *  https://github.com/microsoft/hyperspace/issues/370
+   *  Also add require for hyperspace project version when we introduce breaking change.
+   */
+  def create(name: String,
+            derivedDataset: CoveringIndex,
+            content: Content,
+            source: Source,
+            properties: Map[String, String] = Map()): IndexLogEntry = {
```
imback82, line 633: Let's not create a default value for this, similar to `IndexLogEntry` constructor, at least for this PR.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -610,6 +616,29 @@ object IndexLogEntry {
   val VERSION: String = "0.1"
 
   def schemaString(schema: StructType): String = schema.json
+
+  /*
```
imback82, line 620:
```suggestion
  /**
```

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -72,14 +128,7 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
...
 
-  test("IndexLogEntry spec example") {
-    val schemaString =
-      """{\"type\":\"struct\",
-        |\"fields\":[
-        |{\"name\":\"RGUID\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}},
-        |{\"name\":\"Date\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}}]}
-        |""".stripMargin.replaceAll("\r", "").replaceAll("\n", "")
-
+  test("IndexLogEntry spec example, missing project Version.") {
```
imback82, line 131: Can you assert on this condition, "missing project version"?

imback82, line 131:
```suggestion
  test("IndexLogEntry spec example, missing project version.") {
```

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -184,60 +233,231 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
...
+         |  },
+         |  "version" : "0.1",
+         |  "id" : 0,
+         |  "state" : "ACTIVE",
+         |  "timestamp" : 1578818514080,
+         |  "enabled" : true
+         |}""".stripMargin
+
+    val actual = JsonUtils.fromJson[IndexLogEntry](jsonString)
+    assert(!actual.equals(expected))
```
imback82, line 460: Can you assert on this condition "project version mismatch"?

imback82 (commented):
LGTM (minor comments), thanks @thugsatbay!

@sezruby Could you take one more look?

imback82:
@thugsatbay, I pushed the changes directly to your branch. The diff is now very small - the new tests in `IndexLogEntryTest` were reverted and I just added tests making sure that the hyperspace version is added correctly to the log entry.

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
imback82: reverted

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
imback82: reverted

imback82: reverted

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: fixed

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: fixed

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
imback82: fixed

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -103,6 +103,7 @@ object IndexConstants {
   // Indicate whether the source file format is parquet.
   private[hyperspace] val HAS_PARQUET_AS_SOURCE_FORMAT_PROPERTY = "hasParquetAsSourceFormat"
   // Indicate index log version.
+  private[hyperspace] val HYPERSPACE_VERSION_PROPERTY: String = "hyperspaceVersion"
```
sezruby, line 106:
```suggestion
  // Indicate Hyperspace version.
  private[hyperspace] val HYPERSPACE_VERSION_PROPERTY: String = "hyperspaceVersion"
  // Indicate index log version.
```

sezruby (commented):
LGTM thanks!

imback82:
Thanks @sezruby for the review, merging to master!

@thugsatbay Feel free to create a follow up PR if you are not satisfied with any of the changes made.

PR #371, opened 2021-03-02
[WIP] Java version fix

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a
 
Set Java 8 for Scala 11 - default java version is changed in Ubuntu 20.04
https://github.com/actions/virtual-environments/blob/ubuntu20/20210219.1/images/linux/Ubuntu2004-README.md
 

No

build

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,6 +11,10 @@ jobs:
     pool:
       vmImage: 'ubuntu-latest'
     steps:
+      - script: |
+          echo "##vso[task.setvariable variable=JAVA_HOME]$(JAVA_HOME_8_X64)"
+          echo "##vso[task.setvariable variable=PATH]$(JAVA_HOME_8_X64)\bin;$(PATH)"
+        displayName: 'Set Java version'
       - script: sbt ++2.11.12 clean
         displayName: 'Running $sbt clean'
       - script: sbt ++2.11.12 update


azure-pipelines.yml
```diff
@@ -69,6 +75,17 @@ jobs:
...
+        inputs:
+          versionSpec: '2.7.18'
+          addToPath: true
+      - task: JavaToolInstaller@0
+        displayName: 'Set Java version'
+        inputs:
+          versionSpec: '8'
+          jdkArchitectureOption: 'x64'
+          jdkSourceOption: 'PreInstalled'
       - script: sbt ++2.11.12 clean
```
sezruby, line 84: @imback82 in run-tests.py, it names hyperspace package as:
```
package = "com.microsoft.hyperspace:hyperspace-core_2.12:" + version
```
But here it's 2.11. Is there any reason for this?

imback82: Probably a byproduct of https://github.com/delta-io/delta/blob/4277443703c5ab59a567c1e80189bbcdb7495817/python/run-tests.py#L67. We can fix this if this doesn't work.

azure-pipelines.yml
```diff
@@ -11,6 +11,12 @@ jobs:
     pool:
       vmImage: 'ubuntu-latest'
```
imback82, line 12: Do we need to move to ubuntu-20.04 now? Or can we simply set the `vmImage` to 18?

sezruby: That is the other fix.  I changed java & python ver. to keep this 'ubuntu-latest' to detect any other change later.
If it's unnecessary, it's ok to change it to 18

azure-pipelines.yml
sezruby: it's working anyway but I was just curious.

azure-pipelines.yml
imback82: Can we set this to ubuntu18 since internal build is (will be) on ubuntu18? You can leave other changes if they work fine on 18.

sezruby: Let's keep other changes for reference. Thanks!

azure-pipelines.yml
```diff
@@ -69,6 +75,17 @@ jobs:
...
```
imback82, line 76: change this?

imback82: same for Build_2_12

imback82 (approved):
LGTM, thanks @sezruby!

andrei-ionescu:
@imback82 & @sezruby Thanks for fixing it up!

PR #374, opened 2021-03-04
Support concurrent Spark sessions 

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Currently, Hyperspace keeps only one HyperspaceContext for each Hyperspace object.
So if an app uses multiple concurrent Spark sessions using one Hyperspace object, a new Hyperspace context is continuously created for each request.  (workaround: use different hyperspace object for each session)

With this PR, Hyperspace maintains contexts for multiple spark sessions with one hyperspace object.

Yes, Hyperspace contexts are reused when multiple sessions

Unit test


src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -166,18 +168,20 @@ class Hyperspace(spark: SparkSession) {
 }
 
 object Hyperspace {
-  private lazy val context = new ThreadLocal[HyperspaceContext]
+  // Maintain HyperspaceContext for each Spark session, since Hyperspace depends on
+  // the session's properties such as configs, etc.
+  private[hyperspace] lazy val contexts =
+    mutable.WeakHashMap[SparkSession, ThreadLocal[HyperspaceContext]]()
```
imback82, line 174: If the key is `SparkSession` and you are synchronizing on `contexts`, do you still need the value to be `ThreadLocal`?

Is this to support the scenario where two different threads share the same spark session? But even in this case, it seems that it will keep resetting if two threads call `getContext` back to back?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -16,6 +16,8 @@
 
 package com.microsoft.hyperspace
 
+import scala.collection.{mutable, _}
```
imback82, line 19: what is `_` here? It may be better to just spell it out for `scala.collection` case.

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
sezruby, line 174: @imback82 is this ThreadLocal required? if there are several threads holding the same spark session, then yes?
I'm bit confused about previous use - `ThreadLocal` in `object`. If that is working properly, is it ok to use the previous ver. when each thread has its own spark session?

+ I tested weakPtr behavior (with System.gc + waiting), but it seems Spark (sharedState?) holds the reference of each session even after I removed the reference in test code; `WeakHashMap` might be ineffective.

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
sezruby: You mean ` scala.collection.mutable.WeakHashMap[` this way?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
sezruby: > Is this to support the scenario where two different threads share the same spark session? 

I'm not sure about thread safety. One thread per session limitation might be good to have.

> But even in this case, it seems that it will keep resetting if two threads call getContext back to back?

I thought thread local variables stored in each thread context, so it would be ok..

imback82: > I thought thread local variables stored in each thread context, so it would be ok..

But you are sharing a key in a map that's not thread local?

imback82: > One thread per session limitation might be good to have.

We need to store more info then?

imback82: So, `SparkSession` created from `newSession` can never be GC-ed? If that's the case, we don't need to use `WeakHashMap`.

sezruby: Then how about (Thread.currentThread().getId, SparkSession) as a key?

imback82: I think we can't "enforce" if the thread id is in the key. We can put it in the value `(Thread.currentThread().getId, HyperspaceContext)`, and throw an exception if current thread id is mismatching (and do not need to worry about `ThreadLocal` 😄 ).

Also, we need to limit the map size to say 16? To prevent the scenario where the user application can keep calling `newSession`. We can just throw an exception if it goes over the limit. WDYT?

sezruby: Yep I was waiting more than 30 mins, but no gc-ed. I guess it's because of statistics or history info. 
Then I'll replace it with ConcurrentHashMap

sezruby: Yea - number of sessions limitation might be required but it can be configurable. 
And we could GC in Hyperspace (for every 5 mins?) - by ` spark.sparkContext.isStopped` and remove it. WDYT?

imback82: I think we can just go simple and throw an exception if it goes over the limit (and the exception message can suggest to increase the limit if necessary).

imback82: sgtm

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -166,18 +171,30 @@ class Hyperspace(spark: SparkSession) {
...
+    if (!ctx.isDefined) {
+      contexts.synchronized {
+        if (contexts.size >= HyperspaceConf.maxSupportedSparkSessions(spark)) {
+          throw HyperspaceException("Too many Spark sessions uses Hyperspace." +
+            "Please increase the config: spark.hyperspace.maxSessions")
+        }
+        contexts.put(spark, (Thread.currentThread().getId, new HyperspaceContext(spark)))
+      }
+    } else if (ctx.get._1 != Thread.currentThread().getId) {
+      throw HyperspaceException(s"Hyperspace does not support multiple threads " +
```
sezruby, line 193: @imback82 
I guess this is too restricted.. How about writing some warning log and creating a new context in this case?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,31 @@ class Hyperspace(spark: SparkSession) {
...
-    if (ctx == null || !ctx.spark.equals(spark)) {
-      // Ensure the Spark session the caller provides is the same as
-      // the one HyperspaceContext is using because Hyperspace depends on the
-      // session's properties such as configs, etc.
-      context.set(new HyperspaceContext(spark))
+    val ctx = contexts.get(spark)
+    if (!ctx.isDefined) {
+      contexts.synchronized {
+        if (contexts.size >= HyperspaceConf.maxSupportedSparkSessions(spark)) {
+          throw HyperspaceException("Too many Spark sessions uses Hyperspace." +
```
imback82, line 187:
```suggestion
          throw HyperspaceException("Too many Spark sessions are using Hyperspace. " +
```

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,31 @@ class Hyperspace(spark: SparkSession) {
...
+            "Please increase the config: spark.hyperspace.maxSessions")
+        }
+        contexts.put(spark, (Thread.currentThread().getId, new HyperspaceContext(spark)))
```
imback82, line 190: Store `Thread.currentThread().getId` in a variable?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,31 @@ class Hyperspace(spark: SparkSession) {
...
+      }
+    } else if (ctx.get._1 != Thread.currentThread().getId) {
+      logWarning(s"Hyperspace is not thread safe for threads using one Spark session. " +
+        "Please be aware of it. Current thread id: ${Thread.currentThread().getId}, " +
```
imback82, line 194: What is the implication of potentially sharing the `HyperspaceContext` by threads if we already know that it's not thread safe? Isn't it better to fail faster here instead of failing somewhere else?

src/test/scala/com/microsoft/hyperspace/index/plananalysis/ExplainTest.scala
```diff
@@ -36,6 +36,7 @@ class ExplainTest extends SparkFunSuite with HyperspaceSuite {
 
   override def beforeAll(): Unit = {
     super.beforeAll()
+    Hyperspace.contexts.clear
```
imback82, line 39: should we expose a `def` to clear instead of exposing `contexts` directly?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,31 @@ class Hyperspace(spark: SparkSession) {
...
+        s"Previous thread id: ${ctx.get._1}")
+      contexts.put(spark, (Thread.currentThread().getId, ctx.get._2))
     }
-
-    context.get()
+    contexts.get(spark).get._2
```
imback82, line 198: Instead of looking up again, you can do the following?
```scala
    val ctx = contexts.get(spark)
    if (!ctx.isDefined) {
      contexts.synchronized {
        if (contexts.size >= HyperspaceConf.maxSupportedSparkSessions(spark)) {
          throw HyperspaceException("")
        }
        val newCtxt = new HyperspaceContext(spark)
        contexts.put(spark, (Thread.currentThread().getId, newCtxt))
        newCtxt
      }
    } else {
      if (ctx.get._1 != Thread.currentThread().getId) {
        logWarning(s"")
        contexts.put(spark, (Thread.currentThread().getId, ctx.get._2))
      }
      ctx.get._2
    }
```

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -52,6 +52,10 @@ object IndexConstants {
   val INDEX_FILTER_RULE_USE_BUCKET_SPEC = "spark.hyperspace.index.filterRule.useBucketSpec"
   val INDEX_FILTER_RULE_USE_BUCKET_SPEC_DEFAULT = "false"
 
+  // Max. number of Spark sessions that Hyperspace supports.
+  val MAX_NUM_SPARK_SESSIONS = "spark.hyperspace.maxSessions"
```
imback82, line 56: `maxNumAllowedConcurrentSparkSessions`?

src/test/scala/com/microsoft/hyperspace/HyperspaceTest.scala
```diff
@@ -85,4 +89,39 @@ class HyperspaceTest extends HyperspaceSuite {
...
+    val newCtx = Hyperspace.getContext(newSession)
+    assert(newCtx.indexCollectionManager.indexes.count == 0)
+    val ctx2 = Hyperspace.getContext(spark)
+
+    // Check HyperspaceContext instances.
+    assert(ctx.eq(ctx2))
+    assert(ctx.ne(newCtx))
+
+    // Test using a different thread.
+    val f: Future[HyperspaceContext] = Future {
```
imback82, line 118: no type info for a local var. You can also simplify as following if you want:
```scala
Future(Hyperspace.getContext(newSession)).onComplete {
  case Success(newCtx2) => assert(newCtx.eq(newCtx2))
  case Failure(e) => assert(false, e.getMessage)
}
```

src/test/scala/com/microsoft/hyperspace/HyperspaceTest.scala
```diff
@@ -85,4 +89,39 @@ class HyperspaceTest extends HyperspaceSuite {
...
+      Hyperspace.getContext(newSession)
+    }
+
+    f onComplete {
+      case Success(newCtx2) => assert(newCtx.eq(newCtx2))
+      case Failure(e) => assert(false, e.getMessage)
```
imback82, line 124:
```suggestion
      case Failure(e) => fail(e.getMessage)
```

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
imback82: It will be useful to print `contexts.size` as well.

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
sezruby: checkAnswer won't work if we throw an exception here. No idea how to fix it and one possible scenario:
- setup & check in main thread
- run with a worker thread

imback82: are you still blocked on this?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,40 @@ class Hyperspace(spark: SparkSession) {
...
+            "Please increase the config `spark.hyperspace.maxNumAllowedConcurrentSparkSessions`. " +
+            s"Current limit: ${contexts.size}")
+        }
+        val newCtx = new HyperspaceContext(spark)
+        contexts.put(spark, (threadId, newCtx))
+        newCtx
+      }
+    } else {
+      if (ctx.get._1 != threadId) {
+        logWarning(s"Hyperspace is not thread safe for threads using one Spark session. " +
```
imback82, line 198:
```suggestion
        logWarning("Hyperspace is not thread safe for threads using one Spark session. " +
```

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,40 @@ class Hyperspace(spark: SparkSession) {
...
+          s"Please be aware of it. Current thread id: $threadId, Previous thread id: ${ctx.get._1}")
+        contexts.put(spark, (threadId, ctx.get._2))
+      }
+      ctx.get._2
     }
+  }
 
-    context.get()
+  def clearHyperspaceContexts(): Unit = {
```
imback82, line 206: just `clear` seems fine? But, I am also fine if you think `clearHyperspaceContexts` is better.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -52,6 +52,11 @@ object IndexConstants {
   val INDEX_FILTER_RULE_USE_BUCKET_SPEC = "spark.hyperspace.index.filterRule.useBucketSpec"
   val INDEX_FILTER_RULE_USE_BUCKET_SPEC_DEFAULT = "false"
 
+  // Max. number of Spark sessions that Hyperspace allows.
```
imback82, line 55:
```suggestion
  // The max number of concurrent Spark sessions that Hyperspace allows.
```

src/main/scala/com/microsoft/hyperspace/util/HyperspaceConf.scala
```diff
@@ -97,6 +97,13 @@ object HyperspaceConf {
         "avro,csv,json,orc,parquet,text")
   }
 
+  def maxAllowedSparkSessions(spark: SparkSession): Int = {
```
imback82, line 100:
```suggestion
  def maxNumAllowedConcurrentSparkSessions(spark: SparkSession): Int = {
```

src/test/scala/com/microsoft/hyperspace/index/plananalysis/ExplainTest.scala
```diff
@@ -36,6 +36,7 @@ class ExplainTest extends SparkFunSuite with HyperspaceSuite {
 
   override def beforeAll(): Unit = {
     super.beforeAll()
+    Hyperspace.clearHyperspaceContexts()
```
imback82, line 39: Is this needed? It seems that `HyperspaceSuite.beforeAll` will be called, no?

apoorvedave1 (dismissed):
LGTM, thanks 👍

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
imback82: gentle ping

sezruby: I think throwing exception is too restrictive. Users might not know how to fix the issue ..?

imback82: How about this?

> Isn't it better to fail faster here instead of failing somewhere else?

You can't suggest this is not supported? Being restrictive seems better than failing at a random place because Hyperspace is not thread safe?

sezruby: I thought it's user's responsibility to use the API which is not thread safe. 
Different thread id not always means they are running concurrently.. 
I think we need documentation rather than restricting the use case?

sezruby: and I got the following error after throwing an exception here ..  ?
```
[info] - Verify createIndex and refreshIndex on Delta Lake table. *** FAILED ***
[info]   com.microsoft.hyperspace.HyperspaceException: Only creating index over HDFS file based scan nodes is supported.
[info]   at com.microsoft.hyperspace.actions.CreateAction.validate(CreateAction.scala:49)
[info]   at com.microsoft.hyperspace.actions.Action.run(Action.scala:88)
[info]   at com.microsoft.hyperspace.actions.Action.run$(Action.scala:83)
[info]   at com.microsoft.hyperspace.actions.CreateAction.run(CreateAction.scala:30)
[info]   at com.microsoft.hyperspace.index.IndexCollectionManager.create(IndexCollectionManager.scala:47)
[info]   at com.microsoft.hyperspace.index.CachingIndexCollectionManager.create(CachingIndexCollectionManager.scala:78)
[info]   at com.microsoft.hyperspace.Hyperspace.createIndex(Hyperspace.scala:49)
[info]   at com.microsoft.hyperspace.index.DeltaLakeIntegrationTest.$anonfun$new$4(DeltaLakeIntegrationTest.scala:66)
[info]   at com.microsoft.hyperspace.index.DeltaLakeIntegrationTest.$anonfun$new$4$adapted(DeltaLakeIntegrationTest.scala:60)
[info]   at com.microsoft.hyperspace.index.HyperspaceSuite.withTempPathAsString(HyperspaceSuite.scala:120)
```

imback82: > Different thread id not always means they are running concurrently..

But the new structure allows multiple threads accessing the object concurrently? The existing implementation makes sure only one to one mapping via thread local. So, I would keep this condition as it is, and if there are demands for accessing the hyperspace object from multiple threads, we should think about making it thread-safe instead of documenting.

Plus, I think random failures are much worse and hard to debug, esp. failures caused by the thread safety.

sezruby: I agree on how difficult analyzing failures from concurrent threads. 
I added the exception change, then this can be a breaking change for some use case.

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,41 @@ class Hyperspace(spark: SparkSession) {
...
+        }
+        val newCtx = new HyperspaceContext(spark)
+        contexts.put(spark, (threadId, newCtx))
+        newCtx
+      }
+    } else {
+      if (ctx.get._1 != threadId) {
+        // scalastyle:off
+        println("here")
+        throw HyperspaceException("Hyperspace does not support concurrent threads using one " +
```
imback82, line 200: How about: "Hyperspace does not support multiple threads sharing the same Spark session. Please create a Spark session per thread. Current thread id ..."?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,41 @@ class Hyperspace(spark: SparkSession) {
...
+            s"Current limit: ${contexts.size}")
+        }
+        val newCtx = new HyperspaceContext(spark)
+        contexts.put(spark, (threadId, newCtx))
+        newCtx
+      }
+    } else {
+      if (ctx.get._1 != threadId) {
+        // scalastyle:off
+        println("here")
```
imback82, line 199: Shall we remove this?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,41 @@ class Hyperspace(spark: SparkSession) {
...
-      // the one HyperspaceContext is using because Hyperspace depends on the
-      // session's properties such as configs, etc.
-      context.set(new HyperspaceContext(spark))
+    val ctx = contexts.get(spark)
+    val threadId = Thread.currentThread().getId
+    if (!ctx.isDefined) {
+      contexts.synchronized {
+        if (contexts.size >= HyperspaceConf.maxNumAllowedConcurrentSparkSessions(spark)) {
+          throw HyperspaceException("Too many Spark sessions are using Hyperspace. " +
+            "Please increase the config `spark.hyperspace.maxNumAllowedConcurrentSparkSessions`. " +
```
imback82, line 189: you can use `IndexConstants.MAX_NUM_ALLOWED_CONCURRENT_SPARK_SESSIONS`?

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -21,6 +21,7 @@ import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.{DataFrame, QueryTest}
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.execution.datasources._
+import org.apache.spark.sql.internal.SQLConf
```
imback82, line 24: is this needed?

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -48,6 +48,7 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
   }
 
   before {
+    hyperspace = new Hyperspace(spark)
```
imback82, line 51: why is this change needed (guide comment would help)?

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -200,10 +201,10 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
             s"${indexConfig.indexName}/v__=0" +: appendedFiles: _*)
 
           // Verify correctness of results.
-          spark.disableHyperspace()
-          val dfWithHyperspaceDisabled = query()
           spark.enableHyperspace()
           val dfWithHyperspaceEnabled = query()
+          spark.disableHyperspace()
+          val dfWithHyperspaceDisabled = query()
```
imback82, line 207: why is this change needed (guide comment would help)?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
imback82: Added breaking change label so we can document it in the release notes.

imback82:
Could you update the PR description as it seems out of date? Basically, we are preventing hyperspace context from being re-created in certain scenarios, and making sure too many contexts are not created.

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
imback82: btw, even if this is a breaking change, it's alerting the user a possible misuse where hyperspace context is being recreated, I think it's worthwhile.

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
sezruby: Sorry I forgot to remove debug code. 😅

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -242,10 +242,12 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
             spark.disableHyperspace()
             val dfWithHyperspaceDisabled = query()
             val basePlan = dfWithHyperspaceDisabled.queryExecution.optimizedPlan
+            val resultDisabled = dfWithHyperspaceDisabled.collect().toSeq.toSet
             spark.enableHyperspace()
             val dfWithHyperspaceEnabled = query()
             assert(!basePlan.equals(dfWithHyperspaceEnabled.queryExecution.optimizedPlan))
-            checkAnswer(dfWithHyperspaceDisabled, dfWithHyperspaceEnabled)
+            val resultEnabled = dfWithHyperspaceEnabled.collect().toSeq.toSet
+            assert(resultEnabled.equals(resultDisabled))
```
imback82, line 250: is this change still needed? if so, why?

sezruby: In checkAnswer, a new thread tries to build the query plan again - so it causes the exception.
No idea why other `checkAnswer`s have no problem with it..

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -242,9 +242,13 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
...
             val basePlan = dfWithHyperspaceDisabled.queryExecution.optimizedPlan
+            val resultDisabled = dfWithHyperspaceDisabled.collect().toSeq.sortBy(_.hashCode())
             spark.enableHyperspace()
             val dfWithHyperspaceEnabled = query()
-            assert(!basePlan.equals(dfWithHyperspaceEnabled.queryExecution.optimizedPlan))
+            val updatedPlan = dfWithHyperspaceEnabled.queryExecution.optimizedPlan
+            val resultEnabled = dfWithHyperspaceEnabled.collect().toSeq.sortBy(_.hashCode())
+            assert(!basePlan.equals(updatedPlan))
+            assert(resultEnabled.equals(resultDisabled))
             checkAnswer(dfWithHyperspaceDisabled, dfWithHyperspaceEnabled)
```
sezruby, line 252: After executing collect(), checkAnswer doesn't throw the exception :anguished:

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
imback82: oh interesting. could you share the code where a new thread is spawned in `checkAnswer`?

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -165,19 +171,42 @@ class Hyperspace(spark: SparkSession) {
...
+    if (!ctx.isDefined) {
+      contexts.synchronized {
+        if (contexts.size >= HyperspaceConf.maxNumAllowedConcurrentSparkSessions(spark)) {
+          throw HyperspaceException(
+            "Too many Spark sessions are using Hyperspace. Please increase the config `" +
+              IndexConstants.MAX_NUM_ALLOWED_CONCURRENT_SPARK_SESSIONS + "`. " +
+              s"Current limit: ${contexts.size}")
+        }
+        val newCtx = new HyperspaceContext(spark)
+        contexts.put(spark, (threadId, newCtx))
```
imback82, line 194: Should we check the return value of `put` and throw an exception if it returns `Some`? There could be multiple threads sharing the same spark session that hits `if (!ctx.isDefined)` at the same time.

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
sezruby: It's because of broadcast join..

```
[info]          at org.apache.spark.sql.execution.exchange.BroadcastExchangeExec.doExecuteBroadcast(BroadcastExchangeExec.scala:146)
[info]          at org.apache.spark.sql.execution.InputAdapter.doExecuteBroadcast(WholeStageCodegenExec.scala:387)
[info]          at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeBroadcast$1(SparkPlan.scala:144)
[info]          at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:155)
[info]          at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
[info]          at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:152)
[info]          at org.apache.spark.sql.execution.SparkPlan.executeBroadcast(SparkPlan.scala:140)
[info]          at org.apache.spark.sql.execution.joins.BroadcastHashJoinExec.prepareBroadcast(BroadcastHashJoinExec.scala:117)
[info]          at org.apache.spark.sql.execution.joins.BroadcastHashJoinExec.codegenInner(BroadcastHashJoinExec.scala:211)
[info]          at org.apache.spark.sql.execution.joins.BroadcastHashJoinExec.doConsume(BroadcastHashJoinExec.scala:101)
```

failed thread stack:
```
..
[info]          at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.transformDown(LogicalPlan.scala:29)
[info]          at com.microsoft.hyperspace.index.rules.FilterIndexRule$.apply(FilterIndexRule.scala:52)
[info]          at com.microsoft.hyperspace.index.rules.FilterIndexRule$.apply(FilterIndexRule.scala:38)
[info]          at org.apache.spark.sql.catalyst.rules.RuleExecutor.$anonfun$execute$2(RuleExecutor.scala:87)
[info]          at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:126)
[info]          at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:122)
[info]          at scala.collection.immutable.List.foldLeft(List.scala:89)
[info]          at org.apache.spark.sql.catalyst.rules.RuleExecutor.$anonfun$execute$1(RuleExecutor.scala:84)
[info]          at org.apache.spark.sql.catalyst.rules.RuleExecutor.$anonfun$execute$1$adapted(RuleExecutor.scala:76)
[info]          at scala.collection.immutable.List.foreach(List.scala:392)
[info]          at org.apache.spark.sql.catalyst.rules.RuleExecutor.execute(RuleExecutor.scala:76)
[info]          at org.apache.spark.sql.execution.QueryExecution.optimizedPlan$lzycompute(QueryExecution.scala:66)
[info]          at org.apache.spark.sql.execution.QueryExecution.optimizedPlan(QueryExecution.scala:66)
[info]          at org.apache.spark.sql.execution.QueryExecution.sparkPlan$lzycompute(QueryExecution.scala:72)
[info]          at org.apache.spark.sql.execution.QueryExecution.sparkPlan(QueryExecution.scala:68)
[info]          at org.apache.spark.sql.execution.QueryExecution.executedPlan$lzycompute(QueryExecution.scala:77)
[info]          at org.apache.spark.sql.execution.QueryExecution.executedPlan(QueryExecution.scala:77)
[info]          at org.apache.spark.sql.Dataset.withAction(Dataset.scala:3359)
[info]          at org.apache.spark.sql.Dataset.collect(Dataset.scala:2782)
[info]          at org.apache.spark.sql.delta.PartitionFiltering.filesForScan(PartitionFiltering.scala:40)
[info]          at org.apache.spark.sql.delta.PartitionFiltering.filesForScan$(PartitionFiltering.scala:27)
[info]          at org.apache.spark.sql.delta.Snapshot.filesForScan(Snapshot.scala:52)
[info]          at org.apache.spark.sql.delta.files.TahoeLogFileIndex.matchingFiles(TahoeFileIndex.scala:140)
[info]          at org.apache.spark.sql.delta.files.TahoeFileIndex.listFiles(TahoeFileIndex.scala:56)
[info]          at org.apache.spark.sql.execution.FileSourceScanExec.selectedPartitions$lzycompute(DataSourceScanExec.scala:193)
[info]          at org.apache.spark.sql.execution.FileSourceScanExec.selectedPartitions(DataSourceScanExec.scala:190)
[info]          at org.apache.spark.sql.execution.FileSourceScanExec.updateDriverMetrics(DataSourceScanExec.scala:529)
[info]          at org.apache.spark.sql.execution.FileSourceScanExec.inputRDD$lzycompute(DataSourceScanExec.scala:307)
[info]          at org.apache.spark.sql.execution.FileSourceScanExec.inputRDD(DataSourceScanExec.scala:305)
[info]          at org.apache.spark.sql.execution.FileSourceScanExec.inputRDDs(DataSourceScanExec.scala:327)
[info]          at org.apache.spark.sql.execution.FilterExec.inputRDDs(basicPhysicalOperators.scala:121)
[info]          at org.apache.spark.sql.execution.ProjectExec.inputRDDs(basicPhysicalOperators.scala:41)
[info]          at org.apache.spark.sql.execution.WholeStageCodegenExec.doExecute(WholeStageCodegenExec.scala:627)
```

sezruby: Seems we shouldn't throw the exception .. ?

imback82: Hmm, then we have to think about this feature again, since it may hit into thread-safety issue even if the user didn't intend it?

imback82: For example, this could have gone unnoticed if we didn't throw the exception?

sezruby: Seems so. How about ThreadLocal[HashMap[Session, Context]]?

imback82: But that would allow multiple threads share the same Hyperspace context.

imback82: So, basically in Delta code, it does `allFiles.toDF()` and eventually that dataframe is `collect`-ed, which triggers our optimizer. In our optimizer, we have an extractor that calls `Hyperspace.getContext(spark).sourceProviderManager` in `ExtractRelation`. So instead of `case filter @ Filter(condition: Expression, ExtractRelation(relation))`, we could do `case filter @ Filter if isSupportedRelation`, and I think we can avoid the issue.

But, I think this is still hacky and the right way to fix it seems like making the context thread-safe.

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -175,6 +176,10 @@ object Hyperspace extends ActiveSparkSession {
       // the one HyperspaceContext is using because Hyperspace depends on the
       // session's properties such as configs, etc.
       context.set(new HyperspaceContext(spark))
+      if (!ctx.spark.equals(spark)) {
```
sezruby, line 179: @imback82 I'm thinking of this change & what can we do about this. 
I think this is the problem - we access using the active spark session when optimizing query plan, but it can be different from the session which is used to create Hyperspace object.
So.. how about 
1) not resetting hyperspace & just using the previous one even with different active spark session. 
2) if a user wants to use another configs/spark session, they need to redefine Hyperspace object. 

I think this is clearer behavior because we do `val hs = new Hyperspace(spark)`

imback82: > 1. not resetting hyperspace & just using the previous one even with different active spark session.

Will it be ok to silently use a different active session?

PR #375, opened 2021-03-07
Fix Iceberg lineage for windows 

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a
 
Fix for inconsistent result of Iceberg input_file_name() between Windows & Linux

Yes, fix iceberg lineage related bug on Windows

tested on local

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
@@ -129,8 +129,14 @@ class IcebergRelation(spark: SparkSession, override val plan: DataSourceV2Relati
    * @return List of pairs of (file path, file id).
    */
   override def lineagePairs(fileIdTracker: FileIdTracker): Seq[(String, Long)] = {
+    // For Windows,
+    //   original file path: file:/C:/path/to/file
+    //   input_file_name(): C:/path/to/file
+    // For Linux,
+    //   original file path: file:///path/to/file
+    //   input_file_name(): /path/to/file
     fileIdTracker.getFileToIdMap.toSeq.map { kv =>
-      (kv._1._1.replaceAll("^file:/{1,3}", ""), kv._2)
+      (kv._1._1.replaceAll("^file:/{1,2}", ""), kv._2)
     }
   }
 
--- a/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
@@ -124,8 +124,10 @@ class IcebergRelation(spark: SparkSession, override val plan: DataSourceV2Relati
    * Returns list of pairs of (file path, file id) to build lineage column.
    *
    * File paths should be the same format as "input_file_name()" of the given relation type.
+   * input_file_name() could be different depending on the OS and source.
+   *
    * For [[IcebergRelation]], each file path should be in this format:
-   *   `/path/to/file` or `X:/path/to/file`
+   *   `/path/to/file` or `X:/path/to/file` for Windows file system.
    *
    * @param fileIdTracker [[FileIdTracker]] to create the list of (file path, file id).
    * @return List of pairs of (file path, file id).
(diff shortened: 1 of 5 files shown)


src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -123,14 +125,26 @@ class IcebergRelation(spark: SparkSession, override val plan: DataSourceV2Relati
...
+    //   original file path: file:///path/to/file or file:/path/to/file
+    //   input_file_name(): /path/to/file
+    if (System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("windows")) {
+      fileIdTracker.getFileToIdMap.toSeq.map { kv =>
+        (kv._1._1.stripPrefix("file:/"), kv._2)
+      }
+    } else {
+      fileIdTracker.getFileToIdMap.toSeq.map { kv =>
+        (kv._1._1.replaceAll("^file:/{1,3}", "/"), kv._2)
+      }
```
andrei-ionescu, line 147:
```suggestion
    val isWindowsOS =
      System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("windows")
    val replacement = if (isWindowsOS) { "" } else { "/" }
    fileIdTracker.getFileToIdMap.toSeq.map { kv =>
      (kv._1._1.replaceAll("^file:/{1,3}", replacement), kv._2)
```

andrei-ionescu (approved):
Approved plus a suggestion

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -123,14 +125,26 @@ class IcebergRelation(spark: SparkSession, override val plan: DataSourceV2Relati
    *
    * File paths should be the same format as "input_file_name()" of the given relation type.
    * For [[IcebergRelation]], each file path should be in this format:
-   *   `/path/to/file`
+   *   `/path/to/file` or `X:/path/to/file`
```
imback82, line 128: `X:` is only for Windows filesystem (e.g., `file:/`), but not applicable for remote storage (e.g., `abfs://`)? If so, could you clarify this a bit in the comment? For example, we can say that we need to take care of a special care if the path URI's scheme is `file://` depending on the OS.

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -123,14 +125,26 @@ class IcebergRelation(spark: SparkSession, override val plan: DataSourceV2Relati
...
   override def lineagePairs(fileIdTracker: FileIdTracker): Seq[(String, Long)] = {
-    fileIdTracker.getFileToIdMap.toSeq.map { kv =>
-      (kv._1._1.replaceAll("^file:/{1,3}", "/"), kv._2)
+    // For Windows,
+    //   original file path: file:/C:/path/to/file
+    //   input_file_name(): C:/path/to/file
+    // For Linux,
+    //   original file path: file:///path/to/file or file:/path/to/file
+    //   input_file_name(): /path/to/file
+    if (System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("windows")) {
```
imback82, line 140: Does `Path.WINDOWS` work?

imback82 (commented):
LGTM (few minor comments) and I also verified on Windows. Thanks for the fix, @sezruby!

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
sezruby: I prefer the current one; I think it looks clearer for os specific impl.
Thanks!

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
sezruby: Added some comment.

src/test/scala/com/microsoft/hyperspace/index/HyperspaceSuite.scala
```diff
@@ -111,7 +111,8 @@ trait HyperspaceSuite extends SparkFunSuite with SparkInvolvedSuite {
     // "inheriting conflicting members" issue.
     val path = Utils.createTempDir()
     path.delete()
-    try f(path.toString)
+    val pathStr = new Path(path.toString).toString
```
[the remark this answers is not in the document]
imback82: Could you put the actual example in the comment (and please mention the Iceberg issue/example as well)? Just looking at the code, it's not easy to see how this creates "environment specific" path.

imback82 (approved):

imback82 (approved):

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
andrei-ionescu: I understand your preference... and in the same time I would contradict you saying " it looks clearer for os specific impl" - I don't think there can be anything clearer than `if (isWindowsOS)` construct that I proposed :).

imback82: either approach seems fine to me.

Issue #347, opened 2021-01-31
[PROPOSAL]: Index nested fields


andrei-ionescu:
## Problem description

This design proposal is for adding support for nested fields in both `indexedColumns` and `includedColumns`.

Currently, Hyperspace does not support nested fields like structs or structured arrays. 

Having the possibility to use nested fields with Hyperspace would be of great benefit to everybody.

## Goal

The user should be able to create and index with nested fields and use it in queries. 

For example:

```
hs.createIndex(
  df, 
  IndexConfig(
    "idx_nested", 
    indexedColumns = Seq("nested.nst.field1"), 
    includedColumns = Seq("id", "name", "nested.nst.field2")
  )
)
```

## Proposed solution

Using the same Hyperspace APIs indexes over nested fields should work as for other simple fields.

Given the `nestedDataset` dataset with schema

```
root
 |-- id: integer (nullable = true)
 |-- name: string (nullable = true)
 |-- nested: struct (nullable = true)
 |    |-- field1: string (nullable = true)
 |    |-- nst: struct (nullable = true)
 |    |    |-- field1: string (nullable = true)
 |    |    |-- field2: string (nullable = true)
```

and the following data

```
+---+-----+-----------------+
|id |name |nested           |
+---+-----+-----------------+
|2  |name2|[va2, [wa2, wb2]]|
|1  |name1|[va1, [wa1, wb1]]|
+---+-----+-----------------+
```

This is the way and index on `field1` field with other 2 projection fields can be created

```
hs.createIndex(
  nestedDataset, 
  IndexConfig(
    "idx_nested", 
    indexedColumns = Seq("nested.nst.field1"), 
    includedColumns = Seq("id", "name", "nested.nst.field2")))
```

Adding the support for nested fields impacts the following areas:
- Validate nested column names
- Modify the create index action 
- Modify the filter and rule index functions

## Creating the index

Given the dataset defined above with the listed data, after doing

```
hs.createIndex(
  df, 
  IndexConfig(
    "idx_nested", 
    indexedColumns = Seq("nested.nst.field1"), 
    includedColumns = Seq("id", "name", "nested.nst.field2")))
```

the following dataset will be created

```
root
 |-- nested__nst__field1: string (nullable = true)
 |-- id: integer (nullable = true)
 |-- name: string (nullable = true)
 |-- nested__nst__field2: string (nullable = true)
```

```
+-------------------+---+-----+-------------------+
[...]

PR #379, opened 2021-03-11
Support index creation on nested fields

andrei-ionescu:

 - **Proposal**: #347
 - **Discussion**: #312
 - Fixes: #347

This PR adds support for creating indexes over nested fields (ie: structs). 

Given the `nestedDataset` dataset with schema

```
root
 |-- name: string (nullable = true)
 |-- nested: struct (nullable = true)
 |    |-- field1: string (nullable = true)
 |    |-- nst: struct (nullable = true)
 |    |    |-- field1: string (nullable = true)
 |    |    |-- field2: string (nullable = true)
```

and the following data

```
+---+-----+-----------------+
|id |name |nested           |
+---+-----+-----------------+
|2  |name2|[va2, [wa2, wb2]]|
|1  |name1|[va1, [wa1, wb1]]|
+---+-----+-----------------+
```

Creating the index on nested fields would be:

```
hs.createIndex(
  nestedDataset, 
  IndexConfig(
    "idx_nested", 
    indexedColumns = Seq("nested.nst.field1"), 
    includedColumns = Seq("id", "name", "nested.nst.field2")))
```

The following index dataset will be created

```
root
 |-- __hs_nested.nested.nst.field1: string (nullable = true)
 |-- name: string (nullable = true)
 |-- __hs_nested.nested.nst.field2: string (nullable = true)
```

```
+-----------------------------+---+-----+-----------------------------+
|__hs_nested.nested.nst.field1| id| name|__hs_nested.nested.nst.field2|
+-----------------------------+---+-----+-----------------------------+
|                          wa1|  1|name1|                          wb1|
|                          wa2|  2|name2|                          wb2|
+-----------------------------+---+-----+-----------------------------+
```

It is important to understand that the name of the field of the index column is a non-nested column and it gets prefixed by `__hs_nested.` prefix to recognise that the field in the index data frame is for a nested field (ie: `nested.nst.field1` into `__hs_nested.nested.nst.field1`).

No.

- Existing unit and integration tests for not breaking the existing functionalities.
- Unit and integration test added for the new functionalities.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
@@ -19,7 +19,6 @@ package com.microsoft.hyperspace.actions
 import scala.util.Try
 
 import org.apache.spark.sql.{DataFrame, SparkSession}
-import org.apache.spark.sql.types.StructType
 
 import com.microsoft.hyperspace.{Hyperspace, HyperspaceException}
 import com.microsoft.hyperspace.actions.Constants.States.{ACTIVE, CREATING, DOESNOTEXIST}
@@ -51,7 +50,7 @@ class CreateAction(
     }
 
     // schema validity checks
-    if (!isValidIndexSchema(indexConfig, df.schema)) {
+    if (!isValidIndexSchema(indexConfig, df)) {
       throw HyperspaceException("Index config is not applicable to dataframe schema.")
     }
 
@@ -65,10 +64,13 @@ class CreateAction(
     }
   }
 
-  private def isValidIndexSchema(config: IndexConfig, schema: StructType): Boolean = {
-    // Resolve index config columns from available column names present in the schema.
+  private def isValidIndexSchema(config: IndexConfig, dataFrame: DataFrame): Boolean = {
+    // Resolve index config columns from available column names present in the dataframe.
     ResolverUtils
-      .resolve(spark, config.indexedColumns ++ config.includedColumns, schema.fieldNames)
+      .resolve(
+        spark,
+        config.indexedColumns ++ config.includedColumns,
+        dataFrame.queryExecution.analyzed)
       .isDefined
   }
 
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -19,13 +19,13 @@ package com.microsoft.hyperspace.actions
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.{DataFrame, SaveMode, SparkSession}
 import org.apache.spark.sql.catalyst.plans.logical.LeafNode
-import org.apache.spark.sql.functions.input_file_name
+import org.apache.spark.sql.functions.{col, input_file_name}
 
 import com.microsoft.hyperspace.{Hyperspace, HyperspaceException}
 import com.microsoft.hyperspace.index._
 import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
 import com.microsoft.hyperspace.index.sources.FileBasedRelation
-import com.microsoft.hyperspace.util.{HyperspaceConf, PathUtils, ResolverUtils}
+import com.microsoft.hyperspace.util.{HyperspaceConf, PathUtils, ResolverUtils, SchemaUtils}
 
 /**
  * CreateActionBase provides functionality to write dataframe as covering index.
@@ -127,9 +127,13 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
     val (indexDataFrame, resolvedIndexedColumns, _) =
       prepareIndexDataFrame(spark, df, indexConfig)
 
-    // run job
-    val repartitionedIndexDataFrame =
-      indexDataFrame.repartition(numBuckets, resolvedIndexedColumns.map(df(_)): _*)
+    // Run job
+    val repartitionedIndexDataFrame = {
+      // For nested fields, resolvedIndexedColumns will have flattened names with `.` (dots),
+      // thus they need to be enclosed in backticks to access them as top-level columns.
+      // Note that backticking the non-nested columns is a no-op.
+      indexDataFrame.repartition(numBuckets, resolvedIndexedColumns.map(c => col(s"`$c`")): _*)
+    }
 
     // Save the index with the number of buckets specified.
     repartitionedIndexDataFrame.write
@@ -156,21 +160,21 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       df: DataFrame,
       indexConfig: IndexConfig): (Seq[String], Seq[String]) = {
     val spark = df.sparkSession
-    val dfColumnNames = df.schema.fieldNames
+    val plan = df.queryExecution.analyzed
     val indexedColumns = indexConfig.indexedColumns
     val includedColumns = indexConfig.includedColumns
-    val resolvedIndexedColumns = ResolverUtils.resolve(spark, indexedColumns, dfColumnNames)
-    val resolvedIncludedColumns = ResolverUtils.resolve(spark, includedColumns, dfColumnNames)
+    val resolvedIndexedColumns = ResolverUtils.resolve(spark, indexedColumns, plan)
+    val resolvedIncludedColumns = ResolverUtils.resolve(spark, includedColumns, plan)
 
     (resolvedIndexedColumns, resolvedIncludedColumns) match {
       case (Some(indexed), Some(included)) => (indexed, included)
       case _ =>
         val unresolvedColumns = (indexedColumns ++ includedColumns)
-          .map(c => (c, ResolverUtils.resolve(spark, c, dfColumnNames)))
+          .map(c => (c, ResolverUtils.resolve(spark, Seq(c), plan)))
           .collect { case c if c._2.isEmpty => c._1 }
         throw HyperspaceException(
           s"Columns '${unresolvedColumns.mkString(",")}' could not be resolved " +
-            s"from available source columns '${dfColumnNames.mkString(",")}'")
+            s"from available source columns:\n${df.schema.treeString}")
     }
   }
 
@@ -180,6 +184,9 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       indexConfig: IndexConfig): (DataFrame, Seq[String], Seq[String]) = {
     val (resolvedIndexedColumns, resolvedIncludedColumns) = resolveConfig(df, indexConfig)
     val columnsFromIndexConfig = resolvedIndexedColumns ++ resolvedIncludedColumns
+
+    val escapedIndexedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIndexedColumns)
+    val escapedIncludedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIncludedColumns)
 
     val indexDF = if (hasLineage(spark)) {
       val relation = getRelation(spark, df)
@@ -214,10 +221,13 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
         .select(
           allIndexColumns.head,
           allIndexColumns.tail :+ IndexConstants.DATA_FILE_NAME_ID: _*)
+        .toDF(escapedIndexedColumns ++ escapedIncludedColumns ++ missingPartitionColumns :+
+          IndexConstants.DATA_FILE_NAME_ID: _*)
     } else {
       df.select(columnsFromIndexConfig.head, columnsFromIndexConfig.tail: _*)
+        .toDF(escapedIndexedColumns ++ escapedIncludedColumns : _*)
     }
 
-    (indexDF, resolvedIndexedColumns, resolvedIncludedColumns)
+    (indexDF, escapedIndexedColumns, escapedIncludedColumns)
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -17,7 +17,7 @@
 package com.microsoft.hyperspace.actions
 
 import org.apache.hadoop.fs.Path
-import org.apache.spark.sql.{DataFrame, SaveMode, SparkSession}
+import org.apache.spark.sql.{Column, DataFrame, SaveMode, SparkSession}
 import org.apache.spark.sql.catalyst.plans.logical.LeafNode
 import org.apache.spark.sql.functions.{col, input_file_name}
 
@@ -104,23 +104,6 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
     }
   }
 
-  private def hasParquetAsSourceFormatProperty(
-      relation: FileBasedRelation): Option[(String, String)] = {
-    if (relation.hasParquetAsSourceFormat) {
-      Some(IndexConstants.HAS_PARQUET_AS_SOURCE_FORMAT_PROPERTY -> "true")
-    } else {
-      None
-    }
-  }
-
-  private def hasLineageProperty(spark: SparkSession): Option[(String, String)] = {
-    if (hasLineage(spark)) {
-      Some(IndexConstants.LINEAGE_PROPERTY -> "true")
-    } else {
-      None
-    }
-  }
-
   protected def write(spark: SparkSession, df: DataFrame, indexConfig: IndexConfig): Unit = {
     val numBuckets = numBucketsForIndex(spark)
 
@@ -156,9 +139,26 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
     relations.head
   }
 
+  private def hasParquetAsSourceFormatProperty(
+      relation: FileBasedRelation): Option[(String, String)] = {
+    if (relation.hasParquetAsSourceFormat) {
+      Some(IndexConstants.HAS_PARQUET_AS_SOURCE_FORMAT_PROPERTY -> "true")
+    } else {
+      None
+    }
+  }
+
+  private def hasLineageProperty(spark: SparkSession): Option[(String, String)] = {
+    if (hasLineage(spark)) {
+      Some(IndexConstants.LINEAGE_PROPERTY -> "true")
+    } else {
+      None
+    }
+  }
+
   private def resolveConfig(
       df: DataFrame,
-      indexConfig: IndexConfig): (Seq[String], Seq[String]) = {
+      indexConfig: IndexConfig): (Seq[(String, Boolean)], Seq[(String, Boolean)]) = {
     val spark = df.sparkSession
     val plan = df.queryExecution.analyzed
     val indexedColumns = indexConfig.indexedColumns
@@ -170,8 +170,8 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       case (Some(indexed), Some(included)) => (indexed, included)
       case _ =>
         val unresolvedColumns = (indexedColumns ++ includedColumns)
-          .map(c => (c, ResolverUtils.resolve(spark, Seq(c), plan)))
-          .collect { case c if c._2.isEmpty => c._1 }
+          .map(c => (c, ResolverUtils.resolve(spark, Seq(c), plan).map(_.map(_._1))))
+          .collect { case (c, r) if r.isEmpty => c }
         throw HyperspaceException(
           s"Columns '${unresolvedColumns.mkString(",")}' could not be resolved " +
             s"from available source columns:\n${df.schema.treeString}")
@@ -183,10 +183,12 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       df: DataFrame,
       indexConfig: IndexConfig): (DataFrame, Seq[String], Seq[String]) = {
     val (resolvedIndexedColumns, resolvedIncludedColumns) = resolveConfig(df, indexConfig)
-    val columnsFromIndexConfig = resolvedIndexedColumns ++ resolvedIncludedColumns
+    val columnsFromIndexConfig =
+      resolvedIndexedColumns.map(_._1) ++ resolvedIncludedColumns.map(_._1)
 
-    val escapedIndexedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIndexedColumns)
-    val escapedIncludedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIncludedColumns)
+    val prefixedIndexedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIndexedColumns)
+    val prefixedIncludedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIncludedColumns)
+    val prefixedColumnsFromIndexConfig = prefixedIndexedColumns ++ prefixedIncludedColumns
 
     val indexDF = if (hasLineage(spark)) {
       val relation = getRelation(spark, df)
@@ -215,19 +217,25 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       val dataPathColumn = "_data_path"
       val lineagePairs = relation.lineagePairs(fileIdTracker)
       val lineageDF = lineagePairs.toDF(dataPathColumn, IndexConstants.DATA_FILE_NAME_ID)
+      val prefixedAllIndexColumns = prefixedColumnsFromIndexConfig ++ missingPartitionColumns
 
       df.withColumn(dataPathColumn, input_file_name())
         .join(lineageDF.hint("broadcast"), dataPathColumn)
-        .select(
-          allIndexColumns.head,
-          allIndexColumns.tail :+ IndexConstants.DATA_FILE_NAME_ID: _*)
-        .toDF(escapedIndexedColumns ++ escapedIncludedColumns ++ missingPartitionColumns :+
-          IndexConstants.DATA_FILE_NAME_ID: _*)
+        .select(prepareColumns(allIndexColumns, prefixedAllIndexColumns) :+
+          col(IndexConstants.DATA_FILE_NAME_ID): _*)
     } else {
-      df.select(columnsFromIndexConfig.head, columnsFromIndexConfig.tail: _*)
-        .toDF(escapedIndexedColumns ++ escapedIncludedColumns : _*)
+      df.select(prepareColumns(columnsFromIndexConfig, prefixedColumnsFromIndexConfig): _*)
     }
 
-    (indexDF, escapedIndexedColumns, escapedIncludedColumns)
+    (indexDF, prefixedIndexedColumns, prefixedIncludedColumns)
+  }
+
+  private def prepareColumns(
+      originalColumns: Seq[String],
+      prefixedColumns: Seq[String]): Seq[Column] = {
+    originalColumns.zip(prefixedColumns).map {
+      case (original, prefixed) =>
+        col(original).as(prefixed)
+    }
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -93,8 +93,10 @@ private[actions] abstract class RefreshActionBase(
     val ddColumns = previousIndexLogEntry.derivedDataset.properties.columns
     IndexConfig(
       previousIndexLogEntry.name,
-      SchemaUtils.removePrefixNestedFieldNames(ddColumns.indexed),
-      SchemaUtils.removePrefixNestedFieldNames(ddColumns.included))
+      // As indexed & included columns in previousLogEntry are resolved & prefixed names,
+      // need to remove the prefix to resolve with the dataframe for refresh.
+      SchemaUtils.removePrefixNestedFieldNames(ddColumns.indexed).map(_._1),
+      SchemaUtils.removePrefixNestedFieldNames(ddColumns.included).map(_._1))
   }
 
   final override val transientState: String = REFRESHING
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
@@ -22,6 +22,7 @@ import org.apache.spark.sql.types.{DataType, StructType}
 import com.microsoft.hyperspace.{Hyperspace, HyperspaceException}
 import com.microsoft.hyperspace.actions.Constants.States.{ACTIVE, REFRESHING}
 import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.util.SchemaUtils
 
 /**
  * Base abstract class containing common code for different types of index refresh actions.
@@ -90,7 +91,10 @@ private[actions] abstract class RefreshActionBase(
 
   protected lazy val indexConfig: IndexConfig = {
     val ddColumns = previousIndexLogEntry.derivedDataset.properties.columns
-    IndexConfig(previousIndexLogEntry.name, ddColumns.indexed, ddColumns.included)
+    IndexConfig(
+      previousIndexLogEntry.name,
+      SchemaUtils.removePrefixNestedFieldNames(ddColumns.indexed),
+      SchemaUtils.removePrefixNestedFieldNames(ddColumns.included))
   }
 
   final override val transientState: String = REFRESHING
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
@@ -23,6 +23,7 @@ import com.microsoft.hyperspace.{Hyperspace, HyperspaceException}
 import com.microsoft.hyperspace.index._
 import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
 import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEvent, RefreshIncrementalActionEvent}
+import com.microsoft.hyperspace.util.SchemaUtils
 
 /**
  * Action to refresh indexes with newly appended files and deleted files in an incremental way.
@@ -91,7 +92,7 @@ class RefreshIncrementalAction(
         refreshDF,
         indexDataPath.toString,
         previousIndexLogEntry.numBuckets,
-        indexConfig.indexedColumns,
+        SchemaUtils.prefixNestedFieldNames(indexConfig.indexedColumns),
         writeMode)
     }
   }
--- a/src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
@@ -23,7 +23,6 @@ import com.microsoft.hyperspace.{Hyperspace, HyperspaceException}
 import com.microsoft.hyperspace.index._
 import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
 import com.microsoft.hyperspace.telemetry.{AppInfo, HyperspaceEvent, RefreshIncrementalActionEvent}
-import com.microsoft.hyperspace.util.SchemaUtils
 
 /**
  * Action to refresh indexes with newly appended files and deleted files in an incremental way.
@@ -92,7 +91,8 @@ class RefreshIncrementalAction(
         refreshDF,
         indexDataPath.toString,
         previousIndexLogEntry.numBuckets,
-        SchemaUtils.prefixNestedFieldNames(indexConfig.indexedColumns),
+        // previousIndexLogEntry should contain the resolved and prefixed field names.
+        previousIndexLogEntry.derivedDataset.properties.columns.indexed,
         writeMode)
     }
   }
@@ -114,10 +114,6 @@ class RefreshIncrementalAction(
         "Index refresh (to handle deleted source data) is " +
           "only supported on an index with lineage.")
     }
-  }
-
-  override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
-    RefreshIncrementalActionEvent(appInfo, logEntry.asInstanceOf[IndexLogEntry], message)
   }
 
   /**
@@ -143,5 +139,9 @@ class RefreshIncrementalAction(
       // New entry.
       entry
     }
+  }
+
+  override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
+    RefreshIncrementalActionEvent(appInfo, logEntry.asInstanceOf[IndexLogEntry], message)
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
@@ -557,10 +557,6 @@ case class IndexLogEntry(
     config.hashCode + signature.hashCode + numBuckets.hashCode + content.hashCode
   }
 
-  def usesNestedFields: Boolean = {
-    SchemaUtils.containsNestedFieldNames(indexedColumns ++ includedColumns)
-  }
-
   /**
    * A mutable map for holding auxiliary information of this index log entry while applying rules.
    */
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
@@ -31,7 +31,7 @@ import org.apache.spark.sql.types.{DataType, StructType}
 
 import com.microsoft.hyperspace.HyperspaceException
 import com.microsoft.hyperspace.actions.Constants
-import com.microsoft.hyperspace.util.PathUtils
+import com.microsoft.hyperspace.util.{PathUtils, SchemaUtils}
 
 // IndexLogEntry-specific fingerprint to be temporarily used where fingerprint is not defined.
 case class NoOpFingerprint() {
@@ -555,6 +555,10 @@ case class IndexLogEntry(
 
   override def hashCode(): Int = {
     config.hashCode + signature.hashCode + numBuckets.hashCode + content.hashCode
+  }
+
+  def usesNestedFields: Boolean = {
+    SchemaUtils.containsNestedFieldNames(indexedColumns ++ includedColumns)
   }
 
   /**
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
@@ -305,8 +305,8 @@ object JoinIndexRule
     val rRequiredIndexedCols = lRMap.values.toSeq
 
     // All required columns resolved with base relation.
-    val lRequiredAllCols = resolve(spark, allRequiredCols(left), leftRelation.plan).get
-    val rRequiredAllCols = resolve(spark, allRequiredCols(right), rightRelation.plan).get
+    val lRequiredAllCols = resolve(spark, allRequiredCols(left), leftRelation.plan).get.map(_._1)
+    val rRequiredAllCols = resolve(spark, allRequiredCols(right), rightRelation.plan).get.map(_._1)
 
     // Make sure required indexed columns are subset of all required columns for a subplan
     require(resolve(spark, lRequiredIndexedCols, lRequiredAllCols).isDefined)
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
@@ -305,8 +305,8 @@ object JoinIndexRule
     val rRequiredIndexedCols = lRMap.values.toSeq
 
     // All required columns resolved with base relation.
-    val lRequiredAllCols = resolve(spark, allRequiredCols(left), lBaseAttrs).get
-    val rRequiredAllCols = resolve(spark, allRequiredCols(right), rBaseAttrs).get
+    val lRequiredAllCols = resolve(spark, allRequiredCols(left), leftRelation.plan).get
+    val rRequiredAllCols = resolve(spark, allRequiredCols(right), rightRelation.plan).get
 
     // Make sure required indexed columns are subset of all required columns for a subplan
     require(resolve(spark, lRequiredIndexedCols, lRequiredAllCols).isDefined)
--- a/src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
@@ -84,12 +84,13 @@ object ResolverUtils {
    * @param spark Spark session.
    * @param requiredStrings List of strings to resolve.
    * @param plan Logical plan to resolve against.
-   * @return Optional Seq of resolved strings if all required strings are resolved. Else, None.
+   * @return Optional sequence of tuples of resolved name string and nested state boolean
+   *         if all required strings are resolved. Else, None.
    */
   def resolve(
       spark: SparkSession,
       requiredStrings: Seq[String],
-      plan: LogicalPlan): Option[Seq[String]] = {
+      plan: LogicalPlan): Option[Seq[(String, Boolean)]] = {
     val schema = plan.schema
     val resolver = spark.sessionState.conf.resolver
     val resolved = requiredStrings.map { requiredField =>
@@ -98,7 +99,12 @@ object ResolverUtils {
         .map { expr =>
           val resolvedColNameParts = extractColumnName(expr)
           validateResolvedColumnName(requiredField, resolvedColNameParts)
-          getColumnNameFromSchema(schema, resolvedColNameParts, resolver).mkString(".")
+          getColumnNameFromSchema(schema, resolvedColNameParts, resolver)
+            .foldLeft(("", false)) { (acc, i) =>
+              val name = Seq(acc._1, i._1).filter(_.nonEmpty).mkString(".")
+              val isNested = acc._2 || i._2
+              (name, isNested)
+            }
         }
         .getOrElse { return None }
     }
@@ -110,8 +116,8 @@ object ResolverUtils {
     expr match {
       case a: Attribute =>
         Seq(a.name)
-      case g @ GetStructField(_, _, Some(name)) =>
-        extractColumnName(g.child) :+ name
+      case GetStructField(child, _, Some(name)) =>
+        extractColumnName(child) :+ name
       case _: GetArrayStructFields =>
         // TODO: Nested arrays will be supported later
         throw HyperspaceException("Array types are not supported.")
@@ -138,19 +144,19 @@ object ResolverUtils {
   private def getColumnNameFromSchema(
       schema: StructType,
       resolvedColNameParts: Seq[String],
-      resolver: Resolver): Seq[String] = resolvedColNameParts match {
+      resolver: Resolver): Seq[(String, Boolean)] = resolvedColNameParts match {
     case h :: tail =>
       val field = schema.find(f => resolver(f.name, h)).get
       field match {
         case StructField(name, s: StructType, _, _) =>
-          name +: getColumnNameFromSchema(s, tail, resolver)
+          (name, true) +: getColumnNameFromSchema(s, tail, resolver)
         case StructField(_, _: ArrayType, _, _) =>
           // TODO: Nested arrays will be supported later
           throw HyperspaceException("Array types are not supported.")
         case StructField(_, _: MapType, _, _) =>
           // TODO: Nested maps will be supported later
           throw HyperspaceException("Map types are not supported")
-        case f => Seq(f.name)
+        case f => Seq((f.name, false))
       }
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
@@ -18,6 +18,11 @@ package com.microsoft.hyperspace.util
 
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.analysis.Resolver
+import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, Expression, ExtractValue, GetArrayStructFields, GetMapValue, GetStructField}
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.types.{ArrayType, MapType, StructField, StructType}
+
+import com.microsoft.hyperspace.HyperspaceException
 
 /**
  * [[ResolverUtils]] provides utility functions to resolve strings based on spark's resolver.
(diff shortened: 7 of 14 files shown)


andrei-ionescu:
@sezruby, @imback82 Here is the first PR that add support for creating indexes over nested fields.

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -65,9 +65,20 @@ class CreateAction(
   }
 
   private def isValidIndexSchema(config: IndexConfig, schema: StructType): Boolean = {
+    // First we flatten the schema. Instead of having struct of leaves
+    // the flatten method will return a list of field names.
+    // The second step is escaping the field names as Parquet does not works well
```
imback82, line 70: Could you elaborate on this "Parquet does not work well"?

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
```diff
@@ -65,9 +65,20 @@ class CreateAction(
...
+    // with field names that contains the `.` (dot).
+    // Given `struct(nested, struct(nst, struct(field1)))`, the fields variable
+    // will contain `Seq("nested__nst__field1")`.
+    val fields = SchemaUtils.escapeFieldNames(SchemaUtils.flatten(schema))
```
imback82, line 74: If the original column has ".", will it be also escaped?

For example, `val df = Seq(1).toDF("a.b")`

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -117,7 +103,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
 
     // run job
     val repartitionedIndexDataFrame =
-      indexDataFrame.repartition(numBuckets, resolvedIndexedColumns.map(df(_)): _*)
+      indexDataFrame.repartition(numBuckets, resolvedIndexedColumns.map(c => col(s"$c")): _*)
```
imback82, line 106: can you explain why this change was required?

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
andrei-ionescu: Given this nested class: 

```
case class Nested(nest: String, v: Int)
```

And this dataset:

```
val df = Seq(("a", "nxn1", Nested("n1", 1))).toDF("name", "nested.nest", "nested")
```

The schema of the dataset is this:

```
root
 |-- name: string (nullable = true)
 |-- nested.nest: string (nullable = true)
 |-- nested: struct (nullable = true)
 |    |-- nest: string (nullable = true)
 |    |-- v: integer (nullable = false)
```

Doing `df.select("nested.nest").limit(1)` will return `n1`, the value of `nest` inside the `Nested` class.
Doing `df.select("`nested.nest`").limit(1)` will return `nxn1`, the value in the top level field called `nested.nest`.

Given the two things above I would say that if the user wants to access a field name that contains a dot and that field is not nested then it should be enclosed by back ticks (` `` `).

src/main/scala/com/microsoft/hyperspace/util/SchemaUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+   * The check is implemented by checking if the field name string contains
+   * the separator defined for nested field expressions.
+   *
+   * See [[NESTED_FIELD_NEEDLE_REGEX]].
+   *
+   * @param field The field nme
+   * @return True if the field name represents a nested field otherwise false.
+   */
+  def isNestedField(field: String): Boolean = {
+    NESTED_FIELD_NEEDLE_REGEX.r.findFirstIn(field).isDefined
```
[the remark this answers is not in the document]
andrei-ionescu: I added the back ticks check. See also my previous comment here: https://github.com/microsoft/hyperspace/pull/379/files#r593052711

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
andrei-ionescu: The `prepareIndexDataFrame` does the things a bit different. First, from the original data frame it has to extract the nested field values to the top level. Second, it gives this new top level field an escaped name so that we can save it properly and retrieve it safely.

So, `indexDataFrame` columns are different from the original `df` data frame and doing `df(_)` with an unknown columns would result it exceptions.

The `resolvedIndexColumns` contains escaped field names and the `indexDataFrame` contains those names.

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
andrei-ionescu: Sorry,  it is a bit misleading and not properly formulated. It is about this bug on `partitionBy` in Apache Spark: https://issues.apache.org/jira/browse/SPARK-18084.

I try writing a better comment.

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
andrei-ionescu: I added the check for backticks and when flattening (`flatten` method) if we find a field name that is not nested but has `.` (dot) in the name it will be enclosed in backticks. This way we can differentiate between the two.

sezruby: What if there's no nested structure (nested -> (nested, v), and call df.select("nest.nest").limit(1)?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -140,13 +126,43 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
-    val dfColumnNames = df.schema.fieldNames
-    val indexedColumns = indexConfig.indexedColumns
-    val includedColumns = indexConfig.includedColumns
+    // Flatten will transform nested field names from `struct(nested, struct(nst, struct(field1)))`
+    // into `Seq("nested.nst.field1")`.
+    val dfColumnNames = SchemaUtils.flatten(df.schema)
+    // The index config will contain the field names as they are given with `.` (dots).
+    // (ie: `Seq("nested.nst.field")`) and they need to be escaped to `nested__nst__field1` as
+    // the index entry know to work with escaped values only.
+    val indexedColumns = SchemaUtils.unescapeFieldNames(indexConfig.indexedColumns)
```
sezruby, line 164: What if a column is not nested and has double underscore in its name?

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
andrei-ionescu: @sezruby if you do that first, the `flatten` method will enclose it with backticks, then `escapeFieldName` will NOT escape it as we check that the field is enclosed in backticks.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -140,13 +126,43 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+
+  private def hasLineageProperty(spark: SparkSession): Option[(String, String)] = {
+    if (hasLineage(spark)) {
+      Some(IndexConstants.LINEAGE_PROPERTY -> "true")
+    } else {
+      None
+    }
+  }
+
+  private def usesNestedFieldsProperty(indexConfig: IndexConfig): Option[(String, String)] = {
```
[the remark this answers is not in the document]
andrei-ionescu: `RefreshAtionBase` does NOT call `getIndexLogEntry`, however `RefreshActionBase` extends `CreateActionBase`. 

The `RefreshAction` is the one calling `getIndexLogEntry` from `CreateActionBase`.

Sorry, but I cannot understand what you're requesting with this comment. Please give more information.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
andrei-ionescu: The `flatten` method encloses with backticks any field names containing `__` and `unescapeFieldName` checks if the field name is backticks enclosed.

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
imback82: Hmm, since we have few questions around the "normalization", have you also considered just prefixing some constant (e.g., `_hyperspace_nested_col.nested.x.y`, where `_hyperspace_nested_col` will be just prepended for a nested column).

BTW, I see bug fixes related to nested fields in OSS Spark regularly, so this is not straightforward to get it right 😄 , so thanks @andrei-ionescu / @sezruby for the discussion.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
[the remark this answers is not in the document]
sezruby: In RefreshAction.scala
```
override def logEntry: LogEntry = getIndexLogEntry(spark, df, indexConfig, indexDataPath)
```
The index config is escaped one, right? 
So need to override `usesNestedFieldsProperty` for RefreshActionBase as RefreshAction extends RefreshActionBase . . ?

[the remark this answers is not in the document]
andrei-ionescu: From what I know `IndexLogEntry` is what the user enters when it creates the index. When an index is created the given column names are NOT escaped. For example:

```
hs.createIndex(
  nestedDataset, 
  IndexConfig(
    "idx_nested", 
    indexedColumns = Seq("nested.nst.field1"), 
    includedColumns = Seq("id", "name", "nested.nst.field2")))
```

src/main/scala/com/microsoft/hyperspace/util/SchemaUtils.scala
```diff
@@ -0,0 +1,199 @@
...
+   * The field names can escaped by enclosing them by backticks to specify
+   * that the `.` (dot) does not mean a nested field.
+   *
+   * See [[NESTED_FIELD_NEEDLE_REGEX]] and [[backticksEnclosed]] method.
+   *
+   * @param field The field name
+   * @return True if the field name represents a nested field otherwise false.
+   */
+  def isNestedField(field: String): Boolean = {
+    !backticksEnclosed(field) &&
```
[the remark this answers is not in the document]
andrei-ionescu: @imback82 The `select` is part of the filter PRs which we agreed to have separate. Please add this comment to that PR and I'll create a UT for it.

On the other hand, the dataset schema of `tbl (id int, point struct<x: float, y: double>)` should be properly handled and an index can be created by doing this:

```
hs.createIndex(
  tbl, 
  IndexConfig(
    "idx_tbl", 
    indexedColumns = Seq("point.x")))
```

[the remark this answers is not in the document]
andrei-ionescu: @imback82 & @sezruby: What's your suggestion on detecting nested vs not-nested fields?

src/main/scala/com/microsoft/hyperspace/actions/CreateAction.scala
andrei-ionescu: I like the prefix suggestion. I'll try implementing it and keeping those `.` (dots). I hope it works.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -140,13 +126,43 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+    if (SchemaUtils.hasNestedFields(indexConfig.indexedColumns ++ indexConfig.includedColumns)) {
```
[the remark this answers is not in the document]
andrei-ionescu: If we use the prefix suggestion this will be easier and clearer without needing to check the data frame.

andrei-ionescu:
@imback82, @sezruby, @rapoth: There are lots of issues on addressing nested fields in Spark. 

Even this example is not working in Spark:
```
case class Prec(s: Double)
case class Point(`coord.x`: Double, `coord.y`: Double, `prec.100`: Prec)
val colName = "__prefixed.precision.`prec.100`.s"
val df = Seq((1, "a", Point(1, 1, Prec(10)))).toDF("id", "name", "precision")
val df2 = df.select("precision.`prec.100`.s").toDF(colName)
df2.select(colName).show
```

It ends up with:

```
org.apache.spark.sql.AnalysisException: cannot resolve '`__prefixed.precision.``prec.100``.s`' given input columns: [__prefixed.precision.`prec.100`.s];;
'Project ['__prefixed.precision.`prec.100`.s]
+- Project [s#26 AS __prefixed.precision.`prec.100`.s#28]
   +- Project [precision#9.prec.100.s AS s#26]
      +- Project [_1#3 AS id#7, _2#4 AS name#8, _3#5 AS precision#9]
         +- LocalRelation [_1#3, _2#4, _3#5]
```

I'm able to do `toDF` with a column name but then I cannot select it.

To access `s` in `df` I have to use ``precision.`prec.100`.s`` in select (it needs `precision.100` to be enclosed by backticks) but that column name brought up to the first level (not nested) cannot contain backticks.

We need to create a two way convention on how to escape the field names and back.

I propose the following...

Given this column ``precision.`prec.100`.s`` it will be escaped to `__hs.precision.-prec.100-.s`.

Steps to escape:
1. Replace `` ` `` (backticks) with `-`
2. Prepend prefix `__hs`
3. Fully enclose it in `` ` `` (backticks)

Steps to unescape:
1. Remove begin and end backticks
2. Remove `__hs` prefix
3. Replace `-` with `` ` `` (backticks)

The ``escapeFieldName("precision.`prec.100`.s")`` will return `__hs.precision.-prec.100-.s`.

src/main/scala/com/microsoft/hyperspace/util/SchemaUtils.scala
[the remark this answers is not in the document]
imback82: > The `select` is part of the filter PRs which we agreed to have separate. Please add this comment to that PR and I'll create a UT for it.

Ah, the example was just for the demonstration. `isNestedField` seems to be used in this PR, and the user can specify the column name in the index config as the example shows.

> What's your suggestion on detecting nested vs not-nested fields?

I think letting Spark to resolve it (https://github.com/microsoft/hyperspace/pull/379#discussion_r594679292) seems like the best way.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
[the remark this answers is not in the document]
sezruby: in CreateActionBase.scala
```scala
    IndexLogEntry(
          indexConfig.indexName,
          CoveringIndex(
            CoveringIndex.Properties(
              CoveringIndex.Properties
                .Columns(resolvedIndexedColumns, resolvedIncludedColumns), <= 
              IndexLogEntry.schemaString(indexDataFrame.schema),
              numBuckets,
              coveringIndexProperties)),
          Content.fromDirectory(absolutePath, fileIdTracker),
          Map())
```

We stored the indexed / included columns as resolved names.

So when refreshIndex, indexConfig.indexedColumn / includedColumn will be resolved one.. 
Could you check the parameter after refresh? It would be good to add the assert after some refresh index tests you created.

src/main/scala/com/microsoft/hyperspace/util/SchemaUtils.scala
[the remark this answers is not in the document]
andrei-ionescu: @imback82 I've seen your comment and I did look at the suggested private method from Spark but that is not really help us, or at least I cannot see how.

We need to know if a given path to a column is nested or not. The method you suggested only resolves a column but doesn't tell us if is nested or not.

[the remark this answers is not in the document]
imback82: You can check the return value of `resolve` and if it's `Attribute`, it's not nested. You can check the link I shared in the comment: https://github.com/apache/spark/blob/9809a2f1c5187205c81542dbdc84b71db535f6e1/sql/core/src/main/scala/org/apache/spark/sql/DataFrameNaFunctions.scala#L358

imback82:
Hmm, how about just not supporting nested columns if their names have "." in the first cut? We can throw an exception for now, and add it later if there is a demand.

So we can try the following:
Escape:
1. Check if a column name given in the config is a nested column by using `df.resolve`.
2. If it is a nested column,
  * Validate if it's supported: `df.resolve` will return `Alias(GetStructField(_))`, and you can traverse it to check if the name at each level contains "." or not. If it does contain, throw an exception.
  * After the validation, prepend `__hs_nested` to the column name (the name should be constructed by traversing GetStructField to make sure it is unquoted).

Unescape:
If the column name has `__hs_nested` as a prefix, just remove the prefix.

andrei-ionescu:
@imback82 I'm not sure what you mean by this:

> Check if a column name given in the config is a nested column by using `df.resolve`.

First, and most important is that the method is **NOT** public API. If we are going towards this route we need to get to upper levels and use public APIs (not marked by `private` or `protected`).

Second, I cannot understand how `resolve` method that returns a `NamedExpression` can tell us if is nested or not. Just calling the method is not enough as far as I know. Is there other process that you have in mind that I'm not aware of?

Putting the two issues aside for a moment, you're suggesting the following steps:

1. Given a new `IndexConfig` with a _"nested construct column name"_ - a field name that contains a `.` (dot) - we need to check this
2. Check the given _"nested construct column name"_ if is resolvable
3. Check the given _"nested construct column name"_ if the resolved parts do contain any `.` (dot) and if so throw exception
4. Transform the given _"nested construct column name"_ into a top level column name (ie: from `nested.field_a.leaf_1` into `` `__hs_nested.nested.field_a.leaf_1` `` (note the required additional backticks).
5. Whenever needed use the prefix to check inside the `IndexLogEntry` that a column is nested without going to the data frame that usually is not available at that level.

Do I get it right?

One other question for all of you @imback82, @sezruby, @rapoth: How can we speed these decisions up? It is taking a day for each back and forth message. Can we somehow speed up the process?

I have some proposals:
1. What's the everyday time interval that you guys are available to review code? I'm asking because I'm thinking to be around on the PR responding to your feedback in a faster way.
2. My timezone is EET (GMT+3). What's your timezone?

andrei-ionescu:
@imback82, @sezruby I started implementing it in another way. I started w/o `flatten` and instead by resolving the nested field constructs with `resolveQuoted` method from `LogicalPlan` - [LogicalPlan.scala#L122-L131](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LogicalPlan.scala#L122-L131). Changes will come soon.

imback82:
> @imback82, @sezruby I started implementing it in another way. I started w/o `flatten` and instead by resolving the nested field constructs with `resolveQuoted` method from `LogicalPlan` - [LogicalPlan.scala#L122-L131](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LogicalPlan.scala#L122-L131). Changes will come soon.

I think that's what's being called from `df.resolve`:  https://github.com/apache/spark/blob/569fb133d09e24e4ed56ed7efff641512d98b01b/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala#L262, which I have been suggesting.

sezruby:
@andrei-ionescu Our timezone is PST.  9am - 2pm PST is 6pm - 11pm EET. So I'll try to do review in the morning.

You can bypass private method in this way:
https://github.com/microsoft/hyperspace/blob/master/src/main/scala/org/apache/spark/sql/hyperspace/utils/package.scala

andrei-ionescu:
@imback82, @sezruby: I pushed a new approach after your suggestion of using Spark's resolve mechanisms. And with this I got rid of the `flatten` method and all related methods for escaping and back. 

The following things were added:
1.  New `ResolverUtils.resolve` method to resolve the field name constructs agains a dataframe plan.
    1. Uses `LogicalPlan.resolveQuoted` method
    2. Fixes the casing (ie: `nesTed.fieLD` into `nested.field`)
    3. When there are field names that contain `.` (dot) and are not nested an exception is thrown
2. New prefix methods in `SchemaUtils`.
3. Unit Tests for all new functionalities.

Please have a look into this approach.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -117,7 +101,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
 
     // run job
     val repartitionedIndexDataFrame =
-      indexDataFrame.repartition(numBuckets, resolvedIndexedColumns.map(df(_)): _*)
+      indexDataFrame.repartition(numBuckets, resolvedIndexedColumns.map(c => col(s"`$c`")): _*)
```
andrei-ionescu, line 104: We need field names to be backticks enclosed as they may contain `.`s (dots).

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -202,10 +203,16 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
         .select(
           allIndexColumns.head,
           allIndexColumns.tail :+ IndexConstants.DATA_FILE_NAME_ID: _*)
+        .toDF(SchemaUtils.prefixNestedFieldNames(allIndexColumns) :+
```
andrei-ionescu, line 206: We do this to transform the selected fields into the prefixed nested ones.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -202,10 +203,16 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
+          IndexConstants.DATA_FILE_NAME_ID: _*)
     } else {
       df.select(columnsFromIndexConfig.head, columnsFromIndexConfig.tail: _*)
+        .toDF(SchemaUtils.prefixNestedFieldNames(columnsFromIndexConfig): _*)
```
andrei-ionescu, line 210: We do this to transform the selected fields into the prefixed nested ones.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -86,7 +87,10 @@ private[actions] abstract class RefreshActionBase(
 
   protected lazy val indexConfig: IndexConfig = {
     val ddColumns = previousIndexLogEntry.derivedDataset.properties.columns
-    IndexConfig(previousIndexLogEntry.name, ddColumns.indexed, ddColumns.included)
+    IndexConfig(
+      previousIndexLogEntry.name,
+      SchemaUtils.removePrefixNestedFieldNames(ddColumns.indexed),
```
andrei-ionescu, line 92: We remove the prefixes as based on these details we later prepare the dataset inside the `prepareIndexDataFrame`.

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -90,7 +91,7 @@ class RefreshIncrementalAction(
         refreshDF,
         indexDataPath.toString,
         previousIndexLogEntry.numBuckets,
-        indexConfig.indexedColumns,
+        SchemaUtils.prefixNestedFieldNames(indexConfig.indexedColumns),
```
andrei-ionescu, line 94: We update the index and for that we need to use the prefixed nested field name.

sezruby (commented):
Few minor comments, generally looks good to me

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -140,25 +123,43 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
     relations.head
   }
 
+  private def hasParquetAsSourceFormatProperty(
```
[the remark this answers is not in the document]
andrei-ionescu: I'll put it back but it is the `scalafmt` that does this arrangement.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -140,25 +123,43 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
     val indexedColumns = indexConfig.indexedColumns
     val includedColumns = indexConfig.includedColumns
-    val resolvedIndexedColumns = ResolverUtils.resolve(spark, indexedColumns, dfColumnNames)
-    val resolvedIncludedColumns = ResolverUtils.resolve(spark, includedColumns, dfColumnNames)
+    val resolvedIndexedColumns = ResolverUtils.resolve(spark, indexedColumns, plan)
+    val resolvedIncludedColumns = ResolverUtils.resolve(spark, includedColumns, plan)
 
     (resolvedIndexedColumns, resolvedIncludedColumns) match {
-      case (Some(indexed), Some(included)) => (indexed, included)
+      case (Some(indexed), Some(included)) =>
```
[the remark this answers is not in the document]
andrei-ionescu: `scalafmt` ... ditto

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -71,4 +76,68 @@ object ResolverUtils {
...
+   * @param requiredStrings List of strings to resolve.
+   * @param plan Logical plan to resolve against.
+   * @return Optional Seq of resolved strings if all required strings are resolved. Else, None.
+   */
+  def resolve(
+      spark: SparkSession,
+      requiredStrings: Seq[String],
+      plan: LogicalPlan): Option[Seq[String]] = {
+
+    def fixParts(parts: Seq[String], outputs: Seq[Attribute]): Seq[String] = {
```
sezruby, line 94: Could you make this function more readable? 
It's bit hard to read..

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -71,4 +76,68 @@ object ResolverUtils {
...
+      var newParts = Seq.empty[String]
+      val h :: t = parts.toList
+      val topLevelField = outputs.find(_.name.compareToIgnoreCase(h) == 0)
+      topLevelField match {
+        case Some(o) =>
+          newParts = newParts :+ o.name
+          var children = o.dataType.asInstanceOf[StructType]
+          t.foreach { e =>
```
sezruby, line 102: how about `foldLeft(children) { (part, children)  => ... }`

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -71,4 +76,68 @@ object ResolverUtils {
...
+              newParts = newParts :+ children.fieldNames(elIdx)
+              children.fields.toSeq(elIdx).dataType match {
+                case s: StructType =>
+                  children = s
+                case _ =>
+              }
+            }
+          }
+        case None =>
+          newParts = h +: t
```
sezruby, line 114: is this a possible case?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -71,4 +76,68 @@ object ResolverUtils {
...
+          case g: GetStructField =>
+            val parts = g.sql // returns backtick enclosed string (ie: `a`.`b`.`c`)
+              .replaceAll("^`(.*)`$", "$1") // strip first and last backticks
+              .split("`\\.`") // split by "`.`"
+            val fixedParts = fixParts(parts, plan.output) // fix the casing
+            val resolvedFieldName = fixedParts.mkString(".") // put back the field name fixed
+            val unsupported = parts.filter(_.contains("."))
+            if (unsupported.nonEmpty) { // the fields with dots should throw exception
+              throw HyperspaceException(s"The following field name construct " +
+                s"$resolvedFieldName contains unsupported parts: ${unsupported.mkString(",")}.")
```
sezruby, line 136: How about this:
`Hyperspace does not support nested column names including '.' (columnName: ${requiredField})` ?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
andrei-ionescu: Not directly. `allIndexColumns != escapedIndexedColumns ++ escapedIncludedColumns`. The `allIndexColumns` contains the columns contained in the index and missing columns.

I will replace it with `escapedIndexedColumns ++ escapedIncludedColumns ++ missingPartitionColumns` which is longer but is ok.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: This not 100% true. We do support indexes over field names - this what are trying to achieve. We just don't support **not nested** fields having field names with dots.

andrei-ionescu: I'll put this message:
```
s"Hyperspace does not support the following access " +
s"field path construct $requiredField because it contains the following " +
s"unsupported field names that contain dots: ${unsupported.mkString(",")}."
```

Let me give an example...

Given a dataset with the following schema:
```
root
 |-- nested: struct (nullable = true)
 |    |-- the.non.nested.part: struct (nullable = true)
 |    |    |-- cnt: integer (nullable = true)
```
Creating an index with:

```
IndexConfig("idx", indexedColumns = Seq("nested.the.non.nested.part.cnt"))
```

Is not supported and the exception message will look like this:

```
Hyperspace does not support the following access field path nested.the.non.nested.part.cnt because it contains the following unsupported field names that contain dots: the.non.nested.part.
```

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: I don't know how that would go. We don't fold anything. For each part we need to find its nested counter part inside the struct nested layers.

Given `nesTED, LeaF, CNT` parts and the nested top level column with 
```
nested: struct (nullable = true)
 |-- leaf: struct (nullable = true)
 |    |-- v: integer (nullable = true)
 |    |-- cnt: integer (nullable = true)
 |    |-- other: string (nullable = true)
```
The `nesTED` will be resolved by finding the top level column name and using that. 

The `LeaF` will go one level into the struct search for it case insensitively and use the name of the struct.

For `CNT` it will go one more level into it the column search case insensitively for it out of the 3 leaves and use the name of the found struct.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: `find` method returns an `Option` so both `Some` and `None` are possible values and the correct matching would be to match against each possibility. 

Now, if you ask me how can we land in this case when a lot of other checks already did take place, the answer is I don't know. Still it is a best practice to cover both cases.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
sezruby: yep I meant for tail part, something like:

```scala
val (fixedParts, _) = t.foldLeft(Seq(o.name), children) => { (part, fixedParts, children) =>
      val elIdx: Int = children.fieldNames.indexWhere { f => f.compareToIgnoreCase(e) == 0 }
      if (elIdx >= 0) {
         // When found add it to the new set of fixed parts
          val fixedPartName = children.fieldNames(elIdx)
          chilchildren.fields.toSeq(elIdx).dataType match {
              case s: StructType => // et it up for the next part
                  (fixedParts :+ fixedPartName, s) 
                case _ => // Nothing to do, cannot go deeper, it must be a leaf
                  (fixedParts :+ fixedPartName, children)
              }
        } else { (fixedParts: + part, children) // or exception? }
}
```

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
sezruby: I think early failure is better when something goes wrong. WDYT?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
sezruby: Thanks for revising it. It looks better 👍

sezruby: Thanks for revising this! it looks better 👍

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: Just falling back to original access path instead of the case fixed ones should be totally ok because the parts are already resolved and found by `resolveQuoted` and `.sql` methods. I would say that in case of any unknown issue in this resolving process it is always possible to go on with already resolved parts.

Given a dataset with the following nested field `struct("my", struct("nested", field("path")))` and `spark.sql.caseSensitive -> false`, using the following access field path `my.neSTed.pAth` it will get resolved to `my.neSTed.pAth` and the `.sql` method with return `` `my`.`neSTed`.`pAth` ``. Now, to comply with the tests you have put in place I am forced to create this `fixParts` method to turn the `Seq("my", "neSTed", "pAth")` into `Seq("my", "nested", "pAth")` and I guess this is due to the fact that we want to create the index dataset with the correct column name otherwise the index will be unusable if the user changes the `spark.sql.caseSensitive` to `true`.

So, are you suggesting to throw a `HyperspaceException` or just `get` and let it throw itself a 
`NoSuchElementException`?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: IMHO your suggestion is harder to read and understand than what I proposed. But I'm willing to change it. 

@imback82 WDYT? 

I'm asking you because I don't want to be in a position to change it back or to something else yet once more.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -71,4 +76,101 @@ object ResolverUtils {
...
+    Some(requiredStrings.map { requiredField =>
+      plan
+        .resolveQuoted(requiredField, spark.sessionState.conf.resolver)
+        .getOrElse { return None }
+        .collectFirst {
+          case a: AttributeReference =>
+            schemaFieldNames.find(_.compareToIgnoreCase(a.name) == 0).getOrElse(a.name)
+          case g: GetStructField =>
+            val parts = g.sql // returns backtick enclosed string (ie: `a`.`b`.`c`)
+              .replaceAll("^`(.*)`$", "$1") // strip first and last backticks
```
imback82, line 160: How about something like the following so that we don't do regex?
```scala
  def resolve(
      spark: SparkSession,
      requiredStrings: Seq[String],
      plan: LogicalPlan): Option[Seq[String]] = {
    val schema = plan.schema
    val resolver = spark.sessionState.conf.resolver
    val resolved = requiredStrings.map { colName =>
      val resolvedColNameParts = plan
        .resolveQuoted(colName, resolver)
        .map(extractColumnName)
        .getOrElse(return None) // We can improve this if you don't want to return in the middle.
      validateResolvedColumnName(colName, resolvedColNameParts)
      getColumnNameFromSchema(schema, resolvedColNameParts, resolver).mkString(".")
    }
    Some(resolved)
  }

  private def extractColumnName(expr: Expression): Seq[String] = {
    expr match {
      case a: Attribute => Seq(a.name)
      case g @ GetStructField(_, _, Some(name)) => extractColumnName(g.child) :+ name
      case Alias(nested: GetStructField, _) => extractColumnName(nested)
    }
  }

  // Validate the resolved column name by checking if nested columns have dots in its field names.
  private def validateResolvedColumnName(
      origCol: String,
      resolvedColNameParts: Seq[String]): Unit = {
    if (resolvedColNameParts.length > 1 && resolvedColNameParts.exists(_.contains("."))) {
      throw HyperspaceException(
        s"Hyperspace does not support the nested column " +
          s"whose name contains dots: $origCol")
    }
  }

  // Given resolved column name parts, return new column name parts using the name in the given
  // schema to use the original name in order to preserve the casing.
  private def getColumnNameFromSchema(
      schema: StructType,
      resolvedColNameParts: Seq[String],
      resolver: Resolver): Seq[String] = resolvedColNameParts match {
    case h :: tail =>
      val field = schema.find(f => resolver(f.name, h)).get
      field match {
        case StructField(name, s: StructType, _, _) =>
          name +: getColumnNameFromSchema(s, tail, resolver)
        case StructField(_, _: ArrayType, _, _) =>
          throw HyperspaceException("Array types are not supported.")
        case StructField(_, _: MapType, _, _) =>
          throw HyperspaceException("Map types are not supported")
        case f => Seq(f.name)
      }
  }
```

andrei-ionescu: @imback82 I'll try out your suggestion but there are some things that are not properly working in the proposed code.

This is what I'll do:
1. Split the code in the those three private methods as they seem to be a nice grouping of actions
2. I will use `.sql` from a `GetStructField` and still fix its case because:
    1. the pattern matching fails in the case when `name` is `None` 
    2. even if the `name` has some value that needs to be case fixed
3. Use the `resolver` once more as suggested to fix the case by extracting the field access paths from schema hoping that the resolver will not keep the wrong casing.

Just for my curiosity, why against regex?

andrei-ionescu: @imback82: One more question...

When doing:
```
resolvedColNameParts.length > 1 && resolvedColNameParts.exists(_.contains("."))
```
the `> 1` is because you want to still support an index on a top level field that is not nested but contains `.` (dot)?

andrei-ionescu:
@imback82: I integrated your feedback and added you as co-author on this commit.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
imback82: > there are some things that are not properly working in the proposed code.

Oh, do you know which part? I ran `ResolverUtilsTest` and they seem fine. One case could be `GetStructField` with its `name` = `None`, but I think we can do the following if needed (but didn't check Spark code when this can be `None`):
```scala
      case g @ GetStructField(_, _, None) =>
        extractColumnName(g.child) :+ g.childSchema(g.ordinal).name
```
> the `> 1` is because you want to still support an index on a top level field that is not nested but contains `.` (dot)?

Yes, 👍🏼

> Just for my curiosity, why against regex?

In this case, it seemed we were doing `Seq("a", "b").mkString(",").split(",").map { ... }` where we could be just doing `Seq("a", "b").map { ... }`; an extreme example, but hopefully you get the idea.

imback82 (commented):
I have a couple of comments, but generally looking fine to me.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -557,6 +557,10 @@ case class IndexLogEntry(
     config.hashCode + signature.hashCode + numBuckets.hashCode + content.hashCode
   }
 
+  def usesNestedFields: Boolean = {
```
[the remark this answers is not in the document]
andrei-ionescu: It will be used later when hybrid scans are implemented. I'll remove it for now and add it later.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -181,6 +185,9 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
     val (resolvedIndexedColumns, resolvedIncludedColumns) = resolveConfig(df, indexConfig)
     val columnsFromIndexConfig = resolvedIndexedColumns ++ resolvedIncludedColumns
 
+    val escapedIndexedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIndexedColumns)
+    val escapedIncludedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIncludedColumns)
+
```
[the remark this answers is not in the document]
andrei-ionescu: It's not straight forward. `select` needs a head and the unpacked tail, while `toDF` unpacks the full collection. Also, a while ago in previous implementation I used the `SchemaUtils.prefixNestedFieldNames(allIndexColumns)` to keep the symmetry but in a review I was suggested to change it to the way it is right now.

I'll try putting it in a nice form.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -181,6 +185,9 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
     val (resolvedIndexedColumns, resolvedIncludedColumns) = resolveConfig(df, indexConfig)
     val columnsFromIndexConfig = resolvedIndexedColumns ++ resolvedIncludedColumns
 
+    val escapedIndexedColumns = SchemaUtils.prefixNestedFieldNames(resolvedIndexedColumns)
```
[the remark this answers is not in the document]
andrei-ionescu: I just reviewed a comment of yours above: https://github.com/microsoft/hyperspace/pull/379#issuecomment-800554698 saying:

>Hmm, how about just not supporting nested columns if their names have "." in the first cut? We can throw an exception for now, and add it later if there is a demand.

So, the top level column even if it's nested it first level name should NOT contain `.` (dot). I guess we should also removed the condition of `> 1` in `ResolverUtils`.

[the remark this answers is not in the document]
imback82: Yea, I think we should still support top-level columns with dots. For example, a top level column "\`x.y\`" should work where as a nested column "nested.\`x.y\`.z" should not be allowed. (thus the >1 check).

With the current code, wouldn't the top level column be also escaped? We can make `resolve` to return a tuple (resolved name, isNestedColumn) and call `prefixNestedFieldNames` only for nested columns.

I think the current hyperspace doesn't support top level column with dots, so we can address this as a separate PR; but let's make sure we are not prefixing top level column with dots.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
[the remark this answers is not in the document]
imback82: I see. If we decided to make `resolve` return `isNestedField` (https://github.com/microsoft/hyperspace/pull/379#discussion_r599159093), we can do `col(config).as(escaped)` only for nested columns, so that `toDF` is not required?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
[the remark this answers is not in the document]
andrei-ionescu: @imback82 Im not sure what you want. First you say that current Hyperspace does NOT support top level column with dots and then you are saying that we should still implement this extra work for a possible future feature that is not yet there. Why not doing it the other way around? I mean why not when adding this support into this code when the feature will be implemented?

I see 2 ways:
1. Keeping the current implementation of Hyperspace, the one that does NOT support dots and change the `> 1` so that it will throw the `HyperspaceException` when dots are encountered.
2. Implement the feature and change the `revolve` methods in `ResolverUtils` so that it supports the tuple output which will result in changes in other parts of the code.

I'll try to implement it but I hope that we will go further on with this PR because until now it did take a lot more than I expected and there are 4 or 5 more PRs coming to have this nested support in Hyperspace.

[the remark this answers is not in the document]
imback82: No, I was suggesting to do it as a separate PR ("so we can address this as a separate PR"). I thought my ask (make `resolve` to return extra info and wrap it with `col`) was a simple change, but it's OK, I will do a follow up to clean this up.

For this PR, we still need to do the following:
1. Instead of removing `>1`, add:
```scala
if (resolvedColNameParts.length == 1 && resolvedColNameParts.head.contains(".")) {
  throw HyperspaceException(
    s"Hyperspace does not support the top-level column whose name contains dots: $origCol")
}
```
2. Clearly document the prerequisite for `prefixNestedFieldName`, since it depends on the check in `resolve`.

andrei-ionescu:
@imback82: Implemented your suggestion with `resolve` method returning a tuple with field name along with its nested state.

Please have a look. I hope this is what you envisioned.

src/test/scala/com/microsoft/hyperspace/util/SchemaUtilsTest.scala
```diff
@@ -0,0 +1,95 @@
...
+    "__hs_nested",
+    "__hs_nested_a",
+    "__hs_nested.a.__hs_nested",
+    "__hs_nested.a.__hs_nested.b",
+    "__hs_nested.a.nested..b",
+    "__hs_nested.a.`g.c`.b",
+    "__hs_nested.a.g-c.b",
+    "a-b")
+
+  test("prefixNestedFieldName") {
```
sezruby, line 50: Could you more elaborate the test name & avoid the same name?

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -143,4 +144,28 @@ class RefreshIncrementalAction(
       entry
     }
   }
+
+  override protected def event(appInfo: AppInfo, message: String): HyperspaceEvent = {
+    RefreshIncrementalActionEvent(appInfo, logEntry.asInstanceOf[IndexLogEntry], message)
+  }
+
+  private def resolveConfigIndexedColumns(
```
sezruby, line 152: I think we could move this function (resolveConfig in CreateActionBase, isValidIndexSchema in CreateAction) to `object IndexConfig` to avoid duplicate code.

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -87,11 +90,13 @@ class RefreshIncrementalAction(
...
       }
 
+      val resolvedIndexedColumns = resolveConfigIndexedColumns(df, indexConfig)
+
       refreshDF.write.saveWithBuckets(
         refreshDF,
         indexDataPath.toString,
         previousIndexLogEntry.numBuckets,
-        indexConfig.indexedColumns,
+        SchemaUtils.prefixNestedFieldNames(resolvedIndexedColumns),
```
sezruby, line 99: How about just using previousIndexLogEntry.derivedDataset.properties.columns.indexed?

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
andrei-ionescu: No need. I removed it altogether.

andrei-ionescu:
@imback82, @sezruby Integrated your feedback. If you have time please have another look. Thanks.

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -104,32 +106,19 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
...
 
-    // run job
-    val repartitionedIndexDataFrame =
-      indexDataFrame.repartition(numBuckets, resolvedIndexedColumns.map(df(_)): _*)
+    // Run job
+    val repartitionedIndexDataFrame = {
+      // For nested fields, resolvedIndexedColumns will have flattened names with `.` (dots),
+      // thus they need to be enclosed in backticks to access them as top-level columns.
+      // Note that backticking the non-nested columns is a no-op.
+      indexDataFrame.repartition(numBuckets, resolvedIndexedColumns.map(c => col(s"`$c`")): _*)
```
sezruby, line 120: are bacticks still required? could you revise the comment?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -71,4 +78,87 @@ object ResolverUtils {
...
+      case Alias(nested: ExtractValue, _) =>
+        extractColumnName(nested)
+    }
+  }
+
+  // Validate the resolved column name by checking if nested columns have dots in its field names.
+  private def validateResolvedColumnName(
+      origCol: String,
+      resolvedColNameParts: Seq[String]): Unit = {
+    if (resolvedColNameParts.length > 1 && resolvedColNameParts.exists(_.contains("."))) {
```
sezruby, line 138: Could you add this condition to here? from @imback82's comment
https://github.com/microsoft/hyperspace/pull/379/files/4f8e3928170fcee0c1813ead6d9ab1521548c278#r599699752 
```scala
if (resolvedColNameParts.length == 1 && resolvedColNameParts.head.contains(".")) {
  throw HyperspaceException(
    s"Hyperspace does not support the top-level column whose name contains dots: $origCol")
}
```

src/main/scala/com/microsoft/hyperspace/util/SchemaUtils.scala
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /**
+   * The method prefixes the nested field names from a map where the keys are
+   * the field names and the values are the nested state of that field.
+   * The field names that are not marked as nested will not be changed.
+   *
+   * See [[prefixNestedFieldName]] method.
+   *
+   * @param fieldNames A [[ListMap]] of field names and nested status.
```
sezruby, line 51: Coudl you add some comments about ResolverUtils.resolve?
e.g.
```suggestion
   * The method prefixes the nested field names from a map where the keys are
   * the field names and the values are the nested state of that field
   * which should be the result of [[ResolverUtils.resolve]].
   * The field names that are not marked as nested will not be changed.
   *
   * See [[prefixNestedFieldName]] method.
   *
   * @param fieldNames A list of (filed name, nested status) resolved by [[ResolverUtils.resolve]]
```

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -71,4 +78,87 @@ object ResolverUtils {
...
+   * @param spark Spark session.
+   * @param requiredStrings List of strings to resolve.
+   * @param plan Logical plan to resolve against.
+   * @return Optional ListMap of tuples of resolved name string and nested state boolean
+   *         if all required strings are resolved. Else, None.
+   */
+  def resolve(
+      spark: SparkSession,
+      requiredStrings: Seq[String],
+      plan: LogicalPlan): Option[ListMap[String, Boolean]] = {
```
sezruby, line 95: Any reason to use ListMap? Seems we don't need "Map"?
How about `Seq[(String, Boolean)]`?

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
andrei-ionescu: The backticks are still required. `repartition` does not work without it. I tried it.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: Not needed anymore. Now, we support top level fields with `.` (dots).

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: Yes. `ListMap` keeps the order of the items inside it. `Seq` of tuple is harder to use. `Map.keys` gives me all the field names. Using `Seq` it would mean to use a `map` call and iterate over all items.

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -91,7 +91,8 @@ class RefreshIncrementalAction(
         refreshDF,
         indexDataPath.toString,
         previousIndexLogEntry.numBuckets,
-        indexConfig.indexedColumns,
+        // should contain the resolved and prefixed field names
```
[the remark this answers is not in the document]
andrei-ionescu: Sorry, I cannot understand the statement above. Can you be more specific?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
sezruby: I guess `keys` also iterates items & needs `toSeq` after keys. As we don't need to consider uniqueness & lookup, I think `Seq` would be better.

src/main/scala/com/microsoft/hyperspace/actions/RefreshActionBase.scala
```diff
@@ -90,7 +91,10 @@ private[actions] abstract class RefreshActionBase(
 
   protected lazy val indexConfig: IndexConfig = {
     val ddColumns = previousIndexLogEntry.derivedDataset.properties.columns
-    IndexConfig(previousIndexLogEntry.name, ddColumns.indexed, ddColumns.included)
+    IndexConfig(
```
sezruby, line 94: Something like 
```
    // As indexed & included columns in previousLogEntry are resolved & prefixed names, 
    // need to remove the prefix to resolve with the dataframe for refresh.
```

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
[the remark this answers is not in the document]
sezruby:
```suggestion
        // previousIndexLogEntry should contain the resolved and prefixed field names.
```
+and I added comment above.. Thanks!

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: @sezruby: I totally disagree, but I'll do what you requested. Bellow are my reason to disagree:

1. The code will be less readable as there will be lots of `.map(_._1)` everywhere. 
2. The `ListMap.keys` is in itself a `Set` so when you do `.keys` it will give you straight away the set collection without any iteration.

andrei-ionescu:
@sezruby, @imback82: Integrated your feedback.

src/test/scala/com/microsoft/hyperspace/util/ResolverUtilsTest.scala
```diff
@@ -0,0 +1,145 @@
...
+
+package com.microsoft.hyperspace.util
+
+import org.apache.spark.SparkFunSuite
+
+import com.microsoft.hyperspace.{HyperspaceException, SparkInvolvedSuite}
+
+class ResolverUtilsTest extends SparkFunSuite with SparkInvolvedSuite {
+
+  test("testResolve against dataframe - simple") {
```
sezruby, line 25: nit: ditto for all test case names, so that they look descriptive.

```suggestion
  test("Verify testResolve against dataframe - simple.") {
```

sezruby (commented):
LGTM, thanks @andrei-ionescu!

src/test/scala/com/microsoft/hyperspace/util/ResolverUtilsTest.scala
andrei-ionescu: I will do this change, although adding `"Verify"` word is redundant. The goal of an UT is to **verify** the function's behaviour.

andrei-ionescu:
@imback82, @sezruby: I integrated your suggestions. Can we move forward with this PR? There are a lot of new PRs that needs to happen to have the support for nested fields fully functional in Hyperspace. Then the second phase will start to happen - to support arrays of nested fields, and maps with nested fields.

src/test/scala/com/microsoft/hyperspace/index/CreateIndexNestedTest.scala
```diff
@@ -0,0 +1,210 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.index
+
+import scala.collection.immutable.ListMap
```
imback82, line 19: not used.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -31,7 +31,7 @@ import org.apache.spark.sql.types.{DataType, StructType}
 
 import com.microsoft.hyperspace.HyperspaceException
 import com.microsoft.hyperspace.actions.Constants
-import com.microsoft.hyperspace.util.PathUtils
+import com.microsoft.hyperspace.util.{PathUtils, SchemaUtils}
```
imback82, line 34: change not needed.

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -305,8 +305,8 @@ object JoinIndexRule
     val rRequiredIndexedCols = lRMap.values.toSeq
 
     // All required columns resolved with base relation.
-    val lRequiredAllCols = resolve(spark, allRequiredCols(left), lBaseAttrs).get
-    val rRequiredAllCols = resolve(spark, allRequiredCols(right), rBaseAttrs).get
+    val lRequiredAllCols = resolve(spark, allRequiredCols(left), leftRelation.plan).get.map(_._1)
```
imback82, line 308: is this change required in this PR?

src/main/scala/com/microsoft/hyperspace/util/SchemaUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+    if (fieldName.contains(".") && !fieldName.startsWith(NESTED_FIELD_PREFIX)) {
+      s"$NESTED_FIELD_PREFIX$fieldName"
+    } else {
+      fieldName
+    }
+  }
+
+  /**
+   * The method prefixes the nested field names from a map where the keys are
+   * the field names and the values are the nested state of that field
```
imback82, line 44: these seem out of sync?

src/main/scala/com/microsoft/hyperspace/util/SchemaUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * The method removes the prefix from a collection of prefixed nested field names.
+   * It returns the original sequence of tuples of field names and nested state.
+   *
+   * The inverse operation is [[prefixNestedFieldNames]].
+   *
+   * @param fieldNames The collection of prefixed field names.
+   * @return A sequence of tuples of field names and nested status.
+   */
+  def removePrefixNestedFieldNames(fieldNames: Seq[String]): Seq[(String, Boolean)] = {
```
imback82, line 84: why do we need to return the nested status? Introduce it when it's needed?

imback82 (approved):
LGTM, thanks @andrei-ionescu.

I have some comments, but to unblock you, I will merge this now, and I will do a follow up PR to address them.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
imback82: > Now, we support top level fields with `.` (dots). So it is good as it is.

Does this work?

```scala
    import spark.implicits._
    Seq(1).toDF("a.b").write.mode("overwrite").parquet("/tmp")
    val df = spark.read.parquet("/tmp")
    hyperspace.createIndex(df, IndexConfig("i", Seq("`a.b`"), Nil))
```

imback82: I should be able to handle this in my follow up

imback82:
> I will do a follow up PR to address them.

Created #393.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
imback82: #393 handles the above case correctly. (we need a test coverage though)

Issue #372, opened 2021-03-03
[PROPOSAL]: Support indexes on nested arrays of struct


andrei-ionescu:
## Problem Statement

Hyperspace does NOT support indexing over columns/fields of type array of struct.

## Background and Motivation

There are use cases where queries like the ones bellow can greatly benefit performance wise from Hyperspace's indexing.

```
df1.filter(array_contains(df1("nested.cnt"), 11)).show
```

```
df1.join(df2, array_contains(df1("nested.cnt"), df2("rank")))
```

## Proposed Solution

This design is based on the work of proposal #347 and #365.

Given the following dataset

```
root
 |-- id: string (nullable = true)
 |-- v: integer (nullable = true)
 |-- nested: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- id: string (nullable = true)
 |    |    |-- cnt: integer (nullable = true)
```

```
+---+---+--------------------+
|id |v  |nested              |
+---+---+--------------------+
|id2|100|[[b1, 1]]           |
|id1|100|[[a1, 1], [a2, 2]]  |
|id3|110|[[c1, 11], [c2, 12]]|
+---+---+--------------------+
```

Anyone should be able to create an index with:

```
hs.createIndex(
  nestedDataset, 
  IndexConfig(
    "idx_arr_nested", 
    indexedColumns = Seq("nested.cnt"), 
    includedColumns = Seq("id")))
```

## Alternatives
N/A

## Known/Potential Compatibility Issues
N/A

## Design

Adding the support for nested fields impacts the following areas:

- Validate nested column names
- Modify the create index action
- Modify the filter and rule index functions

## Creating the index

As seen above the index may be created with already existing Hyperspace APIs.

:warning: ***Important*** :warning: 

*It will NOT be possible to support multiple fields from a nested array due to the way Spark is working right now. This limitation is because arrays need to be exploded to create the proper index and Spark allows `only one generator allowed per select clause`.*

Under the hood the index may be created with something like this:

```
df1.select(explode(df1("nested.cnt"), df1("id")).as("nested__cnt"))
```

The resulted data frame will be:

```
root
 |-- nested__cnt: integer (nullable = true)
 |-- id: string (nullable = true)
```

```
+-----------+---+
|nested__cnt| id|
+-----------+---+
|          1|id2|
|          1|id1|
|          2|id1|
|         11|id3|
|         12|id3|
+-----------+---+
```

[...]

PR #380, opened 2021-03-11
Support filter over indexes on nested fields

andrei-ionescu:

 - **Proposal**: #347
 - **Discussion**: #312
 - **Dependency**: #379 
 - Fixes: #347

This PR adds support for filtering over nested fields using indexes created over nested fields. 

Given the `nestedDataset` dataset with schema

```
root
 |-- name: string (nullable = true)
 |-- nested: struct (nullable = true)
 |    |-- field1: string (nullable = true)
 |    |-- nst: struct (nullable = true)
 |    |    |-- field1: string (nullable = true)
 |    |    |-- field2: string (nullable = true)
```

and the following data

```
+---+-----+-----------------+
|id |name |nested           |
+---+-----+-----------------+
|2  |name2|[va2, [wa2, wb2]]|
|1  |name1|[va1, [wa1, wb1]]|
+---+-----+-----------------+
```

And the following search/filter query

```
df.filter(df("nested.nst.field1") === "wa1").select("id", "name", "nested.nst.field2")
```

The optimized and spark plans without index are

```
Project [id#100, name#101, nested#102.nst.field2]
+- Filter (isnotnull(nested#102) && (nested#102.nst.field1 = wa1))
   +- Relation[id#100,name#101,nested#102] parquet
```

```
Project [id#100, name#101, nested#102.nst.field2]
+- Filter (isnotnull(nested#102) && (nested#102.nst.field1 = wa1))
   +- FileScan parquet [id#100,name#101,nested#102] Batched: false, Format: Parquet, 
        Location: InMemoryFileIndex[file:/..../tableN2], PartitionFilters: [], 
        PushedFilters: [IsNotNull(nested)], ReadSchema:
        struct<id:int,name:string,nested:struct<field1:string,nst:struct<field1:string,field2:string>>>
```

The transformed optimized and spark plans should look like

```
Project [id#1, name#2, nested__nst__field2#3]
+- Filter (isnotnull(nested__nst__field1#0) && (nested__nst__field1#0 = wa1))
   +- Relation[nested__nst__field1#0,id#1,name#2,nested__nst__field2#3]
        Hyperspace(Type: CI, Name: idx_nested, LogVersion: 1)
```

```
Project [id#1, name#2, nested__nst__field2#3]
+- Filter (isnotnull(nested__nst__field1#0) && (nested__nst__field1#0 = wa1))
   +- FileScan parquet [nested__nst__field1#0,id#1,name#2,nested__nst__field2#3] Batched: false, 
        Format: Parquet, Location: InMemoryFileIndex[file:/..../spark_warehouse/indexes/idx_nested/v__=0],
        PartitionFilters: [], PushedFilters: [IsNotNull(nested__nst__field1)], ReadSchema: 
        struct<nested__nst__field1:string,id:int,name:string,nested__nst__field2:string>
```

### Transforming the plan

Filters inside the plan must be modified to accomodate the index schema not the data schema - the flattened schema not the nested field. Instead of accessing the field with `GetStructField(GetStructField(AttributeReference))` it must directly access with `AttributeReference`.

Given the query plan

```
Project [id#100, name#101, nested#102.nst.field2]
+- Filter (isnotnull(nested#102) && (nested#102.nst.field1 = wa1))
   +- Relation[id#100,name#101,nested#102] parquet
```

The filter will be modified from

```
Filter (isnotnull(nested#102) && (nested#102.nst.field1 = wa1))
```

to

```
Filter (isnotnull(nested__nst__field1#0) && (nested__nst__field1#0 = wa1))
```

The projection from

```
Project [id#100, name#101, nested#102.nst.field2]
```

to

```
Project [id#1, name#2, nested__nst__field2#3]
```

The relation from 

```
Relation[id#100,name#101,nested#102] parquet
```

to

```
Relation[nested__nst__field1#0, id#1,name#2, nested__nst__field2#3] Hyperspace(Type: CI, 
  Name: idx_nested, LogVersion: 1)
```

No.

- Existing unit and integration tests for not breaking the existing functionalities.
- Unit and integration test added for the new functionalities.


andrei-ionescu:
@sezruby, @imback82 Here is the second PR that adds support for filtering over nested fields with indexing.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -568,4 +589,56 @@ object RuleUtils {
...
+    fieldNames.foreach { fieldName =>
+      val escapedFieldName = SchemaUtils.prefixNestedFieldName(fieldName)
+      val nestedFields = getNestedFields(index)
+      if (nestedFields.nonEmpty &&
+          nestedFields.exists(i => i.equalsIgnoreCase(escapedFieldName))) {
+        val (parentExpresion, exp) =
+          extractSearchQuery(filter.condition, fieldName)
+        val fieldType = extractTypeFromExpression(exp, fieldName)
+        val attr = extractAttributeRef(exp, fieldName)
+        val exprId = getFieldPosition(index, escapedFieldName)
```
sezruby, line 619: I think `exprId` should be unique. Could  you check `NamedExpression.newExprId` and utilize it if possible?

andrei-ionescu: Yes. I'll look into it.

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
```diff
@@ -0,0 +1,165 @@
...
+   *
+   * @param exp The Spark Catalyst [[Expression]] from which to extract the type.
+   * @param name The field name for which we need to get the type.
+   * @return A Spark SQL [[DataType]] of the given field name.
+   */
+  def extractTypeFromExpression(exp: Expression, name: String): DataType = {
+    val splits = name.split(".")
+    val elem = exp.flatMap {
+      case a: AttributeReference =>
+        if (splits.forall(s => a.name == s)) {
```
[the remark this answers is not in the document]
andrei-ionescu: I don't know what you mean. Any string, even the empty string (`""`), when split it will return at least 1 item. And even for that one item we should iterate over all attributed and find it.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -568,4 +588,56 @@ object RuleUtils {
...
     shuffled
   }
+
+  private def transformProject(project: Project, index: IndexLogEntry): Project = {
+    val projectedFields = project.projectList.map { exp =>
+      val fieldName = extractNamesFromExpression(exp).head
+      val escapedFieldName = SchemaUtils.prefixNestedFieldName(fieldName)
+      val attr = extractAttributeRef(exp, fieldName)
+      val fieldType = extractTypeFromExpression(exp, fieldName)
+      val exprId = getExprId(project, escapedFieldName)
```
sezruby, line 598: Does this work as expected? I think it's because of 'transformDown', transformProject will be called before transforming the relation node. 
Could you add some test for filter index rule?

andrei-ionescu: @sezruby: This functionality is already tested by existing UTs and E2Es. 

This is the correct approach: **Setting the `exprId` on the lowest leaf (aka: relation output) and when transforming each node get that expression id and use it when transforming that node**. So, the same expression id used in the relation is used in the tree otherwise there will be thrown analysis exception saying that the field (aka attribute) could not be found. I cannot set the `exprId` with a value on the relation, then in the filter and projection use other expression id (if I would use `newExprId` again). I have to take it from the relation that I previously created.

For example, given the `__hs_nested.nest.leaf.field`, in the relation I set it to a new number given by `newExprId` (ie: `#701`) and then on the filter to get another `exprId` (ie: `#801`), although at planning it will not crash it will throw exception at runtime.

This is how the correct plan should look like:

```
Project [__hs_nested.nested.nest.field#701]
+- Filter (isnotnull(__hs_nested.nested.nest.field#701) && (__hs_nested.nested.nest.field#701 = wa1))
   +- Relation[__hs_nested.nested.nest.field#701]
```

As you can see that `#701` is everywhere.

Now, regarding the `transformDown` and why it works, I guess the answer is that it is running multiple times.

I'll try switching to `transformUp` and see how it goes.

andrei-ionescu:
re-test

sezruby:
> Is there a mechanism to trigger the build once more on this PR? Something like commenting a keyword to trigger the build again?

You can trigger via "re-run" in `Checks` tab

andrei-ionescu:
@sezruby I don't see any "re-run" in the **Checks** tab. I don't have same right as you. 😃

andrei-ionescu:
@sezruby, @imback82 Can you have another look on the PR so that we can advance with it? Thanks!

sezruby:
Actually I'm waiting for #393 because of the conflict. 
@andrei-ionescu Could you review the change again? Since SchemaUtil is removed in that change, I wonder if it's ok or not.

andrei-ionescu:
@sezruby I didn't know that #393 is a prerequisite for this. If so, let's integrate feedback on that PR and get it merged so that I can advance with this one.

Remember we agreed to split the support for nested fields feature into multiple PRs so it can move faster. So, let's move faster 😁.

andrei-ionescu:
@sezruby, @imback82: I rebased the PR and resolved the conflicts and any issues. Just, FYI: `SchemaUtils` not needed in this PR. Please have another look ASAP.

andrei-ionescu:
@sezruby, @imback82: Ping? It's 2 days since my previous ping and last push. Please have a look on this PR.

andrei-ionescu:
@sezruby, @imback82: Can you re-run the build? It did fail due to:

```
sbt.InvalidComponent: Could not find required component 'xsbti'
```

Also, don't forget to review this PR. Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
```diff
@@ -0,0 +1,174 @@
...
+   * Given an Spark Catalyst [[Expression]], a needle [[Expression]] and a replace
+   * [[Expression]] the method will replace the needle with the replacement into
+   * the parent expression.
+   *
+   * @param parent The parent Spark Catalyst [[Expression]] into which to replace.
+   * @param needle The Spark Catalyst [[Expression]] needle to search for.
+   * @param repl The replacement Spark Catalyst [[Expression]].
+   * @return A new Spark Catalyst [[Expression]].
+   */
+  def replaceInSearchQuery(
```
sezruby, line 104: Could you revise the function name? seems it's irrelevant to search query

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -109,6 +110,45 @@ class FileBasedSourceProviderManager(spark: SparkSession) {
...
   }
 
+  /**
+   * Returns true if the given project is a supported project. If all of the registered
+   * providers return None, this returns false.
+   *
+   * @param project Project to check if it's supported.
+   * @return True if the given project is a supported relation.
+   */
+  def isSupportedProject(project: Project, index: IndexLogEntry): Boolean = {
```
sezruby, line 120: it's more like "hasNestedColumnWithProject" ?

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -109,6 +110,45 @@ class FileBasedSourceProviderManager(spark: SparkSession) {
...
+  }
+
+  /**
+   * Returns true if the given filter is a supported filter. If all of the registered
+   * providers return None, this returns false.
+   *
+   * @param filter Filter to check if it's supported.
+   * @return True if the given project is a supported relation.
+   */
+  def isSupportedFilter(filter: Filter, index: IndexLogEntry): Boolean = {
```
sezruby, line 143: ditto

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -109,6 +110,45 @@ class FileBasedSourceProviderManager(spark: SparkSession) {
...
+   * Returns true if the given project is a supported project. If all of the registered
+   * providers return None, this returns false.
+   *
+   * @param project Project to check if it's supported.
+   * @return True if the given project is a supported relation.
+   */
+  def isSupportedProject(project: Project, index: IndexLogEntry): Boolean = {
+    val indexCols = index.indexedColumns ++ index.includedColumns
+    val resolvedIndexCols = indexCols.map(i => ResolverUtils.ResolvedColumn(i))
+    val hasNestedCols = resolvedIndexCols.exists(_.isNested)
```
sezruby, line 123: How about return early if hasNestedCols is false?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -212,19 +161,106 @@ object ResolverUtils {
...
+   * [[ResolvedColumn]] stores information when a column name is resolved against the
+   * analyzed plan and its schema.
+   *
+   * Outside unit tests, this object should not be created directly, but via the `resolve` function,
+   * or `ResolvedColumn.apply` with a normalized name.
+   *
+   * @param name The column name resolved from an analyzed plan.
+   * @param isNested Flag to denote if this column is nested or not.
+   */
+  private[hyperspace] case class ResolvedColumn(name: String, isNested: Boolean) {
```
sezruby, line 205: Could you move this function up to reduce diff?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -212,19 +161,106 @@ object ResolverUtils {
...
-          throw HyperspaceException("Map types are not supported")
-        case f => Seq(f.name)
+      val fieldOpt = schema.find(f => resolver(f.name, h))
+      fieldOpt match {
+        case Some(field) =>
+          field match {
+            case StructField(name, s: StructType, _, _) =>
+              name +: getColumnNameFromSchema(s, tail, resolver)
+            case StructField(_, _: ArrayType, _, _) =>
+              // TODO: Nested arrays will be supported later
```
sezruby, line 174: Could you create an issue for this & link it? like other TODOs.

sezruby (commented):
Sorry for the delayed review. BTW could you add a simple test for the filter query in E2EHyperspaceRulesTest? Though it's tested in other PRs but need some minimum verifications for this change.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: Support for Arrays of nested fields has the #372 proposal. I'll create a proposal for the maps too.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -111,20 +114,41 @@ object FilterIndexRule
...
+          spark,
+          filterColumns,
+          plan,
+          resolveWithChildren,
+          throwIfNotInSchema = false)
+
+        (resolvedOutputColumnsOpt, resolvedFilterColumnsOpt) match {
+          case (Some(resolvedOutputColumns), Some(resolvedFilterColumns)) =>
+            val candidateIndexes = allIndexes.filter { index =>
+              indexCoversPlan(
```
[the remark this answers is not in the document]
andrei-ionescu: The `resolve` methods used at lines 121 and 127 are different than the one inside `indexCoversPlan`. Inside `indexCoversPlan` it is not used the plan's function to resolve it, it only makes comparison between collections of fields.

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
andrei-ionescu: I will find a better name.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: I will but this is the way `scalafmt` optimises the code.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu: Created #411 for maps.

andrei-ionescu:
@sezruby I integrated your feedback.

The build is still failing with the same previous error:

```
sbt.InvalidComponent: Could not find required component 'xsbti'
```

Can you check your build pipeline?

andrei-ionescu:
@sezruby, @imback82: Can you guys have **another look** and **fix the build** pipeline? It's been a while since this PR is up and I'm doing extra efforts to make the process as smooth as you need. Thanks!

sezruby:
@andrei-ionescu  I made an issue for build failure - https://github.com/actions/virtual-environments/issues/3146 

> I don't think we need another one. Or, is there a specific behaviour that you want to test?

I meant filter index test for nested column. You can change the column schema in an existing test or create a new test for it.

sezruby (commented):
I think not a just simple test, but some test set which is fully covered most of this PR is required as this is not simple enough to skip testing. 
@andrei-ionescu could you add some tests while build problem is fixing? Thanks

src/main/scala/com/microsoft/hyperspace/index/sources/FileBasedSourceProviderManager.scala
```diff
@@ -109,6 +110,49 @@ class FileBasedSourceProviderManager(spark: SparkSession) {
...
   }
 
+  /**
+   * Returns true if the given project is a supported project. If all of the registered
+   * providers return None, this returns false.
+   *
+   * @param project Project to check if it's supported.
+   * @return True if the given project is a supported relation.
+   */
+  def hasNestedColumns(project: Project, index: IndexLogEntry): Boolean = {
```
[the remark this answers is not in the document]
andrei-ionescu: That was the best place I found. Do you have any suggestions?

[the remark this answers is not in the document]
sezruby: Maybe RuleUtils .. ?

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -276,10 +280,29 @@ object RuleUtils {
...
 
-        val updatedOutput = relation.plan.output
-          .filter(attr => indexFsRelation.schema.fieldNames.contains(attr.name))
-          .map(_.asInstanceOf[AttributeReference])
+        val resolvedFields =
+          ResolverUtils.resolve(spark, index.indexedColumns ++ index.includedColumns, relation.plan)
+        val updatedOutput =
+          if (resolvedFields.isDefined && resolvedFields.get.exists(_.isNested)) {
+            indexFsRelation.schema.flatMap { s =>
+              relation.plan.output.find(a => s.name.contains(a.name)).map { a =>
```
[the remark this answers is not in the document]
andrei-ionescu: Modified it to use the `ResolverUtils.ResolvedColumn` construct.

andrei-ionescu:
@sezruby, @imback82: I integrated your feedback and added an extra nested test in the `E2EHyperspaceRulesTest` called `E2E test for filter query on nested columns`.

Please have yet another look to this PR.

**Also, how can we speed things up?** This PR is more than a month old. I expected it to go faster after splitting the previous PR into multiple smaller ones.

sezruby:
@andrei-ionescu Sorry for the delay. but changes are not small or simple; need to spend some time to review it even with small PRs..

BTW... How about this?
original plan => normalized plan (nested filed => __hs_nested.xx) => trying to apply rule =>
1) if an index is applied => rename output
2) not applied => return original plan

In this way, maybe we don't need to change much of our existing code base or nested column related codes are not scattered.

andrei-ionescu:
@sezruby I can try to understand what you are saying. You want to say that in terms of impact the PR is not small - _it has a big impact PR_ - because in terms of number of lines modified in this PR are only 500+ modified lines. 

Even "_it has a big impact PR_" phrase is not true because the impact is minimised by the already present tests that guards the existing functionality of Hyperspace and there are lots of tests - a bit over 300 tests. They may not cover all the edge cases but it minimises the impact.

This second PR is taking more than a month to complete. This is way too much. 

Doing some math here... bringing the full support for nested fields in Hyperspace, with the current pace, it will take 6-12 months for it to land. IMHO **this process can be better**, I've seen it in other bigger projects with heavier features, with bigger PRs. 

@rapoth, @imback82: WDYT? What can be done to make this process faster? I did my part - the part that I could do - what else should I do to speed up the process?

andrei-ionescu:
@sezruby: Regarding this:

>original plan => normalized plan (nested filed => __hs_nested.xx) => trying to apply rule =>
>
>1. if an index is applied => denormalize (rename?) output
>2. not applied => return original plan
>
>In this way, maybe we don't need to change much of our existing code base or nested column related codes are not scattered.

If I understand your text correctly, we are doing exactly what you are saying. Given an index we renamed the output ([R275-R306](https://github.com/microsoft/hyperspace/pull/380/files#diff-8c312528287b51f613809d58a7b14b23c151d65794c099b653605795e0a738edR275-R306)), then we rename the field names in filters ([R310-R311](https://github.com/microsoft/hyperspace/pull/380/files#diff-8c312528287b51f613809d58a7b14b23c151d65794c099b653605795e0a738edR310-R311)) and projections ([R315-R316](https://github.com/microsoft/hyperspace/pull/380/files#diff-8c312528287b51f613809d58a7b14b23c151d65794c099b653605795e0a738edR315-R316)). 

I don't understand your request as for me it seems that the PR is doing exactly what you're requesting.

sezruby:
I meant out side of RuleUtils - normalize query plan from the beginning of apply function in FilterIndexRule & denormalize at the end of the apply function.

andrei-ionescu:
I don't think that is a good approach:
1. It needs to rename the fields in: output, filter and projections.
2. There is already quite some code manipulating the plans in `RuleUtils` so it seems to be the natural place to put it.
3. The same code would be used in hybrid scans and joins.
4. It would require passing through the whole plan twice each time (using `transformUp` and `transformDown` in both `FilterIndexRule` and `RuleUtils`) .
5. `JoinIndexRule` is different than `FilterIndexRule` and both use `RuleUtils` methods.

sezruby:
@andrei-ionescu 
1=> yep, but we could rollback(denormalize) nested columns after applying rules except for from index relation. 
2=> actually.. that's the problem. it's already complicated due to applying indexes & hybrid scan. We don't need to put everything into RuleUtils. I'll also have a plan to refactor hybrid scan because of the complexity.
3, 5=> if we flatten all the nested fields, I guess we don't need to change the code inside the logic of FilterIndexRule or JoinIndexRule or Hybrid Scan.  I currently work on this #403, so we will do flatten/unflatten once after that. (before then, I think it's okay normalize/denormalie for each FilterIndexRule & JoinIndexRule because ..
4=> I think that the benefit (code simplicity) outweighs the multiple iteration of plan. Spark also tries to apply twice (or multiple times) of each rule to confirm there's no transformation from the rule.

andrei-ionescu:
@sezruby 

> 1=> yep, but we could rollback(denormalize) nested columns after applying rules except for from index relation.

Sorry, but I don't understand what you want. Please provide examples with original plan and transformed plan.

> 2=> actually.. that's the problem. it's already complicated due to applying indexes & hybrid scan. We don't need to put everything into RuleUtils. I'll also have a plan to refactor hybrid scan because of the complexity.

I can understand this and I'm ok with splitting the `RuleUtils` big file into smaller ones.

> 3, 5=> if we flatten all the nested fields, I guess we don't need to change the code inside the logic of FilterIndexRule or JoinIndexRule or Hybrid Scan. I currently work on this #403, so we will do flatten/unflatten once after that. (before then, I think it's okay normalize/denormalie for each FilterIndexRule & JoinIndexRule because ..

I don't understand the flatten process you are talking - for me is not clear if you are referring to getting from `nested.field` into `__hs_nested.nested.field` or is other process.

I've seen the your "HyperspaceOneRule" proposal and PR and it looks great. I've seen how it's being split at this moment and that the `RuleUtils` file remains the same. 

But, it is **_NOT FAIR_** to postpone and stall this PR for that one, neither making me address all the code conflicts - although I'm willing to address all of them if your PR gets merged fast (fast == next day).  

> 4=> I think that the benefit (code simplicity) outweighs multiple iterations of plan. Spark also tries to apply twice (or multiple times) of each rule to confirm there's no transformation from the rule.

Sorry, I cannot see the "code simplicity" you are talking about. This may be because I don't understand what you want as I already said.

sezruby:
@andrei-ionescu Ok let me explain more about my suggestion:

queryPlan = original  query plan before transformation

1) nomalizedPlan = flatten + prefix all nested columns in queryPlan
```
       *   Project [id#100, name#101, nested#102.nst.field2]
       *   +- Filter (isnotnull(nested#102) && (nested#102.nst.field1 = wa1))
       *      +- Relation[id#100,name#101,nested#102] parquet
```
to
```
       *   Project [id#100, name#101, __hs_nested.nested.nst.field2#3]
       *   +- Filter (isnotnull(__hs_nested.nested.nst.field1#0) && (__hs_nested.nested.nst.field1#0 = wa1))
       *      +- Relation[__hs_nested.nested.nst.field1#0,id#100,name#101, __hs_nested.nested.nst.field2#3] parquet
```

2) try to apply rules to normalizedPlan, so
```
  override def apply(plan: LogicalPlan): LogicalPlan = {
    ...
    plan transformDown {
```
will be:
```
  override def apply(plan: LogicalPlan): LogicalPlan = {
    val normalizedPlan = normalizeNestedColumns(plan)
    val transformedPlan = normalizedPlan transformDown {
   ..
```

3) after applying rules, 

```
// if there's no transformation, return original queryPlan
if (transformedPlan.equals(normalizedPlan)) return plan

// Otherwise, 
denormalizeNestedColumns(transformedPlan)
```

In denormalizedNestedColumns, we could check if each normalized column is from index relation or not and replace it with the original query plan tree if it's not from the index relation.

===
BTW should we rename the output column or no?
For example:
```
Project [id#1, name#2, __hs_nested.nested.nst.field2#3]
...
``
What is the result column name of " __hs_nested.nested.nst.field2#3"?

**Comment by andrei-ionescu** [comment]
@sezruby 
> BTW should we rename the output column or no?

Yes. We should. I have that in a later PR and forgot about it.

**Comment by andrei-ionescu** [comment]
@sezruby 

> 1. nomalizedPlan = flatten + prefix all nested columns in queryPlan

This requires to go through the whole plan with `transformDown` or `transformUp` and: a) change the relation (schema & attribute refs name and exprId)  b) change the filter (attribute refs name and exprId that needs to match the relation) c) change the project (attribute refs name and exprId that also needs to match the relation one). This is how I understand to get the "normalizedPlan" as you call it.

> 2. try to apply rules to normalizedPlan, so

Given the transformation above, what else is left to apply? I don't understand this "apply rules to normalised plan". Which rules are you referring to? For me a rule is a matching of a plan node that will get transformed.

If `normalizeNestedColumns(plan)` does run a `transformDown` and changes the plan altogether, doing `val transformedPlan = normalizedPlan transformDown` doesn't make any sense. 

> 3. after applying rules,
>
> ```
> if (transformedPlan.equals(normalizedPlan)) return plan
> ```

This does not make any sense either. You should check against the original `plan` not `transformed` vs `normalizedPlan` and return `plan`.

I guess I don't understand what "apply rules" does. The current Hyperspace approach is to FIRST select a suitable index over all the plan and after that is found start the plan transformation based in the columns inside the index. Is your PR - "HyperspaceOneRule" - changing this process?

**Comment by sezruby** [comment]
Sorry about rule"s".. I meant it's FilterIndexRule. I'd like to separate "applying index" and "handling nested columns".

normalizedNestedColumns only transforms nested columns - without considering indexes. 
(though not sure this is possible - creating relation node using different column schema during the transformation)

**Comment by andrei-ionescu** [comment]
@sezruby

> separate "applying index" and "handling nested columns"

What's the use case for this separation? Will there be ever a case where we will do `applyIndex(plan)` and then not iterate over the plan changing attributes where needed?
  
In the context of nested fields `applyIndex(plan)` without transforming the rest of the plan makes no sense and will NOT work.

> normalizedNestedColumns only transforms nested columns - without considering indexes.

IMHO I don't think this is a good approach - to transform the all the nested columns without considering what's in the index. I think the opposite: we should change only what matches for changes, not everything.

From your comments, reading between the lines, I understand that you are NOT happy that supporting nested fields adds so many changes in the current `RuleUtils` file in the `transformPlanToUseIndexOnlyScan` and `transformPlanToUseHybridScan` methods. If this is the case I propose to make the `RuleUtils` a case class and name it `BaseRuleHelper`. Then out of it, create the `NestedFieldsRuleHelper` that will extend it and override the needed methods. `FilterIndexRule` and `JoinIndexRule` should detect the case and when there are nested fields involved, it should use the `NestedFieldsRuleHelper` otherwise fall back to `BaseRuleHelper`.

@sezruby, @imback82 WDYT?

**Comment by sezruby** [comment]
@andrei-ionescu That sounds a good alternative. It's already complicated with source provider & hybrid scan (need to decrypt to read it). Let's modulize it as much as possible.. 
Sorry for the late suggestion, but it came up after reading the code over and over while reviewing.

**Comment by andrei-ionescu** [comment]
@imback82 Are you ok with it? I'm just making sure that I'm not starting working on something that is not agreed by you too. Otherwise my effort may be just wasted time.

**Comment by sezruby** [comment]
@clee704 could you review this PR & above few discussion comments & share some opinion on it?

**Comment by andrei-ionescu** [comment]
@clee704, @imback82, @sezruby Can I have an answer on my question and proposal above, so that I can advance with this PR?

**Comment by sezruby** [comment]
@andrei-ionescu I believe that it would be better than the current approach - if we need to support other plan type, it will be getting bigger & complicated (e.g. nested array, map, ..) and I'm the main reviewer before @imback82, so feel free to proceed it. 
@clee704 is relatively new to Hyperspace so it might need some time to review this change & @imback82 is so busy this week..

**Comment by andrei-ionescu** [comment]
@sezruby Thanks for green light. I understand that your approval should be enough for this PR, once that I change the implementation. I don't want to be in the position that someone else disagrees with it and I have to start over one more. You know: once bitten, twice shy.

**Comment by rapoth** [comment]
@andrei-ionescu Thanks for double-checking! Please go ahead with @sezruby's recommendation - I agree that's a sound proposal and it will help us all in the long run. :)

**Comment by andrei-ionescu** [comment]
@rapoth, @sezruby Thanks for agreement. I'll start grinding on it ASAP.

File: src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala [hunk 13]
```diff
@@ -60,17 +60,16 @@ object JoinIndexRule
         getBestIndexPair(l, r, condition)
           .map {
             case (lIndex, rIndex) =>
+              val ruleHelper = new BaseRuleHelper(spark)
```
andrei-ionescu, line 63: Changes in this file are mainl due to `RuleUtils` changes.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -90,16 +90,17 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
     val indexManager = IndexCollectionManager(spark)
     val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
 
-    val t1Relation = RuleUtils.getRelation(spark, t1ProjectNode).get
-    val t2Relation = RuleUtils.getRelation(spark, t2ProjectNode).get
-    assert(RuleUtils.getCandidateIndexes(spark, allIndexes, t1Relation).length === 3)
-    assert(RuleUtils.getCandidateIndexes(spark, allIndexes, t2Relation).length === 2)
+    val ruleHelper = new BaseRuleHelper(spark)
```
andrei-ionescu, line 93: Changes in here are due to the fact the `RuleUtils` has been renamed and transformed into a class.

andrei-ionescu:
@sezruby: Here is the refactor:

1. Renamed `RuleUtils` into `BaseRuleHelper` and made it a class
2. Created `NestedRuleHelper` that extends `BaseRuleHelper`
3. Added "detection code" into `FilterIndexRule`
4. Kept the new nested field E2E test
5. Added index nested field rename back to original name

Please have a look.

src/main/scala/com/microsoft/hyperspace/index/rules/NestedRuleHelper.scala
```diff
@@ -0,0 +1,501 @@
...
+
+        val indexFsRelation = new IndexHadoopFsRelation(
+          location,
+          new StructType(),
+          newSchema,
+          if (useBucketSpec) Some(index.bucketSpec) else None,
+          new ParquetFileFormat,
+          Map(IndexConstants.INDEX_RELATION_IDENTIFIER))(spark, index)
+
+        val resolvedFields = ResolverUtils.resolve(
```
sezruby, line 79: Could you define some functions (outputFields, transformProject..) and override them rather than overriding whole transformPlanToUse*?
In this way, it's hard to maintain changes in transformPlanToUse*

src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -108,11 +107,12 @@ object JoinIndexRule
   private def isApplicable(l: LogicalPlan, r: LogicalPlan, condition: Expression): Boolean = {
     // The given plan is eligible if it is supported and index has not been applied.
     def isEligible(optRel: Option[FileBasedRelation]): Boolean = {
-      optRel.map(!RuleUtils.isIndexApplied(_)).getOrElse(false)
+      optRel.exists(!BaseRuleHelper.isIndexApplied(_))
     }
 
-    lazy val optLeftRel = RuleUtils.getRelation(spark, l)
-    lazy val optRightRel = RuleUtils.getRelation(spark, r)
+    lazy val ruleHelper = new BaseRuleHelper(spark)
```
sezruby, line 113: Could you let some of utility functions like getRelation in `object RuleUtils`? to avoid unnecessary instance creation

andrei-ionescu: The only method I created as outside of the class instantiation is `BaseRuleHelper.isIndexApplied` and is part of the companion object. 

All other functions do use `spark: SparkSession` inside their logic and  because of it they are now used through instantiation. 

Regarding `getRelation` that also uses `spark` inside it but we can extract that part and provide a `HyperspaceContext` parameter.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -53,19 +54,24 @@ object FilterIndexRule
...
-              // useBucketUnionForAppended as false. If it's true, Hybrid Scan can cause
-              // unnecessary shuffle for appended data to apply BucketUnion for merging data.
-              val transformedPlan =
-                RuleUtils.transformPlanToUseIndex(
-                  spark,
-                  index,
-                  originalPlan,
-                  useBucketSpec = HyperspaceConf.useBucketSpecForFilterRule(spark),
-                  useBucketUnionForAppended = false)
+              // Detect which whether the index contains nested fields.
```
clee704, line 60:
```suggestion
              // Detect whether the index contains nested fields.
```

src/main/scala/com/microsoft/hyperspace/index/rules/NestedRuleHelper.scala
```diff
@@ -0,0 +1,262 @@
...
+   * @return True if the given project is a supported relation.
+   */
+  protected[rules] def hasNestedColumns(filter: Filter, index: IndexLogEntry): Boolean = {
+    val indexCols =
+      (index.indexedColumns ++ index.includedColumns).map(i => ResolverUtils.ResolvedColumn(i))
+    val hasNestedCols = indexCols.exists(_.isNested)
+    if (hasNestedCols) {
+      val filterFields = extractNamesFromExpression(filter.condition).toKeep.toSeq
+      val resolvedFilterFields = filterFields.map(ResolverUtils.ResolvedColumn(_))
+      resolvedFilterFields.exists(i => indexCols.exists(j => j == i || j.name == i.name))
```
clee704, line 243: Is it possible that j.name != i.name but j == i?

src/main/scala/com/microsoft/hyperspace/index/rules/NestedRuleHelper.scala
```diff
@@ -0,0 +1,262 @@
...
+   *
+   * The projection part should become:
+   * {{{
+   *   Project [Date#330, __hs_nested.nested.leaf.cnt#335]
+   * }}}
+   *
+   * @param project The project that needs to be transformed.
+   * @return The transformed project with support for nested indexed fields.
+   */
+  protected[rules] def transformProject(project: Project): Project = {
```
clee704, line 176: Does it work for nested fields in expressions? Example: `select my.nested.field1 + my.nested.field2 from mytab`

src/main/scala/com/microsoft/hyperspace/index/rules/NestedRuleHelper.scala
```diff
@@ -0,0 +1,262 @@
...
+   * The filter part should become:
+   * {{{
+   *   Filter ((__hs_nested.nested.leaf.cnt#335 > 10) &&
+   *           (__hs_nested.nested#.leaf.id#336 = leaf_id9))
+   * }}}
+   *
+   * @param filter The filter that needs to be transformed.
+   * @return The transformed filter with support for nested indexed fields.
+   */
+  protected[rules] def transformFilter(filter: Filter): Filter = {
```
clee704, line 114: seems to be a good candidate for a unit test

src/main/scala/com/microsoft/hyperspace/index/rules/NestedRuleHelper.scala
clee704: Also, it seems to be a good candidate for a unit test

src/main/scala/com/microsoft/hyperspace/index/rules/NestedRuleHelper.scala
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  /**
+   * Returns true if the given project is a supported project. If all of the registered
+   * providers return None, this returns false.
+   *
+   * @param project Project to check if it's supported.
+   * @return True if the given project is a supported relation.
+   */
+  protected[rules] def hasNestedColumns(project: Project, index: IndexLogEntry): Boolean = {
```
clee704, line 209: seems to be a good candidate for a unit test

src/main/scala/com/microsoft/hyperspace/index/rules/NestedRuleHelper.scala
```diff
@@ -0,0 +1,262 @@
...
+    }
+  }
+
+  /**
+   * Returns true if the given filter has nested columns.
+   *
+   * @param filter Filter to check if it's supported.
+   * @return True if the given project is a supported relation.
+   */
+  protected[rules] def hasNestedColumns(filter: Filter, index: IndexLogEntry): Boolean = {
```
clee704, line 236: seems to be a good candidate for a unit test

src/main/scala/com/microsoft/hyperspace/index/rules/NestedRuleHelper.scala
```diff
@@ -0,0 +1,262 @@
...
+   * The method retrieves the expression id for a given field name.
+   *
+   * This method should be mainly used when transforming plans and the
+   * leaves are already transformed.
+   *
+   * @param plan The logical plan from which to get the expression id.
+   * @param fieldName The name of the field to search for.
+   * @return An [[ExprId]] if that could be found in the plan otherwise [[None]].
+   */
+  private def getExprId(plan: LogicalPlan, fieldName: String): Option[ExprId] = {
```
clee704, line 259: seems to be a good candidate for a unit test

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
```diff
@@ -0,0 +1,198 @@
...
+
+object PlanUtils {
+
+  /**
+   * The method extract field names from a Spark Catalyst [[Expression]].
+   *
+   * @param exp The Spark Catalyst expression from which to extract names.
+   * @return A set of distinct field names.
+   */
+  def extractNamesFromExpression(exp: Expression): ExtractedNames = {
```
clee704, line 34: seems to be a good candidate for a unit test

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
```diff
@@ -0,0 +1,198 @@
...
+package com.microsoft.hyperspace.index.rules
+
+import scala.util.Try
+
+import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, BinaryExpression, Expression, GetStructField, IsNotNull, UnaryExpression}
+import org.apache.spark.sql.types.{DataType, StructType}
+
+import com.microsoft.hyperspace.util.ResolverUtils
+
+object PlanUtils {
```
clee704, line 26: The name sounds general, but the actual code seems mostly about nested fields?

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
```diff
@@ -0,0 +1,198 @@
...
+        case u: IsNotNull =>
+          extractNames(u.child, prevExpStrTypes :+ "isNotNull")
+        case u: UnaryExpression =>
+          extractNames(u.child, prevExpStrTypes :+ "unary")
+        case _ =>
+          Set.empty[(String, Seq[String])]
+      }
+    }
+
+    var toRemove = Seq.empty[String]
```
clee704, line 59: This is to remove the unnecessary isnull check, right? If someone without the knowledge that this code is about supporting nested fields, the meaning of this name could be vague.

clee704:
Maybe a dumb question, but why should we rename the fields? Can't we just keep the original nested structures in index data?

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
andrei-ionescu: No. Given this plan:
```
Project [Date#89, nested#94.leaf.cnt AS cnt#336, nested#94.leaf.id AS id#337]
+- Filter ((isnotnull(nested#94) && (nested#94.leaf.cnt > 10)) && (nested#94.leaf.id = leaf_id9))
   +- Relation[Date#89,RGUID#90,Query#91,imprs#92,clicks#93,nested#94] parquet
```
It will be transformed into:
```
Project [Date#380 AS Date#381, __hs_nested.nested.leaf.cnt#379 AS cnt#382, __hs_nested.nested.leaf.id#378 AS id#383]
+- Filter ((__hs_nested.nested.leaf.cnt#379 > 10) && (__hs_nested.nested.leaf.id#378 = leaf_id9))
   +- Relation[__hs_nested.nested.leaf.id#378,__hs_nested.nested.leaf.cnt#379,Date#380] Hyperspace(Type: CI, Name: filterNestedIndex, LogVersion: 1)
```
The `isnotnull(nested#94)` is no longer suitable and is removed in the `transformFilter` method because:
- The `nested` field name is not part of the index (index contains `__hs_nested.nested.leaf.cnt` and `__hs_nested.nested.leaf.id`)
- The `isnotnull` construct checks for a nested field to not be null because trying to access the leaves when it's null would end up in exceptions
- The values stored in the index are not nested, they are flat
- In the example above, we cannot just rename the `isnotnull` attribute reference as we have two fields instead of one (`__hs_nested.nested.leaf.cnt` and `__hs_nested.nested.leaf.id`)

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
andrei-ionescu: The code work for any kind of field. Is needed by the nested fields feature because it needs to do more transformations.

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
andrei-ionescu: I will add tests to all methods in here.

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
sezruby: Can we add both isnotnull(__hs_nested.nested.leaf.cnt) & isnotnull(__hs_nested.nested.leaf.id)? 
Since it seems Spark automatically adds `isnotnull` for project columns and filter condition columns.

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
sezruby: how about `replaceExpression`

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * @param parent The parent Spark Catalyst [[Expression]] into which to replace.
+   * @param needle The Spark Catalyst [[Expression]] needle to search for.
+   * @param repl The replacement Spark Catalyst [[Expression]].
+   * @return A new Spark Catalyst [[Expression]].
+   */
+  def replaceInSearchQuery(
+      parent: Expression,
+      needle: Expression,
+      repl: Expression): Expression = {
```
sezruby, line 129: `replacement` for more clarity?

sezruby:
> Maybe a dumb question, but why should we rename the fields? Can't we just keep the original nested structures in index data?

@clee704 That's a good question. One reason could be that we store index data as "indexed columns" ++ "included columns". For example, if indexed columns are `nested.a, b` and included columns are `nested.c, d`, then it's not possible to define the column schema with that order because of the same field name - `nested`. (But I wonder the order of columns in index data does really matter. )
And maybe another reason might be there're several issues on handling nested columns itself in Spark & to minimize design change in Hyperspace.

src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala
andrei-ionescu: I think we can. I'll try it.

andrei-ionescu:
@sezruby, @clee704: The datasets I use to create indexes over contains deep nested fields with many children.

For example this is a half of the schema of one of my datasets:

```
root
 |-- ts: timestamp (nullable = true)
 |-- id: string (nullable = true)
 |-- xs: struct (nullable = true)
 |    |-- an: struct (nullable = true)
 |    |    |-- ev1_100: struct (nullable = true)
 |    |    |    |-- e1: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |    |-- e2: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |      ...
 |    |    |    |-- e100: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |-- ev101_200: struct (nullable = true)
 |    |    |    |-- e101: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |      ...
 |    |    |    |-- e200: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |-- ev201_300: struct (nullable = true)
 |    |    |    |-- e201: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |      ...
 |    |    |    |-- e300: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |-- ev301_400: struct (nullable = true)
 |    |    |    |-- e301: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |      ...
 |    |    |    |-- e400: struct (nullable = true)
 |    |    |    |    |-- value: double (nullable = true)
 |    |    |-- dim: struct (nullable = true)
 |    |    |    |-- v: struct (nullable = true)
 |    |    |    |    |-- v1: string (nullable = true)
 |    |    |    |    |-- v2: string (nullable = true)
 |    |    |    |      ...
 |    |    |    |    |-- v99: string (nullable = true)
 |    |    |    |    |-- v100: string (nullable = true)
 |    |    |    |-- p: struct (nullable = true)
 |    |    |    |    |-- p1: string (nullable = true)
 |    |    |    |    |-- p2: string (nullable = true)
 |    |    |    |      ...
 |    |    |    |    |-- p_99: string (nullable = true)
 |    |    |    |    |-- p_100: string (nullable = true)
 |    |    |    |-- hiers: struct (nullable = true)
 |    |    |    |    |-- h1: struct (nullable = true)
 |    |    |    |    |    |-- values: array (nullable = true)
 |    |    |    |    |    |    |-- element: string (containsNull = true)
 |    |    |    |    |    |-- delimiter: string (nullable = true)
 |    |    |    |      ...
 |    |    |    |-- lists: struct (nullable = true)
 |    |    |    |    |-- l1: struct (nullable = true)
 |    |    |    |    |    |-- list: array (nullable = true)
 |    |    |    |    |    |    |-- element: struct (containsNull = true)
 |    |    |    |    |    |    |    |-- value: string (nullable = true)
 |    |    |    |    |-- l2: struct (nullable = true)
 |    |    |    |    |    |-- list: array (nullable = true)
 |    |    |    |    |    |    |-- element: struct (containsNull = true)
 |    |    |    |    |    |    |    |-- value: string (nullable = true)
 |    |    |    |      ...
 |-- sch: struct (nullable = true)
 |    |-- se: string (nullable = true)
 |    |-- keys: string (nullable = true)
```

I want to index over the `xs.an.dim.p.p_99` nested field. Are you proposing to keep the whole `xs` struct in the index?

If so, I don't think that is efficient at all and I'll give a few reasons:
1. **Storage cost** — It will take a lot of space
2. **Performance cost** — It will be less usable due to response time when used in queries
3. **Management cost** — It will be harder to maintain up to date

Performance and cost wise I can say that the current approach — _extracting/flattening the nested field_ — is what is needed in Hyperspace even though the implementation code gets more complex.

FYI: I'm currently a bit busy with some other work assignment but I'll get back to integrating your feedback soon.

clee704:
@andrei-ionescu Since we store data in a parquet format, which is efficient in storing nested data in a columnar format, shouldn't it matter? For a parquet file, a dataset with a single column "a" or a single nested column "a.b.c.d.e" should make no difference, I guess.

andrei-ionescu:
It does matter! Even if parquet stores data in an efficient columnar way, there are issues with parquet and nested fields in the way the predicate push down works - or to say it better, it doesn't work. This is the **performance cost**. There is the **storage cost** too that is not negligible either - if you are forced to store half of the dataset in an index. Lastly but not the least, the **management cost** (rebuild index, incremental refresh, etc) will incur additional time that will make index harder update and keep in sync with the data itself - if the index is not easily (fast & efficient) updatable then it's not usable in these _big data fast moving datasets_. These costs are not acceptable for the cases of big nested datasets.

@rapoth, @imback82, @sezruby, @apoorvedave1, @clee704: Here is an article I wrote on what Adobe is doing on big data and what's the scale we work with: https://medium.com/adobetech/high-throughput-ingestion-with-iceberg-ccf7877a413f. Or if you prefer video here is a presentation on the same matter: https://www.dremio.com/subsurface/high-frequency-small-files-vs-slow-moving-datasets.

imback82:
I don't think we can "bucket by" nested fields either. BTW, predicate pushdown for nested fields should be supported in Spark 3.0: https://github.com/apache/spark/commit/cb0db213736de5c5c02b09a2d5c3e17254708ce1

@clee704 is your concerned addressed?

andrei-ionescu:
That's right @imback82! I totally forgot about the bucketing done in Hyperspace. That is not possible over nested fields.

clee704:
> I want to index over the xs.an.dim.p.p_99 nested field. Are you proposing to keep the whole xs struct in the index?

Can't we just store the indexed/included nested fields, ignoring others?

andrei-ionescu:
@sezruby, @clee704, @imback82 I'll get back to this PR very soon. I've been side tracked by another assignment but I'll get back to it integrating the feedback and adding the needed UTs.

andrei-ionescu:
@imback82, @sezruby, @clee704: Sorry for being out of this PR for a few weeks. Now, I got some time and I'm back to it. Here are the changes:

- Rebased it over Spark 3 support and resolved conflicts
- Integrated your feedback
- Added `PlanUtilsTest` and `NestedRuleHelperTest` tests to validate changes on the plan

Please have another look.

src/main/scala/com/microsoft/hyperspace/index/rules/BaseRuleHelper.scala
```diff
@@ -0,0 +1,574 @@
...
+   * If Hybrid Scan is enabled, it compares the file metadata directly, and does not
+   * match signatures. By doing that, we could perform file-level comparison between
+   * index source files and the input files of the given plan. If there are some common
+   * files, the index is considered as a candidate.
+   *
+   * @param indexes List of available indexes.
+   * @param relation Relation with which given indexes are compared.
+   * @return Active indexes built for this plan.
+   */
+  def getCandidateIndexes(
```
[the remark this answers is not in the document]
sezruby: Please address this comment - to reduce the diff

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule.scala
```diff
@@ -53,19 +54,24 @@ object FilterIndexRule
...
-                RuleUtils.transformPlanToUseIndex(
-                  spark,
-                  index,
-                  originalPlan,
-                  useBucketSpec = HyperspaceConf.useBucketSpecForFilterRule(spark),
-                  useBucketUnionForAppended = false)
+              // Detect whether the index contains nested fields.
+              val indexHasNestedColumns = (index.indexedColumns ++ index.includedColumns)
+                .exists(ResolverUtils.ResolvedColumn(_).isNested)
+              val ruleHelper = if (indexHasNestedColumns) {
```
sezruby, line 63: instead of this, how about defining 

```scala
object IndexRelationApplyHelper {
  def apply(index: IndexLogEntry): IndexRelationApplyHelper
}
```

src/main/scala/com/microsoft/hyperspace/index/rules/BaseRuleHelper.scala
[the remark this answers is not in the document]
andrei-ionescu: @sezruby 

1) I rather not put back `getCandidateIndexes` as the method requires spark being given to it. The current implementation is consistent as the methods that don't require spark are in `RuleUtils` while the other are part of the helper instances.
2) There 3000 lined due to extensive tests that I added to the verify the plan transformation. In a way or another someone should review it.
3) "_Frankly speaking_", I already did a lot of extra work complying with your requests of splitting in multiple PRs. Now, once more this request to split PR to this same feature. This is very frustrating for me and looks like you using any opportunity to push this back. I would understand the split into multiple PRs for new features but I though we did come to an agreement at the time we decided to split the feature in 4-5 other PRs.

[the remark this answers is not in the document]
sezruby: 1=>
It will be  refactored out after rule refactoring. Just put it back to reduce the diff...
You can use spark from FilterIndexRule / JoinIndexRule, as before.
Also no need to `getCandidate` function in this class. (IndexPlanApplyHelper ?) - they have different role.

2/3=>
I understand your frustration, but the changes modify wide range of codebase and also core part.
This refactoring was not considered at the time we agreed and that's why this change became huge.
And you know small & explicit changes can be merged quickly, in general.. and splitting PR is also helping to trace the history. 

Sorry but I think we should do that. If you are not up for it, I'll push a refactoring PR(no nested related code) based on your PR, until early next week.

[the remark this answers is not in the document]
andrei-ionescu: @sezruby I want to correct you. The code does not add lots of changes. If you remember - and to help you remember, here is a previous discussion a month back, https://github.com/microsoft/hyperspace/pull/380#issuecomment-820231714 - we agreed to have separate classes for the already existing code and the nested features. So now we have `BaseRuleHelper` and `NestedRuleHelper`. From the old `RuleUtils` I extracted all the methods that need `spark` into the `BaseRuleHelper` and kept the methods that don't need `spark` in `RuleUtils` as static methods. So, from the old `RuleUtils` now we have 2 files: a stripped down `RuleUtils` and the `BaseRuleHelper`. The classes have the code unchanged from the old `RuleUtils`, only the method signatures are changed now that `spark` is a property inside the instance.

The new things are inside `NestedRuleHelper` which extends some methods from `BaseRuleHelper` to accommodate the nested fields support.

This is exactly what we agreed on that comment.

There is no other refactor here than this what I explained above refactor that does NOT make any sense without the next changes.

I propose to merge try merge this and you create the changes that you want on top of it.

@imback82, @rapoth: What is your opinion?

andrei-ionescu:
@imback82, @rapoth: How can we get this moving? It's more than 3 months now...

Issue #206, opened 2020-10-13
Check existence of all index data files before use


sezruby:
### Describe the issue

Currently, we replace the location with the index data files:

```
 plan transformDown {
      case baseRelation @ LogicalRelation(_: HadoopFsRelation, baseOutput, _, _) =>
        val location =
          new InMemoryFileIndex(spark, index.content.files, Map(), None)
        val relation = HadoopFsRelation(
          location,
```

However, this cannot detect the absence of some index data files, which might cause unnoticeable correctness issue.
So we need to check the existence of each index data files for data integrity. 

We recently introduced the incremental index for appended files, so we need to make sure that some dependent old versions should be available.

rapoth:
We can introduce a flag for this check since it might be expensive. For instance: `hyperspace.index.integrityCheck.enabled` and set it to `false` by default.

@apoorvedave1 Can you create an issue for other related tasks as appropriate and link to this issue please?

PR #389, opened 2021-03-23
Check available index versions for Delta Lake time travel query

### LINKED ISSUE #387: Check removed index log versions of time travel query index candidates
Currently, time travel query optimization assumes all index log versions exist. 
However, as most of index applications are based on the latest index log version, some of outdated version might be deleted by a user. 
Therefore, we need to check the removed log versions and filter them out in the candidate log version list in `closestIndexVersion`.

ref:
https://github.com/microsoft/hyperspace/pull/272#r598211313_

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**:  n/a
 - **Dependencies**: n/a

Follow up PR -  https://github.com/microsoft/hyperspace/pull/272#discussion_r598329940

Fixes #387

Check the existence of index data directories for each index log version before using them in Delta Lake time travel query. 
To do so, added below APIs:
```scala
// in IndexLogManager
  /** Returns index log versions whose state is in the given states */
  def getIndexVersions(states: Seq[String]): Seq[Int]

// in IndexManager
/**
   * Get index log version ids of the given index that match any of the given states.
   *
   * @param indexName Name of the index.
   * @param states List of index states of interest.
   * @return Index log versions.
   */
  def getIndexVersions(indexName: String, states: Seq[String]): Seq[Int]
```

No

Unit test


src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -162,12 +180,6 @@ class IndexCollectionManager(
     }
   }
 
-  override def getIndex(indexName: String, logVersion: Int): Option[IndexLogEntry] = {
```
sezruby, line 165: note: moved to above

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -163,6 +163,17 @@ class Hyperspace(spark: SparkSession) {
...
   }
+
+  /**
+   * Get index metadata and detailed index statistics for a given index and log version.
+   *
+   * @param indexName Name of the index to get stats for.
+   * @param logVersion Index log version to retrieve.
+   * @return Index metadata and statistics as a [[DataFrame]].
+   */
+  def index(indexName: String, logVersion: Int): DataFrame = {
```
imback82, line 174: https://github.com/microsoft/hyperspace/pull/272#discussion_r598329940 was actually asking to add an API to `IndexManager` to return available index versions. For example, `def getAvailableIndexVersions(indexName: String): Seq[Int]` so that versions can be used for `def getIndex(indexName: String, logVersion: Int)`. Does it make sense?

We can also update the existing `def index(indexName: String): DataFrame` to return all available versions of index instead of exposing a new `def index(indexName: String, logVersion: Int)` (I don't think we need to expose this as public). Btw, you don't have to address this in this PR since I am not sure if we need it or not. We can just concentrate on updating `IndexLogManager` in this PR.

sezruby: >#272 (comment) was actually asking to add an API to IndexManager to return available index versions. For example, def getAvailableIndexVersions(indexName: String): Seq[Int] so that versions can be used for def getIndex(indexName: String, logVersion: Int). Does it make sense?

Yep then we could fix #387 with the API. I will update this PR.

>We can also update the existing def index(indexName: String): DataFrame to return all available versions of index instead of exposing a new def index(indexName: String, logVersion: Int) (I don't think we need to expose this as public). Btw, you don't have to address this in this PR since I am not sure if we need it or not. We can just concentrate on updating IndexLogManager in this PR.

I think it might be good to expose this API since we expose the log version in plan. A user might want to check the used log version.

imback82: > I think it might be good to expose this API since we expose the log version in plan. A user might want to check the used log version.

If we want to expose the log versions, we can change `index` to return log entries of multiple versions (add an optional param to return last N entries). Since we return `DataFrame`, the user can just filter by the version if needed; I don't think it's necessary to expose this new API, which will require us to expose another API to return available versions.

sezruby: @imback82 BTW, currently if an index is deleted & vacuumed and then a new index with the same index name is created, log files are not removed & the log number is continued. I think this can be a problem. WDYT?

imback82: Good point. It was intentional to keep the log files around (compliance reason), but I think we could provide an option to remove log files.

imback82: Feel free to create an issue so we can track it. Thanks.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -206,7 +214,7 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
         if (equalOrLessThanLastIndex == versions.size - 1) {
           // The given table version is equal or larger than the latest index's.
           // Use the latest version.
-          index
```
sezruby, line 209: @imback82 `getLatestLog()` in getIndexes doesn't guarantee the availability & ACTIVE state. Should we also fix that?

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -138,6 +138,24 @@ class IndexCollectionManager(
...
+  }
+
+  override def getAvailableIndexVersions(indexName: String): Seq[Int] = {
+    withLogManager(indexName) { logManager =>
+      logManager.getActiveIndexVersions().filter { id =>
+        val log = logManager.getLog(id).get
+        val indexDataPaths = log.asInstanceOf[IndexLogEntry].indexContentDirectoryPaths
+        val hadoopConf = spark.sessionState.newHadoopConf()
+        val fs = fileSystemFactory.create(new Path(indexDataPaths.head), hadoopConf)
+        indexDataPaths.forall(path => fs.exists(new Path(path)))
```
imback82, line 154: This is to handle the case where index is vacuumed? How about we move the check to `getActiveIndexVersions` where you stop iterating if you see vacuum operation?

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -44,6 +44,9 @@ trait IndexLogManager {
   /** Returns the latest LogEntry whose state is STABLE */
   def getLatestStableLog(): Option[LogEntry]
 
+  /** Returns Active index log versions */
+  def getActiveIndexVersions(): Seq[Int]
```
imback82, line 48: Does it make sense to pass `states: Seq[String]` to make it more generic? `def getIndexVersions(states: Seq[String]): Seq[Int]`?

imback82, line 48: I noticed that we are using "id" to refer the "version". Should we change it to "id"? WDYT?

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
imback82: You mean https://github.com/microsoft/hyperspace/blob/40c52431b7220ed892f6781efef578af6b826fa3/src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala#L120-L126
?

For the state, it will be handled `.filter(index => states.isEmpty || states.contains(index.state))`? What's the current behavior (error message), if the index data doesn't exist?

sezruby: No errors, but just apply the invalid index version; which results in empty result.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
sezruby: It's for vacuumed or manually removed directories (by storage services).

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
sezruby: I think version is more general term to represent it, but the code base is `id`.. might be good to be consistent.
But in the plan, we print it as `LogVersion: ` ?

sezruby: It's because I didn't expect any use case for other states; but will fix it.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
sezruby: > For the state, it will be handled .filter(index => states.isEmpty || states.contains(index.state))? What's the current behavior (error message), if the index data doesn't exist?

Yep, but it cannot return any previous version of each index if the state of the latest version is not in `states`.

imback82: good catch. if that's the case, we should definitely fix it.

sezruby: But because of that, we could impl gold standard w/o actual data. 
I created the issue last year #206, but it's a backlog due to low importance.  (as Spark also doesn't check the file existence while reading it, though Spark check that when df creation..)
Maybe it would be good if we could check the data directories. 
@imback82 WDYT?

imback82: Ah OK. I think just checking directories is not sufficient. We probably need to either check all files, or check nothing based on config?

> Yep, but it cannot return any previous version of each index if the state of the latest version is not in `states`.

Sorry, I didn't fully get this part. It would return `Nil` in this case, which is fine?

sezruby: > Sorry, I didn't fully get this part. It would return Nil in this case, which is fine?

if a user cancel the spark job in the middle of optimize (or refresh), the latest state remains "optimizing". 
However, we could use the previous log version if it's eligible.

sezruby: > Ah OK. I think just checking directories is not sufficient. We probably need to either check all files, or check nothing based on config?

But what about this PR? checking all files for every version should burden.

imback82: Checking just data directory doesn't seem any better than not checking at all. I am fine with not checking at all in this PR. And we can do a follow up to introduce a config for "nocheck", "directory_check", and "file_check" in one shot.

I thought this PR can be just exposing versions, or am I missing something?

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -138,6 +138,18 @@ class IndexCollectionManager(
...
 
+  override def getIndex(indexName: String, logVersion: Int): Option[IndexLogEntry] = {
+    withLogManager(indexName) { logManager =>
+      logManager.getLog(logVersion).map(_.asInstanceOf[IndexLogEntry])
+    }
+  }
+
+  override def getAvailableIndexVersions(indexName: String): Seq[Int] = {
+    withLogManager(indexName) { logManager =>
+      logManager.getIndexVersions(Seq(Constants.States.ACTIVE))
```
imback82, line 149: Actually, can we have `def getIndexVersions(states: Seq[String]): Seq[Int]` instead of `getAvailableIndexVersions`? "Available" <=> "ACTIVE" doesn't seem to match nicely. (sorry for the wrong suggestion in the beginning). The caller `closestIndex` can decide which state it wants to retrieve.

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -68,7 +68,9 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
       dfFromSample.write.format("delta").save(dataPath)
 
       val deltaDf = spark.read.format("delta").load(dataPath)
-      hyperspace.createIndex(deltaDf, IndexConfig("deltaIndex", Seq("clicks"), Seq("Query")))
+      withSQLConf(IndexConstants.INDEX_LINEAGE_ENABLED -> "true") {
```
imback82, line 71: why is this required now?

imback82 (commented):

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
sezruby: Since I added the condition above:
https://github.com/microsoft/hyperspace/pull/389/files/b938556191c3dcceab0ec784fa26defc5b02804b#diff-ccc638dd1aab8f054c6d13cdfafeea9c1634646e90dd999b491e4e43a18d9efbR188

It's possible to allow append-only hybrid scan if possible, but the condition in `closestIndex` will become more complicated. So just limit the case for now.

imback82: Hmm, so before this PR, append-only hybrid scan was supported? I am curious which change in this PR triggered this new requirement.

sezruby: Might be supported. The index was not applied because of the conditions in getHybridScanCandidate if the version has deleted files (compared to the given relation) - before this PR. 

I found that I was writing a test for directory check change.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -194,9 +200,13 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
         .get
         .asInstanceOf[IndexLogEntry]
     }
-    // TODO: Currently assume all versions of index data exist.
-    //  Need to check and remove candidate indexes.
-    //  See https://github.com/microsoft/hyperspace/issues/387
+
+    // Get available index log versions.
+    val availableLogVersions =
```
imback82, line 205: `activeLogVersions`? and I think you can remove the comment unless you want to explain "why" you are getting active logs.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -194,9 +200,13 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
         .get
         .asInstanceOf[IndexLogEntry]
     }
-    // TODO: Currently assume all versions of index data exist.
-    //  Need to check and remove candidate indexes.
-    //  See https://github.com/microsoft/hyperspace/issues/387
```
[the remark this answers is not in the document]
sezruby: I think we could close the issue now as index data existence is a different issue.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -184,8 +185,18 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
    * @return IndexLogEntry of the closest version among available index versions.
    */
   override def closestIndex(index: IndexLogEntry): IndexLogEntry = {
+
```
imback82, line 188:
```suggestion
```

imback82 (approved):

imback82 (approved):

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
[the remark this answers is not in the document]
imback82: sgtm

PR #390, opened 2021-03-25
[WIP] PartitionEliminationFilterIndexRule implementation

apoorvedave1:
In this PR we introduce FilterIndexRule for PartitionEliminationNonCoveringIndex. Rule Algorithm:
1. Identify if an index contains index columns which can improve point lookups or range queries on data source
2. Query the index using spark to identify list of data files which could satisfy the query
3. Redirect original query to this subset of data files instead of complete list of files.

 - **Tracking Issue**: If you expect any [subjective discussions](https://google.github.io/eng-practices/review/developer/small-cls.html) around this pull request, please consider opening a tracking issue and link to the PR. Write N/A, if this pull request is self-contained.
 - **Parent Issue**: Link to the issue that captures the overall plan. Write N/A, if this is a stand-alone pull request with a tracking issue OR self-contained pull request.
 - **Dependencies**: Links to issues you depend on for this pull request to work. Write N/A, if no dependencies.
   - Issue 1
   - Issue 2

No

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/PEFilterIndexRule.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/PEFilterIndexRule.scala
@@ -19,6 +19,7 @@ package com.microsoft.hyperspace.index.rules
 import org.apache.hadoop.fs.Path
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.expressions.{And, AttributeReference, BinaryComparison, EqualTo, Expression, UnaryExpression}
 import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan}
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, InMemoryFileIndex, LogicalRelation}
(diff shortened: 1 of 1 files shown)


src/main/scala/com/microsoft/hyperspace/index/rules/PEFilterIndexRule.scala
```diff
@@ -0,0 +1,161 @@
...
+
+    returnVal
+  }
+
+  def getFilteredFiles(filter: Filter, index: IndexLogEntry): Seq[Path] = {
+    val condition = getIndexCompatibleCondition(filter.condition, index)
+
+    val filteredDf =
+      spark.read
+        .parquet(index.content.files.map(_.toString): _*)
```
sezruby, line 100: I thought we build an index of (value => file id list) instead of full scan of covering index.
Could you measure the perf of optimize phase with 1TB TPCH dataset?

src/main/scala/com/microsoft/hyperspace/index/rules/PEFilterIndexRule.scala
```diff
@@ -0,0 +1,161 @@
...
+        .where(condition.sql)
+        .select(IndexConstants.DATA_FILE_NAME_ID)
```
sezruby, line 102:
```suggestion
        .select(IndexConstants.DATA_FILE_NAME_ID)
        .distinct
```

apoorvedave1: thanks

src/main/scala/com/microsoft/hyperspace/index/rules/PEFilterIndexRule.scala
apoorvedave1: this would not work if we want to merge Covering and partition elimination index.

sezruby: We cannot deliver this rule without a proper performance validation as it seems expensive. Could you run with TPCH 1TB 100k chunk dataset and share the result? 

1) w/o FEFilterIndexRule - explain time & query execution time
2) w/ PEFilterindexRule - explain time & query execution time

PR #392, opened 2021-03-27
Correctly pass HadoopConfiguration to Content.fromDirectory

### LINKED ISSUE #391: Unable to change hyperspace system path to a non-default file system
### Describe the issue

Changing hyperspace location to non-default file system using ```"spark.hyperspace.system.path"``` fails for non-default file systems. This is reprodicible, for example, on a databricks cluster where dbfs is the default file system and user tries to set index root to wasb or abfs

### To Reproduce

1. Databricks cluster (dbfs file system)
2. spark.conf.set("spark.hyperspace.system.path", "wasb:// or abfs://"
3. hs.createIndex(..

```

java.io.IOException: No FileSystem for scheme: wasb
	at org.apache.hadoop.fs.FileSystem.getFileSystemClass(FileSystem.java:2660)
	at org.apache.hadoop.fs.FileSystem.createFileSystem(FileSystem.java:2667)
	at org.apache.hadoop.fs.FileSystem.access$200(FileSystem.java:94)
	at org.apache.hadoop.fs.FileSystem$Cache.getInternal(FileSystem.java:2703)
	at org.apache.hadoop.fs.FileSystem$Cache.get(FileSystem.java:2685)
	at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:373)
	at org.apache.hadoop.fs.Path.getFileSystem(Path.java:295)
```

### Expected behavior

no error

### RCA
```
val pathStr = "wasb://" + blob + "@" + storageName + ".blob.core.windows.net/hyperspaceIndexes" 
val path = new Path(pathStr)
path.getFileSystem(new Configuration)              >> FAILS WITH THE ABOVE ERROR
path.getFileSystem(spark.sessionState.newHadoopConf())  >> WORKS FINE (also mentioned in this PR #327 
```
This issue happens in https://github.com/microsoft/hyperspace/blob/master/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala#L197-L199
Need to replace ``` new Configuration``` with ```spark.sessionState.newHadoopConf()``` to make sure spark finds all file systems.

imback82:
Fixes #391.

This PR proposes to fix the issue reported in #391. We missed passing the Hadoop configuration to `Content.fromDirectory` in `CreateActionBase`. This PR removes the default value for the Hadoop configuration in `Content.fromDirectory` since it's error-prone.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -59,6 +59,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       indexConfig: IndexConfig,
       path: Path,
       versionId: Int): IndexLogEntry = {
+    val hadoopConf = spark.sessionState.newHadoopConf()
     val absolutePath = PathUtils.makeAbsolute(path, spark.sessionState.newHadoopConf())
     val numBuckets = numBucketsForIndex(spark)
 
@@ -96,7 +97,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
               IndexLogEntry.schemaString(indexDataFrame.schema),
               numBuckets,
               coveringIndexProperties)),
-          Content.fromDirectory(absolutePath, fileIdTracker),
+          Content.fromDirectory(absolutePath, fileIdTracker, hadoopConf),
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -60,7 +60,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
       path: Path,
       versionId: Int): IndexLogEntry = {
     val hadoopConf = spark.sessionState.newHadoopConf()
-    val absolutePath = PathUtils.makeAbsolute(path, spark.sessionState.newHadoopConf())
+    val absolutePath = PathUtils.makeAbsolute(path, hadoopConf)
     val numBuckets = numBucketsForIndex(spark)
 
     val signatureProvider = LogicalPlanSignatureProvider.create()
--- a/src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
@@ -141,18 +141,16 @@ class OptimizeAction(
     // the list of newly created index files.
     val hadoopConf = spark.sessionState.newHadoopConf()
     val absolutePath = PathUtils.makeAbsolute(indexDataPath.toString, hadoopConf)
-    val newContent =
-      Content.fromDirectory(absolutePath, fileIdTracker, hadoopConfiguration = hadoopConf)
+    val newContent = Content.fromDirectory(absolutePath, fileIdTracker, hadoopConf)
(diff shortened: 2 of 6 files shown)


src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -59,6 +59,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
       indexConfig: IndexConfig,
       path: Path,
       versionId: Int): IndexLogEntry = {
+    val hadoopConf = spark.sessionState.newHadoopConf()
     val absolutePath = PathUtils.makeAbsolute(path, spark.sessionState.newHadoopConf())
```
apoorvedave1, line 63:
```suggestion
    val absolutePath = PathUtils.makeAbsolute(path, hadoopConf)
```

src/main/scala/com/microsoft/hyperspace/actions/OptimizeAction.scala
```diff
@@ -141,18 +141,16 @@ class OptimizeAction(
...
       properties = previousIndexLogEntry.derivedDataset.properties
         .copy(
-          properties =
-              Hyperspace
-                .getContext(spark)
-                .sourceProviderManager
-                .enrichIndexProperties(
-                  previousIndexLogEntry.relations.head,
-                  prevIndexProperties + (IndexConstants.INDEX_LOG_VERSION -> endId.toString))))
+          properties = Hyperspace
```
imback82, line 148: scalafmt changed this block. Since this PR is small enough, I let this change in.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -87,8 +88,8 @@ object Content {
   def fromDirectory(
       path: Path,
       fileIdTracker: FileIdTracker,
+      hadoopConfiguration: Configuration,
       pathFilter: PathFilter = PathUtils.DataPathFilter,
-      hadoopConfiguration: Configuration = new Configuration,
```
imback82, line 91: Let's just remove the default value since this is error prone.

src/test/scala/com/microsoft/hyperspace/index/E2EHyperspaceRulesTest.scala
```diff
@@ -286,10 +286,12 @@ class E2EHyperspaceRulesTest extends QueryTest with HyperspaceSuite {
...
-            spark.sql("CREATE TABLE t1 (c1 string, c3 string) USING PARQUET " +
-              s"LOCATION '$nonPartitionedDataPath'")
-            spark.sql("CREATE TABLE t2 (c3 string, c4 int) USING PARQUET " +
-              s"LOCATION '$nonPartitionedDataPath'")
+            spark.sql(
+              "CREATE TABLE t1 (c1 string, c3 string) USING PARQUET " +
+                s"LOCATION '$nonPartitionedDataPath'")
+            spark.sql(
+              "CREATE TABLE t2 (c3 string, c4 int) USING PARQUET " +
+                s"LOCATION '$nonPartitionedDataPath'")
```
imback82, line 294: Again, scalafmt.

apoorvedave1 (approved):
one minor comment, but otherwise LGTM 👍 thanks @imback82

src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
imback82: thanks!

apoorvedave1 (approved):
LGTM, thanks 👍 @imback82

sezruby:
LGTM Thanks!

imback82:
Thanks @apoorvedave1 / @sezruby for the review!

PR #393, opened 2021-03-28
Introduce "ResolvedColumn" to simplify handling resolved columns 

imback82:
This is a follow up to address comments from #379.

This PR also introduces `ResolvedColumn` to capture resolved column info such as name, and a flag to indicate whether it was resolved as nested column or not, etc. The goal of this class is to put quoting, creating `Column`, normalizing nested fields into one place. The class provides the following APIs:
- `normalizedName`: for a nested column, `__hs_nested` is prefixed to the name.
- `toColumn`: For a nested column, it creates an alias with the normalized name. For top-level column, it creates the column quoting the name.
- `toNormalizedColumn`: Creates a column with normalized column.
- apply() to create `ResolvedColumn` from normalized column name, which is useful when you use the name from the index log entry.


src/main/scala/com/microsoft/hyperspace/index/rules/JoinIndexRule.scala
```diff
@@ -305,8 +305,8 @@ object JoinIndexRule
     val rRequiredIndexedCols = lRMap.values.toSeq
 
     // All required columns resolved with base relation.
-    val lRequiredAllCols = resolve(spark, allRequiredCols(left), leftRelation.plan).get.map(_._1)
-    val rRequiredAllCols = resolve(spark, allRequiredCols(right), rightRelation.plan).get.map(_._1)
+    val lRequiredAllCols = resolve(spark, allRequiredCols(left), lBaseAttrs).get
+    val rRequiredAllCols = resolve(spark, allRequiredCols(right), rBaseAttrs).get
```
imback82, line 309: I am reverting this change. This can be reintroduced when `JoinIndexRule` is updated.

src/test/scala/com/microsoft/hyperspace/index/RefreshIndexNestedTest.scala
```diff
@@ -94,14 +94,14 @@ class RefreshIndexNestedTest extends QueryTest with HyperspaceSuite {
 
           // Validate only index records whose lineage is the deleted file are removed.
           val originalIndexDF = spark.read.parquet(s"$systemPath/${indexConfig.indexName}/" +
-              s"${IndexConstants.INDEX_VERSION_DIRECTORY_PREFIX}=0")
```
imback82, line 97: No change in this file other than running scalafmt.

sezruby (commented):
LGTM thanks!

src/test/scala/com/microsoft/hyperspace/index/CreateIndexNestedTest.scala
```diff
@@ -160,25 +160,23 @@ class CreateIndexNestedTest extends HyperspaceSuite with SQLHelper {
...
+          ((indexConfig2.indexedColumns ++ indexConfig2.includedColumns).map(
+            ResolvedColumn(_, isNested = true).normalizedName) ++
             Seq(IndexConstants.DATA_FILE_NAME_ID) ++ partitionKeys).sorted)
     }
   }
 
   test("Check lineage in index records for non-partitioned data.") {
     withSQLConf(IndexConstants.INDEX_LINEAGE_ENABLED -> "true") {
-      hyperspace.createIndex(nonPartitionedDataDF, indexConfig1)
+      hyperspace.createIndex(nonPartitionedDataDF, indexConfig2)
```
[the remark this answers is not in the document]
imback82: `ResolvedColumn` has an assert that if we are setting `isNested = true`, the column name should contain dots. `indexConfig1` has names both with and without dots, whereas `indexConfig2` only has names with dots. So I changed it to use `indexConfig2` (also same as the test right above - `Check lineage in index records for partitioned data when partition key is not in config.`). Since it's testing if the lineage column is included, it should be OK to change it to `indexConfig2`.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -16,19 +16,73 @@
...
+      if (isNested) {
+        s"$NESTED_FIELD_PREFIX$name"
+      } else {
+        name
+      }
+    }
+
+    // Create a column using the resolved name. Top level column names are quoted, and
+    // nested column names are aliased with normalized names.
+    def toColumn: Column = {
```
andrei-ionescu, line 58: Due to the fact that the method is a publicly visible property I would suggest some ScalaDoc comments/explanations on it - please use `/** */`.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -16,19 +16,73 @@
...
+        // No need to quote the string for "as" even if it contains dots.
+        col(name).as(normalizedName)
+      } else {
+        col(quote(name))
+      }
+    }
+
+    // Create a column using the normalized name. Since the normalized name is already flattened
+    // with "dots", it is quoted.
+    def toNormalizedColumn: Column = col(quote(normalizedName))
```
andrei-ionescu, line 69: Ditto - ScalaDoc

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -16,19 +16,73 @@
...
+
+    private def quote(name: String) = s"`$name`"
```
andrei-ionescu, line 71: Can there be a case where a nested field access path contains quotes and become twice quoted? 

For example, given the following dataset schema:
```
root
 |-- nest: struct (nullable = true)
 |    |-- n.dot.nested: struct (nullable = true)
 |    |    |-- field: string (nullable = true)
```
In Spark it's correct to access `field` with this construct: ``nest.`n.dot.nested`.field``. 

Can we get into the case of quoting the ``nest.`n.dot.nested`.field`` ending up with `` `nest.`n.dot.nested`.field` ``? Should we check if field does already contain backticks?

Could you add some unit tests for this use case?

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -84,13 +138,13 @@ object ResolverUtils {
...
-   * @return Optional sequence of tuples of resolved name string and nested state boolean
-   *         if all required strings are resolved. Else, None.
+   * @return Optional sequence of ResolvedColumn objects if all required strings are resolved.
+   *         Else, None.
    */
   def resolve(
       spark: SparkSession,
       requiredStrings: Seq[String],
-      plan: LogicalPlan): Option[Seq[(String, Boolean)]] = {
+      plan: LogicalPlan): Option[Seq[ResolvedColumn]] = {
```
andrei-ionescu, line 147: 👍

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
```diff
@@ -16,19 +16,73 @@
...
+    // with "dots", it is quoted.
+    def toNormalizedColumn: Column = col(quote(normalizedName))
+
+    private def quote(name: String) = s"`$name`"
+  }
+
+  private[hyperspace] object ResolvedColumn {
+    private val NESTED_FIELD_PREFIX = "__hs_nested."
+
+    def apply(normalizedColumnName: String): ResolvedColumn = {
```
andrei-ionescu, line 77: Ditto - Scaladoc

src/test/scala/com/microsoft/hyperspace/index/CreateIndexNestedTest.scala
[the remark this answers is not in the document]
andrei-ionescu: Shouldn't we test if the functionality work correct with both types of column names in one of the `indexedColumns` or `includedColumns`? The `indexConfig1` does contain `Seq("Date", "nested.leaf.cnt")` in the included columns property while `indexConfig2` doesn't have this mixed column types. We should have such tests that verifies that this mixed cased does work.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
andrei-ionescu, line 77: Can we add some UTs to verify the `ResolvedColumn.apply` functionality? There are some conditions in `apply` method that should be covered by tests.

andrei-ionescu (commented):
Nice addition with this `ResolvedColumn` and thanks for taking time adding it. 

I reviewed the code changes and I want to rise the following comments/concerns:
1. The quoting (backticking) mechanism in some use cases - see my example on the `quote` method in `ResolvedColumn` class.
2. Not having UT with mixed column types - see my comment on the change from `indexConfig1` to `indexConfig2`.
3. Missing UTs for `ResolvedColumn.apply` method.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
imback82: As the following comment says, the `name` passed to the `ResolvedColumn` should be the result of `resolve`: "Outside unit tests, this object should not be created directly, but via the `resolve` function". So, it shouldn't receive the example you posted. But, we could add more asserts here though.

src/main/scala/com/microsoft/hyperspace/util/ResolverUtils.scala
imback82: `assert(col2 === ResolvedColumn(col2.normalizedName))`?

src/test/scala/com/microsoft/hyperspace/index/CreateIndexNestedTest.scala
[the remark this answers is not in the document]
imback82: `Index creation with nested indexed and included columns.` should already handle `indexConfig1`? (I can update it make the coverage better)

sezruby:
@imback82 @andrei-ionescu Can we merge this change and then address the unresolved comment w/ #380? so that @andrei-ionescu can update the PRs based on `ResolvedColumn`.

imback82:
Sorry for the delay. I will address pending comments and merge early tomorrow.

imback82 (commented):
Thanks @sezruby / @andrei-ionescu for the review, merging to master!

PR #396, opened 2021-03-29
Update scalastyle to allow 2021 in the copyright notice

### LINKED ISSUE #382: Scala style does NOT allow 2021 Copyright
We are now in 2021. Scala style check (`sbt clean compile`) should allow this:

```
Copyright (2021) The Hyperspace Project Authors.
...
```

Instead it fails with: 

```
[error] /Users/aionescu/github/hyperspace/src/main/scala/com/microsoft/hyperspace/index/rules/PlanUtils.scala: Header does not match expected text
```

clee704:

 - **Tracking Issue**: #382 
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

Fixes #382

According to my research, it is best to include the creation year of the file in the copyright notice. This change allows writing 2021 in the header.

No

Locally checked with scalastyle

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -92,7 +92,7 @@ This file is divided into 3 sections:
  */
 
 \E)?\Q/*
- * Copyright (\E\d{4}(?:-\d{4})?\Q) The Hyperspace Project Authors.
+ * Copyright (\E(?:2020|2021)\Q) The Hyperspace Project Authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -92,7 +92,7 @@ This file is divided into 3 sections:
  */
 
 \E)?\Q/*
- * Copyright (2020) The Hyperspace Project Authors.
+ * Copyright (\E\d{4}(?:-\d{4})?\Q) The Hyperspace Project Authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


andrei-ionescu (dismissed):

scalastyle-config.xml
```diff
@@ -92,7 +92,7 @@ This file is divided into 3 sections:
  */
 
 \E)?\Q/*
- * Copyright (2020) The Hyperspace Project Authors.
+ * Copyright (\E\d{4}(?:-\d{4})?\Q) The Hyperspace Project Authors.
```
imback82, line 95: This will not catch invalid years?

clee704: I think the purpose of having a code linter is mainly to catch human errors, and having a strict validation here seems like overkill. But if you want we can make it something like 20\d{2} to accept any 20xx year (although we can't still prevent invalid years like 2099 or 2099-2010), or make it accept only years known to valid for now (2020|2021|2020-2021) and change the format every year.

imback82: I definitely prefer catching this kind of error via scalastyle (why waste time in PR?). We can just do `(2020|2021)` and I don't think it's an overkill to update this every year.

clee704: After some research it seems sufficient to use something like (2020|2021|...|<current year>). When you create a new file, you just write the year at the moment. No need to update old files.

clee704: By the way, if you really want to be strict, then we should prevent people using past year when creating new files and updating the year of old files - currently we can't catch those errors

andrei-ionescu: I'm also ok with adding new years as we get into a new year. Also, I would not want to have ranges (ie: `2019-2021`). The year of first publishing should be enough.

sezruby:
LGTM Thanks!

andrei-ionescu (approved):
LGTM, Thanks!

imback82 (approved):
LGTM, thanks @clee704!

PR #398, opened 2021-03-30
Add a state barrier for getLatestStableLog

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

In getLatestStableLog, stop searching the previous logs if there's creating or vacuuming state.
Since logs before creating or vacuuming state are irrelevant to the current index and invalid.

Yes

Unit test


src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -99,6 +99,10 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration =
           val entry = getLog(id)
           if (entry.isDefined && Constants.STABLE_STATES.contains(entry.get.state)) {
             return entry
+          } else if (entry.isDefined && (entry.get.state.equals(Constants.States.CREATING) ||
+                     entry.get.state.equals(Constants.States.VACUUMING))) {
+            // Not consider unrelated logs before creating or vacuuming state.
+            return None
           }
```
imback82, line 106: Can the following be simplified now?

https://github.com/microsoft/hyperspace/blob/c900cf56997c42873358e1260050a322f8fe5a0c/src/main/scala/com/microsoft/hyperspace/actions/CancelAction.scala#L44-L53

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -99,6 +99,10 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration =
           val entry = getLog(id)
           if (entry.isDefined && Constants.STABLE_STATES.contains(entry.get.state)) {
             return entry
+          } else if (entry.isDefined && (entry.get.state.equals(Constants.States.CREATING) ||
```
imback82, line 102: move the logic under `if (entry.isDefined)`?

imback82 (commented):
So, this seems to be affecting you are running `getLatestStableLog` in the middle of create/vacuum (whether they are running/failed), right?

I need @apoorvedave1's review once he's back.

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -99,6 +99,10 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration =
...
+                     entry.get.state.equals(Constants.States.VACUUMING))) {
+            // Not consider unrelated logs before creating or vacuuming state.
```
apoorvedave1, line 104: "Do not consider..."

apoorvedave1:
minor comments otherwise LGTM 👍 thanks @sezruby

src/test/scala/com/microsoft/hyperspace/index/IndexLogManagerImplTest.scala
```diff
@@ -166,6 +166,27 @@ class IndexLogManagerImplTest
...
+    val fs = path.getFileSystem(new Configuration)
+
+    FileUtils.createFile(
+      fs,
+      new Path(path, s"$HYPERSPACE_LOG/8"),
+      JsonUtils.toJson(getEntry("ACTIVE")))
+    FileUtils.createFile(
+      fs,
+      new Path(path, s"$HYPERSPACE_LOG/10"),
+      JsonUtils.toJson(getEntry("VACUUMING")))
```
imback82, line 180: Don't you need another assert here so that we check both `VACUUMING` and `CREATING`?

imback82 (commented):

imback82 (approved):

Issue #400, opened 2021-03-30
Content.rec(..) in LogEntry is recursive function, but does not use tail rec optimisation


dmytroDragan:
### Content.rec(..) in LogEntry is recursive function, but does not use tail rec optimisation

```
private def rec[T](
                      prefixPath: Path,
                      directory: Directory,
                      func: (FileInfo, Path) => T): Seq[T] = {
    val files = directory.files.map(f => func(f, prefixPath))
    files ++ directory.subDirs.flatMap { dir =>
      rec(new Path(prefixPath, dir.name), dir, func)
    }
  }
```

It is a potential issue which in some day could give StackOverflowError.
I could quickly fix it.

PR #402, opened 2021-03-30
Refactor Content.rec(..) function to support tail-optimisation

dmytroDragan:

 - **Tracking Issue**: #400
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

Refactor Content.rec(..) function to support tail-optimisation

No

Existing unit tests should be enough, because it is just refactoring without changing functionality.


andrei-ionescu (commented):
Thanks for the PR. A couple of things to address:

1. Please use `scalafmt` for formatting - see https://github.com/microsoft/hyperspace/blob/master/docs/coding-guidelines/scala-coding-style.md.
2. Please add some unit tests on the `rec` method.

dmytroDragan:
Hi @andrei-ionescu ,
1. Sure (might make sense to add also scalafmt version for alignment)
2.  There are already tests for Content class like "Content.files api lists all files from Content object" which rely on it, but I could move rec function to Content object and add more.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -111,6 +102,43 @@ object Content {
...
+              directory: Directory,
+              func: (FileInfo, Path) => T): Seq[T] = {
+    @tailrec
+    def recAcc[A](
+                   dirMap: List[(Path, Seq[Directory])],
+                   func: (FileInfo, Path) => A,
+                   acc: Seq[A] = Seq.empty): Seq[A] = {
+      dirMap match {
+        case Nil => acc
+        case (curPrefixPath, curDirs) :: xs =>
```
sezruby, line 125: Could you explain why it's `xs`?

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -240,6 +240,47 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
...
+
+    val expected =
+      Seq("file:/f0", "file:/a/f1", "file:/a/f2", "file:/a/b/f3", "file:/a/b/f4")
+        .map(new Path(_)).toSet
+
+    val actual = res.toSet
+    assert(actual.equals(expected))
+  }
+
+  test("Content.recFilesApply return empty list for directories without files ") {
```
sezruby, line 272:
```suggestion
  test("Content.recFilesApply returns empty list for directories without files.") {
```

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -240,6 +240,47 @@ class IndexLogEntryTest extends SparkFunSuite with SQLHelper with BeforeAndAfter
     assert(actual.sourceFilesSizeInBytes == 200L)
   }
 
+  test("Content.recFilesApply apply function to all files ") {
```
sezruby, line 243: ditto

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
dmytroDragan: I have started with common unapplied list ```x::xs``` and simplified head later)
xs is plural for x.

I will rename it `otherDirs`  for readability.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -111,6 +102,43 @@ object Content {
...
+   * @tparam T
+   * @return
+   */
+  def recFilesApply[T](
+              prefixPath: Path,
+              directory: Directory,
+              func: (FileInfo, Path) => T): Seq[T] = {
+    @tailrec
+    def recAcc[A](
+                   dirMap: List[(Path, Seq[Directory])],
```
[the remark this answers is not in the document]
dmytroDragan: You right, it works strangely: some scalafmt versions are ignored by idea (1.5, 2.0.0) , while others are to aggressive - look on latest commit with 2.7.5 scalafmt.
Am I missing anything?

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
dmytroDragan: fixed

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
dmytroDragan: fixed

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -111,6 +102,43 @@ object Content {
...
     }
   }
+
+  /**
+   * Apply `func` to each file in directory recursively.
+   * @param prefixPath Root prefix
+   * @param directory Root directory
+   * @param func function which would apply to current prefix and file
+   * @tparam T
+   * @return
```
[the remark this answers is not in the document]
dmytroDragan: done

sezruby:
@dmytroDragan Sorry but, could you revert unnecessary changes from scalafmt to reduce the diff? It can be handled with a different PR (would be appreciated if you could push a new PR!), so that we could review & merge only required change for tailrec.

dmytroDragan:
Hi @sezruby! 
I can revert changes from scalafmt, but it makes a usage of scalafmt useless in first place.

I`m using 2.7.5 (latest) scalamft with dev conf:
```
# The following configs are taken from https://github.com/apache/spark/blob/master/dev/.scalafmt.conf
align = none
align.openParenDefnSite = false
align.openParenCallSite = false
align.tokens = []
optIn = {
  configStyleArguments = false
}
danglingParentheses = false
docstrings = JavaDoc
maxColumn = 98

# The following are specific to Hyperspace.
importSelectors = singleLine
```
If anyone gets different result for this files (like old once), please reach out to me and we could try to figure out.

If no one has time for it... well, I will revert unnecessary parts right now just to not block this PR.

andrei-ionescu:
@dmytroDragan, Here is my `scalafmt` setup in IntelliJ Idea (I have `1.5.1` version):

hyperspace_scalafmt

I usually do `Command+Shift+Alt+L` to reformat the file and use the following options in the dialog box:

hyperspace_reformat

The formatting is correct, the only annoying thing is that it order the functions by name too. I will be great if you could double check the diffs before pushing the changes.

dmytroDragan:
Thank you @andrei-ionescu. I have rechecked my configuration and applied result formatting.
I have created PR with just scalafmt formated code for this 2 files, so reviewing it could be easier:
https://github.com/microsoft/hyperspace/pull/412

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -45,31 +46,20 @@ case class Content(root: Directory, fingerprint: NoOpFingerprint = NoOpFingerpri
...
-    files ++ directory.subDirs.flatMap { dir =>
-      rec(new Path(prefixPath, dir.name), dir, func)
-    }
+    recFilesApply(new Path(root.name), root, createFileInfoWithPrefix).toSet
   }
 }
 
 object Content {
 
+  val createFileInfoWithPrefix: (FileInfo, Path) => FileInfo = (f, prefix) =>
```
sezruby, line 60: just impl. in `fileInfos` - there's no other use case.

dmytroDragan: Correct. I`m just not keen of formatted:
```
      (
          f,
          prefix) =>
        FileInfo(new Path(prefix, f.name).toString, f.size, f.modifiedTime, f.id)).toSet
```
see: https://github.com/microsoft/hyperspace/pull/412/files#diff-5c50f07a41423f4cb5bfe60ac7c30b9c534c4bfc71eeb03d43040047a3994e83

Moving it to separate function makes it more readable.

dmytroDragan:
Th problem was with 1.5.1 vs 2.7.5 scalafmt.
Fixed version is done as new PR:
https://github.com/microsoft/hyperspace/pull/417

PR #403, opened 2021-04-01
HyperspaceOneRule prototype

sezruby:

 - **Tracking Issue**: 
 - **Parent Issue**: 
 - **Dependencies**: 
 
TBD

HyperspaceOneRule prototype for quick design review
Will be delivered by other PRs.

Yes query plan may change

Perf: confirmed there's no noticeable regression in TPCH bench - 1TB 100k chunk dataset

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
+++ b/src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
@@ -40,7 +40,9 @@ class Hyperspace(spark: SparkSession) {
    * @param indexConfig the configuration of index to be created.
    */
   def createIndex(df: DataFrame, indexConfig: IndexConfig): Unit = {
-    indexManager.create(df, indexConfig)
+    withHyperspaceDisabled(spark) {
+      indexManager.create(df, indexConfig)
+    }
   }
 
   /**
@@ -49,7 +51,9 @@ class Hyperspace(spark: SparkSession) {
    * @param indexName the name of index to delete.
    */
   def deleteIndex(indexName: String): Unit = {
-    indexManager.delete(indexName)
+    withHyperspaceDisabled(spark) {
+      indexManager.delete(indexName)
+    }
   }
 
   /**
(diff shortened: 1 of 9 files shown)


sezruby:
@imback82 This is a working version of prototype for both Filter & Join. If you don't have any significant concern on this pr, I'll create small PRs based on this PR. Thanks!

src/main/scala/com/microsoft/hyperspace/Hyperspace.scala
```diff
@@ -40,7 +40,9 @@ class Hyperspace(spark: SparkSession) {
    * @param indexConfig the configuration of index to be created.
    */
   def createIndex(df: DataFrame, indexConfig: IndexConfig): Unit = {
-    indexManager.create(df, indexConfig)
+    withHyperspaceDisabled(spark) {
```
sezruby, line 43: NOTE: enabled hyperspace during the index creation can be a problem because of the following code in HyperspaceOneRule
```scala
 val indexManager = Hyperspace
      .getContext(spark)
      .indexCollectionManager
```

andrei-ionescu: Sorry, but I still don't understand. 

1. Shouldn't `indexManager.getIndexes` return the same answer in all cases?
2. How does the index creation work if the context is not set through `getContext`? Where will that new index be created, in which location?

sezruby: Ok.. I've analyzed the exact cause. `getIndexes` in HyperspaceOneRule causes caching indexes while createIndex.
So cached `indexes` doesn't have the newly created one. 
```
override def create(df: DataFrame, indexConfig: IndexConfig): Unit = {
    clearCache()
    super.create(df, indexConfig)
    clearCache() // this can be another solution
  }
```

But I think it's also good to disable Hyperspace during creating or refreshing indexes, to avoid any unintended plan change.

sezruby:
This is merged to master via other PRs. Please refer #405

PR #404, opened 2021-04-01
Refactor hashing and fingerprint functions

### LINKED ISSUE #271: Revisit FileBasedSignatureProvider for a possible performance optimization
could you also hide this feature behind a flag which defaults to false? 
My concern is I think this would be unnecessary calculation for regular scenarios. In any case, if signature doesn't match, RuleUtils will make sure that if index can be used, it will be used right? I am just wondering why bother O(NlogN) for sorting always for a rare scenario which will still work without this change(assuming hybrid scan is enabled eventually by default)?

Alternately if we still want to fix this issue, could you avoid sorting? Instead you could convert fileInfos to set and then create fingerprint. Set will ensure the order of iteration is unique for a unique collection of elements. That way we can still achieve O(N).

cc @imback82

_Originally posted by @apoorvedave1 in https://github.com/microsoft/hyperspace/issues/268#issuecomment-732451580_

clee704:
Fixes #271.

With newly added classes for fingerprinting, now it is easy to make a
fingerprint from various types of data.

FingerprintBuilder can be used to build a fingerprint from data.
FileBasedRelation.signature now takes it as an argument, so the
implementations don't have to do the hashing themselves. They can just
use the provided builder.

For unorderd combining, one can use bitwise XOR to combine multiple
fingerprints. This way, the order becomes irrelavant.

 - **Tracking Issue**: #271 
 - **Parent Issue**: N/A
 - **Dependencies**: N/A


src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedRelation.scala
```diff
@@ -42,13 +42,12 @@ class DefaultFileBasedRelation(spark: SparkSession, override val plan: LogicalRe
...
    * Computes the signature of the current relation.
    */
-  override def signature: String = plan.relation match {
+  override def signature(fb: FingerprintBuilder): Option[Fingerprint] = plan.relation match {
     case HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _) =>
-      val result = filesFromIndex(location).sortBy(_.getPath.toString).foldLeft("") {
-        (acc: String, f: FileStatus) =>
-          HashingUtils.md5Hex(acc + fingerprint(f))
-      }
-      result
```
clee704, line 51: This code is the reason why we should change the return type to Option. Previously it can return an empty string, which is interpreted by the calling side as "the signature is/cannot be not computed". Because it is hard to know just looking at the method signature, it is better to make it explicit here by returning None in such cases.

src/main/scala/com/microsoft/hyperspace/index/IndexSignatureProvider.scala
```diff
@@ -30,9 +30,10 @@ import com.microsoft.hyperspace.util.HashingUtils
  * If the plan does not comply with [[FileBasedSignatureProvider]] or [[PlanSignatureProvider]]
  * requirements for signature computation, then no signature will be provided for the plan.
  */
-class IndexSignatureProvider extends LogicalPlanSignatureProvider {
-  private val fileBasedSignatureProvider = new FileBasedSignatureProvider
-  private val planSignatureProvider = new PlanSignatureProvider
+class IndexSignatureProvider(fbf: FingerprintBuilderFactory)
```
andrei-ionescu, line 33: Please update the scaladoc. It should contain the `@param fbf` explaining its role.

src/main/scala/com/microsoft/hyperspace/index/IndexSignatureProvider.scala
```diff
@@ -41,10 +42,10 @@ class IndexSignatureProvider extends LogicalPlanSignatureProvider {
    * @return signature, if both [[FileBasedSignatureProvider]] and [[PlanSignatureProvider]]
    *         can generate signature for the logical plan; Otherwise None.
    */
-  def signature(logicalPlan: LogicalPlan): Option[String] = {
+  def signature(logicalPlan: LogicalPlan): Option[Fingerprint] = {
```
andrei-ionescu, line 45: Ditto - Update the scaladoc.

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
```diff
@@ -18,27 +18,24 @@ package com.microsoft.hyperspace.index
...
-import com.microsoft.hyperspace.util.HashingUtils
+import com.microsoft.hyperspace.util.fingerprint.{Fingerprint, FingerprintBuilder, FingerprintBuilderFactory}
 
 /**
  * [[PlanSignatureProvider]] provides signature for a logical plan based on
  * the type of operators in it.
  * A plan needs to have at least one operator so its signature can be generated.
  */
-class PlanSignatureProvider extends LogicalPlanSignatureProvider {
+class PlanSignatureProvider(fbf: FingerprintBuilderFactory) extends LogicalPlanSignatureProvider {
```
andrei-ionescu, line 28: Ditto - update scaladoc. Explain the new `fbf` parameter.

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
```diff
@@ -18,27 +18,24 @@ package com.microsoft.hyperspace.index
...
    * @return signature if there is at least one operator in the plan; Otherwise None.
    */
-  def signature(logicalPlan: LogicalPlan): Option[String] = {
-    var signature = ""
-    logicalPlan.foreachUp(p => signature = HashingUtils.md5Hex(signature + p.nodeName))
-    signature match {
-      case "" => None
-      case _ => Some(signature)
-    }
+  def signature(logicalPlan: LogicalPlan): Option[Fingerprint] = {
```
andrei-ionescu, line 36: Ditto - update scaladoc for the return type.

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
```diff
@@ -18,27 +18,24 @@ package com.microsoft.hyperspace.index
...
+    val fb: FingerprintBuilder = fbf.create
+    logicalPlan.foreachUp(node => fb.add(node.nodeName))
+    Some(fb.build())
```
andrei-ionescu, line 39: Something looks wrong here. The method seems to return always `Some`. If so then let's remove the `Option` and return the `Fingerprint`.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedRelation.scala
```diff
@@ -42,13 +42,12 @@ class DefaultFileBasedRelation(spark: SparkSession, override val plan: LogicalRe
   /**
    * Computes the signature of the current relation.
    */
-  override def signature: String = plan.relation match {
+  override def signature(fb: FingerprintBuilder): Option[Fingerprint] = plan.relation match {
```
andrei-ionescu, line 45: Ditto - update scaladoc explaining the new `fb` parameter.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -36,9 +37,9 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
   /**
    * Computes the signature of the current relation.
    */
-  override def signature: String = plan.relation match {
+  override def signature(fb: FingerprintBuilder): Option[Fingerprint] = plan.relation match {
     case HadoopFsRelation(location: TahoeLogFileIndex, _, _, _, _, _) =>
-      location.tableVersion + location.path.toString
+      Some(fb.add(location.tableVersion).add(location.path.toString).build())
```
andrei-ionescu, line 42: If it's always retuning only `Some` let's remove it and return `Fingerprint` directly.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
```diff
@@ -36,9 +37,9 @@ class DeltaLakeRelation(spark: SparkSession, override val plan: LogicalRelation)
   /**
    * Computes the signature of the current relation.
    */
-  override def signature: String = plan.relation match {
+  override def signature(fb: FingerprintBuilder): Option[Fingerprint] = plan.relation match {
```
andrei-ionescu, line 40: Ditto - please update scaladoc.

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -47,11 +46,11 @@ class IcebergRelation(spark: SparkSession, override val plan: DataSourceV2Relati
   /**
    * Computes the signature of the current relation.
    */
-  override def signature: String = plan.source match {
+  override def signature(fb: FingerprintBuilder): Option[Fingerprint] = plan.source match {
```
andrei-ionescu, line 49: Ditto.

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
```diff
@@ -47,11 +46,11 @@ class IcebergRelation(spark: SparkSession, override val plan: DataSourceV2Relati
...
     case _: IcebergSource =>
       val table = loadIcebergTable
       val snapshotId = plan.options.getOrElse("snapshot-id", table.currentSnapshot().snapshotId())
-      snapshotId + table.location()
+      Some(fb.add(snapshotId.toString).add(table.location).build())
```
andrei-ionescu, line 53: Ditto - if only `Some` is returned then let's drop it and return the `Fingerprint`.

src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
```diff
@@ -57,8 +58,11 @@ trait FileBasedRelation extends SourceRelation {
    *
    * This API is used when the signature of source needs to be computed, e.g., creating an index,
    * computing query plan's signature, etc.
+   *
+   * If it is not possible to compute the signature (e.g. there are no files left),
+   * the implementation might return None.
    */
-  def signature: String
+  def signature(fb: FingerprintBuilder): Option[Fingerprint]
```
andrei-ionescu, line 65: Please add the new parameter in the scaladoc with `@param`.

andrei-ionescu (commented):
1. Please update the documentation on methods where new parameters were added.
2. In the cases where only `Some` is returned let's drop it.

src/main/scala/com/microsoft/hyperspace/index/IndexSignatureProvider.scala
clee704: I think it doesn't need an update, since only the type is changed and the previous description can still apply.

src/main/scala/com/microsoft/hyperspace/index/IndexSignatureProvider.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
clee704: Ditto - the change from String to Fingerprint doesn't invalid the previous description.

src/main/scala/com/microsoft/hyperspace/index/PlanSignatureProvider.scala
clee704: The interface dictates that the implementation may return None, and it is indeed the case for FileBasedSignatureProvider. It's just that this implementation doesn't return None.

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedRelation.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
clee704: Since it is an overriding method, you can't change the return type.

src/main/scala/com/microsoft/hyperspace/index/sources/delta/DeltaLakeRelation.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/sources/iceberg/IcebergRelation.scala
clee704: Ditto

src/main/scala/com/microsoft/hyperspace/index/sources/interfaces.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/util/fingerprint/FingerprintBuilderFactory.scala
```diff
@@ -0,0 +1,49 @@
...
+class MD5FingerprintBuilderFactory extends FingerprintBuilderFactory {
+  override def create: FingerprintBuilder = {
+    new FingerprintBuilder(DigestUtils.getMd5Digest)
+  }
+}
+
+/**
+ * SHA1-based fingerprint builder. The fingerprint has 20 bytes (160 bits).
+ */
+class SHA1FingerprintBuilderFactory extends FingerprintBuilderFactory {
```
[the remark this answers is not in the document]
clee704: I left it just in case we need it. Since this is a breaking change anyway, I might just change the hash function from MD5 to SHA1.

sezruby (commented):
Could you resolve the conflict & share some simple benchmark result (createIndex / apply index with >100k source files?) Thanks!

andrei-ionescu:
@clee704, @sezruby

If this is a **breaking change** that will make the already created indexes useless, I would like to throw in some common sense questions that should be addressed in a way or another:

1. Any migration tool to migrate the existing indexes to the new fingerprinting method?
2. If the answer is NO to the question above, is there a plan to have such a tool?
3. If the answer is NO do we at least know how many are using OSS Hyperspace right now and let them know before about this fact?
4. In the case of a user trying to use an old index with this new version how would he know that the problem is this breaking change? Is there an exception thrown or a message or something?
5. Is there a process defined somewhere for cases when breaking changes are introduced?
6. How will this breaking change be documented for the users?

I know these are a lot of questions and as good engineers we need to have a bit of "client-centric" mind to create a useful and easy-to-use product.

@rapoth, @apoorvedave1, @pirz, @imback82 WDYT?

andrei-ionescu (dismissed):
LGTM, one note though: it seems that there are a couple of places where `signature` method returns `Some` always. I know that the `signature` interface defines the return as `Option` but maybe is a good time to think about it. Maybe that `Option` can be removed from the interface too.

rapoth:
> @clee704, @sezruby
> 
> If this is a **breaking change** that will make the already created indexes useless, I would like to throw in some common sense questions that should be addressed in a way or another:
> 
> 1. Any migration tool to migrate the existing indexes to the new fingerprinting method?
> 2. If the answer is NO to the question above, is there a plan to have such a tool?
> 3. If the answer is NO do we at least know how many are using OSS Hyperspace right now and let them know before about this fact?
> 4. In the case of a user trying to use an old index with this new version how would he know that the problem is this breaking change? Is there an exception thrown or a message or something?
> 5. Is there a process defined somewhere for cases when breaking changes are introduced?
> 6. How will this breaking change be documented for the users?
> 
> I know these are a lot of questions and as good engineers we need to have a bit of "client-centric" mind to create a useful and easy-to-use product.
> 
> @rapoth, @apoorvedave1, @pirz, @imback82 WDYT?

Thanks @andrei-ionescu for rasing these questions! Here's my take on this:

> 1. Any migration tool to migrate the existing indexes to the new fingerprinting method?
> 2. If the answer is NO to the question above, is there a plan to have such a tool?

My thinking is that we should not block on a migration utility for this upcoming release. However, I requested @clee704 to write-up a design proposal for how such a migration utility would be designed along with all the dimensions we would have to cover. @clee704 should be able to get to that soon.  Once it passes vote, it will be open for one of the folks to take up in the coming weeks. 

That being said, if you are affected by this breaking change or one of your users/customers is, please let us know. Currently, I have already provided a heads-up to the customers I know who are using Hyperspace in production and they were fine with it.

> 4. In the case of a user trying to use an old index with this new version how would he know that the problem is this breaking change? Is there an exception thrown or a message or something?

@clee704 @sezruby Could we verify what happens and add the appropriate error message + steps they would have to take to overcome the problem?

> 5. Is there a process defined somewhere for cases when breaking changes are introduced?

No but now is probably a good time to discuss and document this. In general, as long as we do not announce a v1.0, we do not guarantee backward/forward compatibility and do not recommend the code to be used in production. This may not have anything to do with the quality of the code but rather stability (as customers provide us with more and more feature requests, we are trying to establish a good foundation so we can avoid major breaking changes in the future). 

As we get closer to v1.0, we will follow semantic versioning guidelines (i.e., everything within a major version will be forward/backward compatible but across major versions, we may break things but we will provide a migration utility by then).

> 6. How will this breaking change be documented for the users?

It will be documented in the release notes of this specific version and the frequently-asked-questions section. Suggestions are welcome!

andrei-ionescu:
@rapoth 

Thanks for putting up your thoughts.

> My thinking is that we should not block on a migration utility for this upcoming release. 

I agree that we should not block this PR by not having the migration tool - this is acceptable now as Hyperspace is at the beginning. It would be great though to have at least a ticket/issue/proposal/feature to describe the gap and put together a plan to fill the gap. I’m glad we are on the same page.

> if you are affected by this breaking change or one of your users/customers is, please let us know

Currently no major impact. I do have a few indexes created on big datasets 15TB and more or which I will be forced to recreate when you upgrading.

rapoth:
> > My thinking is that we should not block on a migration utility for this upcoming release.
> 
> I agree that we should not block this PR by not having the migration tool - this is acceptable now as Hyperspace is at the beginning. It would be great though to have at least a ticket/issue/proposal/feature to describe the gap and put together a plan to fill the gap. I’m glad we are on the same page.

Thanks @andrei-ionescu! I've requested @clee704 to do this. Once he wraps up his current work, he will write up a proposal . Here's the feature request I just opened #420 - please feel free to add anything else I may have missed directly in that issue.

clee704:
If this signature change is the only breaking change of v0.5, then it should be easy to update the signatures of indices from old systems as we have a version string in the serialized index log entries. This will be addressed as part of #420. Perhaps a simple migration utility to address this could be part of v0.5.

clee704:
Spark was too slow to create such many files, so I did measurement for the `signature` method of `DefaultFileBasedRelation` with fake files.

Below are the elapsed times of the `signature` method with varying number of files (averages of 10 repetitions).

| # Files | Before | After | Speedup |
| -: | -: | -: | -: |
| 10,000 | 65.7 ms | 32.4 ms | 51% |
| 100,000  | 581 ms | 249 ms | 57% |
| 1,000,000 | 7.59 s | 2.32 s | 69% |

The results can vary depending on many other factors such as file path lengths and the format of file names. In this measurement I assumed a single root directory whose path length is 64 and random file names of length 67 (which is the default length of parquet files in Spark).

Here's the source for the test:

```scala
import scala.util.Random

import org.apache.hadoop.fs.{FileStatus, Path}
import org.apache.log4j.{Level, Logger}
import org.apache.spark.sql.SparkSession
import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation, PartitioningAwareFileIndex}
import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
import org.apache.spark.sql.types.StructType

import com.microsoft.hyperspace.index.sources.default.DefaultFileBasedRelation
import com.microsoft.hyperspace.util.fingerprint.MD5FingerprintBuilderFactory

class PerformanceTest extends SparkFunSuite {
  test("DefaultFileBasedRelation.signature performance") {
    for (numFiles <- Seq(10000, 100000, 1000000)) {
      val repetition = 10
      val measurements = (1 to repetition).map(_ => runTest(numFiles))
      println(s"numFiles = $numFiles")
      println(s"max = ${measurements.max / 1000000.0} ms")
      println(s"min = ${measurements.min / 1000000.0} ms")
      println(s"avg = ${measurements.sum / measurements.length / 1000000.0} ms")
    }
  }

  def runTest(numFiles: Int): Long = {
    val fileBasedRel = getDefaultFileBasedRelation(numFiles)
    val start = System.nanoTime()
    val sig = fileBasedRel.signature(fb)
    // val sig = fileBasedRel.signature
    val end = System.nanoTime()
    assert(sig.isDefined)
    // assert(sig.nonEmpty)
    end - start
  }

  def getDefaultFileBasedRelation(numFiles: Int): DefaultFileBasedRelation = {
    val files = (1 to numFiles).map(_ => getRandomFileStatus())
    val location = FakeFileIndex(files)(spark)
    val parquet = new ParquetFileFormat()
    val st = new StructType()
    val rel = HadoopFsRelation(location, st, st, None, parquet, Map())(spark)
    val logicalRel = LogicalRelation(rel)
    new DefaultFileBasedRelation(spark, logicalRel)
  }

  def getRandomFileStatus(): FileStatus = {
    val length = rand.nextLong()
    val modificationTime = rand.nextLong()
    val name = getRandomAlphanumericString(67)
    val path = new Path(s"/$dir", name)
    new FileStatus(length, false, 3, 128 * 1024 * 1024, modificationTime, path)
  }

  def getRandomAlphanumericString(length: Int): String = {
    val chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
    val s = new Array[Char](length)
    for (i <- 0 until length) s(i) = chars(rand.nextInt(chars.length))
    new String(s)
  }

  Logger.getLogger("org").setLevel(Level.OFF)
  lazy val spark = SparkSession.builder().master("local").getOrCreate()
  lazy val rand = new Random(42)
  lazy val dir = getRandomAlphanumericString(64)
  lazy val fb = new MD5FingerprintBuilderFactory().create

  case class FakeFileIndex(files: Seq[FileStatus])(sparkSession: SparkSession)
    extends PartitioningAwareFileIndex(sparkSession, Map(), None) {
    override def allFiles(): Seq[FileStatus] = files
    override def partitionSpec() = ???
    override def leafFiles = ???
    override def leafDirToChildrenFiles = ???
    override def rootPaths = ???
    override def refresh() = ???
  }
}
```

src/main/scala/com/microsoft/hyperspace/index/sources/default/DefaultFileBasedRelation.scala
```diff
@@ -41,14 +41,15 @@ class DefaultFileBasedRelation(spark: SparkSession, override val plan: LogicalRe
...
     case HadoopFsRelation(location: PartitioningAwareFileIndex, _, _, _, _, _) =>
-      val result = filesFromIndex(location).sortBy(_.getPath.toString).foldLeft("") {
-        (acc: String, f: FileStatus) =>
-          HashingUtils.md5Hex(acc + fingerprint(f))
-      }
-      result
+      val initialFingerprint = fb.build()
+      var fingerprint = initialFingerprint
+      filesFromIndex(location).foreach(fingerprint ^= createFingerprint(_, fb))
```
sezruby, line 51: not tested code
```suggestion
      val fingerprint = filesFromIndex(location).foldLeft(initialFingerprint){ (fp, file) => 
        fp ^ createFingerprint(file, fb))
     }
```

src/main/scala/com/microsoft/hyperspace/index/FileBasedSignatureProvider.scala
```diff
@@ -19,44 +19,39 @@ package com.microsoft.hyperspace.index
...
-   * @param logicalPlan logical plan of data frame.
-   * @return fingerprint, if the logical plan has supported relations; Otherwise None.
-   */
-  private def fingerprintVisitor(logicalPlan: LogicalPlan): Option[String] = {
+  def signature(logicalPlan: LogicalPlan): Option[Fingerprint] = {
     val provider = Hyperspace.getContext.sourceProviderManager
-    var fingerprint = ""
+    val fb: FingerprintBuilder = fbf.create
+    var updated = false
     logicalPlan.foreachUp {
```
sezruby, line 47: not tested code
```suggestion
val hasSupportedRelations = logicalPlan.collect {
case l: LeafNode if provider.isSupportedRelation(l) =>
        fingerprint ++= provider.getRelation(l).signature
        provider.getRelation(l).signature(fb).foreach(fb.add(_))
        true  
}

if (hasSupportedRelation.nonEmpty()) Some(fb.build()) else None
```

src/main/scala/com/microsoft/hyperspace/index/LogicalPlanSignatureProvider.scala
```diff
@@ -36,15 +40,17 @@ trait LogicalPlanSignatureProvider {
...
    */
-  def signature(logicalPlan: LogicalPlan): Option[String]
+  def signature(logicalPlan: LogicalPlan): Option[Fingerprint]
 }
 
 /**
  * Factory object for LogicalPlanSignatureProvider.
  */
 object LogicalPlanSignatureProvider {
+  private val fbf: FingerprintBuilderFactory = new MD5FingerprintBuilderFactory
```
sezruby, line 50: How about adding "algorithm" in IndexLogEntry? for clarity

```
"fingerprint" : {
          "properties" : {
            "signatures" : [ {
              "algorithm":"md5",
              "provider" : "com.microsoft.hyperspace.index.IndexSignatureProvider",
              "value" : "5269971142c4be0c144c61130cde5543"
            } ]
          },
          "kind" : "LogicalPlan"
        }
```

PR #412, opened 2021-04-09
Issue400 just scalafmt formated

dmytroDragan:

 - **Tracking Issue**: #400
 - **Parent Issue**:  N/A
 - **Dependencies**: N/A

Files (which were changed in https://github.com/microsoft/hyperspace/pull/402) were missing scalafmt formatting.
As result PR-402 is hard to review due to mix of issue fix and scalafmt formatting.

To resolve it, the simplest approach is to apply scalafmt formatting in first place. 
After merging this PR PR-402 should be unblocked.

No

No need


sezruby:
Could you confirm that this was from `hyperspace/dev/.scalafmt.conf`?

sezruby (commented):
Generally looks good to me (if it's formatted with `hyperspace/dev/.scalafmt.conf`), but I'm not sure about the rearrangement part. (maybe it's intelliJ default?) cc @imback82 do we have any arrangement rule?

dmytroDragan:
Hi @sezruby!  It's formatted with hyperspace/dev/.scalafmt.conf

sezruby:
@dmytroDragan could you revert changes from "rearrangement"? 
I think the current arrangement looks better especially for `IndexLogEntry`, tag related functions / variable.
As we don't have a specific rule for arrangement for now, let's let that part so that we could merge this change quickly.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -323,8 +322,8 @@ case class FileInfo(name: String, size: Long, modifiedTime: Long, id: Long) {
   override def equals(o: Any): Boolean = o match {
     case that: FileInfo =>
       name.equals(that.name) &&
-      size.equals(that.size) &&
-      modifiedTime.equals(that.modifiedTime)
+        size.equals(that.size) &&
```
andrei-ionescu, line 325: This is not scalafmt. I applied styling on the code and here it looks like this:

```
  override def equals(o: Any): Boolean = o match {
    case that: FileInfo =>
      name.equals(that.name) &&
      size.equals(that.size) &&
      modifiedTime.equals(that.modifiedTime)
    case _ => false
  }
```

No extra spaces.

Please check your `scalafmt`.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -350,10 +349,11 @@ case class CoveringIndex(properties: CoveringIndex.Properties) {
   val kindAbbr = "CI"
 }
 object CoveringIndex {
-  case class Properties(columns: Properties.Columns,
-    schemaString: String,
-    numBuckets: Int,
-    properties: Map[String, String])
+  case class Properties(
```
andrei-ionescu, line 352: Same here. No changes required. Here is how this part looks after I applied `scalafmt` reformat command:

```
object CoveringIndex {
  case class Properties(columns: Properties.Columns,
    schemaString: String,
    numBuckets: Int,
    properties: Map[String, String])

  object Properties {
    case class Columns(indexed: Seq[String], included: Seq[String])
  }
}
```

No changes.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -377,9 +377,7 @@ object LogicalPlanFingerprint {
  * @param appendedFiles Appended files.
  * @param deletedFiles Deleted files.
  */
-case class Update(
-    appendedFiles: Option[Content] = None,
-    deletedFiles: Option[Content] = None)
+case class Update(appendedFiles: Option[Content] = None, deletedFiles: Option[Content] = None)
```
andrei-ionescu, line 380: Same here. I don't have any changes.

Please check you formatting settings.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -493,21 +494,20 @@ case class IndexLogEntry(
...
-              relations.head.copy(
-                data = relations.head.data.copy(
-                  properties = relations.head.data.properties.copy(
-                    update = Some(
-                      Update(
-                        appendedFiles =
-                          Content.fromLeafFiles(appended.map(toFileStatus), fileIdTracker),
-                        deletedFiles =
-                          Content.fromLeafFiles(deleted.map(toFileStatus), fileIdTracker)))))))))))
+    copy(source = source.copy(plan = source.plan.copy(properties = source.plan.properties.copy(
```
andrei-ionescu, line 497: Neither here - I don't have any changes after applying the reformatting.

andrei-ionescu (changes requested):
I did reviewing after 4-5 times where I did check how `scalafmt` reformatting is applied in my IDE and I can say that all these changes are not needed.

If we would add these changes, that it would mean that when I would apply reformatting on this file on my side, I'll be back to the original version with lots of formatting changes.

I recommend the you double check your formatting settings.

As I already posted on another of your PRs...

>Here is my `scalafmt` setup in IntelliJ Idea (I have `1.5.1` version):
>
>hyperspace_scalafmt
>
>I usually do `Command+Shift+Alt+L` to reformat the file and use the following options in the dialog box:
>
>hyperspace_reformat

If you use IntelliJ IDEA we can review other settings too so that you get it right. I've been through similar issues where applying formatting but after a few retries I managed to make it work and have as less possible changes into PRs due to formatting.

dmytroDragan:
Th problem was with 1.5.1  vs 2.7.5 scalafmt.
https://github.com/microsoft/hyperspace/pull/418

PR #417, opened 2021-04-14
Refactor Content.rec(..) function to support tail-optimisation

dmytroDragan:

 - **Tracking Issue**: #400
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

Refactor Content.rec(..) function to support tail-optimisation

No

Added unit tests


src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -111,6 +102,43 @@ object Content {
...
+   * Apply `func` to each file in directory recursively.
+   *
+   * @param prefixPath Root prefix
+   * @param directory Root directory
+   * @param func Function which would apply to current prefix and file
+   * @tparam T
+   * @return Result list of applying function to all files
+   */
+  def recFilesApply[T](
+                        prefixPath: Path,
```
sezruby, line 116: Could you try this if you are using intellij?
Setting => editor => coding style => scala
- formatter -> scalafmt
- configuration -> `dev\.scalafmt.conf`

ctrl + alt + l

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -111,6 +102,43 @@ object Content {
...
+  /**
+   * Apply `func` to each file in directory recursively.
+   *
+   * @param prefixPath Root prefix
+   * @param directory Root directory
+   * @param func Function which would apply to current prefix and file
+   * @tparam T
+   * @return Result list of applying function to all files
+   */
+  def recFilesApply[T](
```
sezruby, line 115:
```suggestion
  private[hyperspace] def recFilesApply[T](
```

btw is there any reason to move the function to here `object Content`?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -111,6 +102,43 @@ object Content {
...
+
+          val newLevels = curDirs
+            .filter(_.subDirs.nonEmpty)
+            .map(dir => (new Path(curPrefixPath, dir.name), dir.subDirs))
+
+          recAcc(otherDirs ++ newLevels, func, curAcc ++ acc)
+      }
+    }
+
+    recAcc(List(prefixPath -> Seq(directory)), func)
```
sezruby, line 140: nit: for consistency? 
```suggestion
    recAcc(List((prefixPath, Seq(directory))), func)
```

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -111,6 +102,43 @@ object Content {
...
+  /**
+   * Apply `func` to each file in directory recursively.
+   *
+   * @param prefixPath Root prefix
+   * @param directory Root directory
+   * @param func Function which would apply to current prefix and file
+   * @tparam T
+   * @return Result list of applying function to all files
+   */
+  private[hyperspace] def recFilesApply[T](
```
sezruby, line 115: Could you move this function back to `case class Content`? as this is a private function that should be only used in `case class Content`.

dmytroDragan: Hi, sorry for delay. I consider it as just static function, so it could live separately from class. It also simplify testing

clee704: I'm fine with this being a generic utility function, but:

1) The name `recFilesApply` seems unusual. Maybe `applyToFilesRecursively`?
2) The meaning of "prefix" here seems unclear. What is the "root prefix" and what is the "current prefix"? Without looking at the code, it is hard to catch the meaning of the parameters.

dmytroDragan: Thanks @clee704. Valid point, I have inherited `prefixPath` from original function. I have changed it to just `path` and added a short description why we need it

PR #418, opened 2021-04-14
Issue 400 scalafmt

dmytroDragan:

 - **Tracking Issue**: #400
 - **Parent Issue**:  N/A
 - **Dependencies**: N/A

Files (which were changed in https://github.com/microsoft/hyperspace/pull/417) were missing scalafmt formatting.
As result PR-417 is hard to review due to mix of issue fix and scalafmt formatting.

To resolve it, the simplest approach is to apply scalafmt formatting in first place. 
After merging this PR PR-417 should be unblocked.

No

No need


andrei-ionescu (commented):
These methods being moved around will cause substantial conflicts in other PRs. Is this really necessary? Can we check what's the impact on the current opened PRs?

sezruby:
@andrei-ionescu All changes are from rearrangement and I think it looks no problem. Which change are you concerned?

imback82:
Hmm, I don't think we should enforce rearrangement with scalafmt since it is not consistent with other files. Can you please turn if off?

When I run scalafmt on `IndexLogEntry.scala`, I see "No lines changed: content is already properly formatted". @sezruby / @andrei-ionescu what do you see?

dmytroDragan:
OK, I have turned it off. i guess it would be good to add a note about it to contribution guide.
This PR in this case is obsolete. I will update  #417

imback82:
Thanks @dmytroDragan, will update the contribution guide.

Issue #334, opened 2021-01-26
[PROPOSAL]: Gold Standard


apoorvedave1:
**Describe the problem**
A clear and concise description of what the problem is e.g., *I have this scenario and Hyperspace does not work [...]*
#282 Gold Standard

Design for plan validation for TPCDS queries, with features for extensibility to more data sets and various tpcds configs

**Describe your proposed solution**
A clear and concise sketch of the solution you want to propose.
Class Diagram:
[class diag.pdf](https://github.com/microsoft/hyperspace/files/5871578/class.diag.pdf)

rapoth:
Thank you! Can you follow the [outline from here](https://github.com/microsoft/hyperspace/blob/master/docs/design/TEMPLATE.md) please? I think you have most of it already.

imback82:
Can you explain `tpcdsBasic` vs. `tpcdsOther`?

apoorvedave1:
Thanks @rapoth , yeah I will fix it before merging.

the general idea is we will have lot's of approved simplified plans for different configs. Just as an example, with and without hybrid scan enabled, we will have very different "simplifiedApprovedPlans". so I used `tpcdsBasic` meaning `tpcds with config 1` and `tpcdsOther` meaning `tpcds with config 2`

e.g.
```
tpcdsBasic/
  approvedPlans/
    q1simplified.txt
tpcdsOther/
  approvedPlans/
    q1simplified.txt
```

tpcdsBasic/.../q1Simplified.txt:
```
Project..
  Filter..
    FileScan: <HyperspaceIndex>
```

tpcdsOther/.../q1Simplified.txt:
```
Project..
  BucketUnion
    Filter
      FileScan: <Unindexed files identified by hybrid scan>
    Filter..
      FileScan: <HyperspaceIndex>
```

apoorvedave1:
@imback82 @pirz @thugsatbay @sezruby please take a look at the latest changes for gold standard

PR #419, opened 2021-04-15
Remove linefeed check from gold standard plan comparison

apoorvedave1:
Linux and Windows have different notions of end of line (CRLF vs LF, \r\n vs \n) which causes gold standard tests to fail when the golden file is created on one setup and tested on another.

 - **Parent Issue**: #334

Remove linefeeds before comparing plans.

no

https://github.com/microsoft/hyperspace/pull/376 sanity check will be added to build process and will start passing for windows.


andrei-ionescu (approved):
LGTM

sezruby:
LGTM thanks!

imback82 (approved):
LGTM, thanks @apoorvedave1!

Issue #405, opened 2021-04-02
[PROPOSAL]: HyperspaceOneRule


sezruby:
NOTE: This proposal is outdated and most of PRs are merged. Please refer the [PRs](#implementation)

## Problem Statement
Currently, Hyperspace supports 1 type of index - non-clustered CoveringIndex. 
And there're 2 optimizer rule - FilterIndexRule, JoinIndexRule. 

However, as we're going to introduce another types of indexes (e.g. bloom filter index, .. etc), it's hard to apply different type of indexes only using sequence of optimizer rules. Because
1) rules are applied in order, so latter rules cannot be applied to the plan which was transformed by former rules 
2) each rule cannot consider other available indexes

## Background and Motivation
TBD

## Proposed Solution
Define one rule and handle all indexes in the rule - HyperspaceOneRule
#### Step1) Index candidate collection => only using the relation
  - column schema check
  - signature (source dataset) check
  - filter same index type with less coverage
#### Step2) Score based index application
  - backtracking w/ memoization => try all possible combinations
  - keep transform plan & estimation cost
  - score estimation
    - index type
    - shuffle count
    - number of indexes
    - Hybrid scan / non-hybrid scan
   
### Structure overview

![image](https://user-images.githubusercontent.com/51077614/115628173-1466e500-a2b5-11eb-86a2-55c053ee8ba9.png)

### Batches
FilterIndexRule & JoinIndexRule can be represented as a "Batch". It consists of sequence of `HyperspacePlanCheck`. 

```scala
trait HyperspaceBatch extends ActiveSparkSession {
  val planChecks: Seq[HyperspacePlanCheck]
  def applyIndex(plan: LogicalPlan, indexes: Map[LogicalPlan, Seq[IndexLogEntry]]): LogicalPlan
  def score(plan: LogicalPlan, indexes: Map[LogicalPlan, Seq[IndexLogEntry]]): Int
  final def apply(
      plan: LogicalPlan,
      indexes: Map[LogicalPlan, Seq[IndexLogEntry]]): (LogicalPlan, Int) = {
    if (indexes.isEmpty) {
      return (plan, 0)
    }
    val candidateIndexes = planChecks.foldLeft(indexes) { (pti, check) =>
      check(plan, pti)
    }
    if (candidateIndexes.nonEmpty) {
      (applyIndex(plan, candidateIndexes), score(plan, candidateIndexes))
    } else {
      (plan, 0)
    }
  }
}
```

#### NoOpBatch
[...]

PR #427, opened 2021-04-21
Add ApplyHyperspace rule definitions

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: #405 
 - **Dependencies**: n/a

Add HyperspaceOneRule related definitions. This change introduces several objects & traits for HyperspaceOneRule. Implementation is almost empty & will be added other PRs.

- `object ApplyHyperspace`
  - `object CandidateIndexCollector`
    - `trait SourcePlanIndexFilter`
  - `class ScoreBasedIndexApplication`
    - `trait HyperspaceRule` => FilterIndexRule, JoinIndexRule
      - `trait QueryPlanIndexFilter`
      - `trait RankerIndexFilter`  
    - `object NoOpRule extends HyperspaceRule`

No

will be added in another PR


src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
```diff
@@ -0,0 +1,82 @@
...
+    } else {
+      (plan, 0)
+    }
+  }
+}
+
+/**
+ * No-op rule for traversal.
+ */
+object NoOpHSRule extends HyperspaceRule {
```
sezruby, line 74: Any better naming here? `HSRule` - this is the best I can think of.

imback82: I am fine with just `NoOpRule`.

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceCheck.scala
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package com.microsoft.hyperspace.index.rules
+
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+
+import com.microsoft.hyperspace.ActiveSparkSession
+import com.microsoft.hyperspace.index.IndexLogEntry
+
+trait HyperspaceCheck {
```
imback82, line 24: any reason we are prefixing `Hyperspace` for these traits?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceCheck.scala
```diff
@@ -0,0 +1,45 @@
...
+   * @return Failure reason if an index is excluded in the check.
+   */
+  def reason: String
+}
+
+/**
+ * HyperspaceCheck used in CandidateIndexCollector.
+ */
+trait HyperspaceSourceCheck extends HyperspaceCheck with ActiveSparkSession {
+  def apply(plan: LogicalPlan, indexes: Seq[IndexLogEntry]): Seq[IndexLogEntry]
```
imback82, line 35: Could you add some comments on the return value? Is this also filtering out the given indexes? Then it's doing more than "checking"? Maybe "filter" is a better name?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceCheck.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+
+import com.microsoft.hyperspace.ActiveSparkSession
+import com.microsoft.hyperspace.index.IndexLogEntry
+
+trait HyperspaceCheck {
+  /**
+   * @return Failure reason if an index is excluded in the check.
+   */
+  def reason: String
```
imback82, line 28: I see that source check, and plan check take "indexes". If multiple indexes are filtered, will you aggregate them into a single reason?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceCheck.scala
sezruby: To avoid any confusion from general terms, but I'm okay with omitting Hyperspace prefix.

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceCheck.scala
sezruby: WhyNot will work with a given index name or none. 
If the index or all indexes are filtered by a check, then we can show reason.

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceCheck.scala
sezruby: Yes mainly filtering out indexes. But there could be reordering indexes or checking plan (filter node check or join node check - though it also returns empty index list if it fails). Maybe `Filter` is better?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala
```diff
@@ -0,0 +1,116 @@
...
+ * Apply Hyperspace indexes based on the score of each index application.
+ */
+class ScoreBasedIndexApplication {
+  // TODO: FilterIndexRule :: JoinIndexRule :: Nil
+  val rules = NoOpRule :: Nil
+
+  // Map for memoization.
+  val scoreMap: mutable.HashMap[LogicalPlan, (LogicalPlan, Int)] = mutable.HashMap()
+
+  def recApply(
```
clee704, line 71: nit: private? (maybe anything here should be private except `apply()`)

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala
```diff
@@ -0,0 +1,116 @@
...
+
+  def apply(
+      plan: LogicalPlan,
+      allIndexes: Seq[IndexLogEntry]): Map[LogicalPlan, Seq[IndexLogEntry]] = {
+    val planToIndexes = initializePlanToIndexes(plan, allIndexes)
+    planToIndexes.flatMap {
+      case (node, candidateIndexes) =>
+        Some(node, sourceFilters.foldLeft(candidateIndexes) { (indexes, filter) =>
+          val res = filter(node, indexes)
+          res
```
clee704, line 55: simply `filter(node, indexes)`?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala
```diff
@@ -0,0 +1,116 @@
...
+        }).filterNot(_._2.isEmpty)
```
clee704, line 56: `filter(_._2.nonEmpty)`?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala
```diff
@@ -0,0 +1,116 @@
...
+      .getContext(spark)
+      .indexCollectionManager
+    val allIndexes = indexManager.getIndexes(Seq(Constants.States.ACTIVE))
+    if (allIndexes.isEmpty) {
+      plan
+    } else {
+      try {
+        val candidateIndexes = CandidateIndexCollector(plan, allIndexes)
+        val res = new ScoreBasedIndexApplication().apply(plan, candidateIndexes)
+        res
```
clee704, line 109: `new ScoreBasedIndexApplicate()....`?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala
```diff
@@ -0,0 +1,116 @@
...
+      } catch {
+        case _: Exception =>
+          plan
```
clee704, line 112: Which exceptions is this code expecting here? Can we silently ignore any exceptions here? Shouldn't we at least leave something for the user so that the problem can be fixed? Are lines 99-102 okay to throw an exception?

sezruby: Thanks, added log msg. Currently we ignore the exception during the index application / plan transformation, so as to minimize the effect to the app. 
I think we could throw the exception from getIndexs here, because it's mostly a permission or configuration issue.

sezruby:
`sbt clean` started to fail again 😫

clee704:
LGTM, thanks!

clee704 (approved):

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceFilter.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import com.microsoft.hyperspace.ActiveSparkSession
+import com.microsoft.hyperspace.index.IndexLogEntry
+
+trait HyperspaceFilter {
+
+  /**
+   * @return Failure reason for filtered out indexes.
+   */
+  def reason: String
```
imback82, line 29: I am not still sure about this abstraction since this should be called after `apply` - which is defined in a trait that extends this - is called, and make the implementation have a mutable state.

How about the following (pseudo code)?
```scala
trait IndexFilter[A] {
  def apply(plan: LogicalPlan, entries: Seq[A]): (Seq[A], Option[String]) // Option[String] is the reason.
}

trait SourceFilter extends IndexFilter[IndexLogEntry]
trait PlanFilter extends IndexFilter[(LogicalPlan, Seq[IndexLogEntry])] // I don't think `Map` is required.
```

sezruby: But `reason` will be only used in whyNot API and I don't think it's proper that the return type has the reason string.
The reason string can be empty or non-empty depending on the target index of whyNot API. 

Could you explain more about `the implementation have a mutable state`?

imback82: Meaning it will need to have `var`.

With the current approach:
```scala
val myFilter = MyFilter(...)
println(myFilter.reason) // empty
val filtered1 = myFilter.apply(...))
val filtered2 = myFilter.apply(...))
println(myFilter.reason) // hopefully the reason is from the latest apply, implementor has to make sure to "clear" the reason.
```

Suggested approach if you don't think the reason string is not "proper" in the return type:
```scala
val myFilter = MyFilter(...)
myFilter.filter(...)
val (filtered1, reason1) = myFilter.filterWithReason(...) // no confusion here
val (filtered2, reason2) = myFilter.filterWithReason(...)
```

sezruby: Oh, seems I didn't share impl. of whyNot API in my mind. It looks like:
```scala
val filteredIndexes = MyFilter(..)
if (!filteredIndexes.contains(targetIndex)) {
  println(MyFilter.reason)
}
...
```

Actually it's `val reason` rather than `def reason`.. and maybe we can do the following if you prefer this structure:

```scala 
def apply(targetIndex, plan, entries): (Seq[A], Option[String]) = {
  val filtered = apply(plan, entries)
  val reasonStr = if (filtered.contains(targetIndex)) { None }
    else { Some(reason) }
  (filtered, reasonStr)
}
```

imback82: > Actually it's `val reason` rather than `def reason`

How can you do this? Don't you need to update the `MyFilter.reason` whenever you call `apply` since the reason can be different?





> maybe we can do the following if you prefer this structure:

Maybe this will be better because I think a reason should map to "one target" (I think I brought this up in a different thread). Again, we can have a separate `def` that returns a reason, but I am not sure about having `def reason` as an API due to the concerns I shared above (check the comment in my example above).

sezruby: The reason string will be always same for each filter. Otherwise it will ruin readability of filter code by many of if-not-throw(or set msg) - I tried but gave up to support different reasons because of that.

imback82: Ah OK, then it makes sense. I thought we were doing "many of if-not-throw(or set msg)"

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceFilter.scala
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package com.microsoft.hyperspace.index.rules
+
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+
+import com.microsoft.hyperspace.ActiveSparkSession
+import com.microsoft.hyperspace.index.IndexLogEntry
+
+trait HyperspaceFilter {
```
imback82, line 24: Look like you are filtering out indexes, so `IndexFilter`?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceFilter.scala
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * @return Failure reason for filtered out indexes.
+   */
+  def reason: String
+}
+
+/**
+ * Filter used in CandidateIndexCollector.
+ */
+trait SourceFilter extends HyperspaceFilter with ActiveSparkSession {
```
imback82, line 35: This sounds like we are filtering the source plan whereas we are filtering indexes. How about `IndexFilterOnSourcePlan` and `IndexFilterOnQueryPlan`? (assuming you are ok with `IndexFilter`)

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceFilter.scala
clee704: Actually, my first impression was the same as @imback82. Shouldn't the message different for each invocation? I'm not sure how setting messages for different reasons can *ruin* the readability that much, and moreover, is it really more important than the user experience? For example, with the current implementation, users are forced to examine their queries to find simple mistakes such as missing columns, and this could be time consuming depending in the size of the query and the tables. This is just an example and this can be more serious - maybe the current approach is sufficient for the current filters, but there could be filters in the future for which a generic class-wide error message is near useless.

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
```diff
@@ -0,0 +1,83 @@
...
+import com.microsoft.hyperspace.index.IndexLogEntry
+
+/**
+ * Interface of exclusive type of indexes.
+ */
+trait HyperspaceRule extends ActiveSparkSession {
+
+  /**
+   * Sequence of conditions to apply indexes. Each filter contains conditions and
+   * filter out candidate indexes based on the conditions. The order of the sequence does matter.
```
imback82, line 31: Can you update the comment with "The order of the sequence does matter because ..."?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
```diff
@@ -0,0 +1,83 @@
...
+  def applyIndex(plan: LogicalPlan, indexes: Map[LogicalPlan, Seq[IndexLogEntry]]): LogicalPlan
+
+  /**
+   * Calculate the score of the index application.
+   *
+   * @param plan Original query plan.
+   * @param indexes Selected indexes.
+   * @return Score of selected indexes.
+   */
+  def score(plan: LogicalPlan, indexes: Map[LogicalPlan, Seq[IndexLogEntry]]): Int
```
imback82, line 51: Question. I see the following below:
```scala
(applyIndex(plan, candidateIndexes), score(plan, candidateIndexes))
```
Are you trying to apply `candidateIndexes` again to plan inside `score`? Don't you need to utilize the result of `applyIndex`?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala
```diff
@@ -0,0 +1,115 @@
...
+    val provider = Hyperspace.getContext(spark).sourceProviderManager
+    plan.collect {
+      case l: LeafNode if provider.isSupportedRelation(l) =>
+        (l.asInstanceOf[LogicalPlan], indexes)
+    }.toMap
+  }
+
+  def apply(
+      plan: LogicalPlan,
+      allIndexes: Seq[IndexLogEntry]): Map[LogicalPlan, Seq[IndexLogEntry]] = {
```
imback82, line 49: I see `Map[LogicalPlan, Seq[IndexLogEntry]]` in many places. Should we create a type to represent it?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceFilter.scala
sezruby: https://github.com/microsoft/hyperspace/pull/403/files#diff-c6cbe018cfa141e77ee664c5a8fc680f1bd2d6827ca16c2368e34c812b2c7ab0R27 
This commented out code was from that try. 
```def withWhyNotCheck(indexes: Seq[IndexLogEntry], msg: String)```

For each condition & msg inside a filter, call this func with "filtered indexes" and proper msg. 
That was somewhat bothering me as I needed to split one condition into several statements which can be done with one statement & define multiple "filtered" indexes only for this API call. 
If it's not for whyNot API -disabled whyNotAPI-, it just makes more func calls and to avoid that overhead, we need some duplicate code for the API which is also not acceptable to me..

But I agree that user experience is more important.

Comment (clee704) [reply src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceFilter.scala:29]: I think splitting one condition into several can actually benefit the reader of the filter code too, unless it's extremely verbose or trivial.

For the performance, we can make it a macro that sets the message if whyNot API is enabled, or does nothing otherwise. I'm not sure whether it can be done with Scala's macros but we can try if the performance overhead is significant.

File: src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala (hunk 14 above)
Comment (sezruby) [reply src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala:51]: I assume that `candidateIndexes` at the point is the final candidate which means *all* `candidateIndexes` should be "applicable" to the given plan. I split the function for better structure & function role for readability.

Comment (imback82) [reply src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala:51]: Shouldn't `score` use the same apply logic as `applyIndex`? Do you have a sample implementation?

If I understand the intention of `score` correctly, it seems that you can inject `Scorer` trait into `applyIndex` instead, and invoke it whenever an index is applied.

File: src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceFilter.scala (hunk 10 above)
Comment (imback82) [reply src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceFilter.scala:29]: Thanks @sezruby / @clee704 for the discussion. Let's go with the current approach and evolve it if needed/required.

File: src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala [hunk 16]
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Apply Hyperspace indexes based on the score of each index application.
+ */
+class ScoreBasedIndexApplication {
+  // TODO: FilterIndexRule :: JoinIndexRule :: Nil
+  private val rules = NoOpRule :: Nil
+
+  // Map for memoization.
+  private val scoreMap: mutable.HashMap[LogicalPlan, (LogicalPlan, Int)] = mutable.HashMap()
```
imback82, line 68: Shall we add a comment what each `LogicalPlan` represents in the `HashMap`?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
```diff
@@ -0,0 +1,83 @@
...
+trait HyperspaceRule extends ActiveSparkSession {
+
+  /**
+   * Sequence of conditions to apply indexes. Each filter contains conditions and
+   * filter out candidate indexes based on the conditions. The order of the sequence does matter.
+   */
+  val planFilters: Seq[PlanFilter]
+
+  /**
+   * Transform the plan to use the selected indexes.
```
imback82, line 36: Let's also mention that these indexes are after applying filters. It would be good if you could define some terms: candidate indexes, selected indexes, etc.

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
```diff
@@ -0,0 +1,83 @@
...
+   * @return Transformed plan to use the selected indexes.
+   */
+  def applyIndex(plan: LogicalPlan, indexes: Map[LogicalPlan, Seq[IndexLogEntry]]): LogicalPlan
+
+  /**
+   * Calculate the score of the index application.
+   *
+   * @param plan Original query plan.
+   * @param indexes Selected indexes.
+   * @return Score of selected indexes.
```
imback82, line 49: This comment is a bit confusing and conflicts with the comment above "Calculate the score of the index application". Are we scoring "selected indexes" or the "index application"?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
sezruby: https://github.com/microsoft/hyperspace/pull/403/files#diff-c3a2b337d86b3cc05c4ac4cdafa48c332221c03b38df5cd6a22e315541334d08R516 - example for JoinIndex

> If I understand the intention of score correctly, it seems that you can inject Scorer trait into applyIndex instead, and invoke it whenever an index is applied.

We might need to consider left & right index together (though the sample doesn't); so I thought it would be better to calculate the score with a different function.

imback82: I guess my uber question is is there a relationship b/w `applyIndex` and `score` or are they totally orthogonal?

Also, I see the following in the sample code for `score`: `val candidateIndex = indexes.head._2.head`. So we don't care which index is actually applied?

sezruby: Yes they are orthogonal. `applyIndex` should be able to apply those indexes.

imback82: OK, then what do you mean by `Calculate the score of the index application.`? This index application is different from what's happening inside `applyIndex`?

sezruby: I revised it to "score of the selected indexes". It was because evaluating indexes assumes index application.

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  def apply(plan: LogicalPlan, indexes: PlanToIndexesMap): LogicalPlan = {
+    recApply(plan, indexes)._1
+  }
+}
+
+/**
+ * Transform the given plan to use Hyperspace indexes.
+ */
+object HyperspaceOneRule
```
clee704, line 89: nit: The name still bugs me but I can't think of a better name. How about `ApplyHyperspaceIndices` (or `ApplyHyperspaceIndexes`, if you prefer "indexes" over "indices"), as most built-in rule's name is in the imperative form such as `ResolveAliases`, `EliminateUnions`, `PushDownPredicates`, etc.? Also, most rules don't have the suffix "Rule".

clee704 (approved):
LGTM

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceOneRule.scala
sezruby: That sounds fair enough. But compared to that built-in rules, this rule is more complex and will do various roles, so I'm good with either the current one or `ApplyHyperspaceIndexes`.  @imback82 WDYT?

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    val selectedIndexes = filtersOnQueryPlan
+      .foldLeft(indexes) { (pti, filter) =>
+        filter(plan, pti)
+      }
+      .map { kv =>
+        // After applying filtersOnQueryPlan, each candidate relation should have
+        // one candidate index.
+        assert(kv._2.length == 1)
```
imback82, line 68: Should we enforce this with API for now, meaning `QueryPlanIndexFilter` returns `Map[LogicalPlan, IndexLogEntry]`?

src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Transform the given plan to use Hyperspace indexes.
+ */
+object ApplyHyperspace
+    extends Rule[LogicalPlan]
+    with Logging
+    with HyperspaceEventLogging
+    with ActiveSparkSession {
+
+  type PlanToIndexesMap = Map[LogicalPlan, Seq[IndexLogEntry]]
```
imback82, line 95: I am thinking whether we should define the following:
```scala
type PlanToCandidateIndexesMap = Map[LogicalPlan, Seq[IndexLogEntry]]
type PlanToSelectedIndexMap = Map[LogicalPlan, IndexLogEntry]
```
, and clearly define what candidate and selected index mean here. WDYT?

src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
```diff
@@ -0,0 +1,115 @@
...
+          filter(node, indexes)
+        }).filter(_._2.nonEmpty)
+    }
+  }
+}
+
+/**
+ * Apply Hyperspace indexes based on the score of each index application.
+ */
+class ScoreBasedIndexApplication {
```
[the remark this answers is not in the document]
sezruby: But it also transforms the plan 🤔  How about `ScoreBasedIndexPlanOptimizer`?

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -0,0 +1,61 @@
...
+  def reason: String
+}
+
+/**
+ * IndexFilter used in CandidateIndexCollector.
+ */
+trait SourcePlanIndexFilter extends IndexFilter with ActiveSparkSession {
+
+  /**
+   * Filter out candidate indexes for the given source plan.
```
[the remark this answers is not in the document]
sezruby: Right but after thru one filter, it might not `allIndexes` anymore.

src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
sezruby: yep it looks better & consistent 👍

src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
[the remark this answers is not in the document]
imback82: ah ok, sgtm

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
sezruby: We could introduce additional filter which returns PlanToSelectedIndexes if necessary - `RankFilter`?

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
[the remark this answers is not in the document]
imback82: we can define candidate indexes as the ones that survived after running all `SourcePlanIndexFilter`s.

src/main/scala/com/microsoft/hyperspace/index/rules/HyperspaceRule.scala
imback82: yea seems fine to me. so, we will have:
```
all indexes ----> candidate indexes ---> selected indexes -----> best index
             |                       |                      |
      apply all SPIFs         apply all QPIFs       apply Ranker
```
, where
all indexes: all indexes available
candidate indexes: indexes after applying all SPIFs looking at only the source plan. These indexes may not be eligible against the overall query
selected indexes: index after applying all QPIFs looking at the overall query. Any of these selected indexes can be applied against the query, but they are not ranked.
best index: the best index to apply among selected indexes which is ranked by the Ranker.

sezruby: selected indexes sounds bit weird. how about applicableIndexes & best(or selected)Index ?

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Rank best index for the given query plan.
+   *
+   * @param plan Query plan
+   * @param applicableIndexes Map of source plan to applicable indexes
+   * @return Map of source plan to selected index
+   */
+  def apply(
+      plan: LogicalPlan,
+      applicableIndexes: PlanToCandidateIndexesMap): PlanToSelectedIndexMap
```
imback82, line 79: Looks like this is conflicting now? I think we need to revert this back to PlanToIndexesMap, and use "candidateIndexes: PlanToIndexesMap", and "applicableIndexes: PlanToIndexesMap"

PR #428, opened 2021-04-21
Deterministic index selection for filter rule with no hybrid scan

apoorvedave1:
Deterministic index selection for FilterIndexRule with hybrid scan disabled.

 - **Dependent PRs**: https://github.com/microsoft/hyperspace/pull/386

This PR changes the filter index rule selection from picking the `head` of the eligible indexes to picking the index which has
- the minimum size in terms of index file sizes
- lexicographically smallest index name (if there are multiple indexes with minimum size)

no

unit tests

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
@@ -54,7 +54,12 @@ object FilterIndexRanker {
         } else {
           // TODO: Add ranking algorithm to sort candidates.
           //  See https://github.com/microsoft/hyperspace/issues/52
-          Some(candidates.head)
+
+          // Pick the index with minimum size. If indexes with same size are found, pick the
+          // one with lexicographically smaller name. This is required for deterministic selection
+          // of indexes.
+          Some(candidates.minBy(index =>
+            (index.getTagValue(plan, IndexLogEntryTags.INDEX_SIZE_IN_BYTES), index.name)))
         }
     }
   }
--- a/src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
@@ -58,8 +58,7 @@ object FilterIndexRanker {
           // Pick the index with minimum size. If indexes with same size are found, pick the
           // one with lexicographically smaller name. This is required for deterministic selection
           // of indexes.
-          Some(candidates.minBy(index =>
-            (index.getTagValue(plan, IndexLogEntryTags.INDEX_SIZE_IN_BYTES), index.name)))
+          Some(candidates.minBy(index => (index.indexFilesSizeInBytes, index.name)))
         }
     }
   }
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -113,6 +113,11 @@ object RuleUtils {
               IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES,
               commonBytes)
 
+            entry.setTagValue(
+              relation.plan,
+              IndexLogEntryTags.INDEX_SIZE_IN_BYTES,
+              entry.content.fileInfos.foldLeft(0L)(_ + _.size))
+
             // If there is no change in source dataset, the index will be applied by
             // transformPlanToUseIndexOnlyScan.
             entry.setTagValue(
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
@@ -113,11 +113,6 @@ object RuleUtils {
               IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES,
               commonBytes)
 
-            entry.setTagValue(
-              relation.plan,
-              IndexLogEntryTags.INDEX_SIZE_IN_BYTES,
-              entry.content.fileInfos.foldLeft(0L)(_ + _.size))
-
             // If there is no change in source dataset, the index will be applied by
             // transformPlanToUseIndexOnlyScan.
             entry.setTagValue(
--- a/src/test/scala/com/microsoft/hyperspace/index/HyperspaceSuite.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/HyperspaceSuite.scala
@@ -18,12 +18,13 @@ package com.microsoft.hyperspace.index
 
 import java.io.File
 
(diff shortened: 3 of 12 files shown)


src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
```diff
@@ -54,7 +54,12 @@ object FilterIndexRanker {
         } else {
           // TODO: Add ranking algorithm to sort candidates.
           //  See https://github.com/microsoft/hyperspace/issues/52
-          Some(candidates.head)
+
+          // Pick the index with minimum size. If indexes with same size are found, pick the
+          // one with lexicographically smaller name. This is required for deterministic selection
+          // of indexes.
+          Some(candidates.minBy(index =>
+            (index.getTagValue(plan, IndexLogEntryTags.INDEX_SIZE_IN_BYTES), index.name)))
```
sezruby, line 62: Could you use `sourceFilesSizeInBytes` in IndexLogEntry?

apoorvedave1: i think that is for source data, not index data

apoorvedave1: pl. let me know if i am wrong

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
```diff
@@ -113,6 +113,11 @@ object RuleUtils {
               IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES,
               commonBytes)
 
+            entry.setTagValue(
```
sezruby, line 116: This function is for hybrid scan.

src/main/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRanker.scala
sezruby: ok.. then can we just add `lazy val indexFilesSizeInBytes` instead of tag?
since I think we can use this condition for join index ranker.

src/main/scala/com/microsoft/hyperspace/index/rules/RuleUtils.scala
apoorvedave1: sorry I don't understand why we can't use this function. Could you please explain why we need to limit its use to hybrid scan?

apoorvedave1: For gold standard tests, there are no files present on disk so to test this functionality, I will need to create index files. Just trying to avoid that case

sezruby: This change is for "non" hybrid scan, but if hybrid scan is not enabled, this function is not called.

apoorvedave1: oh i see what you are saying, from your previous comment i thought "setTagValue" was created specifically for hybrid scan.

src/test/scala/com/microsoft/hyperspace/index/HyperspaceSuite.scala
```diff
@@ -18,12 +18,13 @@ package com.microsoft.hyperspace.index
 
 import java.io.File
 
+import org.apache.hadoop.conf.Configuration
```
sezruby, line 21: nit: --

src/test/scala/com/microsoft/hyperspace/index/HyperspaceSuite.scala
```diff
@@ -18,12 +18,13 @@ package com.microsoft.hyperspace.index
...
 import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkFunSuite
 import org.apache.spark.util.hyperspace.Utils
 
 import com.microsoft.hyperspace.{Hyperspace, SparkInvolvedSuite}
-import com.microsoft.hyperspace.util.FileUtils
+import com.microsoft.hyperspace.util.{FileUtils, PathUtils}
```
sezruby, line 27: nit: --

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleSuite.scala
```diff
@@ -99,7 +103,7 @@ trait HyperspaceRuleSuite extends HyperspaceSuite {
       spark)
 
   def getIndexRootPath(indexName: String): Path =
-    new Path(systemPath, indexName)
+    PathUtils.makeAbsolute(new Path(systemPath, indexName), new Configuration)
```
sezruby, line 106: Is the change in this file required?

src/test/scala/com/microsoft/hyperspace/index/rankers/FilterIndexRankerTest.scala
```diff
@@ -45,16 +45,36 @@ class FilterIndexRankerTest extends HyperspaceRuleSuite {
...
+    // Index with only 1 file of size 10
+    val ind1 = createIndexLogEntry(
+      "ind1",
+      Seq(t1c1),
+      Seq(t1c2),
+      tempPlan,
+      writeLog = false,
+      filenames = Seq("f1.parquet", "f2.parquet"))
+
+    // Index with only 2 files of total size 20
```
imback82, line 58: comments don't seem to match up?

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleSuite.scala
```diff
@@ -27,17 +28,18 @@ import com.microsoft.hyperspace.HyperspaceException
 import com.microsoft.hyperspace.actions.Constants
 import com.microsoft.hyperspace.index._
 import com.microsoft.hyperspace.index.Hdfs.Properties
+import com.microsoft.hyperspace.util.PathUtils
 
 trait HyperspaceRuleSuite extends HyperspaceSuite {
-  private val filenames = Seq("f1.parquet", "f2.parquet")
```
imback82, line 32: You can change this to `defaultFilenames` and use it as a default parameter.

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleSuite.scala
apoorvedave1: yes without this system path could be relative

src/test/scala/com/microsoft/hyperspace/index/rules/HyperspaceRuleSuite.scala
apoorvedave1: thanks @imback82

imback82 (approved):
LGTM (pending @sezruby's approval), thanks @apoorvedave1!

sezruby:
LGTM Thanks!

Issue #185, opened 2020-10-07
Use TreeNodeTag for index plan identifier in Spark 3.0


sezruby:
**Feature requested**

#178 introduces an identifier for plans which is already modified to apply an index.
The identifier is stored in `options` of `HadoopFsRelation` which is ad hoc approach.

Spark 3.0 provides `TreeNodeTag` feature to keep some values for each plan node.
Therefore, we could utilize the tag for the identifier.

ref) https://github.com/apache/spark/blob/branch-3.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala

**Acceptance criteria** 

**Success criteria**

**Additional context**

PR #437, opened 2021-04-28
ApplyHyperspace rule - CandidateIndexCollector implementation

sezruby:

 - **Tracking Issue**: n/a 
 - **Parent Issue**: #405 
 - **Dependencies**: #427 

Add implementation of CandidateIndexCollector.
It consists of `ColumnSchemaFilter` and `FileSignatureFilter`. 
`FileSignatureFilter` is from `RuleUtils.getCandidateIndexes`

No

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
@@ -53,4 +53,8 @@ object IndexLogEntryTags {
   // for Hybrid Scan.
   val INMEMORYFILEINDEX_HYBRID_SCAN_APPENDED: IndexLogEntryTag[InMemoryFileIndex] =
   IndexLogEntryTag[InMemoryFileIndex]("inMemoryFileIndexHybridScanAppended")
+
+  // WHYNOT_REASON stores reason string for disqualification.
+  val WHYNOT_REASON: IndexLogEntryTag[String] =
+    IndexLogEntryTag[String]("whyNotReason")
 }
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
@@ -54,7 +54,11 @@ object IndexLogEntryTags {
   val INMEMORYFILEINDEX_HYBRID_SCAN_APPENDED: IndexLogEntryTag[InMemoryFileIndex] =
   IndexLogEntryTag[InMemoryFileIndex]("inMemoryFileIndexHybridScanAppended")
 
-  // WHYNOT_REASON stores reason string for disqualification.
-  val WHYNOT_REASON: IndexLogEntryTag[String] =
-    IndexLogEntryTag[String]("whyNotReason")
+  // WHYNOT_REASON stores reason strings for disqualification.
+  val WHYNOT_REASON: IndexLogEntryTag[Seq[String]] =
+    IndexLogEntryTag[Seq[String]]("whyNotReasons")
+
+  // WHYNOT_ENABLED indicates whether whyNotAPI is enabled or not.
+  val WHYNOT_ENABLED: IndexLogEntryTag[Boolean] =
+    IndexLogEntryTag[Boolean]("whyNotEnabled")
 }
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
@@ -16,11 +16,17 @@
 
 package com.microsoft.hyperspace.index.rules
 
+import scala.collection.mutable
+
+import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 
-import com.microsoft.hyperspace.ActiveSparkSession
-import com.microsoft.hyperspace.index.IndexLogEntry
+import com.microsoft.hyperspace.{ActiveSparkSession, Hyperspace}
+import com.microsoft.hyperspace.index.{IndexConstants, IndexLogEntry, IndexLogEntryTags, LogicalPlanSignatureProvider}
+import com.microsoft.hyperspace.index.IndexLogEntryTags.{HYBRIDSCAN_RELATED_CONFIGS, IS_HYBRIDSCAN_CANDIDATE}
 import com.microsoft.hyperspace.index.rules.ApplyHyperspace.{PlanToIndexesMap, PlanToSelectedIndexMap}
+import com.microsoft.hyperspace.index.sources.FileBasedRelation
+import com.microsoft.hyperspace.util.{HyperspaceConf, ResolverUtils}
 
 trait IndexFilter {
 
@@ -63,7 +69,7 @@ trait QueryPlanIndexFilter extends IndexFilter with ActiveSparkSession {
 /**
  * IndexFilter used in ranking applicable indexes.
  */
-trait RankerIndexFilter extends IndexFilter with ActiveSparkSession {
+trait IndexRankFilter extends IndexFilter with ActiveSparkSession {
 
   /**
    * Rank best index for the given query plan.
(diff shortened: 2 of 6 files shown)


src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -74,3 +80,151 @@ trait RankerIndexFilter extends IndexFilter with ActiveSparkSession {
...
+      Some(entry)
+    } else {
+      None
+    }
+  }
+
+  override def reason: String = {
+    "Underlying data is changed. Hybrid Scan Configs: \n" +
+      s"${IndexConstants.INDEX_HYBRID_SCAN_ENABLED}: ${HyperspaceConf.hybridScanEnabled(spark)}, " +
+      s"${IndexConstants.INDEX_HYBRID_SCAN_APPENDED_RATIO_THRESHOLD}: " +
```
imback82, line 225: Hmm. It's possible that an index was not selected b/c of the threshold constraint was not met. How would this `reason` help in that case?

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -74,3 +80,151 @@ trait RankerIndexFilter extends IndexFilter with ActiveSparkSession {
...
+        val isCandidate = isAppendAndDeleteCandidate || isAppendOnlyCandidate
+        if (isCandidate) {
+          entry.setTagValue(
+            relation.plan,
+            IndexLogEntryTags.COMMON_SOURCE_SIZE_IN_BYTES,
+            commonBytes)
+
+          // If there is no change in source dataset, the index will be applied by
+          // RuleUtils.transformPlanToUseIndexOnlyScan.
+          entry.setTagValue(
```
imback82, line 207: Let's clearly document in the public API if we are modifying the entry. Also, should this be done at the `SourcePlanIndexFilter` level or do we need a different abstraction?

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
imback82: Maybe, another way to achieve a better message is by using the tag (we can set the filter reason tag to the index log entry, for example)

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
sezruby: These tags are for caching result, so I guess it's hard to do this in somewhere else.

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
imback82: If we go with the log entry tagging approach, we should handle it now, no? e.g., the static `reason` doesn't make sense anymore.

sezruby: hmm if so we don't need to split a rule into multiple filters, thus I'd like to keep `reason` in somehow. 🤔

imback82: Having multiple filters seems fine to me even without `reason`?

sezruby: But no explicit reason/condition for splitting them 🤔ok let's just remove it for now.

sezruby:
@clee704 Could you also have a look at this change? Most of changes are from "getCandidateIndexes".

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
imback82: I think the class name seems sufficient for the separation of concerns. Can you actually tag with the reason since this PR contains the actual filter implementations? I would like to confirm whether this approach makes sense.

sezruby: @imback82 updated PR - I'll add some tests for that msg but please have a look

imback82: I took a quick look at it and the new approach looks fine to me. I will do another round of review. Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -16,24 +16,42 @@
...
-trait IndexFilter {
+trait IndexFilter extends ActiveSparkSession {
 
-  /**
-   * @return Failure reason for filtered out indexes.
-   */
-  def reason: String
+  protected def setReasonTag(condition: Boolean, reasonString: => String)(
+      implicit keyPair: (LogicalPlan, IndexLogEntry)): Unit = {
+    if (!condition && Hyperspace.getContext(spark).whyNotEnabled) {
```
sezruby, line 35: `Hyperspace.getContext(spark).whyNotEnabled` not sure this would be ok or no

sezruby: or we could mark the flag via index log entry tag 😂

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntryTags.scala
```diff
@@ -53,4 +53,8 @@ object IndexLogEntryTags {
   // for Hybrid Scan.
   val INMEMORYFILEINDEX_HYBRID_SCAN_APPENDED: IndexLogEntryTag[InMemoryFileIndex] =
   IndexLogEntryTag[InMemoryFileIndex]("inMemoryFileIndexHybridScanAppended")
+
+  // WHYNOT_REASON stores reason string for disqualification.
+  val WHYNOT_REASON: IndexLogEntryTag[String] =
+    IndexLogEntryTag[String]("whyNotReason")
```
clee704, line 59: Maybe `IndexLogEntryTag[Seq[String]]` instead of `IndexLogEntryTag[String]`?

clee704: As a side note, not specific to this PR - I don't like the IndexLogEntryTag approach because it bloats IndexLogEntry, makes it mutable, and IndexLogEntry being cached can unnecessarily keep information from earlier query executions which might not be needed again. And maybe I'm wrong, but how does it work when there are multiple concurrent query executions and they reference the same index log entry?

Instead of IndexLogEntryTag, having an execution context object containing such information specific to each execution would be better. It makes IndexLogEntry immutable and allows something that cannot be done now - like caching common information which doesn't depend on IndexLogEntry, such as source data signature.

sezruby: It would be ok since it's only tagged when whyNot flag is enabled.
And the API will be usually called from notebooks, so that might be ok if we clear the tag properly. + hyperspace object is currently ThreadLocal so it won't affect another concurrent query.

In the previous structure - 2 different rules - it's hard to share the context between rules as they are called from Spark optimizer, so we introduced IndexLogEntryTag to cache the execution context.
(fyi Spark3 support Plan https://github.com/microsoft/hyperspace/issues/185)

One good thing of IndexLogEntryTag is it can be shared between different queries having the same source plan key. 

We could introduce another execution context as you said, after this refactoring, but let's keep the index log entry tag approach for faster refactoring.

imback82 (approved):
LGTM thanks @sezruby!

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -31,11 +31,11 @@ trait IndexFilter extends ActiveSparkSession {
...
   protected def setReasonTag(condition: Boolean, reasonString: => String)(
       implicit keyPair: (LogicalPlan, IndexLogEntry)): Unit = {
-    if (!condition && Hyperspace.getContext(spark).whyNotEnabled) {
-      val (plan, index) = keyPair
+    val (plan, index) = keyPair
+    if (!condition && index.getTagValue(IndexLogEntryTags.WHYNOT_ENABLED).getOrElse(false)) {
       val prevReason =
-        index.getTagValue(plan, IndexLogEntryTags.WHYNOT_REASON).map(_ + "; ").getOrElse("")
-      index.setTagValue(plan, IndexLogEntryTags.WHYNOT_REASON, prevReason + reasonString)
+        index.getTagValue(plan, IndexLogEntryTags.WHYNOT_REASON).getOrElse(Seq.empty)
```
clee704, line 37: nit: could be Nil?

clee704 (approved):
LGTM

PR #421, opened 2021-04-16
Cross library build for Spark 2/3

clee704:
Fixes #407

### What is the context for this pull request?

 - **Tracking Issue**: #407 
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

### What changes were proposed in this pull request?

Support for Spark 3 in addition to Spark 2. The build produces three JARs for the following combinations:

- Spark 3.0 (3.0.1) / Scala 2.12
- Spark 2.4 (2.4.2) / Scala 2.12
- Spark 2.4 (2.4.2) / Scala 2.11

Note that Spark 3 doesn't support Scala 2.11.

### Does this PR introduce _any_ user-facing change?

The artifact name is changed from `hyperspace-core_<scala version>-<hyperspace version>.jar` to `hyperspace-core_spark<spark version>_<scala version>-<hyperspace version>.jar`, where `<spark version>` is either `2.4` or `3.0`.

### How was this patch tested?

Existing unit tests

PR #438, opened 2021-04-29
Update .gitignore for IntelliJ on Windows

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

.gitingore update for IntelliJ on windows after #421.
Since IntelliJ git cannot handle symlink properly, files by symlink are shown under "unversioned files":

image

No

tested with IntelliJ on windows

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,7 @@ build.log
 
 # macOS related files
 .DS_Store
+
+# for IntelliJ
+/spark2.4/src/*
+/spark3.0/src/*


clee704 (approved):
You should be able to use Git on Windows even if the symlinks are replaced, because we mark those replaced symlinks as `--assume-unchanged`. I don't like including IDE-specific things in the repo, especially when it's for a workaround that shouldn't be needed, but can't find a better way due to the lack of access to Windows machines.

imback82 (approved):
LGTM, thanks @sezruby!

PR #442, opened 2021-05-14
ApplyHyperspace rule - FilterIndexRule

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: #405 
 - **Dependencies**: #437 

NOTE: please check the last commit until #437 is merged.

Refactored FilterIndexRule based on IndexFilter. 
Renamed HPFilterIndexRule (HP = Hash Partitioned)

`ApplyHyperspace` will be substitute `JoinIndexRule` & `FilterIndexRule`  when it's delivered completely.

No

Unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
@@ -32,8 +32,7 @@ import com.microsoft.hyperspace.telemetry.HyperspaceEventLogging
  * Collect candidate indexes for each source plan.
  */
 object CandidateIndexCollector extends ActiveSparkSession {
-  private val sourceFilters
-    : Seq[SourcePlanIndexFilter] = ColumnSchemaFilter :: FileSignatureFilter :: Nil
+  private val sourceFilters = ColumnSchemaFilter :: FileSignatureFilter :: Nil
 
   private def initializePlanToIndexes(
       plan: LogicalPlan,
@@ -68,7 +67,7 @@ object CandidateIndexCollector extends ActiveSparkSession {
  */
 class ScoreBasedIndexPlanOptimizer {
   // TODO: FilterIndexRule :: JoinIndexRule :: Nil
-  private val rules: Seq[HyperspaceRule] = NoOpRule :: Nil
+  private val rules = NoOpRule :: Nil
 
   // Map for memoization. The key is the logical plan before applying [[HyperspaceRule]]s
   // and its value is a pair of best transformed plan and its score.
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule_disabled.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule_disabled.scala
@@ -119,8 +119,8 @@ object FilterRankFilter extends IndexRankFilter {
  * filter predicate.
  */
 object FilterIndexRule_disabled extends HyperspaceRule {
-  override val filtersOnQueryPlan
-    : Seq[QueryPlanIndexFilter] = FilterPlanNodeFilter :: FilterColumnFilter :: Nil
+  override val filtersOnQueryPlan: Seq[QueryPlanIndexFilter] =
+    FilterPlanNodeFilter :: FilterColumnFilter :: Nil
 
   override val indexRanker: IndexRankFilter = FilterRankFilter
 
@@ -153,14 +153,15 @@ object FilterIndexRule_disabled extends HyperspaceRule {
       .getOrElse {
         relation.allFileInfos.foldLeft(0L) { (res, f) =>
           if (candidateIndex.sourceFileInfoSet.contains(f)) {
-            (res + f.size) // count, total bytes
+            res + f.size // count, total bytes
           } else {
             res
           }
         }
       }
 
-    // TODO enhance scoring
-    (50 * (commonBytes * 1.0f / relation.allFileSizeInBytes)).round
+    // TODO: Enhance scoring function.
+    //   See https://github.com/microsoft/hyperspace/issues/444
+    (50 * (commonBytes.toFloat / relation.allFileSizeInBytes)).round
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule_disabled.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule_disabled.scala
@@ -1,0 +1,166 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.index.rules
+
+import org.apache.spark.sql.catalyst.analysis.CleanupAliases
+import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Expression}
+import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan, Project}
+
+import com.microsoft.hyperspace.index.IndexLogEntryTags
+import com.microsoft.hyperspace.index.rankers.FilterIndexRanker
+import com.microsoft.hyperspace.index.rules.ApplyHyperspace.{PlanToIndexesMap, PlanToSelectedIndexMap}
+import com.microsoft.hyperspace.util.{HyperspaceConf, ResolverUtils}
+
+object FilterPlanNodeFilter extends QueryPlanIndexFilter {
+  override def apply(plan: LogicalPlan, indexes: PlanToIndexesMap): PlanToIndexesMap = {
+    if (indexes.isEmpty) {
+      return Map.empty
+    }
+
+    // FilterPlanNodeFilter looks for below patterns, in ordered manner:
+    //  Pattern-1: Scan -> Filter -> Project
+    //  Pattern-2: Scan -> Filter
+    // Pattern-2 covers the case where project node is eliminated or not present.
+    // An example is when all columns are selected.
+    plan match {
+      case Project(_, Filter(_: Expression, ExtractRelation(relation)))
+          if !RuleUtils.isIndexApplied(relation) =>
+        indexes.filterKeys(relation.plan.equals(_))
+      case Filter(_: Expression, ExtractRelation(relation))
+          if !RuleUtils.isIndexApplied(relation) =>
+        indexes.filterKeys(relation.plan.equals(_))
+      case _ =>
+        Map.empty
+    }
+  }
+}
+
+object FilterColumnFilter extends QueryPlanIndexFilter {
+  override def apply(plan: LogicalPlan, indexes: PlanToIndexesMap): PlanToIndexesMap = {
+    if (indexes.isEmpty) {
+      return Map.empty
+    }
+
+    val (filterColumnNames, projectColumnNames) = plan match {
+      case project @ Project(_, _ @Filter(condition: Expression, ExtractRelation(relation)))
+          if !RuleUtils.isIndexApplied(relation) =>
+        val projectColumnNames = CleanupAliases(project)
+          .asInstanceOf[Project]
+          .projectList
+          .map(_.references.map(_.asInstanceOf[AttributeReference].name))
+          .flatMap(_.toSeq)
+        val filterColumnNames = condition.references.map(_.name).toSeq
+        (filterColumnNames, projectColumnNames)
+
+      case Filter(condition: Expression, ExtractRelation(relation))
+          if !RuleUtils.isIndexApplied(relation) =>
+        val relationColumnNames = relation.plan.output.map(_.name)
+        val filterColumnNames = condition.references.map(_.name).toSeq
+        (filterColumnNames, relationColumnNames)
+      case _ =>
+        (Seq(), Seq())
+    }
+
+    // Filter candidate indexes if:
+    //  1. Filter predicate's columns include the first 'indexed' column of the index.
+    //  2. The index covers all columns from the filter predicate and output columns list.
+    val candidateIndexes =
+      indexes.head._2.filter { index =>
+        val ddColumns = index.derivedDataset.properties.columns
+        withReasonTag("The first indexed column should be in filter columns.") {
+          ResolverUtils.resolve(spark, ddColumns.indexed.head, filterColumnNames).isDefined
+        }(plan, index) &&
+        withReasonTag(
+          "Index does not contain required columns. Required columns: " +
+            s"[${filterColumnNames ++ projectColumnNames}], indexed & included columns: " +
+            s"[${ddColumns.indexed ++ ddColumns.included}") {
+          ResolverUtils
+            .resolve(
+              spark,
+              filterColumnNames ++ projectColumnNames,
+              ddColumns.indexed ++ ddColumns.included)
+            .isDefined
+        }(plan, index)
+      }
+
+    Map(indexes.head._1 -> candidateIndexes)
+  }
+}
+
+object FilterRankFilter extends IndexRankFilter {
+  override def apply(plan: LogicalPlan, indexes: PlanToIndexesMap): PlanToSelectedIndexMap = {
+    if (indexes.isEmpty || indexes.size != 1 || indexes.head._2.isEmpty) {
+      Map.empty
+    } else {
+      val selected = FilterIndexRanker.rank(spark, plan, indexes.head._2).get
+      setRankReasonTag(indexes.head._1, indexes.head._2, selected)
+      Map(indexes.head._1 -> selected)
+    }
+  }
+}
+
+/**
+ * FilterIndexRule looks for opportunities in a logical plan to replace
+ * a relation with an available hash partitioned index according to columns in
+ * filter predicate.
+ */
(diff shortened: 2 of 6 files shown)


src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
```diff
@@ -274,7 +274,6 @@ class RuleUtilsTest extends HyperspaceRuleSuite with SQLHelper {
 
       withIndex("index1") {
         val readDf = spark.read.parquet(dataPath)
-        val expectedCommonSourceBytes = FileUtils.getDirectorySize(new Path(dataPath))
```
andrei-ionescu, line 277: Any reason why this was removed?

andrei-ionescu (commented):
Can you specify which parts of this PR is just code being moved from one place to another? Thanks!

src/main/scala/com/microsoft/hyperspace/index/rules/HPFilterIndexRule.scala
```diff
@@ -0,0 +1,166 @@
...
+    } else {
+      val selected = FilterIndexRanker.rank(spark, plan, indexes.head._2).get
+      setRankReasonTag(indexes.head._1, indexes.head._2, selected)
+      Map(indexes.head._1 -> selected)
+    }
+  }
+}
+
+/**
+ * HPFilterIndexRule looks for opportunities in a logical plan to replace
```
clee704, line 117: Do we have any plan to support covering indexes that are not hash partitioned? I don't see any hash partitioning specific code here or in RuleUtils, but covering index related code. It seems the code don't care about whether the index data is hash partitioned or range partitioned (if such a thing exists), as long as it is laid out efficiently (clustered) regarding the key columns. On the other hand, it won't work for non-covering indexes, as just replacing the source data with index data only works for covering indexes. Therefore naming this Hash Partitioned but not Covering seems misleading.

Actually, I prefer just `FilterIndexRule` because it is the only rule for filters at the moment, and we don't know yet how we will support more types in the future. My second preference would be `CoveringFilterIndexRule`, if you think the index type should be in the name. Lastly, `HashPartitionedCoveringFilterIndexRule` if that information about partitioning is really that important (to me it seems not). Also, I'm against using uncommon acronyms such as HP for naming things.

sezruby: For other index type, we need to add another rule as it might have a different filter condition for the index & different application / score function. We can add index type filter at first.  

> Actually, I prefer just FilterIndexRule because it is the only rule for filters at the moment, and we don't know yet how we will support more types in the future.

We will add it soon and your data-skipping index also requires a new rule as it has different application & score functions.
But as we discussed, it would be better to rename CoveringIndex related things in anoter change. I'll change the name to FilterIndexRule_disabled

src/main/scala/com/microsoft/hyperspace/index/rules/HPFilterIndexRule.scala
```diff
@@ -0,0 +1,166 @@
...
+          if (candidateIndex.sourceFileInfoSet.contains(f)) {
+            (res + f.size) // count, total bytes
+          } else {
+            res
+          }
+        }
+      }
+
+    // TODO enhance scoring
+    (50 * (commonBytes * 1.0f / relation.allFileSizeInBytes)).round
```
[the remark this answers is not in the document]
sezruby: 1. I'll set left child `70` + right child `70` for JoinIndexRule so that it has a higher priority than FilterIndexRule
2. `1.0f` is just for float calculation and one coding habit from my C++ impl. I will try to use toFloat.

src/main/scala/com/microsoft/hyperspace/index/rules/HPFilterIndexRule.scala
```diff
@@ -0,0 +1,166 @@
...
+        relation.allFileInfos.foldLeft(0L) { (res, f) =>
+          if (candidateIndex.sourceFileInfoSet.contains(f)) {
+            (res + f.size) // count, total bytes
+          } else {
+            res
+          }
+        }
+      }
+
+    // TODO enhance scoring
```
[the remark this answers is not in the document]
sezruby: I don't have specific plan for this yet. I'll open an issue for it and maintain the document for scoring.
But for the initial version, we can start from these simple scoring functions.

src/test/scala/com/microsoft/hyperspace/index/rules/RuleUtilsTest.scala
sezruby: it's not used in the test

src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
```diff
@@ -32,8 +32,8 @@ import com.microsoft.hyperspace.telemetry.HyperspaceEventLogging
  * Collect candidate indexes for each source plan.
  */
 object CandidateIndexCollector extends ActiveSparkSession {
-  // TODO: ColumnSchemaFilter :: FileSignatureFilter :: Nil
-  private val sourceFilters: Seq[SourcePlanIndexFilter] = Nil
+  private val sourceFilters
+    : Seq[SourcePlanIndexFilter] = ColumnSchemaFilter :: FileSignatureFilter :: Nil
```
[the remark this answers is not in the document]
sezruby: Seems it's from scalafmt. why do you think it's wrong?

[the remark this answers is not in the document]
andrei-ionescu: My `scalafmt` outputs it differently, something like this:

```
  private val sourceFilters: Seq[SourcePlanIndexFilter] = 
    ColumnSchemaFilter :: FileSignatureFilter :: Nil
```

BTW: The variable is private so you can remove the type. You'll have more free space.

src/main/scala/com/microsoft/hyperspace/index/rules/HPFilterIndexRule.scala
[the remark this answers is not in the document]
andrei-ionescu: 👍 Please add the ticket number to this TODO.

src/main/scala/com/microsoft/hyperspace/index/rules/HPFilterIndexRule.scala
clee704: > For other index type, we need to add another rule as it might have a different filter condition for the index & different application / score function. We can add index type filter at first.

Another possible design is abstracting index functionality into an interface and making the rule work on the interface. I think both designs have their pros and cons.

By the way, now I understand why you named it differently. I think if names collide then it is better to put new things in a new package and rename/move later at once when new things replace old things. It might confuse developers if old and new things, seemingly doing the same thing, are in the same place, although this is only temporary. But I'm okay with either way.

sezruby: Tried it but reverted since I needed to import all things in `rules` package.

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule_disabled.scala
```diff
@@ -0,0 +1,167 @@
...
+    if (indexes.isEmpty) {
+      return Map.empty
+    }
+
+    // FilterPlanNodeFilter looks for below patterns, in ordered manner:
+    //  Pattern-1: Scan -> Filter -> Project
+    //  Pattern-2: Scan -> Filter
+    // Pattern-2 covers the case where project node is eliminated or not present.
+    // An example is when all columns are selected.
+    plan match {
```
sezruby, line 39: This is from `ExtractFilterNode`

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule_disabled.scala
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+object FilterColumnFilter extends QueryPlanIndexFilter {
+  override def apply(plan: LogicalPlan, indexes: PlanToIndexesMap): PlanToIndexesMap = {
+    if (indexes.isEmpty) {
+      return Map.empty
+    }
+
+    val (filterColumnNames, projectColumnNames) = plan match {
```
sezruby, line 58: from `ExtractFilterNode`

src/main/scala/com/microsoft/hyperspace/index/rules/FilterIndexRule_disabled.scala
```diff
@@ -0,0 +1,167 @@
...
+        (Seq(), Seq())
+    }
+
+    // Filter candidate indexes if:
+    //  1. Filter predicate's columns include the first 'indexed' column of the index.
+    //  2. The index covers all columns from the filter predicate and output columns list.
+    val candidateIndexes =
+      indexes.head._2.filter { index =>
+        val ddColumns = index.derivedDataset.properties.columns
+        withReasonTag("The first indexed column should be in filter columns.") {
```
sezruby, line 84: from `FilterIndexRule.indexCoversplan`

clee704 (approved):
LGTM.

By the way, can you share your thoughts and plans for supporting multiple index types? As I said before, it can be done either by 1) general framework-level rules working on indexes through the Index API, thus keeping the Index API as the single interface that implementations should care about, or 2) index-specific rules for each index type.

With option 1), we can carefully choose what to expose and what to delegate to index implementations so that our rules can work without knowing which indexes they are working with. We would have full control over the rules as they are owned by the framework and not an extension point in this design. It will be easier to implement new index types, although some novel index types might require changes in the Index API and the rules.

With option 2), rules become an extension point, and index implementations have to implement their own rules. It will give more freedom to implementations, but we should carefully design the rule API to avoid delegating too much. Otherwise, we could end up with multiple implementations doing the same things in their own ways. Also, we have to design how we should select and apply multiple index-specific rules.

Initially, I was thinking of option 1) as the way to go, but after some thought, option 2) seems good too. It seems harder to do it right, though.

sezruby:
@clee704  I think we could reuse the code at "IndexFilter" level if needed. Separating rules and indexes would be better in terms of 
1) query plan conditions might be different depending on index type so it's hard to define general framework-level interface
2) avoid high complexity / having too much roles in Index

But I think we could add an interface "applyIndex" and move `transformPlanToUseIndex` function to CoveringIndex trait, 
so that all Covering index type can share the implementation.

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -29,16 +29,33 @@ import com.microsoft.hyperspace.util.{HyperspaceConf, ResolverUtils}
...
+   * Append a given reason string to WHYNOT_REASON tag of the index
+   * if the result of the given function is false.
+   *
+   * @param reasonString Informational message in case condition is false.
+   * @param f Function for a condition
+   * @param keyPair A pair of plan and index to tag
+   * @return Result of the given function
+   */
   protected def withReasonTag(reasonString: => String)(f: => Boolean)(
       implicit keyPair: (LogicalPlan, IndexLogEntry)): Boolean = {
```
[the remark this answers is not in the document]
imback82: If you prefer to put the pair at the end, it's fine as long as we are explicit about it.

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -90,6 +107,24 @@ trait IndexRankFilter extends IndexFilter {
...
+   * @param selectedIndex Selected index
+   */
+  protected def setRankReasonTag(
+      plan: LogicalPlan,
+      indexes: Seq[IndexLogEntry],
+      selectedIndex: IndexLogEntry): Unit = {
+    indexes.foreach { index =>
+      setReasonTag(
+        selectedIndex.name.equals(index.name),
+        s"Another candidate index is applied: ${selectedIndex.name}")(plan, index)
```
imback82, line 125: You can remove `s` in line 183.

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -29,16 +29,33 @@ import com.microsoft.hyperspace.util.{HyperspaceConf, ResolverUtils}
...
+  /**
+   * Append a given reason string to WHYNOT_REASON tag of the index if the condition is false and
+   * WHYNOT_ENABLED tag is set to the index.
+   *
+   * @param condition Flag for reason string
+   * @param reasonString Informational message in case condition is false.
+   * @param keyPair A pair of plan and index to tag
+   */
   protected def setReasonTag(condition: Boolean, reasonString: => String)(
       implicit keyPair: (LogicalPlan, IndexLogEntry)): Unit = {
```
imback82, line 41: ditto on `implicit` here, and I don't think we have a caller using `implicit` yet.

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -90,6 +107,24 @@ trait IndexRankFilter extends IndexFilter {
...
   def apply(plan: LogicalPlan, applicableIndexes: PlanToIndexesMap): PlanToSelectedIndexMap
+
+  /**
+   * Set WHYNOT_REASON tag for unselected indexes.
+   *
+   * @param plan Plan to tag
+   * @param indexes Indexes to tag
+   * @param selectedIndex Selected index
+   */
+  protected def setRankReasonTag(
```
imback82, line 118: function name sounds a bit awkward (sounds like you are setting something related to "rank")

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
```diff
@@ -29,16 +29,33 @@ import com.microsoft.hyperspace.util.{HyperspaceConf, ResolverUtils}
 
 trait IndexFilter extends ActiveSparkSession {
 
+  /**
+   * Append a given reason string to WHYNOT_REASON tag of the index if the condition is false and
```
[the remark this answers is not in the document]
imback82: Also, doesn't it make more sense to have `withFilterReasonTag`, `setFilterReasonTag` instead of `setReasonTag` and `withReasonTag`?

src/main/scala/com/microsoft/hyperspace/index/rules/disabled/FilterIndexRule.scala
```diff
@@ -0,0 +1,169 @@
...
+import org.apache.spark.sql.catalyst.analysis.CleanupAliases
+import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Expression}
+import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan, Project}
+
+import com.microsoft.hyperspace.index.IndexLogEntryTags
+import com.microsoft.hyperspace.index.rankers.FilterIndexRanker
+import com.microsoft.hyperspace.index.rules._
+import com.microsoft.hyperspace.index.rules.ApplyHyperspace.{PlanToIndexesMap, PlanToSelectedIndexMap}
+import com.microsoft.hyperspace.util.{HyperspaceConf, ResolverUtils}
+
```
imback82, line 28: remove extra line

src/main/scala/com/microsoft/hyperspace/index/rules/disabled/FilterIndexRule.scala
```diff
@@ -0,0 +1,169 @@
...
+
+object FilterPlanNodeFilter extends QueryPlanIndexFilter {
```
imback82, line 30: Shall we add a description for each filter on what it does?

src/main/scala/com/microsoft/hyperspace/index/rules/disabled/FilterIndexRule.scala
```diff
@@ -0,0 +1,169 @@
...
+          ResolverUtils
+            .resolve(
+              spark,
+              filterColumnNames ++ projectColumnNames,
+              ddColumns.indexed ++ ddColumns.included)
+            .isDefined
+        }(plan, index)
+      }
+
+    Map(indexes.head._1 -> candidateIndexes)
```
imback82, line 102: We have `indexes.head._2` above. Can we do:
```scala
// Consider only the first index because ...
val index = indexes.head
```

src/main/scala/com/microsoft/hyperspace/index/rules/disabled/FilterIndexRule.scala
```diff
@@ -0,0 +1,169 @@
...
+    val candidateIndexes =
+      indexes.head._2.filter { index =>
+        val ddColumns = index.derivedDataset.properties.columns
+        withReasonTag("The first indexed column should be in filter condition columns.") {
+          ResolverUtils.resolve(spark, ddColumns.indexed.head, filterColumnNames).isDefined
+        }(plan, index) &&
+        withReasonTag(
+          "Index does not contain required columns. Required columns: " +
+            s"[${filterColumnNames ++ projectColumnNames}], indexed & included columns: " +
+            s"[${ddColumns.indexed ++ ddColumns.included}") {
```
imback82, line 92: missing closing `]`?, and maybe missing test?

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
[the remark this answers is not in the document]
sezruby: Since I thought specifying (plan, index) for all calls looks a bit verbose, especially there are many withReasonTag in the same function.  So I tried to use implicit.. but I'm fine with the explicit approach

src/main/scala/com/microsoft/hyperspace/index/rules/IndexFilter.scala
sezruby: renamed `setUnselectedReasonTagForRank`

src/main/scala/com/microsoft/hyperspace/index/rules/disabled/FilterIndexRule.scala
sezruby: renamed `indexes: PlanToIndexesMap` => `pti: PlanToIndexes`

PR #453, opened 2021-06-03
Remove throwIfNotExists param in fromDirectory util

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Remove exception type specific code.

Yes, fix the bug when lower api throws non-"FileNotFoundException" exception if the path does not exist

Covered by existing tests

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -100,12 +100,7 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
               IndexLogEntry.schemaString(indexDataFrame.schema),
               numBuckets,
               coveringIndexProperties)),
-          if (absolutePath.getFileSystem(hadoopConf).exists(absolutePath)) {
-            Content.fromDirectory(absolutePath, fileIdTracker, hadoopConf)
-          } else {
-            // Create empty Content if the index content path does not exist.
-            Content(Directory.createEmptyDirectory(absolutePath))
-          },
+          Content.fromDirectory(absolutePath, fileIdTracker, hadoopConf),
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
--- a/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
+++ b/src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
@@ -100,7 +100,12 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager) 
               IndexLogEntry.schemaString(indexDataFrame.schema),
               numBuckets,
               coveringIndexProperties)),
-          Content.fromDirectory(absolutePath, fileIdTracker, hadoopConf),
+          if (absolutePath.getFileSystem(hadoopConf).exists(absolutePath)) {
+            Content.fromDirectory(absolutePath, fileIdTracker, hadoopConf)
+          } else {
+            // Create empty Content if the index content path does not exist.
+            Content(Directory.createEmptyDirectory(absolutePath))
+          },
           Source(SparkPlan(sourcePlanProperties)),
           Map())
 
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
@@ -292,10 +292,10 @@ object Directory {
       path: Path,
       pathFilter: PathFilter,
       fs: FileSystem): Seq[FileStatus] = {
-      val (files, directories) = fs.listStatus(path).partition(_.isFile)
-      // TODO: explore fs.listFiles(recursive = true) for better performance of file listing.
-      files.filter(s => pathFilter.accept(s.getPath)) ++
-        directories.flatMap(d => listLeafFiles(d.getPath, pathFilter, fs))
+    val (files, directories) = fs.listStatus(path).partition(_.isFile)
+    // TODO: explore fs.listFiles(recursive = true) for better performance of file listing.
+    files.filter(s => pathFilter.accept(s.getPath)) ++
+      directories.flatMap(d => listLeafFiles(d.getPath, pathFilter, fs))
   }
 }
 
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
@@ -80,7 +80,7 @@ object Content {
    *                   codebase, which filters out files like _SUCCESS.
    * @return Content object with Directory tree starting at root, and containing all leaf files
    *         from "path" argument. If the given path does not exist, return Content object with
-   *         empty Directory object represents the path.
+   *         empty Directory object that represents the path.
    */
   def fromDirectory(
       path: Path,
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
@@ -16,8 +16,6 @@
 
 package com.microsoft.hyperspace.index
 
-import java.io.FileNotFoundException
-
 import scala.annotation.tailrec
 import scala.collection.mutable.{HashMap, ListBuffer}
 import scala.collection.mutable
@@ -80,8 +78,6 @@ object Content {
    * @param hadoopConfiguration Hadoop configuration.
    * @param pathFilter Filter for accepting paths. The default filter is picked from spark
    *                   codebase, which filters out files like _SUCCESS.
-   * @param throwIfNotExists Throws FileNotFoundException if path is not found. Else creates a
-   *                         blank directory tree with no files.
    * @return Content object with Directory tree starting at root, and containing all leaf files
    *         from "path" argument.
    */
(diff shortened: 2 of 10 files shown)


src/main/scala/com/microsoft/hyperspace/actions/CreateActionBase.scala
```diff
@@ -100,7 +100,12 @@ private[actions] abstract class CreateActionBase(dataManager: IndexDataManager)
               IndexLogEntry.schemaString(indexDataFrame.schema),
               numBuckets,
               coveringIndexProperties)),
-          Content.fromDirectory(absolutePath, fileIdTracker, hadoopConf),
+          if (absolutePath.getFileSystem(hadoopConf).exists(absolutePath)) {
+            Content.fromDirectory(absolutePath, fileIdTracker, hadoopConf)
+          } else {
+            // Create empty Content if the index content path does not exist.
+            Content(Directory.createEmptyDirectory(absolutePath))
```
imback82, line 107: Maybe just move this logic into `Content.fromDirectory`, and update the comment for the function about the behavior when the path doesn't exist?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -80,19 +78,20 @@ object Content {
    * @param hadoopConfiguration Hadoop configuration.
    * @param pathFilter Filter for accepting paths. The default filter is picked from spark
    *                   codebase, which filters out files like _SUCCESS.
-   * @param throwIfNotExists Throws FileNotFoundException if path is not found. Else creates a
-   *                         blank directory tree with no files.
    * @return Content object with Directory tree starting at root, and containing all leaf files
-   *         from "path" argument.
+   *         from "path" argument. If the given path does not exist, return empty Directory.
```
imback82, line 82: you are returning Content with empty directory?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -211,7 +203,9 @@ object Directory {
   }
 
   @tailrec
-  private def createEmptyDirectory(path: Path, subDirs: Seq[Directory] = Seq()): Directory = {
+  private[hyperspace] def createEmptyDirectory(
```
imback82, line 206: Can this be reverted now?

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -362,33 +362,10 @@ class IndexLogEntryTest extends HyperspaceSuite with SQLHelper {
     assert(directoryEquals(actual, expected))
   }
 
-  test("Directory.fromLeafFiles: throwIfNotExist flag throws exception for non-existent" +
```
imback82, line 365: Don't we still need a test case with nonexistent directory?

sezruby: I think below `Directory.fromDirectory where the directory is empty.` is enough

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: It's still required since this is Directory object

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: revised

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -80,19 +78,21 @@ object Content {
...
    * @param pathFilter Filter for accepting paths. The default filter is picked from spark
    *                   codebase, which filters out files like _SUCCESS.
-   * @param throwIfNotExists Throws FileNotFoundException if path is not found. Else creates a
-   *                         blank directory tree with no files.
    * @return Content object with Directory tree starting at root, and containing all leaf files
-   *         from "path" argument.
+   *         from "path" argument. If the given path does not exist, return Content object with
+   *         empty Directory object represents the path.
    */
   def fromDirectory(
```
imback82, line 85: This is also called from `OptimizeAction`. Should we fail fast if the directory doesn't exist in that case? I think your original approach would handle it? (or we can add `failIfPathNotExists`)

sezruby: I think both ways are almost same in terms of call depth ?

imback82: Maintainability wise, it seems better to handle here, but I will leave it to you.

sezruby: OptimizeAction also needs to ignore the error for its begin log - some tests failed without it and I added the fix after that - (https://github.com/microsoft/hyperspace/pull/453/commits/bcdfe9226ca19da1090690031aec490184169bfd).

I think it's okay as long as this is internal use only.

imback82: But it seems weird that optimize() would create an empty directory, doesn't it?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -80,19 +78,21 @@ object Content {
    * @param hadoopConfiguration Hadoop configuration.
    * @param pathFilter Filter for accepting paths. The default filter is picked from spark
    *                   codebase, which filters out files like _SUCCESS.
-   * @param throwIfNotExists Throws FileNotFoundException if path is not found. Else creates a
-   *                         blank directory tree with no files.
    * @return Content object with Directory tree starting at root, and containing all leaf files
-   *         from "path" argument.
+   *         from "path" argument. If the given path does not exist, return Content object with
+   *         empty Directory object represents the path.
```
imback82, line 83:
```suggestion
   *         empty Directory object that represents the path.
```

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -362,40 +362,26 @@ class IndexLogEntryTest extends HyperspaceSuite with SQLHelper {
...
-    val expected = {
-      val nonExistentDirDirectory = Directory(nonExistentDir.getName)
-      createDirectory(nonExistentDir, nonExistentDirDirectory)
-    }
-
-    val actual = Directory.fromDirectory(nonExistentDir, fileIdTracker, throwIfNotExists = false)
-    assert(directoryEquals(actual, expected))
-  }
-
   test("Directory.fromDirectory where the directory is empty.") {
```
imback82, line 365:
```suggestion
  test("Directory.fromDirectory where the directory is empty or nonexistent.") {
```

imback82 (commented):
LGTM, thanks @sezruby!

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -80,19 +78,21 @@ object Content {
    * @param hadoopConfiguration Hadoop configuration.
    * @param pathFilter Filter for accepting paths. The default filter is picked from spark
    *                   codebase, which filters out files like _SUCCESS.
-   * @param throwIfNotExists Throws FileNotFoundException if path is not found. Else creates a
-   *                         blank directory tree with no files.
    * @return Content object with Directory tree starting at root, and containing all leaf files
-   *         from "path" argument.
+   *         from "path" argument. If the given path does not exist, return Content object with
+   *         empty Directory object that represents the path.
```
clee704, line 83: Nit: Although it's easy to guess, the wording here isn't clear about whether the empty directory will be created on the file system. The method name "fromDirectory" sounds like it's a read-only action, but now it can modify the file system and it could be surprising to developers.

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -362,40 +362,26 @@ class IndexLogEntryTest extends HyperspaceSuite with SQLHelper {
...
     val expected = {
       val emptyDirDirectory = Directory(emptyDirPath.getName)
       createDirectory(emptyDirPath, emptyDirDirectory)
     }
 
-    val actual = Directory.fromDirectory(emptyDirPath, fileIdTracker)
-    assert(directoryEquals(actual, expected))
+    {
+      // Test non-existent directory.
+      val actual = Directory.fromDirectory(emptyDirPath, fileIdTracker)
```
clee704, line 375: Is the directory really non-existent? It seems it is created at line 369.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: It's still read-only function; just creating a `Directory` object, not a real directory.

src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
sezruby: ditto it's also `Directory` object

clee704 (approved):
LGTM, thanks!

PR #455, opened 2021-06-05
Upgrade sbt-pgp to 2.1.2

imback82:

Upgrade `sbt-pgp` to 2.1.2 so that GPG signing works correctly.

No

Manually tested by uploading SNAPSHOT JARs.


imback82 (commented):
Merging to unblock myself, thanks!

PR #443, opened 2021-05-17
Refactoring for an extensible Index API

clee704:
### What is the context for this pull request?

 - **Tracking Issue**: #441
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

### What changes were proposed in this pull request?

- Introduce common interfaces for indexes with which Hyperspace can
  manage various types of indexes.
- Adjust IndexStatistics so that implementation-specific fields can be
  added. For instance, included columns are one of such fields now.
- Actions work with generic indexes, not just covering indexes which are
  the only type we support at the moment.
- Existing rules only work with covering indexes. New rules will be
  added along with new index types.

### Does this PR introduce _any_ user-facing change?

- Serialization format of CoveringIndex, and thus IndexLogEntry, is changed.
- The format of IndexStatistics is changed. It means the format of the
  dataframe returned by Hyperspace.indexes is also changed.
- The user interface for users to create/manage indexes is not changed.

### How was this patch tested?

With existing unit tests

PR #456, opened 2021-06-05
[WIP] Introduce conditional compilation macros

clee704:

 - **Tracking Issue**: N/A
 - **Parent Issue**: N/A
 - **Dependencies**: #443

Introduce conditional compilation macros for Spark-version-dependent compilation. Currently, we are relying on Spark-version-specific source directories and they work nicely, but sometimes it's not easy to extract differences into separate files without duplicating a lot of code, introducing performance overhead, and/or impacting the code readability.

For example, in #443, I added annotations for `StructType` serialization/deserialization because without them it doesn't work in Spark 2.4. However, those annotations do unnecessary extra things for Spark 3+. There are ways to handle the issue with version-specific source directories, but using macros is the cleanest: no duplication, no runtime overhead, and readable.

No

Unit test (see [MacroTest.scala](src/test/scala/com/microsoft/hyperspace/shim/MacroTest.scala))


src/main/scala/com/microsoft/hyperspace/index/CoveringIndex.scala
```diff
@@ -0,0 +1,283 @@
...
+import com.microsoft.hyperspace.index.DataFrameWriterExtensions.Bucketizer
+import com.microsoft.hyperspace.shim.{StructTypeConverter, UntilSpark24}
+import com.microsoft.hyperspace.util.ResolverUtils
+import com.microsoft.hyperspace.util.ResolverUtils.ResolvedColumn
+
+case class CoveringIndex(
+    override val indexedColumns: Seq[String],
+    includedColumns: Seq[String],
+    // See schemaJson for more information about the annotation.
+    @JsonDeserialize(converter = classOf[StructTypeConverter.T]) schema: StructType,
```
imback82, line 37: What's the downside of just storing `schemaJson: String` as before? We can validate the given string in `require` if that was the concern. I may have missed other concerns.

clee704: This PR is a follow-up of #443 and this part belongs to that. Please ignore other commits and see the commit 7e0a6693c0618091d56b0485ae05ed8b9494f31c for the changes of this PR.

I think the downsides are as follows:

1. You have to convert the string to `StructType` every time you need it. You can make a lazy val and `@JsonIgnore` it but then it consumes more memory. These overheads might be slim, but why pay more when you don't have to?
2. Escaped JSON strings are ugly and less readable than normal JSON strings. Could be a minor nuisance when debugging.

imback82: > 1. You have to convert the string to `StructType` every time you need it. You can make a lazy val and `@JsonIgnore` it but then it consumes more memory. These overheads might be slim, but why pay more when you don't have to?

How much memory are we talking here? :) I guess we are paying for the unnecessary macros-related code? In #443, it seems we can just use `schemaString` and not worry about separating 2.4 vs. 3.0 in the first place, no?



> 2\. Escaped JSON strings are ugly and less readable than normal JSON strings. Could be a minor nuisance when debugging.

Seems minor to me. There are many unescaping tools available online if readability is really an issue while debugging.

imback82: I will leave it up to you and @sezruby on the final decision. Just checking to see if shim was necessary for this field.

sezruby: Seems code redundancy & complexity for StructTypeConverter outweigh the gain. I think we could introduce these macro when it's really required for expensive conversion/shim (e.g. data file paths?) later.

clee704: Why use tools when you don't have to? :)

These macros are general utilities. If it has any cost, it is amortized as we use it more.

I'm fine delivering this when more needs arise. The current usage is not strictly necessary; it's just a demonstration.

imback82: > Why use tools when you don't have to? :)

Because it is brining in unnecessary complexity to the code base.

clee704: First, it might be "unnecessary" for now but might be necessary in the future. I agree that we can wait until a more strong need arises. Whether something is necessary or unnecessary is subjective and I think it can be necessary.

Second, about complexity - having one more subproject can feel complex for developers unfamiliar with multi-projects SBT builds, but once it's set up, you don't need to know it's there. The macro code may be complex, but the complexity is contained - you don't have to understand the macro code in order to understand the codebase. You can ignore the macro subproject entirely and don't have to use macros for your daily Hyperspace coding.

clee704:
Will revisit later

Issue #441, opened 2021-05-13
[PROPOSAL]: Data Skipping Indexes


clee704:
## Problem Statement

Add support for data skipping indexes.

## Background and Motivation

Hyperspace has been supporting hash-partitioned covering indexes only. Covering indexes are good for some workloads, but we should have more index types to facilitate a broader range of use cases.

## Proposed Solution

Refactor existing code by moving covering index specific code into a single implementation for a new `Index` API, and implement data skipping indexes for the API.

## Alternatives

The proposed solution requires heavy refactoring, as there is a lot of code specific to covering indexes scattered throughout the codebase. Alternatively, we can just make `DataSkippingIndex` without introducing a new API and write many if-elses depending on the type of the index. It might be quick to implement the required features for now in this way, but it will make the code hard to maintain in the long run.

## Known/Potential Compatibility Issues

Regarding data compatibility, it is very likely that this will make the new version of Hyperspace incompatible with indexes created in the older versions. This is unfortunate but can be excused as Hyperspace is still in version zero. If a strong need arises, this can be mitigated by a migration utility which can be designed separately.

The API compatibility can be retained, if we introduce new optional arguments to `IndexConfig`. However, as `includedColumns` doesn't make sense for data skipping indexes, and newly added arguments won't either for covering indexes. Therefore, it might be wise to break compatibility now and make things clean for the future.

### Data Skipping Index Basics

Data skipping indexes can be used to filter out certain files that cannot have the desired values in indexed columns based on statistics or sketches on the values in files. Examples of statistics or sketches include min/max values and bloom filters.

Suppose that we have sample data like below. There is a relation `df` which has two columns, `a` and `b`, and consists of two files, `/data/p0` and `/data/p1`:

`/data/p0`:
| a | b |
|--:|--:|
| 1 | 2 |
| 2 | 4 |
| 6 | 6 |

`/data/p1`:
|  a |  b |
|---:|---:|
|  5 | 10 |
| 10 | 10 |

A min-max data skipping index on a column `a` will look like this:

| path     | min | max |
|:---------|----:|----:|
| /data/p0 |   1 |   6 |
| /data/p1 |   5 |  10 |

Then, when a query like `df.filter($"a" < 4)` is given, we know that we can skip the `/data/p1` file.

### Storing Index Data

The index data can be stored as parquet files like it is done for covering indexes. The difference is here we don't need bucketing. Also, the data size is a lot less because there is only a single row for each file.

[...]

PR #461, opened 2021-06-10
Data Skipping Index Part 2: Basics

clee704:
This PR has dependency on other PRs - if you want to review this PR only, please look at the last commit.

 - **Tracking Issue**: #441 
 - **Parent Issue**: N/A
 - **Dependencies**:
   - #485 
   - #481 

Implement DataSkippingIndex and MinMaxSketch.

The index data looks like this:

| _data_file_id  | min_A_ | max_A_ |
| - | - | - |
| 146 | 1834 | 1845 |
| 763 | 9585 | 9596 |
| 351 | 4409 | 4421 |
| 148 | 1859 | 1870 |
| 758 | 9522 | 9534 |

The following data is index creation times of CoveringIndex and DataSkippingIndex for varying numbers of files for the same dataset (TPC-H 1TB). Index configs are chosen so that they will be useful for TPC-H Q6. Measurements are done on Azure Synapse with 199 executors.

Index configurations:
- CoveringIndex (indexed: l_shipdate, l_discount, l_quantity, included: l_extendedprice)
- DataSkippingIndex (MinMaxSketch/ValueListSketch on l_shipdate/l_discount/l_quantity)

| Index type | 1K files | 10K files | 100K files |
| - | - | - | - |
| CoveringIndex | 3m 5s | 3m 12s | 4m 54s |
| DataSkippingIndex | 27s | 54s | 2m 24s |

Kind of. Users can create data skipping indexes by `hs.createIndex(df, DataSkippingIndexConfig(...))`. However, they will not be utilized because there are no rules for them.

With unit tests


clee704:
TODO: check if existing rules work in the presence of unsupported indexes... (they should, as if there are no such indexes at all)

sezruby:
> TODO: check if existing rules work in the presence of unsupported indexes... (they should, as if there are no such indexes at all)

I think we should add `IndexTypeFilter` at the beginning of the index filter list of FilterIndexRule / JoinIndexRule. WDYT?

sezruby:
Could you add `Implementation` section and write the list of PR-level to do in the proposal? 
So that we could check the remaining work & upcoming changes for Data skipping indexes.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/sketch/BloomFilterSketch.scala
```diff
@@ -0,0 +1,50 @@
...
+ * Sketch based on a bloom filter for a given column.
+ *
+ * Being a probabilistic structure, it is more efficient in terms of the index
+ * data size than [[ValueListSketch]] if the number of distinct values for the
+ * column is large, but can be less efficient in terms of query optimization
+ * than [[ValueListSketch]] due to false positives.
+ *
+ * Users can specify the target false positive rate and the expected number of
+ * distinct values per file. These variables determine the size of the bloom
+ * filters and thus the size of the index data.
```
clee704, line 32: TODO: We can do this semi-automatically.

The expected number of distinct values per file can be directly calculated in Spark. The problem is there is no way to specify this differently for each file in the current implementation, as aggregation functions don't know about the current grouping key (= input_file_name() for our case). As a workaround, we can use the average value, although it won't be ideal for files varying greatly in size or distribution of key values.

For a better user experience, instead of specifying FPP, users should be able to specify the target size of the index. Then we can choose FPP based on the number of disticted values. If the calculated FPP is too high, users should be warned about it.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,132 @@
...
+   */
+  def index(ctx: IndexerContext, sourceData: DataFrame): DataFrame = {
+    val aggregateFunctions = sketches.flatMap { s =>
+      val aggrs = s.aggregateFunctions
+      assert(aggrs.nonEmpty && aggrs.length == s.numValues)
+      aggrs.map(aggr => aggr.as(DataSkippingIndex.normalizeColumnName(aggr.toString)))
+    }
+    val fileNameCol = "input_file_name"
+    val indexData = sourceData
+      .groupBy(input_file_name().as(fileNameCol))
```
sezruby, line 89: Question: 
is this scalable for millions of files? how many tasks will be created?
Could you add some benchmark number (for index creation) using 1TB/100KChunk dataset? (or bigger?)

sezruby, line 89: Could you add an example of index data to the PR description? e.g.
```
val df = spark.read.parquet("/path/to/index/v__=0")
df.limit(10).show(false)
```

clee704: My guess - it won't be slower than creating covering index data.

Creating covering index data - each local node makes a vertical slice of the source data. Then those vertical slices need to be shuffled all around the cluster because of sorting and bucketing.

Creating data skipping index data - each local node does aggregation for the source data. Then those aggregated data needs to be shuffled all around the cluster, but this time the shuffle size will be much smaller.

I'd like to do the measurement, but last time I tried to create big data in Spark it was too slow. Do you have any tips on this?

clee704: Results for TPC-H 1TB, 1K/10K/100K chunks, lineitem, index suitable for Q6, 199 executors
- CoveringIndex (indexed: l_shipdate, l_discount, l_quantity, included: l_extendedprice)
  - 1K chunks: 3m 05
  - 10K chunks: 3m 12
  - 100K chunks: 4m 54
- DataSkippingIndex (MinMaxSketch/ValueListSketch on l_shipdate/l_discount/l_quantity)
  - 1K chunks: 27 sec
  - 10K chunks: 54 sec
  - 100K chunks: 2m 24

clee704: Done

sezruby: Thanks! The result looks great 👍 Seems already have the rule implementation for minmax/value list :)
Could you also share index creation time & index data size for both CoveringIndex & DataSkippingIndex?

clee704: These are the index creation times, not query execution times. Sorry I didn't make it clear 😅

Data skipping indexes are only effective when the indexed columns have some correlation with the partitioning columns, and the data I used was unpartitioned, so DS cannot be used.

Index size was 31 GB for CoveringIndex and 4/13/150 KB for DataSkippingIndex for 1K/10K/100K chunks respectively.

src/test/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndexTest.scala
```diff
@@ -0,0 +1,144 @@
...
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.SaveMode
+import org.apache.spark.sql.functions.{input_file_name, max, min}
+
+import com.microsoft.hyperspace.index.{FileInfo, Index}
+import com.microsoft.hyperspace.index.dataskipping.sketch.{MinMaxSketch, ValueListSketch}
+
+class DataSkippingIndexTest extends DataSkippingSuite {
+  test("kind") {
```
sezruby, line 28: Can we have a better naming convention? 
https://www.puresourcecode.com/dotnet/testing/7-popular-unit-test-naming-conventions/

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  override def canHandleDeletedFiles: Boolean = true
+
+  override def write(ctx: IndexerContext, indexData: DataFrame): Unit = {
+    indexData.write.mode(SaveMode.Overwrite).parquet(ctx.indexDataPath.toString)
+  }
+
+  override def optimize(ctx: IndexerContext, indexDataFilesToOptimize: Seq[FileInfo]): Unit = {
+    val indexData = ctx.spark.read.parquet(indexDataFilesToOptimize.map(_.name): _*)
+    write(ctx, indexData.repartition(1))
```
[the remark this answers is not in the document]
clee704: Yeah, maybe we can drop `repartition(1)` here. The purpose of `optimize` here is to reduce the number of the index data files, but it doesn't have to be a single file.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,141 @@
...
+    // so that the file id column comes first
+    val indexDataWithFileId = indexDataWithFileName
+      .join(fileIdDf.hint("broadcast"), fileNameCol)
+      .drop(fileNameCol)
+    val cols = indexDataWithFileId.columns
+    indexDataWithFileId.select(cols.last, cols.dropRight(1): _*)
+  }
+
+  override def equals(o: Any): Boolean = o match {
+    case that: DataSkippingIndex => sketches == that.sketches
```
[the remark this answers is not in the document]
andrei-ionescu:
```suggestion
    case DataSkippingIndex(sketches, _, _) => sketches == that.sketches
```

src/main/scala/com/microsoft/hyperspace/index/dataskipping/sketch/BloomFilterAgg.scala
```diff
@@ -0,0 +1,82 @@
...
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.catalyst.expressions.aggregate.TypedImperativeAggregate
+import org.apache.spark.sql.types.{BinaryType, DataType, StringType}
+import org.apache.spark.unsafe.types.UTF8String
+import org.apache.spark.util.sketch.BloomFilter
+
+case class BloomFilterAgg(
```
andrei-ionescu, line 28: Please add scaladoc and explain: why this is from the start in the Data Skipping index? I would expect this bloom filter improvement to be added later on as improvement over the data skipping index.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndexConfig.scala
```diff
@@ -0,0 +1,45 @@
...
+
+package com.microsoft.hyperspace.index.dataskipping
+
+import org.apache.spark.sql.DataFrame
+
+import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.index.{IndexConfigTrait, IndexerContext}
+import com.microsoft.hyperspace.index.dataskipping.sketch.Sketch
+
+case class DataSkippingIndexConfig(
```
andrei-ionescu, line 25: Please add scaladoc for this new public API. It is very important to understand how to properly use this new kind of index. Please respond to the following questions: 

1. What's a `sketch` in the context of this index?
2. What's a `sketchOffset` and how can it be used?
3. Is it a 1-1 mapping between a `sketch` and a column?

src/main/scala/com/microsoft/hyperspace/index/dataskipping/sketch/BloomFilterSketch.scala
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param col Column name this sketch is based on
+ * @param fpp Target false positive rate
+ * @param expectedDistinctCountPerFile Expected number of distinct values per file
+ */
+case class BloomFilterSketch(col: String, fpp: Double, expectedDistinctCountPerFile: Long)
```
andrei-ionescu, line 38: Why this is from the start in the Data Skipping index? I would expect this bloom filter improvement to be added later on as improvement over the data skipping index not being in the **Data Skipping Index Part 1** PR.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/sketch/MinMaxSketch.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package com.microsoft.hyperspace.index.dataskipping.sketch
+
+import org.apache.spark.sql.Column
+import org.apache.spark.sql.functions.{max, min}
+
+/**
+ * Sketch based on minimum and maximum values for a given column.
+ */
+case class MinMaxSketch(col: String) extends SingleColumnSketch("MinMax", col) {
```
andrei-ionescu, line 25: Nit: Exlpain the `col` param.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/sketch/MinMaxSketch.scala
```diff
@@ -0,0 +1,29 @@
...
+  override def withNewColumn(newColumn: String): Sketch = copy(col = newColumn)
+  override def aggregateFunctions: Seq[Column] = min(col) :: max(col) :: Nil
+  override def numValues: Int = 2
```
andrei-ionescu, line 28: Could you explain why the `2` value? Does it represents that it will have 2 columns: min & max?

src/main/scala/com/microsoft/hyperspace/index/dataskipping/sketch/SingleColumnSketch.scala
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Base class for sketches which are based on a single column.
+ *
+ * @param sketchName Name of this sketch type to be used for toString
+ * @param col Column name this sketch is based on
+ */
+abstract class SingleColumnSketch(sketchName: String, column: String) extends Sketch {
+  final override def withNewColumns(newColumns: Seq[String]): Sketch = {
+    assert(newColumns.length == 1)
```
andrei-ionescu, line 27: Ditto with assertions... See above.

andrei-ionescu (commented):
Added some feedback. 

The main inconvenience for me was the fact that important pieces of core code do not have scaladoc and I couldn't easily understand what that piece of code or class does.

Please, don't exclude the heavy use cases, aka the datasets with millions of files. Although this first PR may not be the place to address them, I'm just letting you know about the fact that those heavy use cases are real and very close to come.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/sketch/BloomFilterSketch.scala
clee704: Then we can make a ratio-based config and set a default value for it - for example, aim for 0.1% (default, configurable) of the source data size.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: I don't have datasets with 1 million files. Since the index data creation code is fairly simple, I think the results for 1-100K files are enough.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.spark.sql.catalyst.expressions.NamedExpression
+import org.apache.spark.sql.functions.input_file_name
+
+import com.microsoft.hyperspace.index._
+import com.microsoft.hyperspace.index.dataskipping.sketch.Sketch
+
+case class DataSkippingIndex(
+    sketches: Seq[Sketch],
+    sketchOffsets: Seq[Int],
+    override val properties: Map[String, String] = Map.empty)
```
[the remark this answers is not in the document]
clee704: This is NOT overriding the property with an empty map. `override` means that `val properties` is overriding (implementing) the interface. The default value means that you don't have to (you can if you want to) specify the value for `properties` when creating an instance of `DataSkippingIndex`. It is used here because `properties` merely exists to satisfy the interface and not the interest of `DataSkippingIndex`.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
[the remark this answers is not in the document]
clee704: AFAIK Spark can read a single Parquet file in parallel without a problem. A file is stored as multiple blocks in HDFS and in theory, there should be no big difference between reading a single Parquet file and reading multiple Parquet files, except that too many files might bring some overhead.

But I agree that this `repartition(1)` is unnecessarily restricting. What I'd like to do here is reducing the number of files, if there were too many. Let me amend this a little bit.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,141 @@
...
+      val aggrs = s.aggregateFunctions
+      assert(aggrs.nonEmpty && aggrs.length == s.numValues)
+      aggrs.map(aggr => aggr.as(DataSkippingIndex.normalizeColumnName(getColumnName(aggr))))
+    }
+    val fileNameCol = "input_file_name"
+    val indexDataWithFileName = sourceData
+      .groupBy(input_file_name().as(fileNameCol))
+      .agg(aggregateFunctions.head, aggregateFunctions.tail: _*)
+
+    // construct a dataframe to convert file names to file ids
```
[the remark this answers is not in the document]
clee704: Done

src/main/scala/com/microsoft/hyperspace/actions/RefreshIncrementalAction.scala
```diff
@@ -110,15 +113,15 @@ class RefreshIncrementalAction(
...
   override def logEntry: LogEntry = {
-    val index = updatedIndex.getOrElse(previousIndexLogEntry.derivedDataset)
+    val index = updatedIndexOpt.getOrElse(previousIndexLogEntry.derivedDataset)
     val entry =
       getIndexLogEntry(spark, df, previousIndexLogEntry.name, index, indexDataPath, endId)
 
     // If there is no deleted files, there are index data files only for appended data in this
     // version and we need to add the index data files of previous index version.
     // Otherwise, as previous index data is rewritten in this version while excluding
     // indexed rows from deleted files, all necessary index data files exist in this version.
```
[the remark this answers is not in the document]
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/Index.scala
```diff
@@ -144,4 +143,24 @@ trait Index {
...
+   */
+  def equals(that: Any): Boolean
+
+  /**
+   * Returns the hash code for this index.
+   */
+  def hashCode: Int
+}
+
+object Index {
```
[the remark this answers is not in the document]
clee704: This is explained by the docstring attached to `refreshIncremental` which is in the same file.

src/main/scala/com/microsoft/hyperspace/index/CoveringIndex.scala
```diff
@@ -266,25 +270,4 @@ object CoveringIndex {
 
     (indexData, resolvedIndexedColumns, resolvedIncludedColumns)
   }
-
-  private def resolveConfig(
```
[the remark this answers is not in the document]
clee704: Moved to `IndexUtils` to make it available to other Index implementations.

src/main/scala/com/microsoft/hyperspace/index/rules/CoveringIndexFilter.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+
+import com.microsoft.hyperspace.index.CoveringIndex
+import com.microsoft.hyperspace.index.rules.ApplyHyperspace.PlanToIndexesMap
+
+/**
+ * Filters out indexes which are not [[CoveringIndex]].
+ */
+object CoveringIndexFilter extends QueryPlanIndexFilter {
```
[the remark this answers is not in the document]
clee704: Moved to com.microsoft.hyperspace.index.types.covering

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  override def refreshIncremental(
+      ctx: IndexerContext,
+      appendedSourceData: Option[DataFrame],
+      deletedSourceDataFiles: Seq[FileInfo],
+      indexContent: Content): (Index, Index.UpdateMode) = {
+    if (appendedSourceData.nonEmpty) {
+      write(ctx, index(ctx, appendedSourceData.get))
+    }
+    (this, Index.UpdateMode.Merge)
```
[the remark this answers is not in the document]
clee704: Done

src/test/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndexTest.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndexConfig.scala
```diff
@@ -0,0 +1,45 @@
...
+    override val indexName: String,
+    firstSketch: Sketch,
+    moreSketches: Sketch*)
+    extends IndexConfigTrait {
+
```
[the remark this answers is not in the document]
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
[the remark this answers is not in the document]
clee704: Ignoring `properties` is the current behavior. If it's problematic we should create an issue.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,141 @@
...
+      .getIdToFileMapping(relation.pathNormalizer)
+      .toDF(IndexConstants.DATA_FILE_NAME_ID, fileNameCol)
+
+    // drop the file name column and reorder the columns
+    // so that the file id column comes first
+    val indexDataWithFileId = indexDataWithFileName
+      .join(fileIdDf.hint("broadcast"), fileNameCol)
+      .drop(fileNameCol)
+    val cols = indexDataWithFileId.columns
+    indexDataWithFileId.select(cols.last, cols.dropRight(1): _*)
```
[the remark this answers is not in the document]
clee704: Done

sezruby:
No major comment, but I'd like to wait for Part 2 PR just in case you may want to fix index creation code after that.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,181 @@
...
+    copy(properties = newProperties)
+  }
+
+  override def statistics(extended: Boolean = false): Map[String, String] = {
+    Map("sketches" -> sketches.mkString(", "))
+  }
+
+  override def canHandleDeletedFiles: Boolean = true
+
+  override def write(ctx: IndexerContext, indexData: DataFrame): Unit = {
```
sezruby, line 68: Can we have a better naming on this?
This is for create index & refresh full, but seems write is not intuitive compared to refreshIncremental / optimize.
WDYT?

clee704: In the Scaladoc for `Index.write` I've written:

> Writes the index data to the specified path.

and it is precisely what it is supposed to do. We can use other words like "save", "put", or "store". But I think "write" isn't bad. "save" also sounds good.

src/main/scala/com/microsoft/hyperspace/index/IndexUtils.scala
```diff
@@ -55,4 +58,10 @@ object IndexUtils {
             s"from available source columns:\n${df.schema.treeString}")
     }
   }
+
+  /**
+   * Spark UDF to decode the URL-encoded string returned by input_file_name().
+   */
+  lazy val decodeInputFileName = udf(
+    (p: String) => URLDecoder.decode(p.replace("+", "%2B"), "UTF-8"))
```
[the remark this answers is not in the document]
paryoja: From comments in #482 
`input_file_name() returns a URL-encoded string and it doesn't match the file paths stored in the FileIdTracker if there are special characters in the path. CoveringIndex has the same issue and will be fixed in a separate PR.

The plus sign has to be treated specially because the built-in URLDecoder decodes "+" to " ".`

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,185 @@
...
+ * DataSkippingIndex is an index that can accelerate queries by filtering out
+ * files in relations using sketches.
+ *
+ * @param sketches List of sketches for this index
+ * @param properties Properties for this index; see [[Index.properties]] for details.
+ */
+case class DataSkippingIndex(
+    sketches: Seq[Sketch],
+    override val properties: Map[String, String] = Map.empty)
+    extends Index {
```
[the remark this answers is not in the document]
clee704: Option would be better than using special values like -1 or an empty string "". For containers such as Seq or Map, I think it depends. If the distinction between None and an empty container is needed, then you should go with Option, but if you don't care about the difference, then you can use an empty container instead of None.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,183 @@
...
+        fileIdDf.hint("broadcast"),
+        IndexUtils.decodeInputFileName(indexDataWithFileName(fileNameCol)) ===
+          fileIdDf(fileNameCol))
+      .select(
+        IndexConstants.DATA_FILE_NAME_ID,
+        indexDataWithFileName.columns.filterNot(_ == fileNameCol).map(c => s"`$c`"): _*)
+  }
+
+  private def writeImpl(ctx: IndexerContext, indexData: DataFrame, writeMode: SaveMode): Unit = {
+    indexData.cache()
```
sezruby, line 140: Shouldn't we call "unpersist()" after the index data write?

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,183 @@
...
+    val minRowCountPerFileData = indexData
+      .groupBy(spark_partition_id())
+      .count()
+      .agg(min("count"))
+      .collect()
```
sezruby, line 145: Seems spark_partition_id() is non-deterministic. Is it worth to get the min count of partitions here?
Could you explain why don't you use the length of `indexData.inputFiles` here?

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndexConfig.scala
```diff
@@ -0,0 +1,76 @@
...
+      sourceData: DataFrame,
+      properties: Map[String, String]): (DataSkippingIndex, DataFrame) = {
+    val resolvedSketches = ExpressionUtils.resolve(ctx.spark, sketches, sourceData)
+    checkDuplicateSketches(resolvedSketches)
+    val index = DataSkippingIndex(resolvedSketches, properties)
+    (index, index.index(ctx, sourceData))
+  }
+
+  private def checkDuplicateSketches(sketches: Seq[Sketch]): Unit = {
+    val uniqueSketches = mutable.Set[Sketch]()
```
sezruby, line 68: how about 
```suggestion
  if (sketches.length != sketches.toSet.length) {
```

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,183 @@
...
+      deletedSourceDataFiles: Seq[FileInfo],
+      indexContent: Content): (Index, Index.UpdateMode) = {
+    if (appendedSourceData.nonEmpty) {
+      writeImpl(ctx, index(ctx, appendedSourceData.get), SaveMode.Overwrite)
+    }
+    if (deletedSourceDataFiles.nonEmpty) {
+      val spark = ctx.spark
+      import spark.implicits._
+      val deletedFileIds = deletedSourceDataFiles.map(_.id).toDF(IndexConstants.DATA_FILE_NAME_ID)
+      val updatedIndexData = ctx.spark.read
```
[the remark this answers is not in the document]
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,183 @@
...
+        .parquet(indexContent.files.map(_.toString): _*)
+        .join(deletedFileIds, Seq(IndexConstants.DATA_FILE_NAME_ID), "left_anti")
+      val writeMode = if (appendedSourceData.nonEmpty) {
+        SaveMode.Append
+      } else {
+        SaveMode.Overwrite
+      }
+      writeImpl(ctx, updatedIndexData, writeMode)
+    }
+    val updateMode = if (deletedSourceDataFiles.isEmpty) {
```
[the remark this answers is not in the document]
clee704: You can find the info in the Scaladoc for "Index.refreshIncremental".

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: It's computing the minimum value of the row counts of index data files that would be written. Grouping by spark_partition_id() is the same as grouping by the file to be written. `groupBy()` followed by `count()` computes the number of rows per group. `agg(min("count"))` computes the min.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/sketch/Sketch.scala
```diff
@@ -0,0 +1,72 @@
...
+package com.microsoft.hyperspace.index.dataskipping.sketch
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.types.DataType
+
+/**
+ * Represents a sketch specification for data skipping indexes.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "type")
```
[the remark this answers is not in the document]
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndexConfig.scala
clee704: Done

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
sezruby: But can we just use `indexData.inputFiles / HyperspaceConf.DataSkipping.minRecordsPerIndexDataFile` ?
Or we could use `spark.sql.files.maxRecordsPerFile` config though I haven't tested yet.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndexConfig.scala
```diff
@@ -0,0 +1,76 @@
...
+      sourceData: DataFrame,
+      properties: Map[String, String]): (DataSkippingIndex, DataFrame) = {
+    val resolvedSketches = ExpressionUtils.resolve(ctx.spark, sketches, sourceData)
+    checkDuplicateSketches(resolvedSketches)
+    val index = DataSkippingIndex(resolvedSketches, properties)
+    (index, index.index(ctx, sourceData))
+  }
+
+  private def checkDuplicateSketches(sketches: Seq[Sketch]): Unit = {
+    val uniqueSketches = sketches.toSet
```
sezruby, line 68: I found better one. 
```scala
val duplicates = sketches.groupBy(identity).collect { case (x, s) if s.lengthCompare(1) > 0 => x }
```

clee704: Thank you for the suggestion. Actually, I considered that, but I chose the current approach. I think using groupBy gives you no advantage over using toSet - it is slower and requires more memory for the common path (where the next branch is not taken), and it is less intuitive. It's unlikely to see millions of sketches, so performance might be irrelevant here, but readability matters. I'm not saying it's terrible, but it is more like that I don't see a strong need to change it.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: That's basically running lines 152-153 unconditionally, and it can actually *increase* the number of index data files, which I want to avoid.

I don't understand how `maxRecordsPerFile` can be used. Can you elaborate on that?

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndexConfig.scala
sezruby: Fair enough 👍

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
sezruby: >That's basically running lines 152-153 unconditionally, and it can actually increase the number of index data files, which is I want to avoid.

I don't get this. Could you explain more?

Below test code generated 4 parquet files:
```scala
spark.conf.set("spark.sql.files.maxRecordsPerFile", 1)
Seq(1, 2, 3, 4).toDF("a").write.parquet("recordtest")
```

clee704: Here we want to avoid too many files. Suppose that minRecordsPerIndexDataFile = 100 and there are 1000 files with 10 records each. Because 10 < 100, we repartition the data by 1000 / 10 = 100, so that there will be 100 files with 100 records each. It is also possible that 10 files with 1000 records each. In this case, we don't want to repartition.

I've changed `repartition` to `coalesce` and the meaning of the config to make it simpler.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,174 @@
...
+      .select(
+        IndexConstants.DATA_FILE_NAME_ID,
+        indexDataWithFileName.columns.filterNot(_ == fileNameCol).map(c => s"`$c`"): _*)
+  }
+
+  private def writeImpl(ctx: IndexerContext, indexData: DataFrame, writeMode: SaveMode): Unit = {
+    indexData.cache()
+    val minRowCountPerFileConfig =
+      HyperspaceConf.DataSkipping.minRecordsPerIndexDataFile(ctx.spark)
+    val maxNumFiles = toIntExact(math.max(1, indexData.count() / minRowCountPerFileConfig))
```
sezruby, line 145: is there any difference between `indexData.count` and `indexData.inputFiles`? 
if not, we don't need `indexData.cache`

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,174 @@
...
+    val repartitionedIndexData = indexData.coalesce(maxNumFiles)
```
sezruby, line 146: Ok one question: the index data - is it already partitioned? Why are you worry about small files? 
If it's **max**RecordPerFiles, we don't need to care about it.

I think coalesce is not a good idea. it just control the number of files, not the size or number of record for each files. 
And the performance is sometimes worse than repartition.

clee704: Because often data skipping index data is very small and Spark is bad at working with many small files. Without repartition, the tiny index data is stored in too many files like 200 files. It's no good for performance.

I think coalesce is fine - it doesn't increase the number of files and I've verified that it evenly distributes rows over files.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: Using `indexData.inputFiles.length` can silently break the code if we change the way we create index data in the future, whereas `indexData.count()` will be always correct.

sezruby: But `indexData.count()` requires `indexData.cache` which is not free and it will take longer than `inputFiles`. . . ?

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
sezruby: Because of the join, it's partitioned. coalesce merges partitions, not distributing the rows as it's optimized for that purpose. 
So.. as you said performance doesn't matter for small index dataset, I would prefer `repartition(maxNumFiles)`

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -117,4 +117,15 @@ object IndexConstants {
...
+  /**
+   * Target minimum number of records per index data file for data skipping
+   * indexes.
+   *
+   * When index data is stored in files, the number of files to be written is
+   * limited to # of records / this value.
+   */
+  val DATASKIPPING_MIN_RECORDS_PER_INDEX_DATA_FILE =
+    "spark.hyperspace.index.dataskipping.minRecordsPerIndexDataFile"
+  val DATASKIPPING_MIN_RECORDS_PER_INDEX_DATA_FILE_DEFAULT = "100000"
```
paryoja, line 130: Is there any reason why use 100000 as default?

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: Okay, then we need to bring back the branch because it shouldn't increase the number of files.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
clee704: The optimal value would vary depending on data skipping indexes, but we need a default value. Anything bigger than 1 would fit for the purpose. 100000 is chosen so that typical data skipping indexes with MinMaxSketches take at least >1MB per file.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: I think we don't have to worry too much about the performance here, as index creation/refresh are less frequent operations and index data is usually small. `cache()` and `count()` should be fast enough.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
sezruby: I think it's okay to repartition with numPartitions based on maxRowCountPerFiles config as users can adjust it and the default value(100000) looks reasonable. `indexData.count()`  (+ for minRowCountPerFileData groupby/count) incurs spark jobs & caching operation even it's faster enough.
So I would avoid it if possible. cc @imback82 WDYT?

This is my preference:
```scala
    val maxRowCountPerFileConfig =
      HyperspaceConf.DataSkipping.maxRecordsPerIndexDataFile(ctx.spark)
    val maxNumFiles = toIntExact(math.max(1, indexData.inputFiles.length / maxRowCountPerFileConfig))
    val repartitionedIndexData = indexData.repartition(maxNumFiles)
```

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
sezruby: I think we could put the constraint(indexData.count == indexData.inputfiles.length) as an unit test. (+ some comment here)
Dataframe caching is not cheaper & simpler than you think; it involves many components like  optimizer, scheduler, block management .. etc. Of course, problems rarely happens with this kind of small dataset, I'd like to avoid to using it if possible..

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: The whole point of the code here is to optimize the index application time. Index creation/refresh times are also important, but we should not reverse the priority. If it helps the application time, it makes sense to pay the small cost during creation/refresh. Index creation/refresh are maintenance operations, and the code for cache, groupBy, count, all combined, takes only a fraction of the total creation/refresh time. Most of the time is spent during reading the source data and computing sketch aggregation functions.

sezruby: @clee704 I agree with your point, but what I'm trying to say is your approach and suggested approach do the *same* thing, but yours has more additional operations that I'd like to avoid.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: My data says `cache()` doesn't involve noticeable overhead for 1TB/10K files data.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
clee704: No, it's not the same, as the suggested approach fixes the number of files, which can be larger than it was before repartition.

clee704: Anyway, this row count based repartitioning thing was only an approximation for a better approach which is based on the data size. I'll push a new commit.

sezruby: With the default config (100,000) and the shuffle partition config (200), it only happens when it has more than (200  *  100,000) files. Considering smaller shuffle partition number, let say 1, numFiles would be between 1~10 for 100K~1M files. 

And if a user wants to reduce the index file size, maybe because of large BFs, they tries a smaller config val, but it won't work because of the condition.

clee704: That's why I said it was a suboptimal approach. For the best index application performance, index data files should be evenly distributed over files in terms of data size because index data filtering is mostly I/O bound.

src/main/scala/com/microsoft/hyperspace/index/dataskipping/DataSkippingIndex.scala
```diff
@@ -0,0 +1,180 @@
...
+        IndexConstants.DATA_FILE_NAME_ID,
+        indexDataWithFileName.columns.filterNot(_ == fileNameCol).map(c => s"`$c`"): _*)
+  }
+
+  private def writeImpl(ctx: IndexerContext, indexData: DataFrame, writeMode: SaveMode): Unit = {
+    indexData.cache()
+    indexData.count() // force cache
+    // Note: the actual index data size can be smaller due to compression.
+    val indexDataSize = indexData.queryExecution.optimizedPlan.stats.sizeInBytes
+    indexData.unpersist()
```
sezruby, line 145: Seems don't need to cache here; optimizedPlan.stats.sizeInBytes doesn't read from the cached data, but just estimate the plan.

clee704: If you don't cache, the reported size is close to the size of the source data because it doesn't take the aggregations into account. The difference is huge.

Even if you cache, I found that optimizePlan.stats.sizeInBytes is somewhat too rough for the purpose. Please refer to the new commit.

clee704: I understand your concern about `cache()`. I'll check the performance with some real data.

sezruby: @clee704  Can we use the max record limit or `spark.hyperspace.index.dataskipping.numPartitions` config? To me, these looks unnecessary complexity.
In most cases, DS index data file will be stored in 1 file. Even it's not evenly distributed, it's okay. We don't need to measure the exact size; repartition distributes rows by round robin and compression rate will be different. 

The reason I made the comment - https://github.com/microsoft/hyperspace/pull/461#discussion_r656790431 - is.. I just wanted to make it configurable.

clee704: What is so complex about the code here? It's just a few lines of code. It's for the best user experience. If we fix the number of files with such a config, users should set the config value by themselves every time they create an index. Most of the time, they won't know the final index size and should create the index twice. If they set the value to 1 and don't change it, there is a possibility of OOM or slow write because write is done by a single task. Therefore, I cannot accept using such a config. It could be used as a supplementary config in addition to the current implementation, though.

I don't know where you are getting such strong concern for caching, but here are actual numbers:

Data: catalog_sales (from TPC-DS 1TB)
Stored size (parquet files): 138.4 GiB
Row count: 1,439,980,416
Number of files: 1,837

cache | sketch | # index data files | index creation time | index data size
--- | --- | --: | --: | --:
with cache | MinMaxSketch | 1 | 1m 4s | 30 KB
without cache | MinMaxSketch | 1 | 1m 2s | 30 KB
with cache | BloomFilterSketch | 4 | 1m 30s | 1 GB
without cache | BloomFilterSketch | 1 | 1m 40s | 1 GB
without cache | BloomFilterSketch | 4 | 1m 26s | 1 GB

"with cache" indicates the current implementation is used. "without cache" is repartition using the value from a config without caching.

As you can see, the overhead is minimal and less than 5% of the total creation time. The reality would be worse for the "without cache" case - you have to try multiple times to create an index because you don't know the index size, and always repartitioning to a single partition doesn't work for all workloads (OOM/too slow write).

If you still don't think caching isn't right, could you please state more clearly what's so bad about it?

sezruby: File scan is not done by a single task - spark creates multiple partitions to read a file based on partition size configs. (Partition num = task num)

Could you share the cluster size and result using a small cluster like 2 small nodes?

cc @imback82 @andrei-ionescu  @rapoth @paryoja Could you add some comment on this? 
I think cache() for the exact RDD byte size to determine the number of DS index files is unnecessary. Since in most cases it would be 1. 

So I suggested to use maxRecordsPerFile config or index.dataskipping.numPartitions
config which does not need an additional spark job and cache. 

Just my opinion and I'm not saying it's a must; I'd like to hear from other folks.

clee704: The data was for 10 executors / 40 cores. Most of the index creation time comes from reading and aggregating the source data. Using a small cluster won't change the overall results regarding caching. But I'll give the results with 2 executors.

Index data sizes can range from a few KB to tens of GB or more, depending on the number of the source data files, types of sketches, number of distinct values for each column, etc. Repartition by 1 wouldn't work for large index data.

clee704: There is another option: write the index data without any repartitioning, and check if the written index data files are too small (smaller than the configured target index data file size). If it's the case, read the files again to merge them. How about this?

sezruby: Thanks for the update! Without repartition, each file would usually be small. 
Seems there's no major concern from others, I'll approve the change.

sezruby (approved):

PR #464, opened 2021-06-16
Fix index cache bug

### LINKED ISSUE #463: [Bug] Deleted indexes are applied / not shown
### Describe the issue
Deleted indexes are applied / not shown.

### To Reproduce
Run the following code in `spark-shell`.

Scenario 1 - the bug occurs in the current master:
```scala
import com.microsoft.hyperspace._

spark.range(10).toDF("A").write.saveAsTable("T")
val df = spark.read.table("T")
val hs = Hyperspace()
hs.createIndex(df, IndexConfig("IDX", Seq("A"), Nil))
spark.enableHyperspace()
df.filter("A == 1").queryExecution // index applied
hs.deleteIndex("IDX")
hs.indexes.show // IDX is shown as DELETED
df.filter("A == 1").queryExecution // index still applied - bug
```

If we change the order of the last two lines, now deleted indexes are not shown.

Scenario 2 - the bug occurs both in v0.4.0 and the current master:
```scala
import com.microsoft.hyperspace._

spark.range(10).toDF("A").write.saveAsTable("T")
val df = spark.read.table("T")
val hs = Hyperspace()
hs.createIndex(df, IndexConfig("IDX", Seq("A"), Nil))
spark.enableHyperspace()
df.filter("A == 1").queryExecution // index applied
hs.deleteIndex("IDX")
df.filter("A == 1").queryExecution // index not applied
hs.indexes.show // IDX is not shown - bug
```

### Expected behavior
In the first scenario, the index shouldn't have been applied.

In the second scenario, the index should be shown as `DELETED`.

### Environment 
 - OS: MacOS
 - Apache Spark 3.0.2 / 2.4.2

### Workaround
Disable the index cache.

```scala
spark.conf.set("spark.hyperspace.index.cache.expiryDurationInSeconds", "0")
```

clee704:
Fixes #463.

The results of getIndexes depend on the states parameter, but the cache
wasn't using the parameter. If the function was called with a parameter
and then called again with a different parameter, it could return a
wrong result.

 - **Tracking Issue**: #463 
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

Use different cache entries for different parameters.

It fixes the bug.

Unit tests are added.


src/main/scala/com/microsoft/hyperspace/index/CachingIndexCollectionManager.scala
```diff
@@ -59,11 +60,16 @@ class CachingIndexCollectionManager(
    * @return list of indexes whose current state is among desired states
    */
   override def getIndexes(states: Seq[String] = Seq()): Seq[IndexLogEntry] = {
-    indexCache.get().getOrElse {
-      val originalIndexes = super.getIndexes(states)
-      indexCache.set(originalIndexes)
-      originalIndexes
-    }
```
sezruby, line 66: It might be better to cache indexes with all states and return filtered indexes for the given states

clee704: I considered that too but didn't choose it because it relies on the semantics of getIndexes. On the other hand, this version works regardless of how getIndexes works.

However, it seems we have to do as you said due to caching (tagging) in IndexLogEntry.

sezruby (approved):
LGTM Thanks!

PR #467, opened 2021-06-19
Fix ResolverUtils to handle StructType column properly

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Fix ResolverUtils to handle nested StructType column properly.

Without the fix, the following exception is thrown while resolving nested StructType column.

```
// nestedC schema
StructField(nestedC,StructType(
  StructField(id,StringType,true), 
  StructField(leaf,StructType(
    StructField(id,StringType,true), 
    StructField(cnt,IntegerType,true)),
    true)),
  true))
//
scala> hs.createIndex(dff, IndexConfig("nestedIndex2", Seq("nestedC"), Seq("hsh")))
scala.MatchError: List() (of class scala.collection.immutable.Nil$)
  at com.microsoft.hyperspace.util.ResolverUtils$.com$microsoft$hyperspace$util$ResolverUtils$$getColumnNameFromSchema(ResolverUtils.scala:215)
  at com.microsoft.hyperspace.util.ResolverUtils$.com$microsoft$hyperspace$util$ResolverUtils$$getColumnNameFromSchema(ResolverUtils.scala:220)
  at com.microsoft.hyperspace.util.ResolverUtils$$anonfun$1$$anonfun$apply$2.apply(ResolverUtils.scala:174)
  at com.microsoft.hyperspace.util.ResolverUtils$$anonfun$1$$anonfun$apply$2.apply(ResolverUtils.scala:171)
  at scala.Option.map(Option.scala:146)
  at com.microsoft.hyperspace.util.ResolverUtils$$anonfun$1.apply(ResolverUtils.scala:171)
  at com.microsoft.hyperspace.util.ResolverUtils$$anonfun$1.apply(ResolverUtils.scala:168)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
  at scala.collection.immutable.List.foreach(List.scala:392)
  at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
  at scala.collection.immutable.List.map(List.scala:296)
  at com.microsoft.hyperspace.util.ResolverUtils$.resolve(ResolverUtils.scala:168)
  at com.microsoft.hyperspace.actions.CreateAction.isValidIndexSchema(CreateAction.scala:77)
  at com.microsoft.hyperspace.actions.CreateAction.validate(CreateAction.scala:60)
  at com.microsoft.hyperspace.actions.Action$class.run(Action.scala:89)
  at com.microsoft.hyperspace.actions.CreateAction.run(CreateAction.scala:29)
  at com.microsoft.hyperspace.index.IndexCollectionManager.create(IndexCollectionManager.scala:47)
  at com.microsoft.hyperspace.index.CachingIndexCollectionManager.create(CachingIndexCollectionManager.scala:78)
  at com.microsoft.hyperspace.Hyperspace$$anonfun$createIndex$1.apply$mcV$sp(Hyperspace.scala:45)
  at com.microsoft.hyperspace.Hyperspace.withHyperspaceRuleDisabled(Hyperspace.scala:188)
  at com.microsoft.hyperspace.Hyperspace.createIndex(Hyperspace.scala:44)
  ... 54 elided

```

Yes, fixes the bug described above

Unit test

### DIFF
--- a/src/test/scala/com/microsoft/hyperspace/index/CreateIndexNestedTest.scala
+++ b/src/test/scala/com/microsoft/hyperspace/index/CreateIndexNestedTest.scala
@@ -98,8 +98,7 @@ class CreateIndexNestedTest extends HyperspaceSuite with SQLHelper {
       "Indexed columns with wrong case are stored in metadata")
     assert(
       indexes.head
-        .getAs[Map[String, String]]("additionalStats")(
-          "includedColumns") == "__hs_nested.nested.leaf.cnt",
+        .getAs[Map[String, String]]("additionalStats")("includedColumns") == "__hs_nested.nested.leaf.cnt",
       "Included columns with wrong case are stored in metadata")
   }
 
@@ -139,10 +138,7 @@ class CreateIndexNestedTest extends HyperspaceSuite with SQLHelper {
     val dfB = nonPartitionedDataDF.as("B")
     val dfJoin = dfA
       .join(dfB, dfA("Query") === dfB("Query"))
-      .select(
-        dfA("RGUID"),
-        dfA("Query"),
-        dfA("nested.leaf.cnt"))
+      .select(dfA("RGUID"), dfA("Query"), dfA("nested.leaf.cnt"))
     val exception = intercept[HyperspaceException] {
       hyperspace.createIndex(dfJoin, indexConfig1)
     }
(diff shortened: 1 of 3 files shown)


src/test/scala/com/microsoft/hyperspace/index/CreateIndexNestedTest.scala
```diff
@@ -206,4 +203,30 @@ class CreateIndexNestedTest extends HyperspaceSuite with SQLHelper {
...
+    hyperspace.createIndex(nonPartitionedDataDF, indexConfig)
+    hyperspace.createIndex(nonPartitionedDataDF, indexConfig2)
+    assert(hyperspace.indexes.where(s"name = 'index1' ").count == 1)
+    assert(hyperspace.indexes.where(s"name = 'index2' ").count == 1)
+
+    import com.microsoft.hyperspace._
+    spark.enableHyperspace
+
+    {
+      val filter = nonPartitionedDataDF.filter(col("nested").isNotNull).select("nested", "clicks")
```
clee704, line 219: Can you check if this filter works: "nested.id == 'id1'"

clee704 (commented):
Minor point, but I just checked it and it seems whether the StructType column is nested or not doesn't matter. It's just that, regardless they are nested or not, StructType columns don't work when they are specified as indexed or included columns.

src/test/scala/com/microsoft/hyperspace/index/CreateIndexNestedTest.scala
clee704: Indexing StructType columns don't make much sense, so I think it's safer to block it until there are strong use cases. Including StructType columns should be okay and it would be nice if it's supported.

By the way, I think the check for column types should be done by index implementations because different index types might support different column types.

sezruby: Seems StructType column can be used as a join column, so I think it's better not to block it..? That might cause some regression.

clee704: Okay, that seems a valid use case.

clee704 (approved):
LGTM, thanks

PR #468, opened 2021-06-21
Fix ExplainTest failure on Windows

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

ExplainTest fix for Windows. 
The expected string from file system contains "\r\n" instead of just "\n", so the string comparison fails because of remaining "\r".

No, test fix

test fix


clee704 (approved):
LGTM, thanks!

Issue #469, opened 2021-06-22
[FEATURE REQUEST]: Update azure-pipelines.yml to build on Windows


imback82:
**Feature requested**

Update azure-pipelines.yml to add building on Windows.

**Success criteria**

CI will build/test on Windows as well.

imback82:
I have a PR open for this (#376), but it needs to be updated. Anyone can pick this up if interested (just leave a comment in #376 that you will be working on it).

paryoja:
I've made PR for this issue. #470  Please review it.

imback82:
Thanks @paryoja!

PR #470, opened 2021-06-23
Add windows build pipeline

paryoja:

 - **Tracking Issue**: #376  #469

Using Windows environments to test in the azure pipeline.

No

By using the Azure pipeline configured.

### DIFF
--- a/ci/windows_test.yml
+++ b/ci/windows_test.yml
@@ -1,0 +1,29 @@
+parameters:
+  sparkVersion: ''
+  scalaVersion: ''
+
+steps:
+  - task: PowerShell@2
+    displayName: Download Winutils.exe
+    inputs:
+      workingDirectory: $(Build.BinariesDirectory)
+      pwsh: true
+      targetType: inline
+      script: |
+        echo "Download Hadoop utils for Windows."
+        curl -k -L -o hadoop.zip https://github.com/steveloughran/winutils/releases/download/tag_2017-08-29-hadoop-2.8.1-native/hadoop-2.8.1.zip
+        unzip hadoop.zip
+        New-Item -ItemType Directory -Force -Path hadoop\bin
+        cp hadoop-2.8.1\winutils.exe hadoop\bin
+        cp hadoop-2.8.1\hadoop.dll hadoop\bin
+        cp hadoop-2.8.1\hadoop.dll c:\windows\system32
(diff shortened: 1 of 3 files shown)


paryoja:
I added four windows jobs according to linux jobs already exist.
However, windows jobs (~20mins) takes more time than linux's (~12min).
I think it could be a bit annoyed if you wait for your job to be completed.

paryoja:
I think I should do some refactoring so that this pipeline could be easily used after new versions of Scala / Spark are released.

paryoja:
Refactoring Done.

ci/windows_test.yml
```diff
@@ -0,0 +1,32 @@
...
+      pwsh: true
+      targetType: inline
+      script: |
+        echo "Download Hadoop utils for Windows."
+        curl -k -L -o hadoop.zip https://github.com/steveloughran/winutils/releases/download/tag_2017-08-29-hadoop-2.8.1-native/hadoop-2.8.1.zip
+        unzip hadoop.zip
+        New-Item -ItemType Directory -Force -Path hadoop\bin
+        cp hadoop-2.8.1\winutils.exe hadoop\bin
+        cp hadoop-2.8.1\hadoop.dll hadoop\bin
+        cp hadoop-2.8.1\hadoop.dll c:\windows\system32
```
imback82, line 19: oh, this is required?

imback82 (approved):
LGTM (one question), thanks @paryoja!

imback82:
> I added four windows jobs according to linux jobs already exist.
> However, windows jobs (~20mins) takes more time than linux's (~12mins).
> I think it could be a bit annoyed if you wait for your job to be completed.

It should be OK for now, and we can enable Windows builds only for CI builds on master if the speed up is needed.

> I think I should do some refactoring so that this pipeline could be easily used after new versions of Scala / Spark are released.

Thanks for doing this!

ci/windows_test.yml
paryoja: Except for the spark 3.1 build, it is not necessary. (2.4, 3.0 tests success without copying it to system32 dir)

For the spark 3.1 build, it seems that sbt test doesn't reference hadoop/bin/hadoop.dll so it makes the tests fail.
I don't fully understand the reason. Maybe Spark 3.1 doesn't reference HADOOP_HOME to include dynamic libraries.

Issue #373, opened 2021-03-04
JsonMappingException on index create and read


JNKHunter:
Following directions here: 
https://microsoft.github.io/hyperspace/docs/ug-quick-start-guide/

### Describe the issue
CreateIndex job completes but with error below. Subsequent access of indexes result in same error.
```
com.fasterxml.jackson.databind.JsonMappingException: Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
```

### To Reproduce
```
val df = spark.read.parquet("path/to/parquet") //Creates dataframe with columns s,p,o
import com.microsoft.hyperspace._
val hs = new Hyperspace(spark)
import com.microsoft.hyperspace.index._
hs.createIndex(df, IndexConfig("index", indexedColumns = Seq("o","s"), includedColumns = Seq("p")))
```

Job starts and completes but with the following error:
 ```
JsonMappingExpection com.fasterxml.jackson.databind.JsonMappingException: Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
```

Subsequent read of index fails with same error:
```
val indexes = hs.indexes
```
Produces
```
com.fasterxml.jackson.databind.JsonMappingException: Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
```
### Expected behavior
Indexes should be readable and usable without JsonMappingException errors

### Environment 
OS: Linux RHEL
IDE: spark-shell
Apache Spark Version: 2.4.0
Platform: Cloudera CDH 6.3.3

**Comment by andrei-ionescu** [comment]
@JNKHunter Could you provide the parquet schema? 

You could do it with a command like this:

```
parquet-tools schema path/to/parquet
``` 

It would also be very helpful to have the following:
- full stack trace of the error
- a few example rows
- the version of the `jackson-databind` library that is loaded in Cloudera CDH

The exact steps that you followed in your Cloudera CDH 2.6.3 cluster, I did on my local machine and in Databricks cluster with  Runtime 5.5 and no issue there. 

It could be:

1. Character encoding issue due to unsupported char in some Spark actions
2. Library related error

We need more info from your side to have a better understanding of the cause.

andrei-ionescu:
@JNKHunter Any updates on this ticket?

mmuru-w:
I am getting the same exception. Here is the full stack trace.  
`
[...]

PR #472, opened 2021-06-28
Raise Hyperspace Exception with a path information when JSON parse error has occurred

paryoja:

 - **Dependencies**: Links to issues you depend on for this pull request to work. Write N/A, if no dependencies.
   - Issue 1 : #373 
   - Issue 2 : #431

Add error logs of the exact file that raises Exception to analyze the problem more easily.

No

Using UnitTest

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -71,8 +71,16 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration = 
     if (!fs.exists(path)) {
       return None
     }
+
     val contents = FileUtils.readContents(fs, path)
-    Some(LogEntry.fromJson(contents))
+
+    try {
+      Some(LogEntry.fromJson(contents))
+    } catch {
+      case e: Exception =>
+        logError(s"Cannot parse json with file path: ${path}")
+        throw e
+    }
   }
 
   override def getLog(id: Int): Option[LogEntry] = {
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -80,7 +80,7 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration = 
     } catch {
       case e: Exception =>
         throw HyperspaceException(
-          s"Cannot parse json with file path: ${path}, reason: ${e.getMessage}")
+          s"Cannot parse JSON in ${path}: ${e.getMessage}")
     }
   }
 
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
@@ -80,7 +80,7 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration = 
     } catch {
       case e: Exception =>
         throw HyperspaceException(
-          s"Cannot parse json with file path: ${path}, reason: ${e.getMessage()}")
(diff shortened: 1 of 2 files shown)


paryoja:
Since there are some issues regarding parsing Json logging files.
However, I found that it is not easy to pin point the problem since it is not easily reproducible.
So I added additional error log containing the log file path if there occurs exception while parsing the file.

@sezruby @imback82  I have some concerns within this PR.

1. Do you think this change is required? If so, am I following the rules regarding consensus of handling exceptions / logging errors?
2. Should I add unit test ensuring that the log is properly printed and containing the path when there are parsing errors?
3. My Scala formatter also changed lines in 57 / 105 with some indentations. Do you think my formatter setting is wrong? or it wasn't correct before?

sezruby:
1 => good to show the exact file path, but it's better to throw the message with HyperspaceException as users need to check driver log to see the error message. There's no specific rules for logging/exception yet afaik.
2 => good to add a unit test, but it's not mandatory.  
3 => seems it wasn't correct before

clee704:
@paryoja @sezruby Created #473 to handle the wrong formatting issue.

paryoja:
> 1 => good to show the exact file path, but it's better to throw the message with HyperspaceException as users need to check driver log to see the error message. There's no specific rules for logging/exception yet afaik.

Ok. I will change it to HyperspaceException.
BTW, I feel like HyperspaceException covers quite broad area of Exceptions. Is there any plan for make subclasses for exceptions?

> 2 => good to add a unit test, but it's not mandatory.
I will make some unit test also.

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -71,8 +72,16 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration =
...
+
     val contents = FileUtils.readContents(fs, path)
-    Some(LogEntry.fromJson(contents))
+
+    try {
+      Some(LogEntry.fromJson(contents))
+    } catch {
+      case e: Exception =>
+        logError(s"Cannot parse json with file path: ${path}")
+        throw HyperspaceException(s"Cannot parse json with file path: ${path}")
```
sezruby, line 83: Could you append e.getMessage() to the exception message?

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -71,8 +72,16 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration =
...
     }
+
     val contents = FileUtils.readContents(fs, path)
-    Some(LogEntry.fromJson(contents))
+
+    try {
+      Some(LogEntry.fromJson(contents))
+    } catch {
+      case e: Exception =>
+        logError(s"Cannot parse json with file path: ${path}")
```
sezruby, line 82: This seems redundant as we throw the exception

src/main/scala/com/microsoft/hyperspace/index/IndexLogManager.scala
```diff
@@ -71,8 +72,16 @@ class IndexLogManagerImpl(indexPath: Path, hadoopConfiguration: Configuration =
...
+
     val contents = FileUtils.readContents(fs, path)
-    Some(LogEntry.fromJson(contents))
+
+    try {
+      Some(LogEntry.fromJson(contents))
+    } catch {
+      case e: Exception =>
+        throw HyperspaceException(
+          s"Cannot parse json with file path: ${path}, reason: ${e.getMessage}")
```
sezruby, line 83: No specific rule, but for consistency:
```suggestion
          s"Cannot parse JSON in ${path}: ${e.getMessage}")
```

paryoja: Ok. Please let me know if the new title doesn't follow the consistency :)

sezruby (approved):
LGTM thanks!

PR #471, opened 2021-06-24
Documentation: coding style setting with examples

paryoja:
### What is the context for this pull request?

 - **Tracking Issue**: #426 

### What changes were proposed in this pull request?

Add detailed setting for scalafmt with Intellij.

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

Not required.

PR #473, opened 2021-06-28
Reformat files with scalafmt

clee704:
 - **Tracking Issue**: N/A
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

Reformat files with `scalafmt`, as in the following command:

```bash
$ scalafmt -c dev/.scalafmt.conf src
```

No.

Unit test


paryoja:
> $ scalafmt -c dev/.scalafmt.conf src

What if we add this command to pre-commit hook?
Every committers should set their environment to use pre-commit hook so it cannot be mandatory though.

clee704:
I think we can add a guideline in README or somewhere regarding scalafmt, and a git hook is a good one. There can be other ways, though, like running scalafmt automatically in your IDE or vim (using coc-metals).

clee704:
Seems we can use `sbt-scalafmt` to check the formatting. What do you think about adding the check to the build or CI to enforce the formatting? @imback82 @sezruby

paryoja:
> I think we can add a guideline in README or somewhere regarding scalafmt

I mentioned to use 'reformat on save' in the formatting guild committed in #471 but it wasn't obligation so I wonder how many committers read it carefully and set it :( 

sbt-scalafmt could be good. It is the project mentioned in the scalafmt guide. https://scalameta.org/scalafmt/docs/installation.html

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -224,12 +224,10 @@ object Directory {
    * @param files List of leaf files.
    * @param fileIdTracker FileIdTracker to keep mapping of file properties to assigned file ids.
    *                      Note: If a new leaf file is discovered, the input fileIdTracker gets
-    *                     updated by adding it to the files it is tracking.
+   *                     updated by adding it to the files it is tracking.
```
sezruby, line 227:
```suggestion
   *                      updated by adding it to the files it is tracking.
```

sezruby (commented):
Thanks for doing this 👍 Other than the comment, LGTM

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
clee704: Done

sezruby (approved):

clee704:
Opened #476 for enforcing/checking scalafmt during build/CI

PR #474, opened 2021-06-28
Refactoring for an extensible Index API: Part 2

clee704:
### What is the context for this pull request?
 - **Tracking Issue**: #441 
 - **Parent Issue**: N/A
 - **Dependencies**:
   - #473 

### What changes were proposed in this pull request?
- Change the return type of refreshIncremental so that implementations
  can choose what the framework should do after the incremental refresh
  regarding the index data files.
- Move CoveringIndex.resolveConfig to IndexUtils.resolveColumns to make
  it available to other index implementations.

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
With existing tests

PR #475, opened 2021-06-28
Refactoring for an extensible Index API: Part 3

clee704:
 - **Tracking Issue**: #441 
 - **Parent Issue**: N/A
 - **Dependencies**:
   - #473 
   - #474 

- Move CoveringIndex specific code into the
  com.microsoft.hyperspace.index.types.covering package
- Move traits/classes/objects in the
  com.microsoft.hyperspace.index.rules package into their own files
- Add alias for CoveringIndexConfig for backward compatibility

No

With existing tests


src/test/scala/com/microsoft/hyperspace/index/IndexLogEntryTest.scala
```diff
@@ -76,7 +77,7 @@ class IndexLogEntryTest extends HyperspaceSuite with SQLHelper {
          |{
          |  "name" : "indexName",
          |  "derivedDataset" : {
-         |    "type" : "com.microsoft.hyperspace.index.CoveringIndex",
+         |    "type" : "com.microsoft.hyperspace.index.types.covering.CoveringIndex",
```
sezruby, line 80: Can we do `types.CoveringIndex` instead of `types.covering.coveringIndex`? 
hmm.. is that for another type of covering index?

clee704: I'd like to give each index type their own package, like how Spark put data source implementations in the `org.apache.spark.sql.execution.datasources.{parquet, orc, json, csv}` packages.

If this seems verbose, then how about `com.microsoft.hyperspace.indexes.covering`? That is, `com.microsoft.hyperspace.index` is for index-related stuff, and `com.microsoft.hyperspace.indexes` is for index implementations.

Note that data skipping indexes will be put in the `dataskipping` package.

sezruby: I pointed it out since the package string will be shown in index log entry. I think Option 3 looks good because it's shorter.
cc @imback82 @paryoja

sezruby:
Other than the comment, LGTM thanks @clee704 👍 
Could you resolve the conflict?

clee704:
Changed `com.microsoft.hyperspace.index.types.covering` to `com.microsoft.hyperspace.index.covering`.

python/hyperspace/indexconfig.py
```diff
@@ -1,14 +1,17 @@
...
+        :return: CoveringIndexConfig object
 
-        >>> idxConfig = IndexConfig("indexName", ["c1"], ["c2","c3"])
+        >>> idxConfig = CoveringIndexConfig("indexName", ["c1"], ["c2","c3"])
         """
         self.indexName = indexName
         self.indexedColumns = indexedColumns
         self.includedColumns = includedColumns
+
+# TODO(ChungminL): Add deprecation warning
```
sezruby, line 16: Will you resolve this soon? Otherwise please create an issue for this and add its link here.
In any case it's not good to add your alias here

clee704: Actually specifying names for TODO comments is a thing that many developers do for a good reason. To quote the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#TODO_Comments):

> TODOs should include the string TODO in all caps, followed by the name, e-mail address, bug ID, or other identifier of the person or issue with the best context about the problem referenced by the TODO. The main purpose is to have a consistent TODO that can be searched to find out how to get more details upon request. A TODO is not a commitment that the person referenced will fix the problem. Thus when you create a TODO with a name, it is almost always your name that is given.

sezruby: I searched Spark codebase and the first TODO with a name was committed 2 years ago. Let's be consistent with the current TODO policy - an issue link or a follow-up PR soon.

clee704: I didn't know there is a TODO policy - could you share a link or something?

The issue is created as #477.

sezruby: No documentation yet😅 but recently added TODOs have issue links. (via code review) 

I suggest to add the issue link like below, but I'm fine w/o the comment as the issue itself explains well. 

```
# TODO: Add deprecation warning.
#   See https://github.com/microsoft/hyperspace/issues/477
```

sezruby (approved):

Issue #118, opened 2020-08-07
[PROPOSAL]: deleteOldVersions API


sezruby:
**Describe the problem**
After refreshing an index,  the old version of the index remains on the storage. We should keep the old versions to support consistency & isolation of index data, but at some point in time, they're no longer needed.
So it would be good if there's an API to clean up the old versions.
 
**Describe your proposed solution**

API design
```
// API with hyperspace
hs.deleteOldVersions(indexName)
==
// for an index
def deleteOldVersions(indexName: String)

// for multiple indexes
def deleteOldVersions(indexNames: Seq[String])

// for clean-up globally
def deleteOldVersions(indexNames: Seq[String]=Seq(), global: Boolean=false)
```

**Describe alternatives you've considered**
It might be useful to delete a specific version of an index? 
```
def deleteOldVersion(indexName: String, versionID: int)
```

But there's no API to show the list of versions. I think it would be great to provide an API for statistics of an index so that a user can check [ size of index / existing versions / creation time / last used time(from event log).. etc]

**Additional context**

paryoja:
> But there's no API to show the list of versions. I think it would be great to provide an API for statistics of an index so that a user can check [ size of index / existing versions / creation time / last used time(from event log).. etc]

Why this sentences are struck out? Is there any API to show the list of versions? Or is it another feature suggestion?

sezruby:
Yea `hs.index("indexName")` was added after the sentences.

paryoja:
@sezruby Is there any API shows history of that index? `hs.index("indexName")` seems to show only the latest stable log.

sezruby:
No but there's an internal API - `getIndexLogEntry` and you can use `getIndexContentDirectoryPaths`.

BTW I realized that the following description is invalid now. 

hs.deleteOldVersions("indexName", Seq(0, 1, 3)) // remove v__=0, v__=1, v__=3 dir

There are several problem here:
- "version" doesn't mean the index data directory number (v__*)
- one index log entry can refer multiple "v__*" directory, because of incremental refresh
- delta lake time travel query can refer the old version of index data.
- Hyperspace currently doesn't check the existence of index files at query time, for some reason.

So, I would suggest the following to make things simple:
- API:
```
def deleteOldVersions(indexName: String)
hs.deleteOldIndexData("indexName")
```
- remove all the directory that the latest version doesn't refer
- for delta lake source, update the history property in IndexLogEntry, "deltaVersions"
[...]

PR #479, opened 2021-07-12
Add another behavior of VacuumIndex: VacuumIndex will clean up the index when its state is ACTIVE.

paryoja:

 - **Tracking Issue**: #118 

VacuumIndex will clean up old versions index data if it is invoked with the index whose state is ACTIVE.

Yes. Users now can use VacuumIndex api to clear disk files containing the outdated data of an index.

By using unit test.


paryoja:
I just deleted every index except the latest version. 
I am not sure this will make side effects if there is any query dealing with old indexes.
Do we have such cases? @sezruby

src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
```diff
@@ -38,6 +38,8 @@ import com.microsoft.hyperspace.util.FileUtils
 trait IndexDataManager {
   def getLatestVersionId(): Option[Int]
 
+  def getAllVersionIds(): Seq[Int]
```
paryoja, line 41: I added one public api that can list all version ids of indexes in the disk

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -56,6 +56,16 @@ case class Content(root: Directory, fingerprint: NoOpFingerprint = NoOpFingerpri
         FileInfo(new Path(prefix, f.name).toString, f.size, f.modifiedTime, f.id)).toSet
   }
 
+  @JsonIgnore
+  lazy val versionInfos: Set[Int] = {
```
paryoja, line 60: In the log entry, it is hard to find which versions are actually used. 
Maybe it is better to track ids that are currently used in a class variable of IndexLogEntry (or maybe LogEntry).

src/main/scala/com/microsoft/hyperspace/actions/DeleteOldVersionsAction.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  final override def op(): Unit = {
+    // get versions used
+    val indexVersionsInUse: Set[Int] = logEntry match {
+      case indexLogEntry: IndexLogEntry =>
+        indexLogEntry.content.versionInfos
+      case other =>
+        throw HyperspaceException(
+          s"DeleteOldVersions is not defined with log entry class ${other.getClass.getName}")
```
paryoja, line 59: One thing I found is that 
trait LogEntry has no information which versions are used.
So, if someone wants to add another inherited class of LogEntry (a sibling of IndexLogEntry), they should implement something like indexLogEntry.content.versionInfos.

I think it would be better to add a method like `usedVersionIds` to to LogEntry

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
sezruby: Could you utilize `private def getIndexContentDirectoryPaths(entry: IndexLogEntry): Seq[String] = {` in IndexStatistics?
It would be good to move the function to here.

src/test/scala/com/microsoft/hyperspace/actions/VacuumOutdatedDataActionTest.scala
```diff
@@ -0,0 +1,84 @@
...
+  private val mockLogManager: IndexLogManager = mock(classOf[IndexLogManager])
+  private val mockDataManager: IndexDataManager = mock(classOf[IndexDataManager])
+  private val mockIndexLogEntry: IndexLogEntry = mock(classOf[IndexLogEntry])
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    when(mockLogManager.getLatestId()).thenReturn(None)
+  }
+
+  test("validate() passes if old index logs are found with ACTIVE state") {
```
sezruby, line 38: nit: ditto for others
```suggestion
  test("validate() passes if old index logs are found with ACTIVE state.") {
```

src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
```diff
@@ -48,22 +50,28 @@ class IndexDataManagerImpl(indexPath: Path, configuration: Configuration)
...
 
+  override def getLatestVersionId(): Option[Int] = {
+    val ids = getAllVersionIds()
+    if (ids.isEmpty) None else Some(ids.max)
+  }
+
   /**
    * This method relies on the naming convention that directory name will be similar to hive
-   * partitioning scheme, i.e. "root/v__=value/f1.parquet" etc. Here the value represents the
+   * partitioning scheme, i.e. {{{"root/v__=value/f1.parquet"}}} etc. Here the value represents the
```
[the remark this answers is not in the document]
paryoja: Since "__" interpreted as bold (i.e., \_\_abc\_\_ -> __abc__ ), it seems that it is not shown correctly due to 'v\_\_' part.

src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
```diff
@@ -48,22 +50,28 @@ class IndexDataManagerImpl(indexPath: Path, configuration: Configuration)
...
    * version id of the data.
    */
-  override def getLatestVersionId(): Option[Int] = {
+  override def getAllVersionIds(): Set[Int] = {
```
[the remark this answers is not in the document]
paryoja: Do you prefer Seq instead?

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -513,6 +513,43 @@ case class IndexLogEntry(
...
+    directory.subDirs.map(d => s"$prefix${d.name}")
+  }
+
+  /**
+   * This function extracts latest versions of an index.
+   *
+   * @return List of directory paths containing index files for latest index version.
+   */
+  @JsonIgnore
+  lazy val versionInfos: Set[Int] = {
```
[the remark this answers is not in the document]
paryoja: Ok. I will.

src/main/scala/com/microsoft/hyperspace/index/IndexLogEntry.scala
```diff
@@ -513,6 +513,43 @@ case class IndexLogEntry(
     (name, derivedDataset, signature, content).hashCode
   }
 
+  /**
+   * This function extracts paths to top-level directories paths which
+   * contain the latest version index files.
+   *
+   * @return List of directory paths containing index files for latest index version.
+   */
+  def versionDirectories(): Seq[String] = {
```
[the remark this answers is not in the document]
paryoja: I will change the name.

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedDataAction.scala
```diff
@@ -0,0 +1,73 @@
...
+        throw HyperspaceException(
+          s"VacuumOutdatedData is not defined with log entry class ${other.getClass.getName}")
+    }
+
+    // delete version not used
+    dataManager.getAllVersionIds().foreach { id =>
+      if (!indexVersionsInUse.contains(id)) {
+        dataManager.delete(id)
+      }
+    }
```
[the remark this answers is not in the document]
paryoja: Ok. I will.

python/hyperspace/hyperspace.py
```diff
@@ -103,6 +103,16 @@ def vacuumIndex(self, indexName):
         """
         self.hyperspace.vacuumIndex(indexName)
 
+    def vacuumOutdatedDataIndex(self, indexName):
```
clee704, line 106: nit: for conciseness and consistency, a shorter verb would be nice, such as "clean" (or "clean up"), "prune", "trim", or "shrink".

src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
```diff
@@ -48,22 +50,27 @@ class IndexDataManagerImpl(indexPath: Path, configuration: Configuration)
...
    * This method relies on the naming convention that directory name will be similar to hive
-   * partitioning scheme, i.e. "root/v__=value/f1.parquet" etc. Here the value represents the
+   * partitioning scheme, i.e. {{{"root/v__=value/f1.parquet"}}} etc. Here the value represents the
    * version id of the data.
    */
-  override def getLatestVersionId(): Option[Int] = {
+  override def getAllDataVersionIds(): Seq[Int] = {
     if (!fs.exists(indexPath)) {
-      return None
+      return Seq()
```
clee704, line 65: nit: `return Nil`

python/hyperspace/hyperspace.py
sezruby: As we discussed, using a flag in vacuumIndex
```scala
vacuumIndex(indexName: String, exceptLatest: Boolean = False)
```
or I think `shrinkIndex` looks better to me

.gitignore
```diff
@@ -47,3 +46,6 @@ build.log
 /spark2.4/src/*
 /spark3.0/src/*
 /spark3.1/src/*
+
+# in case you have python venv
+/venv
```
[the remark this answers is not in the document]
paryoja: Oh. I will use that instead.

python/hyperspace/hyperspace.py
paryoja: Should I use shrinkIndex?

src/main/scala/com/microsoft/hyperspace/index/IndexDataManager.scala
```diff
@@ -38,6 +38,8 @@ import com.microsoft.hyperspace.util.FileUtils
 trait IndexDataManager {
   def getLatestVersionId(): Option[Int]
 
+  def getAllDataVersionIds(): Seq[Int]
```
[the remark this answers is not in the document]
paryoja: I will change the name.

python/hyperspace/hyperspace.py
clee704: IMHO, putting this feature in the same API function as "vacuumIndex", is like putting "deleteIndex" and "vacuumIndex" in the same function. They perform similar actions, but different enough to be called by different names.

Regarding the word choice, I'm not a native speaker, but it seems "clean (up)" or "prune" is more appropriate here for the feature than "shrink". Maybe we'd better consult a native English speaker on this matter.

imback82: > IMHO, putting this feature in the same API function as "vacuumIndex", is like putting "deleteIndex" and "vacuumIndex" in the same function. They perform similar actions, but different enough to be called by different names.

cc @rapoth who suggested using `vacuumIndex` with a flag.

paryoja:
@sezruby @clee704  

I tried to make a test code to check where FileUtils.delete is actually called with expected arguments.

For example, if there are "file:/a/b/v__=1/part-a", "file:/a/b/v__=1/part-b" in the file system and
`IndexLogEntry.content` contains only the "file:/a/b/v__=1/part-b" then I should check whether deleting "file:/a/b/v__=1/part-a" is executed.


Since I used `FileUtils` to delete file and `FileUtils` is an object, I tried https://github.com/mockito/mockito-scala#mocking-scala-object but I cannot compile correctly with `withObjectMocked` even though I added `src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker` .

Do you have any idea to test this? Maybe I can add helper function in DataManager to delete file but the function will just proxy the delete file something like below and it could be confusing with the existing delete function

```
class IndexDataManagerImpl {
  override def delete(f: Path): Unit = FileUtils.delete(f)
  override def delete(id: Int): Unit = FileUtils.delete(getPath(id))
}
```

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
```diff
@@ -0,0 +1,111 @@
...
+    val filesInUse = logEntry match {
+      case indexLogEntry: IndexLogEntry =>
+        indexLogEntry.content.fileInfos.map { info =>
+          info.name
+        }
+    }
+
+    // Delete unused files
+    dataManager.getAllFilePaths().foreach { path =>
+      if (path.getName.startsWith("part-") && !filesInUse.contains(path.toString)) {
```
sezruby, line 82: All file paths will start with "part-" so I don't think the condition is necessary.

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
```diff
@@ -0,0 +1,111 @@
...
+      if (!indexVersionsInUse.contains(id)) {
+        dataManager.delete(id)
+      }
+    }
+
+    val filesInUse = logEntry match {
+      case indexLogEntry: IndexLogEntry =>
+        indexLogEntry.content.fileInfos.map { info =>
+          info.name
+        }
```
sezruby, line 77:
```suggestion
        }.toSet
```
for quick lookup

python/hyperspace/tests/test_indexmanagement.py
```diff
@@ -89,6 +95,7 @@ def test_index_metadata(self):
         df = self.hyperspace.index('idx1')
         df.show()
 
+
```
sezruby, line 98: nit: --

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  /**
+   * Extracts latest versions of an index.
+   *
+   * @return List of directory paths containing index files for latest index version.
+   */
+  private[actions] def dataVersionInfos(entry: IndexLogEntry): Set[Int] = {
+    // get used versions using the filenames of contents
```
sezruby, line 95: nit: same for other comment
```suggestion
    // Get used versions using the filenames of contents.
```

src/main/scala/com/microsoft/hyperspace/index/IndexManager.scala
```diff
@@ -54,7 +54,9 @@ trait IndexManager {
   def restore(indexName: String): Unit
 
   /**
-   * Does hard delete of indexes marked as `DELETED`. Once vacuumed, an index can't be 'restore'd.
+   * Does hard delete of indexes if it is marked as `DELETED`.
+   * Does clean up of indexes (hard delete of unused indexes) if it is 'ACTIVE'.
```
sezruby, line 58: Could you revise the comment? it would be good these two lines in one sentence.

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
```diff
@@ -84,6 +84,16 @@ case class RestoreActionEvent(appInfo: AppInfo, index: IndexLogEntry, message: S
 case class VacuumActionEvent(appInfo: AppInfo, index: IndexLogEntry, message: String)
     extends HyperspaceIndexCRUDEvent
 
+/**
+ * Deletion of old indexes event. Emitted when vacuum is called on an ACTIVE index.
```
sezruby, line 88:
```suggestion
 * Deletion of old index files event. Emitted when vacuum is called on an ACTIVE index.
```

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Algorithm:
+ *  - Delete every version except the latest versions.
+ */
+class VacuumOutdatedAction(
+    final override protected val logManager: IndexLogManager,
+    dataManager: IndexDataManager)
+    extends Action {
+  override def logEntry: LogEntry = {
+    logManager.getLog(baseId).getOrElse {
```
sezruby, line 38: Maybe you need to update delta lake version info property in index log entry here. 
Could you add a test for time travel query after vacuuming outdated files?

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
paryoja: Hmm. There was a file "_SUCCESS" to indicate success of tasks. Is it ok to delete it?

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
paryoja: Sicne `fileInfos` is `Set[FileInfo]` so it seems to be redundant.

python/hyperspace/tests/test_indexmanagement.py
paryoja: according to PEP8, (common guildline for python formatting)

https://www.python.org/dev/peps/pep-0008/#blank-lines  says that


> Surround top-level function and class definitions with two blank lines. 

And this is default behavior of python formatters such as black formatter.

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
paryoja: Ok. I will make a test for it.

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
paryoja: I've changed it. Thanks :)

src/main/scala/com/microsoft/hyperspace/index/IndexManager.scala
paryoja: How about the new one?

src/main/scala/com/microsoft/hyperspace/telemetry/HyperspaceEvent.scala
paryoja: Fixed it.

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  final override lazy val logEntry: LogEntry = {
+    val relationMetadata = Hyperspace
+      .getContext(spark)
+      .sourceProviderManager
+      .getRelationMetadata(previousIndexLogEntry.relations.head)
+
+    val updatedDerivedDataset = relationMetadata match {
+      case deltaLakeRelationMetadata: DeltaLakeRelationMetadata =>
+        // Reset Delta Lake version mapping.
```
paryoja, line 54: Delta lake version is handled here.

src/main/scala/com/microsoft/hyperspace/actions/VacuumOutdatedAction.scala
paryoja: I added a test. Please check it.

src/main/scala/com/microsoft/hyperspace/index/IndexCollectionManager.scala
```diff
@@ -60,13 +60,23 @@ class IndexCollectionManager(
...
       val hadoopConf = spark.sessionState.newHadoopConf()
       val indexPath = PathResolver(spark.sessionState.conf, hadoopConf)
         .getIndexPath(indexName)
       val dataManager =
         indexDataManagerFactory.create(indexPath, hadoopConf)
-      new VacuumAction(logManager, dataManager).run()
+
+      logManager.getLatestLog() match {
+        case Some(index) if index.state == ACTIVE =>
+          // clean up only if state is ACTIVE.
```
sezruby, line 75: nit
```suggestion
          // Clean up only if state is ACTIVE.
```

src/test/scala/com/microsoft/hyperspace/index/DeltaLakeIntegrationTest.scala
```diff
@@ -457,6 +457,74 @@ class DeltaLakeIntegrationTest extends QueryTest with HyperspaceSuite {
...
+          // delta version 7, index log version 7 (optimize).
+          appendAndRefresh(df, path, indexName, Some("incremental"), tsMap)
+          // delta version 8, index log version 9 (refresh).
+          hyperspace.optimizeIndex(indexName)
+          // delta version 8, index log version 11 (optimize).
+          appendAndRefresh(df, path, indexName, None, tsMap)
+          // delta version 9, index long version 11.
+
+          // calling vacuumIndex on active index deletes outdated data and history.
+          hyperspace.vacuumIndex(indexName)
```
sezruby, line 511: Could you check if outdated files are removed - using listFiles?

paryoja: Ok. I will.

sezruby (approved):
LGTM, thanks @paryoja!

Issue #297, opened 2020-12-15
No module named Hyperspace on AWS EMR


mmchougule:
### Describe the issue
Great project, I am trying to use this in pyspark on AWS EMR. I have added `spark.jars.packages: "com.microsoft.hyperspace:hyperspace-core_2.11:0.1.0"` in my spark config and is used when creating a SparkSession with spark master on `yarn`.
When I do `from hyperspace import Hyperspace` I am getting an error `No module named hyperspace`. I do see that it has downloaded the hyperspace jar but for some reason, my code isn't recognizing this library in python. Any ideas what I could be doing wrong?

### To Reproduce
1. Add hyperspace jar when creating a spark session in jupyter notebook or in the project code
2. Add `from hyperspace import Hyperspace`, you will see the error.

### Expected behavior
I should be able to import Hyperspace so I can use it.

### Environment 
AWS EMR, centos VM

apoorvedave1:
Hi @mmchougule , could you please try the following; 

C:\path\to\hyperspace\jar> pyspark --jars hyperspace-core_2.11-<version>.jar --files hyperspace-core_2.11-<version>.jar
…
Spark session start logs
…
>>> from hyperspace import *
>>> hyperspace = Hyperspace(spark)
>>>

mmchougule:
Hi @apoorvedave1 thanks for getting back to me. This works for me in the command line. I see the issue  I run it in my code or in a jupyter notebook and try to create a SparkSession with spark.jars as the config. Am I doing it wrong? I also tried passing the path of the jar file on the VM. It seems like it did download the jar correctly just not enabling the `hyperspace` package for some reason.
```
spark = (
    SparkSession.builder.master("yarn")
    .config("spark.jars", "com.microsoft.hyperspace:hyperspace-core_2.11:0.3.0")
    .appName("myapp")
    .getOrCreate()
)
 
```

apoorvedave1:
thanks @mmchougule , I am not really sure how to move forward in this case. Could you try replacing hyperspace jar with any other external jar in your environment? 
I am not familiar with setting up jupyter with external jars and just wanted to rule out the possibility that it's a problem with hyperspace jar or just its availability in the environment.

apoorvedave1:
We would also love your contribution if you manage to figure out how to setup Jupyter with hyperspace. Please feel free to add instructions to do it if it works for you!

mmchougule:
[...]

PR #489, opened 2021-08-10
Fixing bug where large index files weren't being read fully

alex-shchetkov:
I ran into an issue where I was unable to use any of the created indexes, due to a Json Parser claiming it encountered invalid chars. 

This was misleading, because the actual issue was that only a portion of the index file was being read.

 - **Tracking Issue**: N/A
 - **Parent Issue**: N/A
 - **Dependencies**: N/A

Changing the FileSystem.read() to a FileSystem.readFully().
This is because using .read() does not always read in the full file. 

This bug fix very likely fixes these:
https://github.com/microsoft/hyperspace/discussions/431
https://github.com/microsoft/hyperspace/issues/373
https://github.com/microsoft/hyperspace/issues/297#issuecomment-747502799 (point #2)

No

I compiled/packaged the code and ran it on an EMR (spark 3.1) cluster to generate a relatively large (8MB in my case) index file in an s3 location
With this change I was able to use the index to run a query.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/util/FileUtils.scala
+++ b/src/main/scala/com/microsoft/hyperspace/util/FileUtils.scala
@@ -117,7 +117,7 @@ object FileUtils {
     val fileInputStream = fileSystem.open(path)
     val len = fileSystem.getFileStatus(path).getLen
     val bytes = new Array[Byte](len.toInt)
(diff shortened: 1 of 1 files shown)


sezruby (approved):
Thanks for providing the fix!

PR #493, opened 2021-09-06
Introduce utility function to analyze min/max data layout

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a
 - **Dependencies**: n/a

Introduce utility function to analyze the data layout of the given source data.

Spark and parquet perform min/max pruning based on statistics, to improve query performance.
This function helps to understand the physical layout of column values by showing some statistics based on min/max for each file.

The function returns html format or text format. 

This is an example of HTML format result:

```scala
displayHTML(DataAnalysisUtils.minMaxAnalysis(df, Seq("colA"), format="html"))
```
quickstart-optimizezorderby-result-colA-0

An example for text result:
```scala
println(DataAnalysisUtils.minMaxAnalysis(df, Seq("colA"), format="text"))
```
```
Min/Max analysis on colA

                  < Number of files (%) >
     +--------------------------------------------------+
100% |                                                  |
     |                                                  |
     |                                                  |
     |                                                  |
 75% |                                                  |
     |                                                  |
     |                                                  |
     |                                                  |
     |                                                  |
 50% |                                                  |
     |                                                  |
     |                                                  |
     |                                                  |
     |                                                  |
 25% |                                                  |
     |   ********  *****     **************             |
     |  **********************************************  |
     |**************************************************|
     |**************************************************|
  0% |**************************************************|
     +--------------------------------------------------+
     Min <-----           colA value           -----> Max

min(colA): 0
max(colA): 9999999
Total num of files: 144
Total byte size of files: 416776899
Max. num of files for a point lookup: 36 (25.00%)
Estimated average num of files for a point lookup: 29.44 (20.44%)
Max. bytes to read for a point lookup: 104095313 (24.98%)
```

The utility function only supports NumericType. 

Yes, provide an utility function for min/max analysis.

via Notebook.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/util/MinMaxAnalysisUtil.scala
+++ b/src/main/scala/com/microsoft/hyperspace/util/MinMaxAnalysisUtil.scala
@@ -1,0 +1,780 @@
+/*
+ * Copyright (2021) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.util
+
+import scala.collection.mutable
+
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute
+import org.apache.spark.sql.catalyst.util.TypeUtils
+import org.apache.spark.sql.functions.{col, input_file_name, max, min}
+import org.apache.spark.sql.types.{StructField, StructType}
+
+import com.microsoft.hyperspace.HyperspaceException
+
+case class MinMaxAnalysisResult(
+    colName: String,
+    maxNumFilesForValueRanges: Seq[Int],
+    minVal: String,
+    maxVal: String,
+    totalNumFiles: Int,
+    totalBytes: Long,
+    maxBytesToLookup: Long) {
+  lazy val maxNumFilesToLookup: Int = {
+    if (maxNumFilesForValueRanges.nonEmpty) {
+      maxNumFilesForValueRanges.max
+    } else {
+      -1
+    }
+  }
+
+  lazy val avgNumFilesToLookup: Double = {
+    maxNumFilesForValueRanges.sum / maxNumFilesForValueRanges.count(_ > 0).toDouble
+  }
+
+  def appendSummary(newline: String, stringBuilder: StringBuilder): Unit = {
+    if (minVal != null) {
+      Seq(
+        s"min($colName): $minVal",
+        s"max($colName): $maxVal",
+        s"Total num of files: $totalNumFiles",
+        s"Total byte size of files: $totalBytes",
+        s"Max. num of files for a point lookup: $maxNumFilesToLookup (" +
+          "%.2f".format((maxNumFilesToLookup / totalNumFiles.toDouble) * 100) +
+          "%)",
+        s"Estimated average num of files for a point lookup: " +
+          "%.2f".format(avgNumFilesToLookup) + " (" +
+          "%.2f".format((avgNumFilesToLookup / totalNumFiles) * 100) + "%)",
+        s"Max. bytes to read for a point lookup: $maxBytesToLookup (" +
+          "%.2f".format((maxBytesToLookup / totalBytes.toDouble) * 100) + "%)")
+        .foreach { line =>
+          stringBuilder.append(line)
+          stringBuilder.append(newline)
+        }
+    } else {
+      Seq(
+        s"min($colName): $minVal",
+        s"max($colName): $maxVal",
+        s"Total num of files: $totalNumFiles",
+        s"Total byte size of files: $totalBytes")
+        .foreach { line =>
+          stringBuilder.append(line)
+          stringBuilder.append(newline)
+        }
+    }
+  }
+}
+
+object MinMaxAnalysisResult {
+  def allNullValueResult(colName: String, numFiles: Int, numBytes: Long): MinMaxAnalysisResult = {
+    MinMaxAnalysisResult(colName, Nil, null, null, numFiles, numBytes, 0)
+  }
+}
+
+trait MinMaxResultWriter {
+  val stringBuilder = new StringBuilder
+  val newline: String
+  val distFigureWidth: Int
+  def appendEmptyLine(): Unit = {
+    stringBuilder.append(newline)
+  }
+  def appendLine(line: String): Unit = {
+    stringBuilder.append(line)
+    stringBuilder.append(newline)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/microsoft/hyperspace/util/DataAnalysisUtils.scala
```diff
@@ -0,0 +1,261 @@
...
+    val dataType = df.schema.fields.find(_.name.equals(colName)).get.dataType
+    val numeric = try {
+      TypeUtils.getNumeric(dataType)
+    } catch {
+      case _: Exception =>
+        throw HyperspaceException(
+          s"minMaxAnalysis only supports Numeric type. colName=$colName, dataType=$dataType")
+    }
+
+    val doubleNumeric = TypeUtils.getNumeric(DoubleType)
```
paryoja, line 118: doubleNumeric seems that it isn't used elsewhere.

src/main/scala/com/microsoft/hyperspace/util/DataAnalysisUtils.scala
```diff
@@ -0,0 +1,278 @@
...
+        throw HyperspaceException(
+          s"minMaxAnalysis only supports Numeric type. colName=$colName, dataType=$dataType")
+    }
+
+    val distFigureLen = if (htmlFormat) 20 else 50
+    val newline = if (htmlFormat) "</br>" else System.lineSeparator()
+    val res = df.inputFiles.par.map { f =>
+      val file = df.sparkSession.read.parquet(f)
+      f -> file.agg(min(col(colName)), max(col(colName))).collect.toSeq.headOption
+    }.toIndexedSeq.filter(_._2.isDefined)
```
clee704, line 122: This is interesting. Currently I'm using `groupBy(input_file_name()).agg(...)` for data skipping indexes but also considering this approach, but with Spark RDD to parallelize. Did you consider using `spark.sparkContext.parallelize` instead of Scala's `.par`? My understanding is that if you use `.par` then Spark can't efficiently schedule jobs because the parallelization is opaque to Spark.

src/main/scala/com/microsoft/hyperspace/util/DataAnalysisUtils.scala
```diff
@@ -0,0 +1,278 @@
...
+
+    /* groupBy shuffles all data, though it can be done with one spark job.
```
clee704, line 124: Are you sure about this? There is no reason that Spark should shuffle "all data" when data is grouped by `input_file_name()`. All that shuffled is just the aggregated data, which is much less than the source data.

src/main/scala/com/microsoft/hyperspace/util/DataAnalysisUtils.scala
```diff
@@ -0,0 +1,278 @@
...
+      stringBuilder.append("*" * (endRangeFixed - startRangeFixed))
+      stringBuilder.append(" " * (distFigureLen - endRangeFixed))
+      stringBuilder.append("|")
+      stringBuilder.append(newline)
+    }
+    stringBuilder.append("+")
+    stringBuilder.append("-" * (distFigureLen))
+    stringBuilder.append("+")
+    stringBuilder.append(newline)
+     */
```
clee704, line 174: I thought this PR is ready to be merged once reviewed? Should this comment be committed?

clee704:
In the textual form, Y-axis has labels "0%" and "100%", whereas there are actual values from 0 to 240 in the HTML form. Why the difference? It would be more consistent if both forms have the same content.

clee704:
The chart produced by the function looks very nice. But how about plotting min-max ranges for each file as we did in previous presentations (optionally sorted by min, max, or file path) in addition to the chart? That would explain the "why" of the chart, and some users like me might like to know more details.

src/main/scala/com/microsoft/hyperspace/util/DataAnalysisUtils.scala
sezruby: current ver - ".par.map" triggers a new spark job for each file with multiple threads. It can be done by multiple tasks if it's a large file. So parallelism is not an issue for this; actually this is faster than groupBy impl. (though not sure why groupBy is slower; maybe it's just because "input_file_name()" function ..?)
One problem is that the impl incurs numerous spark jobs if there're many files which can cause some complexity of spark history ui or other analysis tools. Considering the fact, I would prefer groupBy impl. though it's bit slower than this.

For `spark.sparkContext.parallize`, I'm not sure how I can use it instead of `.par`

src/main/scala/com/microsoft/hyperspace/util/DataAnalysisUtils.scala
sezruby: yea seems spark does the aggregation before groupBy exchange. I'll remove the comment.

sezruby:
> In the textual form, Y-axis has labels "0%" and "100%", whereas there are actual values from 0 to 240 in the HTML form. Why the difference? It would be more consistent if both forms have the same content.
The text form is not that flexible(?) so I wrote %, but in html form I can use 0 to max number of files. 
I agree about that consistency; will try to change html form to 0%-100%

> The chart produced by the function looks very nice. But how about plotting min-max ranges for each file as we did in previous presentations (optionally sorted by min, max, or file path) in addition to the chart? That would explain the "why" of the chart, and some users like me might like to know more details.

I think we can extend the feature later :) I tried to show min-max range of each file first (there's some commented out code for file level  distribution) but it's bit tricky to show it in text format in case there're millions of files.

src/main/scala/com/microsoft/hyperspace/util/DataAnalysisUtils.scala
clee704: `spark.sparkContext.parallelize(df.inputFiles)` gives an RDD than you can work on. It is better than Scala's built-in parallel collections because Spark can efficiently schedule jobs. For Parquet files, you can directly read min/max stats with `ParquetFileReader` without reading the whole files.

```
def collectMinMax(path: String, colName: String): (Any, Any) = {
   // Read min/max using ParquetFileReader
}
spark.sparkContext.parallelize(df.inputFiles).map(collectMinMax(_, colName))
```

Unfortunately, I couldn't find yet how you can populate DF in the map function. The map function is executed in executor nodes so obviously `spark.read` doesn't work. But even if it only works for Parquet files, I think it's worth a try since the speed up would be huge and Parquet is a very popular format. This can be done for other formats supporting statistics such as ORC.

sezruby: I think we can use groupBy / min&max aggregation for now, as it can support most type of data sources. We can optimize the function for delta lake, but for hyperspace, let's keep the current impl. for the initial version and extend the feature later on demand.

clee704:
I think we can split the code into two parts, data/histogram generation and presentation, with the presentation part being specialized for two variants, html and text. It would be easier to evolve the code later if we need to. Until then, it seems just okay for now.

src/main/scala/com/microsoft/hyperspace/util/DataAnalysisUtils.scala
```diff
@@ -0,0 +1,310 @@
...
+      .text("Column value range");
+
+    </script>
+    <body><svg width=500 height=500 id="svg$colName"></svg></body>
+    """
+  }
+
+  def minMaxAnalysis(
+      df: DataFrame,
+      colNames: Seq[String],
```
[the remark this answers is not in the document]
sezruby: I updated the pr to support nested column. Thanks!

sezruby:
@clee704  could you review and vote the PR? Thanks!

clee704 (approved):

PR #504, opened 2021-10-12
Add spark session extension for Hyperspace

### LINKED ISSUE #503: [FEATURE REQUEST]: Enable hyperspace with SparkSessionExtention
Need to add a functionality to enable Hyperspace with `SparkSessionExtention` to avoid change in user code.

**Acceptance criteria** 

Describe what you consider (e.g., functionality requirement) has to be true to mark this feature as complete

- [ ] Enable Hyperspace with SparkSessionExtention

**Additional context**

### Discussed in https://github.com/microsoft/hyperspace/discussions/499

<div type='discussions-op-text'>

<sup>Originally posted by **ashokblend** October  7, 2021</sup>
We have server code which connects to spark jdbc(thrift) server to execute sql.

We can have batch jobs which can create index for our data lake.

But we are not getting how to enable hyperspace for spark thrift server, where we do not have control in initialising sparksession.

So question is how to enable hyperspace in spark thriftserver ?</div>

paryoja:

 - **Tracking Issue**: Fixes #503 

Added a feature to enable Hyperspace with SparkSessionExtention

Yes. Users now can enable Hyperspace as follows
```
spark-shell -c spark.sql.extensions=com.microsoft.hyperspace.HyperspaceSparkSessionExtension
```
or
```
val spark = SparkSession
       .builder()
       .appName("...")
       .master("...")
       .config("spark.sql.extensions", "com.microsoft.hyperspace.HyperspaceSparkSessionExtension")
       .getOrCreate()
```

Manually with spark-shell. If an automated test is required, I will add it.

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
+++ b/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
@@ -20,6 +20,9 @@ import org.apache.spark.sql.{SparkSession, SparkSessionExtensions}
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
 
+import com.microsoft.hyperspace.index.execution.BucketUnionStrategy
+import com.microsoft.hyperspace.index.rules.ApplyHyperspace
+
 /**
  * An extension for Spark SQL to activate Hyperspace.
  *
@@ -55,13 +58,27 @@ class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
     }
   }
 
-  type RuleBuilder = SparkSession => Rule[LogicalPlan]
   override def apply(extensions: SparkSessionExtensions): Unit = {
     extensions.injectOptimizerRule { sparkSession =>
       // Enable Hyperspace to leverage indexes.
-      sparkSession.enableHyperspace()
+      HyperspaceSparkSessionExtension.addOptimizationsIfNeeded(sparkSession)
       // Return a dummy rule to fit in interface of injectOptimizerRule
       DummyRule
+    }
+  }
+}
+
+object HyperspaceSparkSessionExtension {
+  def addOptimizationsIfNeeded(sparkSession: SparkSession): Unit = {
+    if (!sparkSession.sessionState.experimentalMethods.extraOptimizations.contains(
+        ApplyHyperspace)) {
+      sparkSession.sessionState.experimentalMethods.extraOptimizations ++=
+        ApplyHyperspace :: Nil
+    }
+    if (!sparkSession.sessionState.experimentalMethods.extraStrategies.contains(
+        BucketUnionStrategy)) {
+      sparkSession.sessionState.experimentalMethods.extraStrategies ++=
+        BucketUnionStrategy :: Nil
     }
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
+++ b/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
@@ -1,0 +1,68 @@
+/*
+ * Copyright (2020) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace
+
+import org.apache.spark.sql.{SparkSession, SparkSessionExtensions}
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
+
+/**
+ * An extension for Spark SQL to activate Hyperspace.
+ *
+ * Example to run a `spark-submit` with Hyperspace enabled:
+ * {{{
+ *   spark-submit -c spark.sql.extensions=com.microsoft.hyperspace.HyperspaceSparkSessionExtension
+ * }}}
+ *
+ * Example to create a `SparkSession` with Hyperspace enabled:
+ * {{{
+ *    val spark = SparkSession
+ *       .builder()
+ *       .appName("...")
+ *       .master("...")
+ *       .config("spark.sql.extensions", "com.microsoft.hyperspace.HyperspaceSparkSessionExtension")
+ *       .getOrCreate()
+ * }}}
+ */
+
+class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
+
+  /**
+   * If HyperspaceRule is injected directly to OptimizerRule with HyperspaceExtension,
+   * the order of applying rule is different from without HyperspaceExtension
+   * (i.e., explicitly calling enableHyperspace). To make behavior consistently,
+   * current implementation of HyperspaceExtension uses a trick to call enableHyperspace
+   * before rule is applied. Since the interface of injectOptimizerRule should return rule builder,
+   * it returns a dummy rule that does nothing. It may increase overhead slightly
+   * because enableHyperspace is called once for each evaluation of spark plan.
+   */
+  private case class DummyRule() extends Rule[LogicalPlan] {
+    override def apply(plan: LogicalPlan): LogicalPlan = {
+      plan
+    }
+  }
+
+  type RuleBuilder = SparkSession => Rule[LogicalPlan]
+  override def apply(extensions: SparkSessionExtensions): Unit = {
+    extensions.injectOptimizerRule { sparkSession =>
+      // Enable Hyperspace to leverage indexes.
+      sparkSession.enableHyperspace()
+      // Return a dummy rule to fit in interface of injectOptimizerRule
+      DummyRule()
+    }
+  }
+}
--- a/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
+++ b/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
@@ -69,6 +69,13 @@ class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
 }
 
 object HyperspaceSparkSessionExtension {
+
+  /**
+   * Add ApplyHyperspace and BucketUnionStrategy into extraOptimization
+   * and extraStrategies, respectively, to make Spark can use Hyperspace.
+   *
+   * @param sparkSession Spark session that will use Hyperspace
+   */
   def addOptimizationsIfNeeded(sparkSession: SparkSession): Unit = {
     if (!sparkSession.sessionState.experimentalMethods.extraOptimizations.contains(
         ApplyHyperspace)) {
--- a/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
+++ b/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
@@ -61,31 +61,9 @@ class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
   override def apply(extensions: SparkSessionExtensions): Unit = {
     extensions.injectOptimizerRule { sparkSession =>
       // Enable Hyperspace to leverage indexes.
-      HyperspaceSparkSessionExtension.addOptimizationsIfNeeded(sparkSession)
+      sparkSession.addOptimizationsIfNeeded()
       // Return a dummy rule to fit in interface of injectOptimizerRule
       DummyRule
-    }
-  }
-}
-
-object HyperspaceSparkSessionExtension {
-
-  /**
-   * Add ApplyHyperspace and BucketUnionStrategy into extraOptimization
-   * and extraStrategies, respectively, to make Spark can use Hyperspace.
-   *
-   * @param sparkSession Spark session that will use Hyperspace
-   */
-  def addOptimizationsIfNeeded(sparkSession: SparkSession): Unit = {
-    if (!sparkSession.sessionState.experimentalMethods.extraOptimizations.contains(
-        ApplyHyperspace)) {
-      sparkSession.sessionState.experimentalMethods.extraOptimizations ++=
-        ApplyHyperspace :: Nil
-    }
-    if (!sparkSession.sessionState.experimentalMethods.extraStrategies.contains(
-        BucketUnionStrategy)) {
-      sparkSession.sessionState.experimentalMethods.extraStrategies ++=
-        BucketUnionStrategy :: Nil
     }
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
+++ b/src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
@@ -38,7 +38,6 @@ import org.apache.spark.sql.catalyst.rules.Rule
  *       .getOrCreate()
  * }}}
  */
-
 class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
 
   /**
@@ -50,7 +49,7 @@ class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
    * it returns a dummy rule that does nothing. It may increase overhead slightly
    * because enableHyperspace is called once for each evaluation of spark plan.
    */
-  private case class DummyRule() extends Rule[LogicalPlan] {
+  private case object DummyRule extends Rule[LogicalPlan] {
     override def apply(plan: LogicalPlan): LogicalPlan = {
       plan
     }
@@ -62,7 +61,7 @@ class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
       // Enable Hyperspace to leverage indexes.
       sparkSession.enableHyperspace()
       // Return a dummy rule to fit in interface of injectOptimizerRule
-      DummyRule()
+      DummyRule
     }
   }
 }
--- a/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
@@ -19,6 +19,10 @@ package com.microsoft.hyperspace.index
 import org.apache.spark.sql.internal.SQLConf
 
 object IndexConstants {
+  // If it is set as false, Hyperspace will not be applied.
+  val HYPERSPACE_APPLY_ENABLED = "spark.hyperspace.apply.enabled"
+  val HYPERSPACE_APPLY_ENABLED_DEFAULT = "true"
+
   val INDEXES_DIR = "indexes"
 
   // Config used for setting the system path, which is considered as a "root" path for Hyperspace;
--- a/src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
@@ -24,6 +24,7 @@ import com.microsoft.hyperspace.{ActiveSparkSession, Hyperspace}
 import com.microsoft.hyperspace.actions.Constants
 import com.microsoft.hyperspace.index.IndexLogEntry
 import com.microsoft.hyperspace.telemetry.HyperspaceEventLogging
+import com.microsoft.hyperspace.util.HyperspaceConf
 
 /**
  * Transform the given plan to use Hyperspace indexes.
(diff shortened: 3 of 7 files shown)


src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -19,6 +19,11 @@ package com.microsoft.hyperspace.index
 import org.apache.spark.sql.internal.SQLConf
 
 object IndexConstants {
+  // If it is set as true, Hyperspace will not be applied
+  // even when user explicitly called "spark.enableHyperspace()".
+  val APPLY_HYPERSPACE_BYPASS = "spark.hyperspace.applyhyperspace.bypass"
```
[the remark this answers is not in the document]
sezruby: Could you update with forceDisable & rebase the branch?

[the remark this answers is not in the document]
sezruby: Also could you open a PR to update user document for extension? I think we can add some here:
https://microsoft.github.io/hyperspace/docs/ug-quick-start-guide/#enable-hyperspace

src/test/scala/com/microsoft/hyperspace/HyperspaceExtensionTest.scala
```diff
@@ -0,0 +1,199 @@
...
+      empDF
+        .join(deptDF, empDF("deptId") === deptDF("deptId"))
+        .select(empDF("empName"), deptDF("deptName"))
+
+    verifyIndexUsage(
+      eqJoinQuery,
+      getIndexFilesPath(deptIndexConfig.indexName) ++ getIndexFilesPath(empIndexConfig.indexName))
+
+  }
+
```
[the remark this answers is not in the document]
clee704: How about extracting them as a trait? It could be done separately or in this PR.

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,51 @@
...
+ * {{{
+ *    val spark = SparkSession
+ *       .builder()
+ *       .appName("...")
+ *       .master("...")
+ *       .config("spark.sql.extensions", "com.microsoft.hyperspace.HyperspaceSparkSessionExtension")
+ *       .getOrCreate()
+ * }}}
+ */
+class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
```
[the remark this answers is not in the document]
clee704: How about simplifying the enable/disable process by always putting the `ApplyHyperspace` rule in the optimizer and just toggling a flag to control whether the rule is applied or not? The overall picture would be like this:

1. Users are required to include `com.microsoft.hyperspace.HyperspaceSparkSessionExtension` in `spark.sql.extensions` to apply Hyperspace rules.
    1. See below for backward compatibility.
    2. `HyperspaceSparkSessionExtension` is responsible for putting `ApplyHyperspace` at the end of `experimentalMethods.extraOptimizations`. This can be done as the current PR, which is a little bit hacky but works. Ideally, there should be a way for extensions to access `extraOptimizations`. We can propose this change to Spark if we want.
2. Users use a config `spark.hyperspace.apply.enabled` to enable/disable the Hyperspace rule application.
    1. The value is `true` by default. Therefore, just setting `spark.sql.extensions` correctly is enough to use Hyperspace.
    2. `ApplyHyperspace` checks this config value and does nothing if it is `false`. 
    3. We might extend this mechanism to allow users to enable/disable individual rules, e.g. disable applying data skipping indexes by setting `spark.hyperspace.apply.dataskipping.enabled` to `false`, or including the rule name in `spark.hyperspace.apply.excludedRules` (for this to be usable, rule names should be user-friendly and comprehensible).
3. For backward compatibility, continue providing `sparkSession.enable/disableHyperspace()`, although they should be deprecated in favor of the new mechanism.
    1. `enableHyperspace()` puts `ApplyHyperspace` at the end of `extraOptimizations` if it's not already there and sets `spark.hyperspace.apply.enabled` to `true`.
    2. `disableHyperspace()` sets `spark.hyperspace.apply.enabled` to `false`. It doesn't remove the rule.

src/test/scala/com/microsoft/hyperspace/HyperspaceExtensionTest.scala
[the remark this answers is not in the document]
paryoja: I will do it as a separate PR

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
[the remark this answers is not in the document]
paryoja: @clee704 Thanks for the suggestion. To check whether I understand correctly, in the above suggestion, the existing program users don't have to change their code but they should at least add `com.microsoft.hyperspace.HyperspaceSparkSessionExtension` to their spark session, right?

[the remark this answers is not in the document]
clee704: No, they don't have to, until we deprecate and remove `sparkSession.enable/disableHyperspace()`.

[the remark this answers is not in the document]
clee704: As for the naming of the configs, how about `spark.hyperspace.optimizer.*`, as `apply` or `applyHyperspace` might be unclear to users?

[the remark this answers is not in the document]
paryoja: > puts ApplyHyperspace at the end of extraOptimizations if it's not already there

Oh I understand it now. I missed the above part :)

Ok. I will make change as you suggested.

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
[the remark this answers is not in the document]
sezruby: Sorry but could you split the document PR? As we will merge it after release. Thanks!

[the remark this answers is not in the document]
paryoja: Of course. I will split it to another PR!

src/main/scala/com/microsoft/hyperspace/util/HyperspaceConf.scala
```diff
@@ -25,6 +25,22 @@ import com.microsoft.hyperspace.index.IndexConstants
...
+   */
+  def hyperspaceApplyEnabled(spark: SparkSession): Boolean = {
+    spark.conf
+      .get(
+        IndexConstants.HYPERSPACE_APPLY_ENABLED,
+        IndexConstants.HYPERSPACE_APPLY_ENABLED_DEFAULT)
+      .toBoolean
+  }
+
+  def setHyperspaceApplyEnabled(spark: SparkSession, apply: Boolean): Unit = {
```
paryoja, line 40: In HyperspaceConf, all functions were getter functions. Is it intended?
Since I needed a getter function somewhere, I made it here.

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+class HyperspaceSparkSessionExtension extends (SparkSessionExtensions => Unit) {
+  type RuleBuilder = SparkSession => Rule[LogicalPlan]
+  override def apply(extensions: SparkSessionExtensions): Unit = {
+    extensions.injectOptimizerRule { sparkSession =>
+      // Enable Hyperspace to leverage indexes.
+      sparkSession.enableHyperspace()
+      // returns a dummy rule to fit in interface of injectOptimizerRule
```
sezruby, line 62: nit:
```suggestion
      // Return a dummy rule to fit in interface of injectOptimizerRule.
```

src/main/scala/com/microsoft/hyperspace/index/IndexConstants.scala
```diff
@@ -19,6 +19,11 @@ package com.microsoft.hyperspace.index
 import org.apache.spark.sql.internal.SQLConf
 
 object IndexConstants {
+  // If it is set as true, Hyperspace will not be applied
```
sezruby, line 22:
```suggestion
  // If it is set as false, Hyperspace will not be applied
```

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,66 @@
...
+ * NOTE: If HyperspaceRule is injected directly to OptimizerRule with HyperspaceExtension,
+ * the order of applying rule is different from without HyperspaceExtension
+ * (i.e., explicitly calling enableHyperspace). To make behavior consistently,
+ * current implementation of HyperspaceExtension uses a trick to call enableHyperspace
+ * before rule is applied. Since the interface of injectOptimizerRule should return rule builder,
+ * it returns a dummy rule that does nothing. It may increase overhead slightly
+ * because enableHyperspace is called once for each evaluation of spark plan.
+ */
+
+class DummyRule extends Rule[LogicalPlan] {
```
sezruby, line 50:
```suggestion
case class DummyRule extends Rule[LogicalPlan] {
```

then you don't need `new` in `new DummyRule`

src/main/scala/com/microsoft/hyperspace/package.scala
```diff
@@ -29,42 +30,53 @@ package object hyperspace {
...
+      }
       sparkSession
     }
 
     /**
-     * Plug out Hyperspace-specific rules.
+     * Set `IndexConstants.HYPERSPACE_APPLY_ENABLED` as false
+     * to stop ApplyHyperspace from applied to Spark plans.
+     * Note that `extraOptimizations` can contains ApplyHyperspace in it,
+     * even after Hyperspace is disabled.
```
sezruby, line 57:
```suggestion
     * Disable Hyperspace indexes.
     *
     * Set `IndexConstants.HYPERSPACE_APPLY_ENABLED` as false
     * to stop applying Hyperspace indexes.
```
I don't think we should expose the impl detail in API comment. Maybe we can move the detailed comment inside the function.

src/main/scala/com/microsoft/hyperspace/package.scala
```diff
@@ -29,42 +30,53 @@ package object hyperspace {
   implicit class Implicits(sparkSession: SparkSession) {
 
     /**
-     * Plug in Hyperspace-specific rules.
+     * Plug in Hyperspace-specific rules and set `IndexConstants.HYPERSPACE_APPLY_ENABLED` as true.
```
sezruby, line 33:
```suggestion
     * Enable Hyperspace indexes.
     *
     * Plug in Hyperspace-specific rules and set `IndexConstants.HYPERSPACE_APPLY_ENABLED` as true.
```

src/main/scala/com/microsoft/hyperspace/package.scala
```diff
@@ -29,42 +30,53 @@ package object hyperspace {
...
-        experimentalMethods.extraOptimizations.filterNot(ApplyHyperspace.equals)
-      experimentalMethods.extraStrategies =
-        experimentalMethods.extraStrategies.filterNot(BucketUnionStrategy.equals)
+      HyperspaceConf.setHyperspaceApplyEnabled(sparkSession, false)
       sparkSession
     }
 
     /**
      * Checks if Hyperspace is enabled or not.
+     * Note that Hyperspace is enabled when:
```
sezruby, line 68:
```suggestion
     *
     * Note that Hyperspace is enabled when:
```

src/main/scala/com/microsoft/hyperspace/util/HyperspaceConf.scala
sezruby: It's usually set by user so there is no setter function here so far.
I'm okay with this location or even not using setter.

src/test/scala/com/microsoft/hyperspace/HyperspaceExtensionTest.scala
```diff
@@ -0,0 +1,199 @@
...
+    def eqJoinQuery(): DataFrame =
+      empDF
+        .join(deptDF, empDF("deptId") === deptDF("deptId"))
+        .select(empDF("empName"), deptDF("deptName"))
+
+    verifyIndexUsage(
+      eqJoinQuery,
+      getIndexFilesPath(deptIndexConfig.indexName) ++ getIndexFilesPath(empIndexConfig.indexName))
+
```
sezruby, line 129: nit: --

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,66 @@
...
+ * NOTE: If HyperspaceRule is injected directly to OptimizerRule with HyperspaceExtension,
+ * the order of applying rule is different from without HyperspaceExtension
+ * (i.e., explicitly calling enableHyperspace). To make behavior consistently,
+ * current implementation of HyperspaceExtension uses a trick to call enableHyperspace
+ * before rule is applied. Since the interface of injectOptimizerRule should return rule builder,
+ * it returns a dummy rule that does nothing. It may increase overhead slightly
+ * because enableHyperspace is called once for each evaluation of spark plan.
+ */
+
+case class DummyRule() extends Rule[LogicalPlan] {
```
sezruby, line 50: Can we move the class inside HyperspaceSparkSessionExtension? and make it a private class?

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,66 @@
...
+ * {{{
+ *    val spark = SparkSession
+ *       .builder()
+ *       .appName("...")
+ *       .master("...")
+ *       .config("spark.sql.extensions", "com.microsoft.hyperspace.HyperspaceSparkSessionExtension")
+ *       .getOrCreate()
+ * }}}
+ *
+ * NOTE: If HyperspaceRule is injected directly to OptimizerRule with HyperspaceExtension,
```
sezruby, line 41: Please move this note inside the class; or maybe above DummyRule, so that it cannot be exposed for extension class comment. I think this is too much detail for users.

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+package com.microsoft.hyperspace
+
+import org.apache.spark.sql.{SparkSession, SparkSessionExtensions}
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
+
+/**
+ * An extension for Spark SQL to activate Hyperspace.
```
sezruby, line 24: Can we move the comment right above the extension class?

src/main/scala/com/microsoft/hyperspace/package.scala
```diff
@@ -29,42 +30,56 @@ package object hyperspace {
   implicit class Implicits(sparkSession: SparkSession) {
 
     /**
-     * Plug in Hyperspace-specific rules.
+     * Enable Hyperspace indexes
```
sezruby, line 33: nit:
```suggestion
     * Enable Hyperspace indexes.
```

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
paryoja: Oh. I see. Thank you for the detailed guide.

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,68 @@
...
+ * {{{
+ *    val spark = SparkSession
+ *       .builder()
+ *       .appName("...")
+ *       .master("...")
+ *       .config("spark.sql.extensions", "com.microsoft.hyperspace.HyperspaceSparkSessionExtension")
+ *       .getOrCreate()
+ * }}}
+ */
+
```
sezruby, line 41: nit: --

sezruby (commented):
LGTM, thanks @paryoja!
@clee704 Could you have another look and approve the PR?

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * If HyperspaceRule is injected directly to OptimizerRule with HyperspaceExtension,
+   * the order of applying rule is different from without HyperspaceExtension
+   * (i.e., explicitly calling enableHyperspace). To make behavior consistently,
+   * current implementation of HyperspaceExtension uses a trick to call enableHyperspace
+   * before rule is applied. Since the interface of injectOptimizerRule should return rule builder,
+   * it returns a dummy rule that does nothing. It may increase overhead slightly
+   * because enableHyperspace is called once for each evaluation of spark plan.
+   */
+  private case class DummyRule() extends Rule[LogicalPlan] {
```
clee704, line 53: nit: `case object DummyRule` avoids object creation

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,68 @@
...
+    override def apply(plan: LogicalPlan): LogicalPlan = {
+      plan
+    }
+  }
+
+  type RuleBuilder = SparkSession => Rule[LogicalPlan]
+  override def apply(extensions: SparkSessionExtensions): Unit = {
+    extensions.injectOptimizerRule { sparkSession =>
+      // Enable Hyperspace to leverage indexes.
+      sparkSession.enableHyperspace()
```
clee704, line 63: According to the new model, `enableHyperspace` only exists for backward compatibility. I think it's better to factor out the rule insertion code out of this method and invoke the method here and from `enableHyperspace`.

src/main/scala/com/microsoft/hyperspace/index/rules/ApplyHyperspace.scala
```diff
@@ -42,10 +43,13 @@ object ApplyHyperspace
...
 
   override def apply(plan: LogicalPlan): LogicalPlan = {
-    if (disableForIndexMaintenance.get) {
+    if (!HyperspaceConf.hyperspaceApplyEnabled(spark)) {
       return plan
     }
 
+    if (disableForIndexMaintenance.get) {
+      return plan
+    }
```
clee704, line 52: nit: can combine into a single `if`

src/main/scala/com/microsoft/hyperspace/package.scala
```diff
@@ -29,42 +30,56 @@ package object hyperspace {
...
+      HyperspaceConf.setHyperspaceApplyEnabled(sparkSession, false)
       sparkSession
     }
 
     /**
      * Checks if Hyperspace is enabled or not.
      *
+     * Note that Hyperspace is enabled when:
+     * 1) `ApplyHyperspace` exists in extraOptimization
+     * 2) `BucketUnionStrate` exists in BucketUnionStrategy and
```
clee704, line 73: nit: `BucketUnionStrategy` in extraStrategies

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
paryoja: Do you mean this part?
```
      if (!sparkSession.sessionState.experimentalMethods.extraOptimizations.contains(
          ApplyHyperspace)) {
        sparkSession.sessionState.experimentalMethods.extraOptimizations ++=
          ApplyHyperspace :: Nil
      }
      if (!sparkSession.sessionState.experimentalMethods.extraStrategies.contains(
          BucketUnionStrategy)) {
        sparkSession.sessionState.experimentalMethods.extraStrategies ++=
          BucketUnionStrategy :: Nil
      }
```
Where should I put this code because `package object hyperspace` is quite customer side interfaces, so not sure if it is ok to create a function like

```
package object hyperspace {

  /**
   * Hyperspace-specific implicit class on SparkSession.
   */
  implicit class Implicits(sparkSession: SparkSession) {

    def enableHyperspace(): SparkSession = {
      HyperspaceConf.setHyperspaceApplyEnabled(sparkSession, true)
      addOptimizationsIfNeeded()
      sparkSession
    }

    private def addOptimizationsIfNeeded(): Unit = {
      if (!sparkSession.sessionState.experimentalMethods.extraOptimizations.contains(
          ApplyHyperspace)) {
        sparkSession.sessionState.experimentalMethods.extraOptimizations ++=
          ApplyHyperspace :: Nil
      }
      if (!sparkSession.sessionState.experimentalMethods.extraStrategies.contains(
          BucketUnionStrategy)) {
        sparkSession.sessionState.experimentalMethods.extraStrategies ++=
          BucketUnionStrategy :: Nil
      }
    }
  }
}
```

clee704: You can put `addOptimizationsIfNeeded()` in a companion object `HyperspaceSparkSessionExtension` and call the method from here and `enableHyperspace()`.

paryoja: @clee704 I've update the functions so please check whether the implement was done as you expected. Thanks!

paryoja: @clee704  can you check whether I implemented as you expected?

src/main/scala/com/microsoft/hyperspace/HyperspaceSparkSessionExtension.scala
```diff
@@ -0,0 +1,91 @@
...
+
+object HyperspaceSparkSessionExtension {
+
+  /**
+   * Add ApplyHyperspace and BucketUnionStrategy into extraOptimization
+   * and extraStrategies, respectively, to make Spark can use Hyperspace.
+   *
+   * @param sparkSession Spark session that will use Hyperspace
+   */
+  def addOptimizationsIfNeeded(sparkSession: SparkSession): Unit = {
```
sezruby, line 79: I think it's better to locate the function in package.scala

clee704 (approved):

Issue #515, opened 2021-12-09
[PROPOSAL]: ZOrderCoveringIndex


sezruby:
NOTE: This is an issue for tracking Z-order covering index. 

The initial implementation is ready in #495.

## Problem Statement
Currently, CoveringIndex data is stored as bucketed, and it just sorted within a bucket.
So it is usually less effective for filter queries compared to fully sorted data.
In addition, the index is applied when a given query has conditions only with the first indexed column. 

## Background and Motivation
TBD

## Proposed Solution
Introduce Z-order covering index type to optimize filter query for multiple columns.
TBD

## Known/Potential Compatibility Issues
TBD

## Design
TBD

## Implementation
[ ] CoveringIndex refactoring
[ ] Z-address function
[ ] Rules & Filter index

PR #517, opened 2021-12-10
Add ZOrderUDF and ZOrderField for Z-address calculation

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: #515 
 - **Dependencies**: n/a

Add ZOrderUDf and ZOrderField classes for Z-address calculation.

No

unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderField.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderField.scala
@@ -1,0 +1,569 @@
+/*
+ * Copyright (2021) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.hyperspace.index.zordercovering
+
+import java.util.BitSet
+
+import org.apache.spark.sql.catalyst.util
+import org.apache.spark.sql.types._
+
+import com.microsoft.hyperspace.HyperspaceException
+
+abstract class ZOrderField extends Serializable {
+  val name: String
+  val bitLen: Int
+  def getBitSet(valueAny: Any): BitSet
+  def throwOutOfRangeException(minVal: Any, maxVal: Any, value: Any): Unit = {
+    throw HyperspaceException(
+      "Unexpected operation on ZOrderField. Value out of range: " +
+        s"name=$name, value=$value, min=$minVal, max=$maxVal")
+  }
+
+  protected val emptyBitSet = new BitSet
+  protected def checkEmptySetEligible(valueAny: Any): Boolean = {
+    bitLen == 0 || valueAny == null
+  }
+}
+
+/**
+ * ZOrderField for numeric types based on percentile values of the original column values.
+ *
+ * If the data is skewed, using the original value for z-address cannot significantly affect
+ * the final z-order thus the column values might not be distributed well compared to other
+ * evenly distributed z-order columns. To mitigate the issue, we use the percentile based
+ * bucket index for z-address.
+ *
+ * We assign a bucket index (0 to numDistBuckets - 1) for each value based on its percentile
+ * and use the index value for z-address. Unlike bucket sorting, all values in a bucket will
+ * just have the same bucket index.
+ *
+ * Steps to calculate the bucket index:
+ * 1) define range subsets based on the given approximate percentiles.
+ * 2) get the range subset index that a value is belong to.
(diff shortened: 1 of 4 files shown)


src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderField.scala
```diff
@@ -0,0 +1,569 @@
...
+  }
+
+  protected val emptyBitSet = new BitSet
+  protected def checkEmptySetEligible(valueAny: Any): Boolean = {
+    bitLen == 0 || valueAny == null
+  }
+}
+
+/**
+ * ZOrderField for numeric types based on percentile values of the original column values.
```
MironAtHome, line 43: While reading approach design blueprint in this comment, I am missing the critical component of thinking and design approach, covered indexes make sense in most situation, except, what is called, an lookup, when index is used to calculate row address to read additional data.
It would seem that in such a query case, particularly with large subquery passed in for matching, having unaligned partitions between index and outer rowstore, would result in shuffle, which would either reduce value of index used, or, depending on optimizer, prompt scan of rowstore, ignoring index altogether.
The rule inference for effective colocation of data was long thought out, and it seems like a very nice time to either tag on to this feature, or add it as a V2 for this feature, going forward.

src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderField.scala
```diff
@@ -0,0 +1,569 @@
...
+object ZOrderField {
+  def build(
+      name: String,
+      dataType: DataType,
+      minVal: Any,
+      maxVal: Any,
+      quantiles: Seq[Any] = Nil,
+      quantileEnabled: Boolean = false): ZOrderField = {
+    dataType match {
+      case LongType =>
```
MironAtHome, line 482: does it really have to go through the list, or can the type be extracted from dataType parameter for brevity?

sezruby: I don't think the implementation is inefficient and it would be okay as this is one time cost for each z-order columns.
Could you elaborate more if it still needs to be improved?

src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderField.scala
sezruby: Are you suggesting partitioned covering index e.g. #509? Otherwise I don't really get the point of your comment..

MironAtHome: partitioning is not a "secret souse" in an of itself. By partitioning we don't get much win, unless partition is aligned ( colocated physically ) with outer row, the data row.
In fact this
org.apache.spark.sql.connector.catalog.InMemoryPartitionTable
pushes time to deliver this feature too close, to feel comfortable. I really want us to make it happen ASAP, rather than turn out late to finish line with someone already being past us.

src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderField.scala
clee704: I think @MironAtHome is saying about code repetition. Having a long list of `if` or `case` conditions with similar code modulo some parameter is a typical antipattern. If it were C++, this kind of repetition can be easily factored out using templates, but in Scala, I think it needs more boilerplates and glue code than necessary.

We could have done something like this:
```scala
val builder: {def apply(name: String, min: Any, max: Any): ZOrderField} = dataType match {
  case IntegerType => IntMinMaxZOrderField
  case LongType => LongMinMaxZOrderField
  ...
}
builder(name, min, max)
```

But the actual code is slightly more complex than the simple `apply` function above and we would need to define a factory trait and factories for each class anyway, which seems to be a little too much considering there will be very few, if any, new types to Spark in the future, let alone whether we will support them or not.

I do prefer a modular approach where I would group classes by their differentiating properties (so I would put IntMinMaxZOrderField, IntPercentileZOrderField, and factories for them in IntZOrderField.scala, and so on) and make the combining code like `ZOrderField.build` agnostic about different implementations, meaning that you don't have to change the function when you add/remove implementations. But such a modular approach makes more sense for a larger abstraction. If we need to, we can change to that later.

src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderField.scala
```diff
@@ -0,0 +1,569 @@
...
+          DoublePercentileZOrderField(
+            name,
+            (minVal +: quantiles :+ maxVal).asInstanceOf[Seq[Double]])
+        } else {
+          DoublePercentileZOrderField(
+            name,
+            Seq(minVal.asInstanceOf[Double], maxVal.asInstanceOf[Double]))
+        }
+      case FloatType =>
+        // minVal, maxVal can be Float.
```
clee704, line 525: I think the comment should be the opposite to make sense? When can minVal/maxVal can be `Double`?

clee704 (approved):
I left some comments but none are significant. Please feel free to submit as most code has been already reviewed.

src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderField.scala
sezruby: If it's from `approxQuantile` then the value type is Double. If it's from aggs(min, max), then it's Float type.

PR #518, opened 2021-12-15
Introduce ZOrderCoveringIndex

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**:  #515 
 - **Dependencies**: n/a

Introduce a new covering index type that is Z-ordering dataset.
The current covering index is bucketed by the indexed column and sorted within a bucket. (a file) For filter queries, globally sorted data can be efficient than bucketed + partially sorted dataset.

However, sorting is usually limited to the first sorting column; if we execute a query without conditions of the first sorting column, we should read all files for the query.

Z-order covering index is a basically sorted dataset by Z-address which is derived from values of indexed columns for each row. As a result, rows having similar values can be collocated within a file. With Z-ordered dataset, we could skip some of unnecessary data by min/max pruning.

Usage
```
hs.createIndex(ZOrderCoveringIndex("indexName", Seq("zOrderCol1", "zOrderCol2"), Seq("includedCol1", "includedCol2"))
```

Yes, the PR introduces a new index type.

unit test

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderCoveringIndexConfig.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderCoveringIndexConfig.scala
@@ -53,7 +53,7 @@ case class ZOrderCoveringIndexConfig(
       resolvedIndexedColumns.map(_.normalizedName),
       resolvedIncludedColumns.map(_.normalizedName),
       indexData.schema,
-      HyperspaceConf.ZOrderCovering.approxSourceBytesPerPartition(ctx.spark),
+      HyperspaceConf.ZOrderCovering.targetSourceBytesPerPartition(ctx.spark),
       properties)
     (index, indexData)
   }
--- a/src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderCoveringIndexConfig.scala
+++ b/src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderCoveringIndexConfig.scala
@@ -1,0 +1,60 @@
+/*
+ * Copyright (2021) The Hyperspace Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


sezruby:
@clee704 Could you review the PR? I'll update user document & a notebook for it.

src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderCoveringIndexConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+        sourceData,
+        indexedColumns,
+        includedColumns,
+        IndexUtils.hasLineageColumn(properties))
+
+    val index = ZOrderCoveringIndex(
+      resolvedIndexedColumns.map(_.normalizedName),
+      resolvedIncludedColumns.map(_.normalizedName),
+      indexData.schema,
+      HyperspaceConf.ZOrderCovering.approxSourceBytesPerPartition(ctx.spark),
```
clee704, line 56: Just thinking out loud.

At first, I thought using `numBucketsForIndex` for the number of index files would be better for simplicity and consistency, as the difference between hash bucketed covering index and Z-order covering index is how index files are ordered and everything else can remain the same.

But that will make the number of index files kind of fixed to 200 files, as many users won't care about the numBuckets config. It can be an issue if the data is very large and 200 files are too few for efficient handling. But at the same time, it's also an issue for hash bucketed covering index. So we either rely on the user to change the config appropriately for their data or need to find a better way.

With the current code, small data (< 10 GB) won't get much boost with Z-order index, as there will be too few index files.

Anyway, it would be nice if we decide on a way to determine the number of files, whether it is the old way (numBuckets) or new (target size per file), and apply it consistently for both hash bucketed covering index and Z-order covering index.

src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderFilterIndexRule.scala
```diff
@@ -0,0 +1,150 @@
+/*
```
clee704, line 1: Just curious, what's the difference between these rules and rules for hash bucketed covering index? If they are the same, can't we reuse the code?

src/main/scala/com/microsoft/hyperspace/index/zordercovering/ZOrderCoveringIndexConfig.scala
clee704: Ah, I forgot about the importance of exact numbers for bucketing. Then I think it's okay to have separate parameters.

clee704 (approved):
LGTM, thanks!

PR #521, opened 2022-01-14
Support MinMaxAnalysisUtil.analyzeIndex API

sezruby:

 - **Tracking Issue**: n/a
 - **Parent Issue**: n/a

Add MinMaxAnalysisUtil.analyzeIndex API to analyze min/max data layout of index data. It only supports covering index type.

This PR introduces the following API:
```scala
// spark session, index name, result format
displayHTML(MinMaxAnalysisUtil.analyzeIndex(spark, "zordertestindex3", "html"))

// spark session, index name, target column names, result format
displayHTML(MinMaxAnalysisUtil.analyzeIndex(spark, "zordertestindex3", Seq("colA"), "html"))
```

Without it, users may need to use a dummy filter query (like the following code) to check the index data and it's only for the selected index.

```scala
spark.enableHyperspace
val query = df.filter(col("colA").isNotNull) // Add filter condition for applying the index

displayHTML(MinMaxAnalysisUtil.analyze(query, Seq("colA", "colB"), format = "html")) 
```

Yes

via notebook

### DIFF
--- a/src/main/scala/com/microsoft/hyperspace/util/MinMaxAnalysisUtil.scala
+++ b/src/main/scala/com/microsoft/hyperspace/util/MinMaxAnalysisUtil.scala
@@ -25,7 +25,8 @@ import org.apache.spark.sql.catalyst.util.TypeUtils
 import org.apache.spark.sql.functions.{col, input_file_name, max, min}
 import org.apache.spark.sql.types.{StructField, StructType}
 
-import com.microsoft.hyperspace.HyperspaceException
+import com.microsoft.hyperspace.{Hyperspace, HyperspaceException}
+import com.microsoft.hyperspace.index.IndexLogEntry
 
 case class MinMaxAnalysisResult(
     colName: String,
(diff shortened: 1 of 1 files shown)


src/main/scala/com/microsoft/hyperspace/util/MinMaxAnalysisUtil.scala
```diff
@@ -777,4 +778,31 @@ object MinMaxAnalysisUtil extends MinMaxAnalysis {
...
     analyze(df, colNames, "text")
   }
+
+  private def latestIndexEntry(spark: SparkSession, indexName: String): IndexLogEntry = {
+    val idxManager = Hyperspace.getContext(spark).indexCollectionManager
+    val latestVer = idxManager.getIndexVersions(indexName, Seq("ACTIVE")).max
+    val indexEntry = idxManager.getIndex(indexName, latestVer).get
+
+    if (!Seq("CoveringIndex", "ZOrderCoveringIndex").contains(indexEntry.derivedDataset.kind)) {
+      throw HyperspaceException(s"Does not support index type: ${indexEntry.derivedDataset.kind}")
```
paryoja, line 788: What if this is checked in the analyzeIndex? And is it possible to check supported index types more systemically?
If someone implement a new type of "CoveringIndex", it might be hard to come here to add it in the supported type.

